PR #470, opened 2015-01-02
New prospect fpr resources

### LINKED ISSUE #337: Check resource walking code to see if this API method works.
This post by harvester seems to indicate that there exists an API method that might replace our need for a complex fuel line walk for stage: values:

http://forum.kerbalspaceprogram.com/threads/99200-Devnote-Tuesdays-The-Building-Buildings-Edition?p=1525416&viewfull=1#post1525416

erendrake:
Resolves #337


src/kOS.Safe.Test/ListValueTest.cs
```diff
@@ -62,17 +62,16 @@ public void CanGetIndex()
             var length = InvokeDelegate(list, "LENGTH");
             Assert.AreEqual(4,length);
 
-            IIndexable indexable = list;
```
erendrake: These are changed because the ListValue class correctly implements IList directly now so 

    var list = new ListValue();
    list[0] = foo;

and the rest of basic list stuff you can now do directly.

src/kOS.Safe/Encapsulation/Enumerator.cs
```diff
@@ -0,0 +1,42 @@
+using System.Collections;
```
erendrake: moved to its own file, unchanged for the most part

src/kOS.Safe/Encapsulation/ListValue.cs
```diff
@@ -1,56 +1,119 @@
...
 using System.Text;
-using System.Diagnostics;
-
-using kOS.Safe.Encapsulation.Suffixes;
-using kOS.Safe.Properties;
 
 namespace kOS.Safe.Encapsulation
 {
-    public class ListValue : Structure, IIndexable
+    public class ListValue<T> : Structure, IList<T>
```
erendrake: this a new generic list that the old ListValue extends

src/kOS.Safe/Encapsulation/ListValue.cs
```diff
@@ -202,74 +250,42 @@ private string ListDump()
...
-        public void SetIndex(int index, object value)
-        {
-            list[index] = value;
-        }
-
-        #endregion IIndexable Members
     }
 
-    public class Enumerator : Structure
+    public class ListValue : ListValue<object>
```
erendrake: The original listvalue that the unit tests are still going against and most of the code still uses. These list changes were a little off the reservation.

src/kOS.Safe/Encapsulation/SuffixedList.cs
```diff
@@ -0,0 +1,21 @@
+using System.Collections.Generic;
```
erendrake: This class is the reason i did this list work. This will only let you make a list if its type is a structure that can be used by users. I would have made this the case everywhere but we do serve quite a few value types with lists and that wouldnt work.

src/kOS/Function/BuildList.cs
```diff
@@ -38,7 +40,7 @@ public override void Execute(SharedObjects shared)
                     list = ListValue.CreateList(shared.VolumeMgr.CurrentVolume.GetFileList());
                     break;
                 case "volumes":
-                    list = ListValue.CreateList(shared.VolumeMgr.Volumes);
+                    list = ListValue.CreateList(shared.VolumeMgr.Volumes.Values.ToList());
```
erendrake: I dont think this was working at all before. it was returning a dictionary

src/kOS/Suffixed/ActiveResourceValue.cs
```diff
@@ -0,0 +1,32 @@
+﻿using kOS.Safe.Encapsulation;
```
erendrake: We now have 3 different classes for dealing with resources. I had active and aggregate as one but there was almost no code sharing between them so i thought this made more sense.

src/kOS/Suffixed/AggregateResourceValue.cs
```diff
@@ -1,44 +1,41 @@
...
-                case "CAPACITY":
-                    return capacity;
-                case "PARTS":
-                    return parts;
-            }
-            return base.GetSuffix(suffixName);
+            AddSuffix("NAME", new Suffix<string>(() => name));
+            AddSuffix("AMOUNT", new Suffix<double>(() => amount));
+            AddSuffix("CAPICITY", new Suffix<double>(() => capacity));
+            AddSuffix("PARTS", new Suffix<SuffixedList<PartValue>>(() => parts));
```
erendrake: before this change we were returning a list of ksp parts that would be useless. Because we have so many list suffixes i think i will make a new suffix type to make

    Suffix<SuffixedList<PartValue>>

into
    ListSuffix<PartValue>

src/kOS/Suffixed/Part/PartModuleFields.cs
```diff
@@ -340,17 +340,17 @@ private ListValue AllThings()
             ListValue fields  = AllFields(FORMATTER);
             ListValue events  = AllEvents(FORMATTER);
             ListValue actions = AllActions(FORMATTER);
-            for (int i = 0; i < fields.Count ; ++i)
```
erendrake: this was a simple refactor that came up when ListValue grew its new interface

src/kOS/Suffixed/Part/PartValue.cs
```diff
@@ -161,5 +162,15 @@ public override int GetHashCode()
         {
             return !Equals(left, right);
         }
+
+        public static SuffixedList<PartValue> PartsToList(IEnumerable<global::Part> parts, SharedObjects shared)
```
erendrake: This provides the new type list. I should have added it to PartValueFactory.

src/kOS/Suffixed/StageValues.cs
```diff
@@ -1,28 +1,54 @@
...
         {
-            this.vessel = vessel;
+            this.shared = shared;
+
+            InitializeSuffixes();
+        }
+
+        private void InitializeSuffixes()
+        {
+            AddSuffix("RESOURCES", new Suffix<SuffixedList<ActiveResourceValue>>(GetResourceManifest));
```
erendrake: this is a new feature that is not reflected in the docs, I think ill add a few more to this class and make it its own structure in the docs. for now it is not that different from what we already have so im ok with just not mentioning it for now

src/kOS/Suffixed/StageValues.cs
```diff
@@ -1,28 +1,54 @@
...
         }
 
-        private object GetResourceOfCurrentStage(string resourceName)
+        private double? GetResourceOfCurrentStage(string resourceName)
         {
-            var activeEngines = VesselUtils.GetListOfActivatedEngines(vessel);
-            var total = Utils.ProspectForResource(resourceName, activeEngines);
-            return Math.Round(total, 2);
+            var resource = shared.Vessel.GetActiveResources().FirstOrDefault(r => string.Equals(r.info.name, resourceName, StringComparison.OrdinalIgnoreCase));
+            return resource == null ? null : (double?) Math.Round(resource.amount, 2);
```
erendrake: This is the real crux of what needed to change for #337 

GetConnectedResources was actually not a very easy tool to use with all of its arguments and once i found this and it worked i just decided to use it for this stage class. The GetConnectedResources returned a collection of the actual part resources which will likely be useful when we get to fuel transfer.

src/kOS/Utilities/Utils.cs
```diff
@@ -334,6 +247,18 @@ public static string GetCodeFragment(List<Opcode> codes)
 
             return codeFragment.Aggregate(string.Empty, (current, s) => current + (s + "\n"));
         }
-                
+
+        public static double ProspectForResource(Vessel.ActiveResource resource, List<Part> activeEngines)
```
erendrake: This should be removed, I just missed it.

src/kOS/Suffixed/StageValues.cs
erendrake: If we want to keep STAGE:RESOURCES i would like to add SHIP:RESOURCES to keep them looking the same

src/kOS.Safe/Encapsulation/ListValue.cs
```diff
@@ -1,56 +1,119 @@
...
+            AddSuffix("INSERT", new TwoArgsSuffix<int, T>((index, toAdd) => internalList.Insert(index, toAdd)));
+            AddSuffix("REMOVE", new OneArgsSuffix<int>(toRemove => internalList.RemoveAt(toRemove)));
+            AddSuffix("CLEAR", new NoArgsSuffix(() => internalList.Clear()));
+            AddSuffix("LENGTH", new NoArgsSuffix<int>(() => internalList.Count));
+            AddSuffix("ITERATOR", new NoArgsSuffix<Enumerator>(() => new Enumerator(internalList.GetEnumerator())));
+            AddSuffix("COPY", new NoArgsSuffix<ListValue<T>>(() => new ListValue<T>(this)));
+            AddSuffix("CONTAINS", new OneArgsSuffix<bool, T>(item => internalList.Contains(item)));
+            AddSuffix("SUBLIST", new TwoArgsSuffix<ListValue, int, int>(SubListMethod));
+            AddSuffix("EMPTY", new NoArgsSuffix<bool>(() => !internalList.Any()));
+            AddSuffix("DUMP", new NoArgsSuffix<string>(ListDumpDeep));
```
Dunbaratu: Sigh.  A lot of times automated re-indenting and re-spacing development tools actually get in the way of good spacing by preventing you from making stuff that lines up in columns.

src/kOS.Safe/Encapsulation/ListValue.cs
Dunbaratu: Uhhh... so you're allowed to have two entirely different classes that share the same identifier name but differ only in the fact that one is templated with <> and the other is not?  Ugh.  that's ugly.

src/kOS.Safe/Encapsulation/SuffixedList.cs
Dunbaratu: I like the idea in general of making it so when a user does LIST(), they get a list that can only contain kOS objects and can't contain underlying C# objects that have no kOS wrapper.  But that general idea requires that the FIRST step be to make it so we wrapper around all the primitives like string, float, int, etc.  Then you can make a List that requires its members to be of type kOSObject, where kOSObject is the base all our wrappers derive from.

But this... this is only halfway there and I'm not sure if that's worth the work because it would just have to be gutted and done over again later when the full solution happens.

What was the rationale for WHY we needed a list that is only allowed to contain suffixed items, but don't need one to contain other kOS objects.  What was the use case this was trying to solve?

src/kOS/Function/BuildList.cs
Dunbaratu: I agree.  There's no way that could have been usable.. except that you could probably print it and thus you'd see its ToString() in the terminal.  I'm not sure what the ToString() of a dictionary is.  maybe it dumps the values out in a way that a user might have made sense of?  But anyway, yeah I agree with this change.

src/kOS/Suffixed/AggregateResourceValue.cs
```diff
@@ -1,44 +1,41 @@
...
 namespace kOS.Suffixed
 {
     public class AggregateResourceValue : Structure
     {
         private readonly string name;
+        private readonly SharedObjects shared;
         private double amount;
         private double capacity;
-        private readonly ListValue parts;
+        private readonly SuffixedList<PartValue> parts;
```
Dunbaratu: Now it's my turn to turn the tables on you and ask you to defend redundancy.

What was wrong with ListValue<PartValue>?  What was the incentive for making a special kind of derivative of ListValue which is only templated for suffixable objects as its elements, when you still have to go and be more specific anyway and supply the exact type of suffixable object (in this case PartValue) anyway like this?

src/kOS/Suffixed/AggregateResourceValue.cs
Dunbaratu: I'm not sure I understand what the meaning of :PARTS is even supposed to have been here, given that Aggregate resource, I assume, is a thing like liquidfuel, oxidizer, etc.  In that case I don't get what it means to say "these are the parts that are inside this ship's liquidfuel".

src/kOS/Suffixed/Part/PartValue.cs
Dunbaratu: Again, same question.  What would have been supposedly inferior about doing this:

public static **ListValue**< PartValue > PartsToList(IEnumerable< global::Part > parts, SharedObjects shared)

src/kOS/Suffixed/StageValues.cs
Dunbaratu: Hmm.... It occurs to me reading this that our new development standard that says all changes must be accompanied by user docs to do with them doesn't really account for the case of a feature that is being *deliberately* left undocumented for now but is intended to be exposed later on.

I agree about making STAGE:RESOURCES and SHIP:RESOURCES congruent.

src/kOS/Suffixed/StageValues.cs
Dunbaratu: Is the meaning of GetActiveResources just the ones in the current stage plus the ones that are stage-less?  (stage-less meaning things like ElectricCharge where all stages can always access the entire full ship's supply).

src/kOS/Utilities/Utils.cs
```diff
@@ -63,90 +63,6 @@ public static bool IsValidRotation(Quaternion quaternion)
...
-            {
-                if (DEBUG_WALK) Debug.Log(indent + "- Part is target of a fuel line, traversing fuel line upstream.");
-                ret += ProspectForResource(resourceName, fuelLine.parent, lines, rDepth + 1, ref visited);
-            }
-
-            if (DEBUG_WALK) Debug.Log(indent + "Sum from this branch of the recurse tree is " + ret);
-            return ret;
-#pragma warning restore 162
-        }
-
```
Dunbaratu: It took me a lot of work to make the ugly mess above work, but I'm very happy to see it go away.  It was built on a very fragile assumption about SQUAD keeping the same data structure for fuel transference working the same way, which was never going to happen because the way it worked was kind of dumb in some ways.  (Specifically how fuel hoses worked) and therefore we should have expected SQUAD would clean it up later, which they just did.

src/kOS.Safe/Encapsulation/SuffixedList.cs
erendrake: The reason i went on this tear was because I found yet another place were we were returning a KSP type that was useless to the user but because it was wrapped up in the ListValue you couldnt see it.

If i could have said "ISuffixed and primitives and strings" i likely would have. Dispensing with the other two list types.

src/kOS/Suffixed/AggregateResourceValue.cs
erendrake: The only usage of this right now is in the 

    LIST RESOURCES IN rlist.

So something like 
    
    SET rPartList TO rlist[0]:PARTS.

would give you all of the parts on the craft that had that resource. Ill let you judge if that is cool enough to warrant inclusion.

src/kOS/Suffixed/Part/PartValue.cs
erendrake: you are likely right, i started this wanting to only allow some small subset to be suffixes.

* ISuffixed
* string
* double ( maybe float too)
* int

when sleep deprived chris realized that you cannot specify that kind of constraint in C# he decided that it would shake out in the PR review :)

src/kOS/Suffixed/StageValues.cs
erendrake: I agree that it should be in here. ill add SHIP:RESOURCES and the docs

src/kOS/Suffixed/Part/PartValue.cs
Dunbaratu: While I like the idea of eventually making it so that we are using a sort of ListValue that enforces that only kOS objects are ever seen by a kerboscript, I don't really think it can work as a halfway solution like this, and trying to do it will just mean more work later if we have to rip it out and re-do it, thus my comment.  I think the path to doing what you're trying for here would be this:

Step 1: Make a base class (or an interface - I'm thinking interface so we can multiply inherit), let's call it KOSUserObject, from which ALL objects kerboscript sees derives from.  When a user does SET X TO 1. They don't get a native Integer, they get a KOSNumber which is a subclass of KOSUserObject.  When a user does SET S TO "HELLO" they don't get a native string, they get a KOSString, which is a subclass of KOSUserObject, and so on.

Step 2: Enforce this in the stack and in the dictionary - making it only accept a limited set of objects - all of type KOSObject.  A KOSObject is either a KOSUserObject or a KOSMachineObject - UserObjects being the stuff script see and MachineObjects being stuff the VM needs to exist on the stack, like delegates and the like.

Step 3: make it so that the LIST() that scripts use is only allowed to contain objects of type KOSUserObject.

src/kOS/Suffixed/StageValues.cs
erendrake: Yes, it gives you the available resources for all active parts. 

I believe if you goto your resource panel and click "stage only" this is where the resource panel gets its numbers.

src/kOS/Suffixed/AggregateResourceValue.cs
erendrake: The reason, like i mentioned above, was to make one that we could use for everything and would let us enforce at compiletime that it was a return value that would be useful for scripters. because that failed I have to admit it is not cool enough to exist and i will remove it.

Dunbaratu: Fair enough.  And like I said I'm in favor of the idea as a whole, but it requires a bit more major a refactor than this to really work right for its intended goal.

I do wonder if there may be some other way to design it that would make it possible to actually implement the idea in a "slow" refactor that can be done piecemeal but without needing to remove work.

I am in favor of compile-time enforcement of making it impossible to give a kerbalscript program access to a non-kOS underlying object.  That's why I was thinking of places like the variable dictionary and the stack as being the natural places to start.

erendrake: alright, that list type is removed, thank you very much for the review.

PR #476, opened 2015-01-03
Waypoints: fixes #475

### LINKED ISSUE #475: Ability to read the new FinePrint waypoints from KSP 0.90.
Let people get waypoint data so they can drive rovers and planes to the desired locations of contracts.

Given that issue #444 is for the purpose of getting 3D data from a GeoCoordinates, all this needs to do is return GeoCoordinate data and then the PR for #444, when it gets implemented, should be able to take it from there to get 3D spatial information.

Dunbaratu:
I was able to get useful data out of this.
  Maybe we can merge it before 0.16 because it already works?

### DIFF
--- a/src/kOS/Function/Suffixed.cs
+++ b/src/kOS/Function/Suffixed.cs
@@ -1,8 +1,11 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Linq;
 using kOS.Safe.Encapsulation;
 using kOS.Safe.Function;
 using kOS.Suffixed;
 using kOS.Utilities;
+using FinePrint;
 
 namespace kOS.Function
 {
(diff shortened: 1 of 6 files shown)


src/kOS/Function/Suffixed.cs
```diff
@@ -308,4 +311,51 @@ public override void Execute(SharedObjects shared)
...
+        {
+            // ReSharper disable SuggestUseVarKeywordEvident
+            ListValue<WaypointValue> returnList = new ListValue<WaypointValue>();
+            // ReSharper ensable SuggestUseVarKeywordEvident
+
+            WaypointManager wpm = WaypointManager.Instance();
+            if (wpm == null)
+                shared.Cpu.PushStack(returnList); // When no waypoints exist, there isn't even a waypoint manager at all.
+
+            List<Waypoint> points = wpm.AllWaypoints();
```
erendrake: if wpm is null here it will push onto the stack and throw an exception.

src/kOS/Function/Suffixed.cs
```diff
@@ -308,4 +311,51 @@ public override void Execute(SharedObjects shared)
...
+            FinePrint.WaypointManager wpm = WaypointManager.Instance();
+            if (wpm == null)
+            {
+                shared.Cpu.PushStack(null); // When no waypoints exist, there isn't even a waypoint manager.
+                // I don't like returning null here without the user being able to test for that, but
+                // we don't have another way to communicate "no such waypoint".  We really need to address
+                // that problem once and for all.
+            }
+
+            Waypoint point = wpm.AllWaypoints().FirstOrDefault(p => String.Equals(p.name, pointName,StringComparison.OrdinalIgnoreCase));
```
erendrake: Same as before should we be returning after we push the stack?

src/kOS/Function/Suffixed.cs
Dunbaratu: Ah yes you are correct.
change it to 
```
    if (wpm == null)
    {
        shared.Cpu.PushStack(returnList); // When no waypoints exist, there isn't even a waypoint manager at all.
        return;
    }
```
This was copied from code that was not originally a kOS function.  That line didn't push to the CPU stack, it just did a return directly.  I changed it to a push to the stack but forgot that this would mean it falls through and continues on.

src/kOS/Function/Suffixed.cs
Dunbaratu: Same as before - yes.  This was because I moved the code from a normal method into a kOS function and changed the return to a PushStack and forgot that it would fall through.

PR #480, opened 2015-01-03
finished documentation on the new resource classes

erendrake:
also removed a few suffixes that i added that in hindsight dont really add much value

### DIFF
--- a/doc/source/bindings.rst
+++ b/doc/source/bindings.rst
@@ -114,6 +114,8 @@ Any other resources that you have added using other mods should be
 query-able this way, provided that you spell
 the term exactly as it appears in the resources window.
 
+You can also get a list of all resources, either in SHIP: or STAGE: with the :RESOURCES suffix. 
+
 ALT ALIAS
 ---------
 
--- a/src/kOS/Suffixed/AggregateResourceValue.cs
+++ b/src/kOS/Suffixed/AggregateResourceValue.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 using kOS.Safe.Encapsulation;
 using kOS.Safe.Encapsulation.Suffixes;
 using kOS.Suffixed.Part;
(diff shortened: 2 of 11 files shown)


doc/source/bindings.rst
```diff
@@ -114,6 +114,8 @@ Any other resources that you have added using other mods should be
 query-able this way, provided that you spell
 the term exactly as it appears in the resources window.
 
+You can also get a list of all resources, either in SHIP: or STAGE: with the :RESOURCES suffix. 
+
```
erendrake: this could use a better example now that i see it all lonely in the file

src/kOS/Suffixed/AggregateResourceValue.cs
```diff
@@ -44,9 +45,8 @@ public override string ToString()
             return string.Format("SHIPRESOURCE({0},{1},{2})", name, amount, capacity);
         }
 
-        public static ListValue PartsToList(IEnumerable<global::Part> parts, SharedObjects shared)
```
erendrake: this class had two methods that did almost exactly the same task but had different return types. this unifies them for DRY

src/kOS/Suffixed/SingleResourceValue.cs
```diff
@@ -19,11 +19,9 @@ private void InitializeSuffixes()
             AddSuffix("NAME", new Suffix<string>(() => partResource.resourceName));
             AddSuffix("AMOUNT", new Suffix<double>(() => partResource.amount));
             AddSuffix("DENSITY", new Suffix<double>(() => partResource.info.density));
-            AddSuffix("UNITCOST", new Suffix<double>(() => partResource.info.unitCost));
             AddSuffix("CAPACITY", new Suffix<double>(() => partResource.maxAmount));
-            AddSuffix("TWEAKABLE", new Suffix<bool>(() => partResource.isTweakable));
-            AddSuffix("FLOWMODE", new Suffix<string>(() => partResource.flowMode.ToString()));
-            AddSuffix("FLOWSTATE", new SetSuffix<bool>(() => partResource.flowState, value =>
+            AddSuffix("TOGGLEABLE", new Suffix<bool>(() => partResource.isTweakable));
+            AddSuffix("ENABLED", new SetSuffix<bool>(() => partResource.flowState, value =>
```
erendrake: after looking at this again i realized that i unitcost was an oddball because we never show any cost information in kOS

flowstate flowmode and tweakable were flow-through properties of the resource class and on a second look i dont like the way they would be presented.

src/kOS/Suffixed/WaypointValue.cs
```diff
@@ -8,14 +8,14 @@ namespace kOS.Suffixed
 {
```
erendrake: All of the changes to this file were done to avoid non private fields and fix a spelling mistake. i am sneaking it in here, because im bad.

Dunbaratu: This is preferable to your previous practice of turning things private that I meant to be public - that annoyed me to no end.  I left them open so it's possible to make further derivatives of this later for other kinds of waypoints if we want to expand this.

PR #481, opened 2015-01-04
Add AGX support

SirDiazo:
Add AG11 through AG250 get/set commands.
Does not support group names
Requires AGX Version 1.29 or newer
No clue on how documentation should update for this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,5 +9,3 @@ Addon release thread: http://forum.kerbalspaceprogram.com/threads/68089-0-23-kOS
 
 Addon development thread: http://forum.kerbalspaceprogram.com/threads/68096-kOS-Autopilot
 
-
-This fork for adding Action Group Extended support.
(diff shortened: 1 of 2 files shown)


SirDiazo:
Okay, I'm new to Git but I think I've done this correctly in how I merged the changes to my Develop branch to make them available to this PR.

The only outstanding issue is the "Script Trigger" action, but that does not affect anything in this PR.
Moving the conversation from https://github.com/SirDiazo/KOS/pull/1

Here is the entire part module that I then use a MM script to add to every part that has the kOSProcessor module. https://github.com/SirDiazo/AGExt/blob/master/AGExt/kOSVoidAction
It is an empty place holder action to trick AGX into using the action group so it can be monitored by kOS scripts, it doesn't actually do anything. As this only has a point when AGX is installed, I was thinking it should be part of the AGX install, not the kOS install.

Thoughts?

PR #497, opened 2015-01-07
Gimbal struct

### LINKED ISSUE #498: Gimbal info should be exposed.
because the gimbal information is not exposed in the gui. its hard to account for it in fly by wire systems.

erendrake:
fixes #498

just adds gimbals to the group of well known modules. first requested by @ferram4.

you should now be able to get a list of engines and then find the gimbal modules for each one.

### DIFF
--- a/doc/source/structures/vessels/gimbal.rst
+++ b/doc/source/structures/vessels/gimbal.rst
@@ -1,12 +1,12 @@
-.. _dockingport:
+.. _gimbal:
 
-DockingPort
-===========
+Gimbal
+======
 
-Some of the Parts returned by :ref:`LIST PARTS <list command>` will be of type :struct:`DockingPort`.
+Many engines in KSP have thrust vectoring gimbals and in ksp they are their own module
 
 
-.. structure:: DockingPort
+.. structure:: Gimbal
 
     .. list-table::
         :header-rows: 1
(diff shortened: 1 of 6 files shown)


erendrake:
Opps, i knew there was a reason i didnt put this up earlier. i forgot the docs :P dont merge just yet

erendrake:
Docs added, ready to pull.

doc/source/structures/vessels/gimbal.rst
```diff
@@ -0,0 +1,131 @@
...
+
+.. attribute:: DockingPort:TARGETABLE
+
+    :type: boolean
+    :access: Get only
+
+    True if this part can be picked with ``SET TARGET TO``.
+
+
+
```
Dunbaratu: The above file looks like the contents for dockingport.rst.  It doesn't talk about gimbals at all.

src/kOS/Suffixed/Part/GimbalValue.cs
```diff
@@ -0,0 +1,35 @@
...
+                else
+                {
+                    gimbal.FreeGimbal();
+                }
+            }, "Is the Gimbal free to travel?"));
+            AddSuffix("RANGE", new Suffix<float>(() => gimbal.gimbalRange ,"The Gimbal's Possible Range of movement"));
+            AddSuffix("RESPONSESPEED", new Suffix<float>(() => gimbal.gimbalResponseSpeed, "The Gimbal's Possible Rate of travel"));
+            AddSuffix("PITCHANGLE", new Suffix<float>(() =>  gimbal.gimbalLock ? 0 : gimbal.gimbalAnglePitch, "Current Gimbal Pitch"));
+            AddSuffix("YAWANGLE", new Suffix<float>(() =>  gimbal.gimbalLock ? 0 : gimbal.gimbalAngleYaw, "Current Gimbal Yaw" ));
+            AddSuffix("ROLLANGLE", new Suffix<float>(() => gimbal.gimbalLock ? 0 : gimbal.gimbalAngleRoll, "Current Gimbal Roll"));
```
Dunbaratu: Why are you telling the user the angles are zero when the gimbal is locked?  What if you unlock it, move it, and then re-lock it?  Just because it's locked *now* doesn't mean it was never moved in the past.

erendrake: when you lock the gimbal, KSP instantly sets the rotation back to the initial gimbal state on the next fixed update. 

as far as i know there is no way to 'lock' a gimbal to anything other to the inital state.

erendrake: Just to clarify, it seems that is what is going on. There is a lot of internal witchcraft going on that i cannot see.

Dunbaratu: Okay well that seems a bit ridiculous but even if it does work that way then won't that mean that gimbal.gimbalAnglePitch/Yaw/Roll would just return zero once the game snaps them back to center?  it seems to me that if the base game snaps then to center then it would be safer to "hang" off of that behavior rather than mimicking it ourselves, in case squad changes it later.

Dunbaratu: On another note, is there a way to return the gimbaled orientation as a kOS DIrection?  That might thematically fit the rest of kOS well.

Dunbaratu: Well, I'll trust you when you say the gimbal angles have to be manually zeroed when the gimbal is locked, but it seems weird to me.  I'll go ahead and merge this in.  I would like to see about the possibility of returning the orientation info as a *Direction*.  I think that would be nice as it would activate all the other nice things you can do with *Direction*s in the kOS API.

erendrake:
I agree about directions. There are some quaniterons (sp) in that module, I
should have exposed them I just didn't make the time 😛

PR #499, opened 2015-01-09
AGX Round 2

SirDiazo:
-Add AGX doc, with 2 pictures
-Add PartsInGroup and PartModulesInGroup methods to AGX bindings.

Only the external reflection call to AGX has been added, nothing done on making them visible to the player yet.


SirDiazo:
Just saw that apparently this PR has conflicts. Will leave it as is for now, let me know if there is something I need to change.

Issue #501, opened 2015-01-10
Compare Part


andrey-zakharov:
I've failed with simple case:
```
SET p to ship:rootpart.
print p.
print ship:rootpart.
print (p = ship:rootpart).
```
with 
> print ( p = SHIP:ROOTPART ).
>          ^
> [LOG 16:04:42.266] System.NullReferenceException: Object reference not set to an instance of an object

Ok. lets check it by UID:
```
print ( p:UID = SHIP:ROOTPART:UID ).
```

Wuah:
> print ( p:UID = SHIP:ROOTPART:UID ).
>               ^
> 
> [LOG 16:05:14.893] System.NotImplementedException: Can't operate types System.UInt32 and System.UInt32

erendrake:
Confirmed that this script does cause the problem. 

@Dunbaratu the error is in the function because the node that gets popped is apparently null? looked at the visit code and what gets pushed is an expr. 

actually the following fails as well. i am no longer thinking that this is a simple equality issue. 

    print body("kerbin") = body("kerbin").

    [LOG 07:23:06.351] System.NullReferenceException: Object reference not set to an instance of an object
  at kOS.Function.FunctionPrint.Execute (kOS.SharedObjects shared) [0x00000] in <filename unknown>:0 
  at kOS.Function.FunctionManager.CallFunction (System.String functionName) [0x00000] in <filename unknown>:0 
  at kOS.Execution.CPU.CallBuiltinFunction (System.String functionName) [0x00000] in <filename unknown>:0 
  at kOS.Safe.Compilation.OpcodeCall.Execute (ICpu cpu) [0x00000] in <filename unknown>:0 
  at kOS.Execution.CPU.ExecuteInstruction (kOS.Execution.ProgramContext context) [0x00000] in <filename unknown>:0 
  at kOS.Execution.CPU.ContinueExecution () [0x00000] in <filename unknown>:0 
  at kOS.Execution.CPU.Update (Double deltaTime) [0x00000] in <filename unknown>:0

erendrake:
and the fragment. 

    [Log]: Code Fragment
    File                 Line:Col IP   opcode operand
    ----                 ----:--- ---- --------------------- 
    interpreter history    18:5   0055 push null 
    interpreter history    18:5   0056 call run() 
    interpreter history    20:5   0057 push $eq 
    interpreter history    20:16  0058 push kerbin 
    interpreter history    20:16  0059 call body() 
[...]

PR #507, opened 2015-01-11
structure can now == and +

erendrake:
This is a possible solution to #501. More of a point of discussion. 

I believe we know enough about structures to give it the "==" operator. the + is not all that useful and i am not opposed to removing it.

### DIFF
--- a/src/kOS.Safe/Encapsulation/Structure.cs
+++ b/src/kOS.Safe/Encapsulation/Structure.cs
@@ -123,7 +123,19 @@ namespace kOS.Safe.Encapsulation
 
         public virtual object TryOperation(string op, object other, bool reverseOrder)
         {
-            return null;
+            if (op == "==")
+            {
+                return Equals(other);
+            }
+            if (op == "+")
+            {
+                return ToString() + other;
+            }
+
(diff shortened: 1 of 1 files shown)


Dunbaratu:
@erendrake, it looks like we were both working on this problem simultaneously.  Because you said you'd unassign yourself from #501, I had a go at fixing this myself last night and just made a pull request a few seconds ago this morning before I saw this one.

I think I like my approach a bit better because I did it as a deeper equals check and not just a reference equals check.  Structure subclasses are not forced to implement IEquatable, so Equals() doesn't seem reliable to me.  It looks to me like what you did, while it would make the exception go away, would still claim that ```body("kerbin") <> body("kerbin")``` because it's not looking inside the structures at their contents so two copies of the same thing still get treated as unequal.

erendrake:
sorry, didnt mean to double team it. i just has an idea this morning and because it is so little code i thought i would put it up to talk about.

this does work with 

    body("kerbin") = body("kerbin")

because bodytarget has overridden equals 

https://github.com/KSP-KOS/KOS/blob/develop/src/kOS/Suffixed/BodyTarget.cs#L117-L143

for many structures, reference is what we would want to use and this allows us to be DRY in those cases.

PR #508, opened 2015-01-11
fixes #501 and fixes #459

### LINKED ISSUE #459: object equality fails on wrapper objects.
```
set aaa to body("Duna").
set bbb to body("Duna").
if bob <> aaa { print "they are unequal". } else { print "they are equal". }.
```
These tests seem broken.  I think the problem is that when we construct a wrapper around a KSP object, we're not implementing an equals checker method, so two wrappers around the same native object still report as unequal.  We should *mandate* that all kOS objects implement an equals method, and then go back and populate them properly:

i.e. two PartValue's wrapped around the same inner Part object should be reported as equal.  Two VesselTarget's wrapped around the same inner vessel object should be reported as equal, etc.

### LINKED ISSUE #501: Compare Part

Dunbaratu:
Both problems were caused by how equality comparison worked, or didn't,
between structures.

There's a lot of files changed, but only because each individual file
had a small change in it to add one short KOSEquals method.

### DIFF
--- a/src/kOS.Safe.Test/StructureTest.cs
+++ b/src/kOS.Safe.Test/StructureTest.cs
@@ -31,6 +31,11 @@ namespace kOS.Safe.Test
             {
                 AddSuffix(name,suffix);
             }
+            
+            public override bool KOSEquals(object other)
+            {
+                return false; // dummy because it's required for all structures.
+            } 
         }
 
         [TestFixtureSetUp]
--- a/src/kOS.Safe/Compilation/Calculator.cs
+++ b/src/kOS.Safe/Compilation/Calculator.cs
@@ -33,6 +33,8 @@ namespace kOS.Safe.Compilation
             if (argument2 is float) argument2 = Convert.ToDouble(argument2);
 
             if (argument1 is int) intCount++;
+            if (argument1 is Int32) intCount++;
+            if (argument1 is Int64) intCount++;
             if (argument1 is double) doubleCount++;
             if (argument1 is string) stringCount++;
             if (argument1 is ISuffixed) specialCount++;
@@ -50,7 +52,7 @@ namespace kOS.Safe.Compilation
             if (boolCount > 0) return new CalculatorBool();
             if (specialCount > 0) return new CalculatorStructure();
 
-            throw new NotImplementedException(string.Format("Can't operate types {0} and {1}", argument1.GetType(), argument2.GetType()));
+            throw new KOSBinaryOperandTypeException(argument1,"perform opeations between", "and", argument2);
         }
     }
 
--- a/src/kOS.Safe/Encapsulation/ConstantValue.cs
+++ b/src/kOS.Safe/Encapsulation/ConstantValue.cs
@@ -1,5 +1,6 @@
 using System;
 using kOS.Safe.Encapsulation.Suffixes;
+using kOS.Safe.Exceptions;
 
 namespace kOS.Safe.Encapsulation
 {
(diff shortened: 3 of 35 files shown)


Dunbaratu:
I think that reference equality is utterly useless in a language like kerboscript where we haven't enabled the ability for the users to actually look at the references themselves, nor the ability to check for nulls, and where they aren't directly in charge of (or even aware of since the logic is hidden away inside the C# code) whether something was constructed as a new cloned copy of something or was a reference to the original thing.

That's why I'm reluctant to design the system such that we can unwittingly accidentally use the underlying reference-equals for a Structure as the sort of equality the users see with the equals operator, without having taken a moment to consciously think about whether it's the right thing to do or not and made the deliberate *conscious* decision to expose it in that one structure's case.  

I'd much rather have  the default behavior for the equals operator in kerboscript be "exception! kerboscript doesn't implement equality checking for of this sort of object", than have the default behavior be "we will silently allow you to use a type of equals that doesn't really work for what you're trying to do, and only after lots of experimentation will you find out that it's useless", which is for the most part what a reference equals will be to the kerboscript programmer.

This is why I was arguing against making reference-equals be the default inherited behavior.  I think most of the time if we can't think of a meaningful override for equality, what that really means is that the right answer is "equality checking doesn't work in this case" rather than the right answer being "lets pretend by default that reference equality is the right answer".  There may be a few cases where exposing reference equality might be the right answer, but I think they'll be the exception, not the rule, and thus should require deliberate thought before using that technique (deliberate thought rather than inherited behavior).

For example, I'd be fine with allowing the inheriting of the kerboscript equality behavior (instead of making it an abstract method requiring an override), if and only if what it was inheriting from was something we invent in the Structure base class and NOT something we inherit from C# 's underlying behavior itself.  My fear about hanging the script's meaning of equals onto the underlying C#'s meaning of equals is that it means it would probably break our OWN C# code if we were to try to change what equality means to the kerboscript code.  Imagine if we want to make it impossible for a script to perform the equality check on structures we know it doesn't make sense for, by throwing an exception in the equals checker for that structure.  If we did that inside an Equals() method that C# itself, and all the .NET libraries make use of, it would screw up any attempt to stick that structure into a collection of some sort.  But if we had our own KOSEquals(), then we *could* throw an exception in it and not break how we use it in C# itself.

I guess that's really the reason for my wanting to make a separate KOSEquals method in the first place.  So that we can make it possible to have a rule for what equality means to a KOS script that doesn't also end up altering what equality means to the C# code under the hood.

I guess I got really distracted into the argument about inherited versus explicit behavior when all along the REAL problem I was having was not that it was inherited, but that it's inheriting from the *wrong place*.  It's inheriting from the underling C#'s meaning rather than from some default kerboscript meaning that I'd like to make it possible to differ from the default meaning.

Ideally, because reference equals is useless to a kerboscript program most of the time, I'd like the default KOSEquals behavior to be to throw an exception warning the user that equality checking was never implemented for this kind of Structure.  To do that requires making it a different method from C#'s Equals() check, as that needs to work cleanly and default to a reference equals check for everything in the .NET collections libraries to work right.

Dunbaratu:
Should this be closed and unmerged, since the other PR dealt with it differently?

erendrake:
I had left it open because i wanted to go farther with getting rid of reference types. for now i guess we should close it

Issue #510, opened 2015-01-15
Can't access more than one PartModule of the same name


jwvanderbeck:
On parts that have multiple PartModule entries with the same name, there appears to be no way to access a specific one.

Two solutions. 
1. Provide a LIST of actual PARTMODULEs rather than just STRINGs
2. Allow an index to be passed with the name in GETMODULE.

The former is better , the latter probably the easiest and more backwards compatible.

Dunbaratu:
Ugh .  It's ridiculous that you're allowed to have two partmodules with the same name in the first place.

Between the two solutions I think I'd prefer 2.  Given how stuff works behind the scenes, #1 is a bit of an expensive operation , which is why it wasn't done that way in the first place.

jwvanderbeck:
The PR I submitted does just that.  I added a GETMODULEINDEX method.  It is actually quite nice to use, especially when used interactively as you can print a list, which nicely shows all the indices.

erendrake:
Thank you @jwvanderbeck for the PR :)

PR #511, opened 2015-01-15
Add new PART:GETMODULEINDEX function

jwvanderbeck:
Adds a new function to get a PartModule at a specified index.  Allows
accessing PartModules that share the same name.

Address issue #510


Dunbaratu:
Would the name GETMODULE*BY**INDEX make more sense?  "Get module index" sounds like it will return a number, 0, or 1, or 2, etc - the index of the modules.

jwvanderbeck:
I Can. kOS style seems to more terse than verbose so I kept it as short as I could reasonably make it.  I have no problem making it GETMODULEBYINDEX though if that is preferred.

jwvanderbeck:
Heh I was **just** about to sit down and change this function.  Sorry I didn't get to it sooner.  Been busy with other things.

Issue #509, opened 2015-01-13
Support all 6 aiming points.


Dunbaratu:
I propose we add these directions

* ```SHIP:NORMAL```
* ```SHIP:ANTINORMAL```
* ```SHIP:RADIALOUT``` which is really just an alias for ```SHIP:UP```
* ```SHIP:RADIALIN``` aliased by ```SHIP:DOWN```

Possibly we should add SRF versions of them as well (as they will rotate slightly when converted into the surface-relative frame if you're not traveling in an equatorial path.).

These would all be very easy one-liner calculations, and in fact are already easy calculations in kerboscript but there's no real reason to make things more obfuscated than they need to be by hiding them and forcing people to get them from cross-products.

### Rationale:

I was thinking when KSP 0.90 came out about the new aiming points on the navball - (anti)normal, radial (in/out) - and how you need a good updated probe core before you can access all of them (the early cores can't, just like a low-experience meat pilot can't).

It seemed the right thing to do is just add these to kOS, but In trying to make kOS adhere to the theming of KSP 0.90 I was thinking it wouldn't be right to always allow scripts to access these points if the probe core can't do them.

Here's why I waffled on this: If you disallow the NAMED points, you're not really preventing the kOS script from using them - just demanding that the script do an additional math operation.  Once you have ```SHIP:PROGRADE``` and ```SHIP:UP``` then the remaining other 4 directions are calculable anyway in script code:  Normal is the cross product of the two.  Antinormal is (-1)*normal. radial out is just another word for SHIP:UP and radial in is therefore just (-1)*SHIP:UP.

And even if you deny use of ```SHIP:UP```, because that's really radial out, you can still calculate it as the opposite direction of BODY:POSITION.  If you deny use of ```SHIP:PROGRADE``` you can still calculate it as the direction of ```ship:velocity:orbit`` which even if you deny that you can still calculate it from watching your delta position over delta time.  (or, more accurately from the BODY's delta position because we coerce all positions to be origined at the ship.)

The point is that the only way to prevent kos scripts from getting all 6 points before it's really ready in the tech tree would be to deny them really basic info about everything and rendering the script useless.

Therefore we may as well just say "screw it" and accept that this is one way in which kOS breaks the assumptions about KSP 0.90 and actually makes a thing easier than KSP does.

erendrake:
I agree completely, they are already 'available' to scripters so making it more complicated is only accomplishing making it more complicated :)

dewiniaid:
[...]

PR #512, opened 2015-01-16
Enable built in autopilot

hvacengi:
This pull request implements using the built in SAS autopilot introduced in KSP version 0.90.  It uses the existing toggleflybywire function.  To toggle a feature on or off, pass the name and the bool state to the function (i.e. "toggleflybywire("prograde", True).").

In order for the existing function to work properly, some small modifications were required on the compiler (to properly allow a return value to be popped).  The actual implementation itself resides in the FlightControlManager class.  I tried to make the operation resemble the existing implementation for steering operations as much as possible.  Since the base KSP autopilot is handling the actual steering, there is no implementation of a callback function in the FlightTrlParam class.  When enabling a feature the implementation does disable a locked steering command, or any of the other built in autopilot features, and the "unlock steering." command will disable any of the built in autopilot settings.

Also implemented is enabling/disabling of SAS damping.  I cannot tell if this is actually doing anything, honestly, but further testing may show it provides some use.

In the steering helper, the "lock steering" function will now use the vessel's SAS autopilot to lock to the heading if SAS is turned on, rather than the adapted mechjeb implementation that is traditional.  In my experience, there are some times where the original implementation is better, other instances where the autopilot implementation is better, and still more instances where neither is robust enough.  I like this option as it gives users two paths to engage the autopilot without requiring a full PID implementation and without making the autopilot too good.  

I did not want to get into re-writing anything language wise, so these features all fit into the existing code structure.  They can of course be moved to another structure, if it makes it would fit better with the overall structure of the language or if it would simply be a better implementation.  I will leave these modifications to those who maintain the kos language base however.  This is not based on the most recent version of the development branch, but it does not conflict with any of the changes made while I worked on the code.  I did not want to run into a conflict with one of the other features being implemented causing a bug.


hvacengi:
It looks like I may have spoken too soon for "unlock steering." clearing the new autopilot settings.  It doesn't look like ToggleFlyByWire is being called if the steering is not locked to a direction.  All other functionality works fine, so the code could still be merged so long as the documentation doesn't say that "unlock steering" works on these functions.  The alternative is that you have to know what the last setting was in order to disable it.

Dunbaratu:
Using the SAS instead of having to override it and turn it off would be a welcome addition.

hvacengi:
`lock steering to prograde` will behave normally, until you call `toggleflybywire("antinormal",True)` at which point it will disable the lock steering command as part of the logic to enable antinormal.  SAS is automatically enabled when toggleflybywire enables a built in autopilot mode, so the SAS ON command will do nothing.  Following it with a `toggleflybywire("antinormal", False)` command will reset the SAS to stability mode.  It might be a good idea to include "stability" as a mode.

The code that I added to the steeringhelper class is what enables steering to arbitrary directions using the built in sas.  After the target rotation is determined, there is a check to see if SAS is enabled.  If so, it uses the AutoPilot.SAS.LockHeading method to steer the ship to whatever arbitrary direction you choose, and then it returns before the normal pitch/yaw/roll is calculated and added to the control.

I would actually much prefer the same method of enabling these methods as you describe.  I think it's cleaner and matches the syntax from the rest of kOS better.  But I wasn't sure how to implement the script language portion of the code.  If you can figure out how to implement the language part, I can adapt the implementation to use FlightCtrlParam so that "steering" would either update the direction using SteeringHelper or toggle the built in sas mode only.

What about using something like:
```
lock sas to sascontrol:prograde.
sas on.
```
This would allow keeping sas independent from steering, and give a natural break between defined directions and sascontrol.  That prevents a conflict with your sugestion from #509 where we would end up having directions that match the sascontrol modes.  I'd rather support both the built in sas and the existing implementation of steering, since they both perform better in some situations.

hvacengi:
Also, I added a commit that includes checking to see if the pilot/probe can use the sas mode and throws an error if not.

hvacengi:
ADHD strikes again: I forgot to say that I still can't figure out how to make this change the ui icon colors for the different modes.  I'm going to keep looking into it, but I didn't want that to hold up adding the feature.

SirDiazo:
i do the icons for my Actions Everywhere mod.

Here's the action group the player can call to set the command:
https://github.com/SirDiazo/ActsEverywhere/blob/master/ActsEW/ActsEW/Stock.cs#L159

and here's the actual code to set them:
https://github.com/SirDiazo/ActsEverywhere/blob/master/ActsEW/ActsEW/Stock.cs#L243

hvacengi:
Thanks @SirDiazo! I don't know how you managed to find that code, but I'm glad you did.  I've added it to the existing code so that the icons now work.  I already modified the flow to throw a descriptive error when you try to set the SAS to a mode that is unavailable (though I may want to add a note that a empty target may also be the cause).

I think that I may have come up with an idea on how to implement this as a separate lock statement, but I may need a little help with setting up the constants (I honestly haven't looked at that part yet).  I'll take a swing at it this weekend and see if I can get it implemented well enough.  It kind of feels wrong to rip out everything else I've already got in the pull request, but I really like the idea of `lock sas to sascontrol:prograde.` being the applicable command.

erendrake:
@hvacengi I want to have a small release today and with the conversation not quite resolved i will leave this unmerged for now. If you feel differently let us know.

hvacengi:
I hope you were able to make your release yesterday.  I was out of town for the whole day, and I didn't get a chance to really check email until I found a break here at work.  I'm fine with leaving this out for now, I think we can really add some polish to it before releasing.  I've got my path mostly mapped out and I've started to revise the implementation.  There should be an updated commit later tonight (EST) with moving the implementation out of the toggleflybywire function, adding a new selectsasmode function, and implementing the new selection system.  After that's all done, I'm going to work on replicating the implementation for `lock steering to [value]` into `lock sas to [value]`.  I'm not sure how long that will take me to implement, but it should be done by the end of the week for sure, possibly tonight or tomorrow if I have no snags.

hvacengi:
I found a small amount of time at lunch to work on the implementation.  The autopilot code is now in it's own method an kOS function `selectautopilotmode([mode string]).` which appears to have successfully decoupled it from toggleflybywire.  Calling the function directly from the script seems to work fine.

I can't seem to make up my mind on the syntax we should use.  Instead of making the decision myself and forcing all of you to live with it, I wanted to get your input.  I've narrowed it down to these options, with [value] being either a string (ie "prograde") or a suffixed constant (ie sascontrol:prograde):
```
lock sas to [value].
lock autopilot to [value].
set autopilot to [value].
```
I get the impression that @Dunbaratu maintains much of the syntax side of the mod, maybe he has a recommendation.  Right now I'm leaning towards using set instead of lock.

I also ran into a bug with RemoteTech, maybe @erendrake can help with that one.  If there is no connection available to the probe, the built in autopilot will not function.  I did test this after fixing the reference to RemoteTech.API.API.  I'm not sure of the inner workings of RemoteTech, so I don't really know how to fix this.  Maybe it needs to be a feature/bug fix in RemoteTech itself, I'm not sure.

hvacengi:
Last night I committed changes that added a setter and getter for "autopilot" in the FlightControlManager.  The implementation now supports the following forms of input:
```
selectautopilotmode("prograde").
set autopilot to "prograde".
set autopilot to "". //will reset to stability without forcing sas on
```
Please comment if you think the syntax needs to be changed (particularly if you have a way to cut the character count).

There are only two issues remaining prior to release:
1. RemoteTech still prevents the built in autopilot control.  The more I look at it, the more I think this is a bug with RT an would be better handled in that project.  I'm not sure how RT blocks control, so I'm not really sure how to fix it.
2. The command 'unset autopilot.` does not work.  This is because the setter is a BoundVariable which the unset command ignores.  I think the best workaround would be to add a defaulted false parameter when adding a setter that would mark a BoundVariable as IsNullable, and then let the unset command set Nullable bound variables to Null without removing the variable.

Dunbaratu:
We should really decide how this goes and merge it or something.  I've stayed mostly out of it since autopilot code has been more @erendrake's thing.  It does really feel weird to me though, syntax wise.

Could this work better?
```
   set sasmode to "prograde". // or "retrograde" or "antinormal", etc
   set sas on.  set sas off. // The same commands as before.
```
It just seems like this is more in-theme to how stuff works now.  Right now using LOCK STEERING or using raw SHIP:CONTROL tend to both fight hard with SAS and tend to be mutually exclusive with them.  So the "feel" that SAS is something *different* to the other modes of kOS seems to already be present.  Therefore just make use of THAT instead of calling this SAS work "autopilot" which is a rather ambiguous term - everything kOS does is autopilot stuff.

This would dovetail into what we already have:
Want to use LOCK STEERING or SHIP:CONTROL for raw steering?  Then leave the SAS off.
Want to use SAS? Turn SAS on.
Want to use SAS in one of the newfangled ways that KSP 0.90 made possible, instead of its old stability-only way?  Then set the mode of SAS with the bound variable SASMODE.  Changing this doesn't CAUSE SAS to turn on, but just changes what SAS does when turned on.

I think the above design would dovetail best into what's there now, and would fit what the users see when flying manually.  ("I've clicked the antinormal sas mode button.  Now when I toggle SAS with the 'T' key, it will end up trying to point antinormal.")

hvacengi:
I think all of those points make a lot of sense.  The only reason I used the term autopilot was the internal class naming (technically we toggle using the autopilot member not the SAS component) but since the user never sees the internal code, that could cause a major logical break.  

I'm renaming the setter, and removing the code that enables SAS when you set the mode.  I will verify that there is no error if the mode is set while SAS is turned off.  Should this still force disabling a "steering" flight control parameter?  That is to say should this code result in disabling steering control?
```
lock steering to mydir.
set sasmode to "prograde".
sas on.
```

Dunbaratu:
(your reply ninja'd a large major edit I did to mine.  I had a lot more to say afterward, if you want to go have a look again.)

Dunbaratu:
(Better idea for suggestion #2.  Instead of a field on Direction to track how it got constructed, we just subclass Direction 6 different ways.  Prograde is a subclass of Direction, Antinormal is a subclass of Direction, RadialIn is a subclass of Direction, and so on. They all behave *exactly* the same as Direction and have no additional code changes in them whatsoever, other than the fact that doing this gives us the ability to then have C# code that does stuff like ```if (dir is Prograde) { stuff.... }.```.  Everywhere else the code deals with a Direction it doesn't care that Prograde is a subclass - it just treats it as a normal Direction.  But in the SASMODE code it does care, so it checks just the class derivation to see if it got constructed as a raw Direction or as a Prograde subclass direction,

hvacengi:
I don't know why they don't automatically show edits on these pages since they do show new posts...

The idea of subclassing direction makes a lot more sense to me than trying to modify the direction class with a flag to determine if it matches.  It gets a little more messy with target and stability.  I like simple, so I don't want to force the addition of classes and bound variables just for the sake of this little addition.  If you add the bound variables anyways, it would pretty simple and low overhead to add the subclasses.  I'll have to do a quick test, but I think I modified the code to be a little more flexible which will let us use strings for now add support for new class types later.  While I prefer to keep it simple, I also subscribe to the Gonzo theory ("Why not take them both?"), so I really like the idea of supporting both routes.  

How does a lock work?  Does it automatically try to set the variable on every update cycle?  Or does the lock mechanism itself detect the change in value?  If it forces a set on every update, I would discourage using lock for the sas unless the user needs to prevent it from being changed by using the UI.  Otherwise, a lock just adds overhead to the processing load.  Either way, it will be good that both set and lock will work, since that would be the user's expectation.

Dunbaratu:
I can work on making a PR that has the full 6 directions (It could even have a TARGET one, even thought that should just be the same as othership:position:direction, it would still be useful for the sake of having these named subclasses of direction.)

That's not too hard and it's a thing I wanted to do anyway.  Then you could build off of that.  The only ugly thing is that in *reality*, when you say PROGRADE that's really just an alias for SHIP:PROGRADE, and so I'd LIKE to be able to say SET SASMODE TO SHIP:PROGRADE but we'd have to be careful to not allow SET SASMODE TO SOMEOTHERSHIP:PROGRADE, as that won't work in the underlying KSP API.  I'll work it out.  It's okay.  I should have some time this evening to have a look at it (in several hours from now).

As for how locks work, there's really 2 different kinds, but the mod tries abstracting this away from the user.

In both kinds of LOCK, in reality the expression the compiler builds on the righthand side is built as an actual function call in the underlying opcodes.

For example, if you do this:
```SET X to y + 1.```
Then you end up with a set of opcodes that might look like so:
```
address | pseudo-machine-language
-------   --------
   0123   push x
   0124   push y
   0125   push 1
   0126   add
   0127   store
```
(The 'add' adds the top two things on the stack and replaces them with the answer (the top two things being the y and the 1).  Then the 'store' pops the next two things from the stack - taking the topmost item as the value (the result of the add operation), and the item just underneath it being the name of the variable to store it into. (the x))

When you replace the SET with LOCK, for this:
```LOCK X TO y + 1.```
Then it builds the very same opcodes, but places them up at the top of the program chunk, remembers their location, and makes them a subroutine call, ending up with something more like this:
```
address | pseudo-machine-language
-------   --------
   0054   push y
   0055   push 1
   0056   add
   0057   return // top thing on stack now when returning is now the result of y + 1.
```
Then from then on any time someone tries to refer to the variable 'x', the reference is actually replaced with a call to this routine stored at location 0054.
So for example, this:
```set z to 10 * x + 2.```
will actually look something like this then, because 'x' is now itself a hidden function call:
```
address | pseudo-machine-language
-------   --------
   0311   push z
   0312   push 10
   0313   call 0054 // 0054 is where the 'x' function is stored in this example (see above)
   0314   mult
   0315   push 2
   0316   add
   0317   store
```
So the idea is that when you LOCK a variable, you don't REALLY ever actually store the variable's result in a variable.  Instead every time you try to read the variable, you call a subroutine that will calculate the expression live on the fly, and leave the resulting value atop the stack.

(It's a lot like a 'get' property in C#, really.  It may LOOK like ```theList.Count``` is just a dumb passive field, but it's really resulting in a small function call every time you mention it.)

Now as to the two different types of LOCKs in kOS, well, both work the way I just described, but differ as follows:

- Most locks are entirely passive - they NEVER execute the function code until you try reading the "variable"'s value elsewhere in some other expression.  This is nice and efficient because when nobody is looking at the variable, you're not paying the expense of recalculating it.
- But some locks' function code do actually get re-run each and every Update() - those are the ones that deal directly with flying:  STEERING, and THROTTLE for example.  This is because, effectively, kOS *itself* is reading the value again and again.  For these types of locks, the expression code does get run over and over and over every update so it needs to be efficient or it bogs you down.

**Comment by hvacengi** [comment]
Thanks for explaining the lock logic, that makes a lot of sense.  So if I read it right, not every instance of bound variable will be set on every update.  Which means the lock will really only be referenced when you read the value out of the sasmode variable.   So locking sasmode will not prevent user input from overriding the value.  Would it override reading the value of sasmode?  Right now I have the getter setup to return the ToString value for the current autopilot setting.  I don't think that we want a lock to return the value that was set initially instead of the current value, if that would even work.

I'll be working on the code this evening myself.  I have to try and finish doing some real work that pays the bills.

**Comment by Dunbaratu** [comment]
The more I think about it, the more it seems as if having a new class for *every* built-in direction is probably silly, given that the whole point is to make them identical to the base class.  What may make more sense is to just have a field or property flagging how it was made.

In every way it is a normal direction.  It just remembers whether it got made with the keyword "antinormal" or "prograde", etc. or "other" ("other" meaning it's user-defined and not from a built-in).

What your code would do, is utterly ignore the actual direction contents (the quaternion or euler angles), and instead just query whether or not the Direction got made from a special keyword, and if so, which one.

In other words, where you do something like this:
```
switch (modeName) // modeName being a string
{
  case "prograde" : stuff; break;
  case "retrograde" : stuff; break;
  case "antinormal" : stuff; break;
  etc...
}
```
You would be able to replace that with this:
```
switch (sasDirection.Keyword) // sasDirection is a Direction, and .Keyword is the new thing I'd be adding to it.
{
  case PROGRADE : stuff; break;
  case RETROGRADE : stuff; break;
  case ANTINORMAL : stuff; break;
  .. etc...
  default:
      throw new exception here about how you can't use sasmode with arbitrary homemade directions, only named ones;
}
```
Thus when you say:
``` set sasmode to ship:antinormal.```
The direction created by ship:antinomal will have its .Keyword set to ANTINORMAL so you know that's what was meant.

You don't acually *care* what the real value of the direction is, or whether it has become stale over time and the direction has moved, because all you're really doing is using the fact that it was coming from the 'antinormal' keyword to decide what to do.

What this would allow, I hope, is for people to use the SAME thing for both SAS usage and non-SAS usage (like LOCK STEERING).

It would make it so that LOCK STEERING TO SHIP:ANTINORMAL would work *and* SET SASMODE TO SHIP:ANTINORMAL would work too.  The secret hidden from the user would be the fact that LOCK STEERING is ignoring the fact that the direction's keyword is ANTINORMAL and is just looking at the resulting rotation angle the constructor made (just like it does now), while SET SASMODE would be doing the exact opposite, and ignoring the actual rotation angle and just looking at the keyword it was made with.

**Comment by Dunbaratu** [comment]
My head is getting a bit fuzzy as the evening goes on (I got up early this morning), so I think it's safest for me to hold off and not make my PR for my side of this until tomorrow.  It's an important enough change that I should approach it with a clear logical head, not a groggy sleep-deprived one.

**Comment by hvacengi** [comment]
I finally got around to looking at this edit again today (stupid space station taking up whatever time I don't spend on real life).  I've pushed a commit with my latest code.  It works with the strings just fine and is ready for supporting directions (or any other type for that matter) when they get updated with some form of keyword.  I'm going to shift my focus on to some bugs I've noticed (I'll be submitting reports tonight after I did a little deeper) and then a feature I want to implement (hard drives, again a report will be supported tonight).  I'll keep this as an independent branch so that I can easily jump back in and make more modifications if you see fit.

As far as documenting functionality, I would encourage the following caveats:
1) Your mileage may vary.  Some times SAS control is much more stable than the built in steering, and other times the opposite is true.  This seems closely related to the relationship between the vessel mass and the amount of control authority.  Try different craft designs to get a good feel for how the two options compare.
2) RemoteTech is not currently compatible with the built in SAS modes while out of signal.  The control may still function after signal is lost, but time warp and reloading the vessel will stop the functionality.

I haven't looked at the documentation files to see what it would take, but I would be happy to work on a documentation entry if you need.

Point 1 is exactly the reason that I like the direction implementation that @Dunbaratu suggested.  It would be very fluid to switch between steering and sasmode with those directions.  We might want to give a little more thought to it though.  Since the sasmode uses the selected navball mode to determine the reference frame, it could lead to confusion if `ship:prograde` points at the orbital velocity while 500m off the ground with the navball showing surface velocity.

**Comment by Dunbaratu** [comment]
I got wrapped up in a big kOS edit project right after I said that message last time.  Sorry I never got around to doing what I said I'd do "tomorrow" 15 days ago.   My other big project is coming to a finish so I might be able to get back to it soon (I won't make the mistake of specifying a date this time though).

**Comment by hvacengi** [comment]
@Dunbaratu , please don't take that as me trying to call you out or any thing.  I didn't even open visual studio until that day, so I had done nothing myself.  I just wanted to let you all know that I had it at a point where it is functioning and that I will add direction support as soon as it is ready.  When I found out I couldn't make it work easily with remote tech, it fell off of my radar as a major feature (I find kOS at it's strongest when coupled with Remote Tech).

Whenever you get the directions figured out, let me know.  If you and @erendrake want to wait for that support to put it in the core, that's fine.  But this way you can merge the request in if you want temporary support, or wait if you don't want to have the implementation change.

**Comment by Dunbaratu** [comment]
Sorry for being slow to respond to this.  We'll probably have to push it out to a version after this one, just because there's too many things already happening in the next release as it is and we want to get it out.

**Comment by hvacengi** [comment]
No worries in the timing.  When we reach the point where you're ready to add it, we can revisit.

File: src/kOS/Binding/FlightControlManager.cs [hunk 1]
```diff
@@ -156,6 +160,88 @@ private bool VesselIsValid(Vessel vessel)
             return vessel != null && vessel.rootPart != null;
         }
 
+        public void SelectAutopilotMode(object autopilotMode)
+        {
+            if (autopilotMode is Direction)
+            {
+                //TODO: implment use of direction subclasses.
```
erendrake: is this where our current steering helper algorithm would get replaced?

src/kOS/Binding/FlightControlManager.cs
```diff
@@ -156,6 +160,88 @@ private bool VesselIsValid(Vessel vessel)
...
+                        string.Format("Cannot set autopilot value, pilot/probe does not support {0}, or there is no node/target", autopilotMode));
+                }
+                currentVessel.Autopilot.SetMode(autopilotMode);
+                //currentVessel.Autopilot.Enable();
+                // change the autopilot indicator
+                ((kOS.Module.kOSProcessor)Shared.Processor).SetSASUI((int)autopilotMode);
+                if (RemoteTechHook.IsAvailable(currentVessel.id))
+                {
+                    Debug.Log(string.Format("kOS: Adding RemoteTechPilot: autopilot For : " + currentVessel.id));
+                    // TODO: figure out how to make RemoteTech allow the built in autopilot control.  This may require modification to RemoteTech itself.
```
erendrake: I have added an issue in RT for this as it will require an API addition.

erendrake:
i started off being unhappy with anyone making the non-raw controls so much better that no one would want to use the raw stuff. After having thought a lot more about it I think that now that it is part of the stock game we should leverage it and make simple steering just that.

My only reservations now would be around documentation of this new feature.

src/kOS/Binding/FlightControlManager.cs
hvacengi: It is where we will implement a parallel route to the current' steering.  @Dunbaratu and I seem to agree that we don't want to choose an exclusive control option.

hvacengi:
@erendrake I actually get and understand where you're coming from on the use of autopilots.  But like you said, it's exposing existing functionality.  I would be happy to provide documentation for the feature, if some one can point me to a decent formatting how-to.

src/kOS/Binding/FlightControlManager.cs
hvacengi: After looking at the code, I'm not even sure it will be possible to fix this in RemoteTech.  Not without known buy-in on all other autopilot enabled mods.  But I only gave it about 2 hours of research.

erendrake: @hvacengi i was going to add an '''AddSanctionedAutopilotModeChange(uid, AutopilotMode)''' method. that way remotetech can change it and know its ok :)

erendrake:
@hvacengi i realize now that i came across as really grumpy in that post. forgive me, i really appreciate you doing the work on this and i hope you stick around and keep helping us make this a better and better mod :)

erendrake:
I merged this against the rules on documentation, ill put my quarter in the jar :) I thought it would be nice to get this in and have people try it out and we can add docs before we release.

hvacengi:
Oh, it didn't come off grumpy. It came off reluctantly accepting. I just
wanted to reinforce you're decision. And I have thick skin anyways ;-)

I'll take a peek at the source of other documentation and see what I can
piece together. It will be a little cleaner if the direction support is
enabled before the next version, since we won't have to explain strings and
then double back and explain directions. But I think it works logically
enough either way.

erendrake:
One thing to remember is that we can always release and just not mention it
till we are ready. So long as we dont break anyone and just add 'hidden'
features its ok :)

PR #516, opened 2015-01-19
specify that the kosnametag is for kos

erendrake:
right now when we add the kosnametag module, we do it in the 'legacy' pass. this is the pass that you fall in if you do not specify what mod the change is for.


erendrake:
@Dunbaratu this is the change i mentioned, its a super low priority

Issue #514, opened 2015-01-18
Make Terminal Resizable


Dunbaratu:
In my work toward trying to implement issue #157, I realized that you can't really do a remote terminal properly if the system presumes all terminals are hardcoded to a known number of rows and columns.  The very nature of a remote terminal is that you can't enforce the terminal's size to be what you want.  It's whatever size the user wants it to be.

Therefore the issue to make the *existing* GUI (not remote) terminal resizable needs to be addressed first.  Until that infrastructure is there, the remote terminal can't really work.

That means addressing all the the following:

* My intent is to expand on the TERMINAL Suffixed class first envisioned in #457, with:
  * TERMINAL:COLUMNS (get/set - setting makes the window change size).
  * TERMINAL:ROWS (get/set - setting makes the window change size).
  * For THIS issue, we only need TERMINAL:COLUMNS and TERMINAL:ROWS.  The other functions envisioned by #457 can be added in as well later (an OPEN/CLOSE feature), as well as maybe moving CLEARSCREEN to TERMINAL: and keeping the old CLEARSCREEN as an alias to that (like how ALTITUDE is an alias for SHIP:ALTITUDE), and adding other stuff  like maybe the ability to set the terminal's title string.  But for now these will not be part of this issue.  They can be done by a future issue.  The scope of this issue will just be to support rows and columns.
* I may support stretching by mousedragging too, but for a first pass it is sufficient to just set it by commands typed in, to prove the infrastructure is working.
* There are really two kinds of resizing - do we want to support both of them?  When the terminal gets bigger or smaller, you can:
  * Change the number of rows and columns, keeping the size of an individual character cell the same.
  * Change the size of the font, keeping the number of rows and columns the same.  (the font is already not mapped pixel-perfect 1:1 - the font image cells are stretched to fit the current terminal size.  So you can easily change the stretch amount to accommodate the bigger/smaller window.)
  * The above things are why I won't support mousedragging for a first pass of this.  The question of how to support both of the above, and what the meaning of a mousedrag to stretch the window should be, needs some more careful thought.

Dunbaratu:
This should be associated with pull request #518 and closed when #518 is merged.

For whatever reason, github doesn't like the keyword "implements" either, which I thought it used to understand.

PR #518, opened 2015-01-21
Implements #514.

Dunbaratu:
This is the work to make a GUI terminal resizable.
There's a lot of messy changes so I'll be annotating the PR with github comments explaining why some changes were made (the comments aren't in the mainline code as comments because they only refer to the reason for the change, rather than being something that would make sense to someone in the future reading it post-change).


Dunbaratu:
Commentary on why image files changed:
(I'm putting these here because I can't make a github "line comment" on an image file like I can on source code changes)

**font_sml.png**: This was changed because I wanted to finally get around to making the curly brace look distinguishable from the parentheses.  The original font made them look way too similar.  I added a bit more wiggle and curl to the curly brace.  Other than that there's no changes.  This change is not strictly necessary for the PR.

**monitor_minimal.png**: I cropped the image to remove the bottom blank grey area.  This was necessary because the image is now being scaled to fit the terminal instead of being hardcoded to be a 1:1 pixel match to the PNG file, and when the image included the bottom border (that was in the picture originally) it meant that large windows had a fatter bottom border than small ones.

src/kOS.Safe/Encapsulation/TerminalStruct.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
+using kOS.Safe.Encapsulation.Suffixes;
+using kOS.Safe;
+
+namespace kOS.Safe.Encapsulation
+{
+    public class TerminalStruct : Structure
```
Dunbaratu: This entire file is new - it's purpose is to allow scripts to say things like:
```print TERMINAL:WIDTH.```
```set TERMINAL:WIDTH to 70.```

File: src/kOS.Safe/Encapsulation/TerminalStruct.cs [hunk 2]
```diff
@@ -0,0 +1,52 @@
...
+        // They may have to change after experimentation.
+        protected const int MINROWS = 3;
+        protected const int MAXROWS = 80;
+        protected const int MINCOLUMNS = 15;
+        protected const int MAXCOLUMNS = 160;
+
+        // TODO: To implement IsOpen, we'd have to make a kOS.Safe interface wrapper around TermWindow first.
+        // That's more than I want to do in this update, I'm leaving it as a TODO for me or someone else:
+        //
+        // protected bool IsOpen { get { return shared.Window.IsOpen(); } set {if (value) shared.Window.Open(); else shared.Window.Close(); } }
```
Dunbaratu: The movement of code between src/kOS and src/kOS.Safe is really not my area of expertise so I was a bit reluctant to make the change described above.  The basic problem was that you can only access shared.Window from kOS, rather than from kOS.Safe.  I'm fairly sure this is fixable by fiddling with the ISharedObjects definition.

src/kOS.Safe/Screen/ScreenBuffer.cs
```diff
@@ -38,6 +44,12 @@ public class ScreenBuffer : IScreenBuffer
         public int RowCount { get; private set; }
         public int ColumnCount { get; private set; }
 
+        /// <summary>Delegate prototype expected by AddResizeNotifier</summary>
+        /// <param name="sb">This screenbuffer telling the callback who it is</param>
+        /// <returns>telling this screenbuffer how many vertical rows to scroll as a result of the resize.</returns>
```
Dunbaratu: This return value isn't implemented in any of the delegates yet (they all just hardcode a return of zero), but the intent is that the delegate can tell the scrollbuffer to scroll itself based upon the SetSize.  This is how we might in the future implement the ability to keep the cursor in view when the commandline subbuffer is made narrower (and thus lines wrap onto more rows, moving the cursor out of view past the end of the terminal.  Right now that isn't fixed until the user types something, and the type-something code forces a scroll to bring the cursor into view.

src/kOS.Safe/Screen/ScreenBuffer.cs
```diff
@@ -20,12 +22,16 @@ public interface IScreenBuffer
...
+        string DebugDump();
     }
+    
 
     public class ScreenBuffer : IScreenBuffer
     {
-        public const int MAX_ROWS = 36;
-        public const int MAX_COLUMNS = 50;
+        private const int DEFAULT_ROWS = 36;
+        private const int DEFAULT_COLUMNS = 50;
```
Dunbaratu: The above names were changed because, well, it's not the MAX - it's just the starting value and the terminal can be made bigger.

src/kOS.Safe/Screen/ScreenBuffer.cs
```diff
@@ -38,6 +44,12 @@ public class ScreenBuffer : IScreenBuffer
         public int RowCount { get; private set; }
         public int ColumnCount { get; private set; }
 
+        /// <summary>Delegate prototype expected by AddResizeNotifier</summary>
+        /// <param name="sb">This screenbuffer telling the callback who it is</param>
+        /// <returns>telling this screenbuffer how many vertical rows to scroll as a result of the resize.</returns>
+        public delegate int ResizeNotifier(IScreenBuffer sb);
+        protected List<ResizeNotifier> notifyees;
```
Dunbaratu: The integer return value isn't implemented in any of the delegates yet (they all just hardcode a return of zero), but the intent is that the delegate can tell the scrollbuffer to scroll itself based upon the SetSize. This is how we might in the future implement the ability to keep the cursor in view when the commandline subbuffer is made narrower (and thus lines wrap onto more rows, moving the cursor out of view past the end of the terminal.) Right now that isn't fixed until the user types something, and the type-something code forces a scroll to bring the cursor into view.

src/kOS.Safe/Screen/ScreenBuffer.cs
```diff
@@ -38,6 +44,12 @@ public class ScreenBuffer : IScreenBuffer
         public int RowCount { get; private set; }
         public int ColumnCount { get; private set; }
 
+        /// <summary>Delegate prototype expected by AddResizeNotifier</summary>
+        /// <param name="sb">This screenbuffer telling the callback who it is</param>
+        /// <returns>telling this screenbuffer how many vertical rows to scroll as a result of the resize.</returns>
+        public delegate int ResizeNotifier(IScreenBuffer sb);
```
Dunbaratu: This integer return value is just hardcoded to return zero in all the delegates that use the system at the moment.
The intent is to allow a delegate to inform the parent ScreenBuffer that it needs to scroll itself based upon the resize operation, to keep the important part in view (the part with the cursor).  Right now that calculation wasn't quite right and it's not strictly necessary for a first pass at this so I left them hardcoded to 0.  (The effect is that at the moment, if you shrink the width such that the current command line gets pushed off the bottom of the screen, you don't scroll the cursor back into view until you start typing, which snaps it back into view.)

src/kOS.Safe/Screen/ScreenBuffer.cs
Dunbaratu: The notifyees are delegates that need to be told whenever this ScreenBuffer gets its size changed.  This is mostly so that the SubBuffers and the TermWindow attached to this ScreenBuffer get told about it and recalculate themselves.  There is no direct reference from SubBuffer to ScreenBuffer, so the SubBuffer normaly can't query its ScreenBuffer to find out its size.  So instead the ScreenBuffer registers the SubBuffer's notifier method with itself when it adds a SubBuffer.

src/kOS.Safe/Screen/ScreenBuffer.cs
```diff
@@ -48,15 +60,18 @@ public int AbsoluteCursorRow
...
-            ColumnCount = MAX_COLUMNS;
+            RowCount = DEFAULT_ROWS;
+            ColumnCount = DEFAULT_COLUMNS;
             InitializeBuffer();
         }
 
         private void InitializeBuffer()
         {
+            buffer.Clear();
             AddNewBufferLines(RowCount);
```
Dunbaratu: It used to be the case that InitializeBuffer was only called at the start of the constructor and never again.  Now it's getting called at other times too, thus the need to explicitly Clear.

src/kOS.Safe/Screen/ScreenBuffer.cs
```diff
@@ -70,15 +85,45 @@ protected void AddNewBufferLines( int howMany = 1)
...
+            ColumnCount = columns;
+            ResizeBuffer();
+            int scrollDiff = 0;
+            foreach (ResizeNotifier notifier in notifyees)
+            {
+                if (notifier != null)
+                    scrollDiff += notifier(this);
+            }
+            ScrollVertical(scrollDiff);
+        }
```
Dunbaratu: The above method is where the meat of the change here comes in.  In the past you only ever initialized the ScreenBuffer once and then it stayed as it was - the same size forever.  But since the ScreenBuffer holds the entire "image" of the terminal (or at least the 2D array of its chars even its blank padding chars), it needs to match the size of the terminal.  Thus it needs to be told to resize itself when the terminal is stretched or shrunk.

src/kOS.Safe/Screen/ScreenBuffer.cs
```diff
@@ -70,15 +85,45 @@ protected void AddNewBufferLines( int howMany = 1)
...
-                InitializeBuffer();
+                char[] newRow = new char[ColumnCount];
+                Array.Copy(buffer[row], 0, newRow, 0, Math.Min(buffer[row].Length, ColumnCount));
+                buffer[row] = newRow;
             }
+            
+            // Add more buffer lines if needed to pad out the rest of the screen:
+            while (buffer.Count - topRow < RowCount)
+                buffer.Add(new char[ColumnCount]);
         }
```
Dunbaratu: And this is how it preserves the contents.  When the terminal size changes. it copies the old data into the new buffer its making, row for row, cutting off old content that was past the end (if shrinking the width or height).

For this first pass, the old content is thrown away when cropped off the new smaller size.  For a more sophisticated future pass it may be possible to wrap long lines when the width is shortened, and unwrap them when it's widened, but doing so requires that the ScreenBuffer record the difference between a "soft" end of line wrap and "hard" return, so it knows where the wrapping happens.  Right now that information is lost after the Print() statement is done.

src/kOS.Safe/Screen/ScreenBuffer.cs
```diff
@@ -20,12 +22,16 @@ public interface IScreenBuffer
...
+    
 
     public class ScreenBuffer : IScreenBuffer
     {
-        public const int MAX_ROWS = 36;
-        public const int MAX_COLUMNS = 50;
+        private const int DEFAULT_ROWS = 36;
+        private const int DEFAULT_COLUMNS = 50;
         
         private int topRow;
```
Dunbaratu: In order to understand some of the algorithm of ScreenBuffer, it's important to remember the meaning of this little "topRow" variable.  The ScreenBuffer is NOT just a buffer of the current terminal view, but it's actually the entire scrollback buffer of everything that got printed to the terminal in the past.  (It would be possible to use it to implement an actual vertical scrollbar for the terminal if we wanted to).  The topRow variable defines what portion of the ScreenBuffer is currently in view by defining which index into buffer is the first row currently shown on the screen.  (the bottom row is therefore at index (topRow + RowCount - 1). )

src/kOS.Safe/Screen/ScreenBuffer.cs
```diff
@@ -258,6 +296,27 @@ public List<char[]> GetBuffer()
...
+                {
+                    char ch = buffer[i][j];
+                    sb.Append((int)ch < 32 ? (" \\"+(int)ch) : (" "+ch) );
+                }
+                sb.Append("]\n");
+            }
+            foreach (SubBuffer sub in subBuffers)
+                sb.Append(sub.DebugDump());
+            return sb.ToString();
+        }
```
Dunbaratu: This isn't being used in the checked-in code because making use of it spams the output log and slows KSP down.  But it was very handy for debugging things and I'd like it left in place so it's still available for any future expansions to be able to debug with it.

src/kOS.Safe/Screen/SubBuffer.cs
```diff
@@ -13,24 +15,116 @@ public class SubBuffer
...
+
+            // Because Buffer is readonly, copy the data from newBuffer into it rather than just resetting the reference to newBuffer:
             Buffer.Clear();
+            for (int i = 0 ; i < newBuffer.Count ; ++i)
+                Buffer.Add(newBuffer[i]);
 
-            for (int row = 0; row < RowCount; row++)
+            // return Fixed ? 0 : scrollDiff; - this logic not quite working - disabled for now.
+            return 0;
+        }
```
Dunbaratu: The logic of SubBuffer.ResizeBuffer() (the method above) is important to pay attention to because it's essentially how we'd do proper line wrapping for the whole ScreenBuffer.  The important thing is that here in SubBuffer we can make the safe assumption that the entire contents of SubBuffer are indeed meant to be one joined together line.  Thus we can invoke that assumption and wrap the lines properly.  That's not an assumption we can make about the ScreenBuffer itself because the information about which rows are run-on continuation rows is lost.

The fact that we can make the assumption here is why this code is so much more complex than ResizeBuffer in ScreenBuffer.  But it also means that when you resize a terminal while a commandline was halfway typed in, that your commandline gets preserved properly and wrapped correctly even when the rest of the terminal doesn't.

src/kOS.Safe/Screen/TextEditor.cs
```diff
@@ -1,6 +1,7 @@
 using System.Collections.Generic;
 using System.Text;
 using kOS.Safe.Utilities;
+using System;
 
 namespace kOS.Safe.Screen
 {
```
Dunbaratu: The weird thing to remember is that TextEditor, despite the name, is really only used now as the base for the commandline editing logic of the interactive interpreter.

src/kOS/Binding/TerminalSettings.cs
```diff
@@ -10,6 +11,7 @@ public override void AddTo(SharedObjects shared)
             shared.BindingMgr.AddGetter("SESSIONTIME", () => shared.Cpu.SessionTime);
             shared.BindingMgr.AddGetter("VERSION", () => Core.VersionInfo);
             shared.BindingMgr.AddGetter("VOLUME:NAME", () => shared.VolumeMgr.CurrentVolume.Name);
+            shared.BindingMgr.AddGetter("TERMINAL", () => new TerminalStruct(shared));
```
Dunbaratu: Incedentally, why on earth were the above exiting 3 things in a file called "TerminalSettings"??  They have nothing to do with terminals.

src/kOS/Screen/TermWindow.cs
```diff
@@ -14,7 +15,8 @@ public class TermWindow : KOSManagedWindow
     {
         private const int CHARSIZE = 8;
         private const string CONTROL_LOCKOUT = "kOSTerminal";
-        private const int CHARS_PER_ROW = 16;
+        private const int FONTIMAGE_CHARS_PER_ROW = 16;
```
Dunbaratu: Renamed to make it clear that the "per row" it's talking about is about how the character pictures are packed into the font image file, not about rows of the terminal.

src/kOS/Screen/TermWindow.cs
```diff
@@ -22,8 +24,10 @@ public class TermWindow : KOSManagedWindow
         private static readonly Color textColorAlpha = new Color(0.45f, 0.92f, 0.23f, 0.5f);
         private static readonly Color textColorOff = new Color(0.8f, 0.8f, 0.8f, 0.7f);
         private static readonly Color textColorOffAlpha = new Color(0.8f, 0.8f, 0.8f, 0.3f);
-        private static readonly Rect closeButtonRect = new Rect(398, 370, 59, 30);
-        
+        private Rect closeButtonRect = new Rect(0, 0, 0, 0); // will be resized later.
```
Dunbaratu: Can't be static because its position is possibly stretchable per terminal

src/kOS/Screen/TermWindow.cs
```diff
@@ -47,13 +52,14 @@ public class TermWindow : KOSManagedWindow
...
             IsPowered = true;
-            windowRect = new Rect(60, 50, 470, 405);
+            windowRect = new Rect(50, 60, 0, 0); // will get resized later in AttachTo().
         }
 
         public void Awake()
         {
             LoadTexture("GameData/kOS/GFX/font_sml.png", ref fontImage);
             LoadTexture("GameData/kOS/GFX/monitor_minimal.png", ref terminalImage);
+            LoadTexture("GameData/kOS/GFX/resize-button.png", ref resizeButtonImage);
```
Dunbaratu: For some reason I cannot fathom, this is getting rendered on the button as just a tiny little red squiggle.  That's not what the image file actually looks like.

src/kOS/Screen/TermWindow.cs
```diff
@@ -364,13 +385,47 @@ void TerminalGui(int windowId)
                 ShowCharacterByAscii((char)1, screen.CursorColumnShow, screen.CursorRowShow, currentTextColor);
             }
             GUI.EndGroup();
+            
+            GUI.Label(new Rect(windowRect.width/2-40,windowRect.height-20,100,10),screen.ColumnCount+"x"+screen.RowCount);
 
+            CheckResizeDrag(); // Has to occur before DragWindow or else DragWindow will consume the event and prevent drags from being seen by the resize icon.
+            GUI.DragWindow();
         }
```
Dunbaratu: I'm becoming a bit unhappy about how long the TerminalGui() method (above) is getting.  But I was unwilling to split it up just yet, until we're certain everything is working just right the way it is.  It can be broken into smaller pieces later.

src/kOS/Screen/TermWindow.cs
```diff
@@ -364,13 +385,47 @@ void TerminalGui(int windowId)
...
+                    // Note I wanted to call this dynamically as it's resizing, but there's some weird issue
+                    // with the timing of it, where the windowRect is temporarily set to a bogus value during the
+                    // time it was trying to run this, and thus it created exception-throwing code unless it waits
+                    // until the drag is done before trying to calculate this.
+                    // The effect the user sees is that the text can float in space past the edge of the window (when
+                    // shrinking the size) until the resize drag is done, and then it redraws itself.
+                    shared.Screen.SetSize(HowManyRowsFit(), HowManyColumnsFit());
+                }
+            }
+        }
```
Dunbaratu: This is quite similar, but not *identical* logic to the resize feature of the source code editor popup window.  It's similar enough that we might consider trying to consolidate the logic together at some point.  It's one of those cases of almost-but-not-quite cut&paste code that makes you think "this is probably an opportunity for a refactor".

src/kOS/Screen/TermWindow.cs
```diff
@@ -397,6 +452,24 @@ internal void AttachTo(SharedObjects shared)
...
+        
+        private int HowManyRowsFit()
+        {
+            return (int)(windowRect.height - 100) / CHARSIZE;
+        }
+
+        private int HowManyColumnsFit()
+        {
+            return (int)(windowRect.width - 65) / CHARSIZE;
```
Dunbaratu: These are meant to be the mapping from the window's pixel size to it's character cell size.

Dunbaratu:
I have discovered one small problem with this PR that I may want to try to address.  It's not the end of the world, because it takes some messing about to actually trigger the bug, but I may follow up this PR with a second one to address the problem.  (In other words, don't let the problem prevent you from pulling the PR.)

The problem is this one:  *If you repeatedly make the window wider, then narrower, then wider, then narrower, about 10-20 times, after you've ever typed in an interactive command*, then the terminal can become semi-permanently unusable because all the text is always scrolled up off the screen where you can't see it.  It can't be fixed until you leave the scene and return (thus causing the terminal, as part of the CPU's SharedObjects, to get destructed and re-constructed, thus resetting the data that caused bug.)

The cause of the bug is this:
The subbuffer that holds the command-line overlay atop the terminal's screenbuffer needs to always be resized to match the terminal's screenbuffer when the terminal is resized.  Currently, that resizing ignores the actual *used* size of the subbuffer, and merely tries to retain the contents of the entire subbuffer, even the extra "null" chars (ascii 0) that occur past the end of the typed text.

The effect is that it does this:
Step 1: Your terminal is currently 50x42 (50 cols, 42 rows) (the starting default size, if I recall).  The command-line subbuffer is sized at 50x1.
Step 2: You widen the terminal so it is now 80x42.  The command-line subbuffer gets padded out to be 80x1 now.
Step 3: You shrink the terminal width so it is now 50x42 again.  The command-line subbuffer, in order to hold just as many characters as it previously did at 80x1, has to wrap and add a row.  Because the subbuffer must be the same width as the screen, it pads the new row out to the entire size of 50 wide even though it only needs to be 1 row of 50 and one row of 30 to hold the 80 char cells.  So the new size is now 50x2.
Step 4: You widen the terminal back to 80x42 again.  Now the 50x2 subbuffer gets padded out to be 80x2.  Therefore it now holds 160 chars even though most of them are pointless null chars.
Step 5: You shrink the terminal back to 50x42 again.  Now the 160 chars of the old 80x2 subbuffer need a new row added again, in fact they need 2 of them because 50x3 isn't big enough as it's only 150 chars.  The new subbuffer size is now 50x4, which is 200 chars.
Step 6: You widen the terminal back to 80x42 again.   Now the previous 50x4 subbuffer gets widened to 80x4 to fit the new wider screen, so now it's got 320 chars in it, most of which are null chars.

...etc...
Repeatedly doing this long enough, you end up with a subbuffer that has grown to the point where it's taller than the 42 rows of the full terminal.  When it's trying to be displayed, the terminal messes up the display.

Fixing it is possible, but I'm leaving it for a future pull request, as it requires fiddling with the up-arrow commandline history, which I'm only just getting into looking at now.  The reason it needs the extra padding null chars space now, is because when you up-arrow to get a previous command back again, and the previous command was bigger than your current one, if your new buffer is smaller than the previous one (because the command you are up-arrowing to was a 2-line command but your current one is a 1-line command) it bombs out trying to fill it into the new, smaller subbuffer.  The way the existing code seems to have gotten around this problem was to just have the rule "the commandline subbuffer never ever gets trimmed back down.  It only grows.  It never shrinks."   Thus if there ever had been a 2-row command in the past, then the subbuffer is always 2-rows long from now on.  If there had ever been a 3-row command in the past, then the subbuffer is always 3-rows long from now on.  To get rid of the phantom extra chars that keep growing in the subbuffer size from all the resizing as I described above, I'll have to tackle that assumption and change it.  It has to become possible to trim down the commandline subbuffer again after a long wrapped command, and re-grow it dynamically again if the previous, longer command is brought back in.

src/kOS.Safe/Encapsulation/TerminalStruct.cs
erendrake: window will be pretty hard to move over on its own because it inherits from MonoBehavior. We could just give it an interface ITermWindow except for AttachTo that needs the kOS sharedobjects object.

src/kOS/Binding/TerminalSettings.cs
erendrake: Who knows :P i am willing to bet i added at least one of them but you are right. they belong somewhere else. so much so that i moved version to universe.

erendrake:
Greate PR, thank you very much for this fine work @Dunbaratu :)

src/kOS.Safe/Encapsulation/TerminalStruct.cs
Dunbaratu: Well, if we decide that having TERMINAL:OPEN (settable and gettable - opens when set true, closes when set false) is a worthy thing, then the other way to allow it could be to just move TerminalStruct.cs from kOS.Safe into kOS.

Issue #389, opened 2014-11-18
LOCK STEERING broken for RCS-only (no torque) ships.


Dunbaratu:
Problem described here:

http://forum.kerbalspaceprogram.com/threads/68089-0-25-kOS-Scriptable-Autopilot-System-v0-15-0-2014-11-16?p=1548099&viewfull=1#post1548099

soulsource:
While I don't know how to fix this, I can meanwhile tell you where the bug probably sits:
The if-block starting at line 170 of Utilities/SteeringHelper.cs

If I understand that code correctly, it currently approximates the maximum possible torque for pitch/yaw of each RCS block by simply multiplying its maximum thrust with the distance to the COM, what is a reasonable approximation for very long ships.
Nevertheless, the contribution of RCS blocks to roll adjustments is completely neglected.

While I don't know how to fix stuff, as I have no idea about C# and even less about the KSP plugin API, I can outline how I would improve (not fix - a fix would require to treat yaw and pitch separately) that calculation:
I'd try to extract vectors for all possible thrust directions of the RCS block, and the direction of the ships roll axis. Then, for each RCS thrust direction, I'd calculate the magnitude of the cross product and of the dot product between thrust direction and roll axis, and add them to the corresponding torque variables (of course: cross product means roll torque, dot product means pitch/yaw torque).

soulsource:
I "fixed" the issue for my own build. Basically, for reasons I don't fully understand, the RCS blocks need to be accessed as PartModule instead of using the RCSModule part type. I also tried to improve the torque prediction slightly, but it's still just a rough approximation, assuming a fully pivoting RCS nozzle... The GetTorque function anyhow has some additional issues (like not taking into account the position of reaction wheels - Huygens-Steiner theorem), if I understand it correctly, so being less than fully exact here will probably not hurt.

As I'm new to git, it might take a few hours until I figure how to fork the KOS repositoy and to submit a pull request, but be assured that you'll see one soon.

soulsource:
Ok, pull request sent.

soulsource:
Commit d8aed05a78745ae0675f89d4e99a3004c874c2c9 brings checks for thruster fuel availability (per part, as not all thrusters need to use the same fuel(s) if mods altering the fuel system are installed), and only takes thruster ports into account that aren't disabled (either individually or globally by using the RCS toggle). 
[...]

PR #520, opened 2015-01-25
Fix for issue #389 - LOCK STEERING broken for RCS-only (no torque) ships...

soulsource:
The RCS parts seem to require accessing their individual PartModule members instead of accessing them through RCSModule.
Also: Approximated torque caused by RCS modules better than before (but still not correct - that would require the whole function to be rewritten).

### DIFF
--- a/src/kOS/Utilities/SteeringHelper.cs
+++ b/src/kOS/Utilities/SteeringHelper.cs
@@ -153,8 +153,11 @@ namespace kOS.Utilities
             var centerOfMass = vessel.findWorldCenterOfMass();
             var rollaxis = vessel.transform.up;
             rollaxis.Normalize ();
+            var pitchaxis = vessel.GetFwdVector ();
+            pitchaxis.Normalize ();
             
-            float pitchYaw = 0.0f;
+            float pitch = 0.0f;
+            float yaw = 0.0f;
             float roll = 0.0f;
 
             foreach (Part part in vessel.parts)
@@ -164,7 +167,8 @@ namespace kOS.Utilities
                 var pod = part as CommandPod;
                 if (pod != null)
                 {
-                    pitchYaw += Math.Abs(pod.rotPower);
+                    pitch += Math.Abs(pod.rotPower);
+                    yaw += Math.Abs(pod.rotPower);
                     roll += Math.Abs(pod.rotPower);
                 }
                 /*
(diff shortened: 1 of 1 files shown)


soulsource:
Please wait with merging this. There's a mistake (forgot to take the absolute value of the dot product in line 203), and I'm currently working on checks for the conditions Dunbaratu mentioned. I'll push another commit soon.

soulsource:
The latest commit should fix the missing absolute value over which I stumbled, it should check if RCS is toggled, if individual thrusters are disabled, and if all required fuels for the RCS are available.

soulsource:
Sorry for spamming, but I couldn't stop playing around with the code, and with a little inspiration from the RCS Build Aid source code I managed to obtain torque values which I think are correct. I compared them to the torque values exposed by Mechjeb, and indeed the numbers are equal for the crafts I checked. The only thing I'm not certain about is the order of the return values - maybe pitch and yaw need to be exchanged... Ah, and asymmetric thruster placement can still cause issues, as currently only positive torques are being counted.

src/kOS/Utilities/SteeringHelper.cs
```diff
@@ -162,31 +167,65 @@ public static Vector3d GetTorque(Vessel vessel, float thrust)
...
+                                    //just counting positive contributions in one direction. This is incorrect for asymmetric thruster placements.
+                                    roll += Mathf.Max(rcsthrust * Vector3.Dot(Vector3.Cross(relCoM, thrustdir.up), rollaxis), 0.0f);
+                                    pitch += Mathf.Max(rcsthrust * Vector3.Dot(Vector3.Cross(Vector3.Cross(relCoM, thrustdir.up), rollaxis), pitchaxis), 0.0f);
+                                    yaw += Mathf.Max(rcsthrust * Vector3.Dot(Vector3.Cross(Vector3.Cross(relCoM, thrustdir.up), rollaxis), Vector3.Cross(rollaxis,pitchaxis)),0.0f);
+                                }
+                            }
+                        }
+                    }
+                }
+                pitch += (float)GetThrustTorque(part, vessel) * thrust;
```
erendrake: GetThrustTorque doesnt actually work because it uses the engine derived classes that are not being used anymore. We need to get a new implementation that uses the gimbal module.

src/kOS/Utilities/SteeringHelper.cs
```diff
@@ -162,31 +167,65 @@ public static Vector3d GetTorque(Vessel vessel, float thrust)
...
                 if (pod != null)
                 {
-                    pitchYaw += Math.Abs(pod.rotPower);
+                    pitch += Math.Abs(pod.rotPower);
+                    yaw += Math.Abs(pod.rotPower);
                     roll += Math.Abs(pod.rotPower);
                 }
-
+                /*
+                //This is probably useless - commented out.
```
erendrake: if it is useless, remove it.

src/kOS/Utilities/SteeringHelper.cs
```diff
@@ -162,31 +167,65 @@ public static Vector3d GetTorque(Vessel vessel, float thrust)
                 var pod = part as CommandPod;
```
erendrake: we dont have CommandPods anymore, they are all ModuleCommand now. This is likely never being hit

erendrake:
Thank you very much for the PR :)

src/kOS/Utilities/SteeringHelper.cs
soulsource: As the code seems to be working on crafts that only have reaction wheels in the command pod, I guess that they are accessed later on, when accessing the ModuleReactionWheel part modules. I guess one can safely remove this obsolete code.

PR #525, opened 2015-01-28
Fixes #523 and Fixes #524

### LINKED ISSUE #523: New waypoint function doesn't work on alpha/beta/gamma sub-sites.
When there are a group of waypoints near each other, named "site foo alpha", "site foo beta", and "site foo gamma", then this fails to work:
```set wp to waypoint("site foo beta").```
(it returns a null into wp).
But this does work:
```set wp to waypoint("site foo").```
even though there are no waypoints on-screen that are labeled as such.  The location returned by this is the central average point between the alpha/beta/gamma sites.  It exists as a waypoint in memory but does not appear anywhere on the UI.

```
    EDIT: THE above sentence isn't true.  I thought it was until
    I got into the details of fixing this.  Actually, what you used to get
    was just the 'alpha' location when you used the term without the name,
    because it was retuning the first waypoint it hit and all the waypoints
    had the same name.  It was NOT returning the average central location.
```

So it seems there is not exactly a 1-to-1 match between the waypoint data structure and the waypoints as shown to the user.  When it's a cluster of waypoints near a center point, the user *sees* 3 different waypoints, but behind the scenes those waypoints are not "full" citizens of the waypoint system and are some kind of different structure that hangs off the parent waypoints.

The fix should be:  Make kOS match the user's view.  To the USER they look like 3 entirely independent waypoints that just happen to be near each other and have similar names, so kOS should abstract away the underlying difference between the data path to reach a root waypoint and a child waypoint, and make them appear to be the same to the user.

### LINKED ISSUE #524: Make waypoints show 3D position.
Allow you to get the position of a waypoint as a 3D position (like Vessel:position or body:position work).

### PR DESCRIPTION
**Pull request by Dunbaratu**
Making waypoints a bit better.

For #523: Fixed the lookup of user-provided waypoint names to be able to find clustered sites.

  It turns out that clustered sites' names as they appear on screen to the user are NOT how they're actually stored internally.  If a site has a name like "Jebadiah's Lament Gamma" on the screen, then in *reality* it's waypoint name is *really* this:

Waypoint.name = "Jebadiah's Lament",
Waypoint.index = 2,
Waypoint.clustered = true.

When  presenting the name to the user, the game **builds** the name by looking up what greek letter name to use for index=2 ("Gamma") and then appending it to the name for display only.  

Since the API provides no nice way to go the other way in this system (decompose a display name to its component parts), nor does it actually store the display name in the class or provide it as a Property built on the fly that I can query with Linq tests, that means I had to make presumptions about the display-name making algorithm and repeat the algorithm (backward) in kOS's own code.  I spoke with the author of Fineprint and the assumptions *are* correct, but because we're not using any built-in method for it, the danger exists that the assumption could become wrong in the future.

One example of a potential problem is that if you use a greek letter term in the name, kOS assumes that cannot happen unless it's a clustered waypoint.  In other words, if there's a waypoint called "FooFiFum beta", and it's not *really* a cluster, but just utterly coincidentally happens to end in the word "beta", kOS won't find it because upon seeing the "beta" it will presume it is a clustered waypoint and look for just the prefix name "FooFiFum".

None of these are problems NOW in KSP 0.90.  They're just gotchas to look for later in future releases.

For #524: Added suffix waypoints:

   :POSITION - 3D vector position of the waypoint.
   :CLUSTERED - boolean for if the waypoint is part of an alpha/beta/gamma... cluster
   :INDEX - if :CLUSTERED is true, is this the zeroth, first, second, etc member of the cluster.

### DIFF
--- a/src/kOS/Suffixed/WaypointValue.cs
+++ b/src/kOS/Suffixed/WaypointValue.cs
@@ -91,20 +91,18 @@ namespace kOS.Suffixed
             return String.Format("A Waypoint consisting of\n" +
                                  "  name= {0}\n" +
                                  "  body= {1}\n" +
-                                 "  latitude= {2}\n" +
-                                 "  longitude= {3}\n" +
+                                 "  geoposition= {2}\n" +
+                                 "  agl= {3}\n" +
                                  "  altitude= {4}\n" +
-                                 "  height= {5}\n" +
-                                 "  isOnSurface= {6}\n" +
-                                 "  worldPosition= {7}\n" +
-                                 "  index= {8}\n" +
-                                 "  clustered= {9}\n",
+                                 "  nearsurface= {5}\n" +
+                                 "  position= {6}\n" +
+                                 "  index= {7}\n" +
+                                 "  clustered= {8}\n",
                                  CookedName(),
                                  WrappedWaypoint.celestialName,
-                                 WrappedWaypoint.latitude,
-                                 WrappedWaypoint.longitude,
+                                 new GeoCoordinates(GetBody(), Shared, WrappedWaypoint.latitude, WrappedWaypoint.longitude),
                                  WrappedWaypoint.altitude, // A location inside the contract range's altitude range - and NOT the edge of it.
-                                 WrappedWaypoint.height,
+                                 BuildSeaLevelAltitude(),
                                  WrappedWaypoint.isOnSurface,
                                  GetPosition(),
                                  WrappedWaypoint.index,
--- a/src/kOS/Suffixed/WaypointValue.cs
+++ b/src/kOS/Suffixed/WaypointValue.cs
@@ -100,7 +100,7 @@ namespace kOS.Suffixed
                                  "  clustered= {8}\n",
                                  CookedName(),
                                  WrappedWaypoint.celestialName,
-                                 new GeoCoordinates(GetBody(), Shared, WrappedWaypoint.latitude, WrappedWaypoint.longitude),
+                                 BuildGeoCoordinates(),
                                  WrappedWaypoint.altitude, // A location inside the contract range's altitude range - and NOT the edge of it.
                                  BuildSeaLevelAltitude(),
                                  WrappedWaypoint.isOnSurface,
--- a/src/kOS/Suffixed/WaypointValue.cs
+++ b/src/kOS/Suffixed/WaypointValue.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using kOS.Utilities;
 using kOS.Safe.Encapsulation.Suffixes;
 using kOS.Safe.Encapsulation;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/kOS/Suffixed/WaypointValue.cs [hunk 1]
```diff
@@ -8,47 +9,80 @@ namespace kOS.Suffixed
 {
     public class WaypointValue : Structure
     {
-        protected Waypoint WayPoint { get; set; }
+        protected Waypoint WrappedWaypoint { get; set; }
```
Dunbaratu: Renamed purely for style reasons: If the type name "Waypoint" establishes the standard that the "P" in "Point" doesn't count as a second word (not capitalized), then that should be adhered to universally, and that would make the property ```WayPoint``` into just ```Waypoint```.  I then renamed ```Waypoint``` to ```WrappedWaypoint``` because although C# allows you to use the exact same identifier (including capitalization) as both a type name and an instance name, I consider that bad practice and try to avoid it, but the capitalization rules the project is operating under don't allow this property to be lowercase named (because it uses capitalization to differentiate fields from methods rather than the standard I'm used to which uses it to differentiate type names from instance names), so I was left with just giving it a totally different name.  I chose ```WrappedWaypoint``` as in "the waypoint being wrapped by this WaypointValue class"

src/kOS/Suffixed/WaypointValue.cs
```diff
@@ -8,47 +9,80 @@ namespace kOS.Suffixed
...
         public WaypointValue(Waypoint wayPoint, SharedObjects shared)
         {
-            WayPoint = wayPoint;
+            WrappedWaypoint = wayPoint;
             Shared = shared;
             AddSuffix("DUMP", new NoArgsSuffix<string>(ToVerboseString)); // for debugging
-            AddSuffix("NAME", new NoArgsSuffix<string>(() => wayPoint.name, "Name of waypoint as it appears on the map and contract"));
-            AddSuffix("BODY", new NoArgsSuffix<BodyTarget>(() => new BodyTarget(GetBody(),shared), "Celestial body the waypoint is attached to"));
+            AddSuffix("NAME", new NoArgsSuffix<string>(() => CookedName(), "Name of waypoint as it appears on the map and contract"));
+            AddSuffix("BODY", new NoArgsSuffix<BodyTarget>(() => new BodyTarget(GetBody(),Shared), "Celestial body the waypoint is attached to"));
```
Dunbaratu: Esoterically, I prefer using the class member here ```Shared``` (capitalized) as opposed to the constructor's argument ```shared``` (lowercase).  Granted, they're the same at this point, but It just seems more appropriate as it makes the code more mobile (it could be moved outside the constructor and still work, this way).

src/kOS/Suffixed/WaypointValue.cs
```diff
@@ -8,47 +9,80 @@ namespace kOS.Suffixed
...
         }
         
         public override string ToString()
         {
-            return String.Format("Waypoint \"{0}\"", WayPoint.name);
+            return String.Format("Waypoint \"{0}\"", CookedName() );
+        }
+        
+        public string CookedName()
+        {
```
Dunbaratu: The purpose of ```CookedName()``` (which is also accessed via the suffix ```:NAME``` is to present the user with the facade that the waypoint's name really is the name they see on the screen.

teleksterling:
With this change, what now happens if a user queries 'Jebadiah's Lament' without the suffix? Is it still the average location, or a null?

Dunbaratu:
> With this change, what now happens if a user queries 'Jebadiah's Lament' without the suffix? Is it still the average location, or a null? 
The intent is to match exactly to what the user can see on-screen.  Therefore in the scenario you describe, there is no such waypoint as "Jebadiah's Lament".  The user can't see it on the screen, no such waypoint has that label at all, so yeah it returns a null to leave the alpha/beta/gamma part of the label off.

src/kOS/Suffixed/WaypointValue.cs
erendrake: I agree, i went ahead a moved the suffix creation into its own method so you have to use the property.

teleksterling:
Fair enough! 
The mention of a central average was just based on my cursory reading of #523. Obviously now that you've delved into the code further, you've uncovered the true (old) behavior - which clearly is of no use to anyone!
Thanks for the response.

PR #530, opened 2015-01-29
fixes #522 with new suffixes for BodyTarget

### LINKED ISSUE #522: Construct GeoCoordinates from position
It would be awfully handy to be able to construct a GeoCoordinates object from an arbitrary position vector. This isn't a showstopper since I can always do the math myself, but it does tend to clutter up the code (especially with no support for user-defined functions).

Dunbaratu:
The changes are pretty self-explanatory - just read my changes to the docs files that come with the PR.

I picked the names GEOPOSITION**OF**() and ALTITUDE**OF**() because I can't overload methods in kOS suffixes, and there already exist zero-args suffixes called GEOPOSITION and ALTITUDE due to the fact that BodyTarget is an Orbitable.  When I tried calling the new suffiix by the same name, the system didn't find it - hitting the zero-args version first and throwing an exception because of the wrong number of args.

### DIFF
--- a/doc/source/structures/celestial_bodies/body.rst
+++ b/doc/source/structures/celestial_bodies/body.rst
@@ -56,6 +56,8 @@ All of the main celestial bodies in the game are reserved variable names. The fo
     :attr:`MU`                       scalar (:math:`m^3 s^{−2}`)
     :attr:`ATM`                      :struct:`Atmosphere`
     :attr:`ANGULARVEL`               :struct:`Direction` in :ref:`SHIP-RAW <ship-raw>`
+    :attr:`GEOPOSITIONOF`            :struct:`GeoCoordinates` in :ref:`SHIP-RAW <ship-raw>`
+    :attr:`ALTITUDEOF`               scalar (m)
     ================================ ============
 
 .. attribute:: Body:NAME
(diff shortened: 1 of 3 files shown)


erendrake:
if this is meant to replace the Geoposition and altitude suffixes but replacement isnt possible i would rather fix that then have two terms.

Dunbaratu:
The thing they provide didn't exist before .  They're not the geoposition of a ship, but the geoposition of an arbitrary vector.  That never existed before.

They really do mean something quite different.  body:geopsition means "treat this body as an orbitable object.  Get its positon on its parent."  (In other words, it's the same thing as SHIP:Geoposiion.  Geoposiion of the thing itself.)

body:geopositionof(vector) means "get the vector's position on THIS body".

Dunbaratu:
In retrospect while overridden suffixes might be nice, now that I think of it this might not be a good place to use it.  These aren't the same meaning but with different data.  These are actually different meanings.

erendrake:
alright, are you happy with these names then?

Dunbaratu:
What names ? I don't see what changes you're proposing.

erendrake:
you said that you named them something different because overriding Orbitable was not possible. so you named them something very similar. Now that we discovered that they are actually pretty different, Are you still happy with having the names be so similar? or would you rather want something like GEOPOSITIONFROMVECTOR or something else.

Dunbaratu:
any name more descriptive would probably be just horribly verbose.  it's already pretty verbose to begin with.  I think I'm okay with it like I proposed it now.

PR #537, opened 2015-01-31
Fixes #536 just moving the attachment point.

### LINKED ISSUE #536: Attachment point of the CX-4181 causes graphic overlap.
I think this might be quickly fixable with just a mere edit of the attachment points in the part.cfg file:

overlap_illustration

Dunbaratu:
New screenshot:

overlap_fix_illustration

Note, it does sometimes make parts seem to hover a bit off the connection, as shown, but then again normal parts do that too a bit (as shown).


Dunbaratu:
It just occurred to me - I wonder what will happen if an old vessel made prior to this change is loaded into memory after the change?  Will it end up with parts stored at the wrong location intersecting each other, or will it recreate them according to the new attachment points?

erendrake:
well the vessels that are in play already wont get updated, the craft in the VAB im not sure. they will very likely get shuffled around with the change.

PR #556, opened 2015-02-09
added HudTxt function

pgodd:
command format hudtxt ( "text"+var, int delay, int style, int size, "color", bool(0|1) mirror in terminal).
e.g.:   hudtxt ("Hello Kerbin!, I'm "+round(alt:radar)+"m above you", 2, 3, 25, "red", 0).

                           in addition you can enter some HTML tags. I haven't explored this one much "<i>Hello world</i>" will produce italics.
Delay = seconds integers only, the amount of time the message will be displayed. New messages within this time will cause message to scroll
Style = choice of styles between 1 and 4 ~ 1=Upper_left 2=Upper_center 3=Upper_right 4=Lower_center   The positions of upper right is actually lower right... comes from KSP code. If an improper choice is made i.e.:5 then defaults and prefixes with *.
Size = ? not really sure what this is in terms of how big the text is. 15 will get you started on any. 150 is huge. Effect varies depending on style
Color = (hurts my Canadian eyes, COLOUR!) this is the same as an HTML color tag. e.g. "red" or "#ff0099ff"
Mirror = mirror the on screen display in the terminal 0=no 1=yes (I want to change this to a real boolean when I figure out how)

That's all I got. The command lines are long, but easily remembered.

### DIFF
--- a/src/kOS/Function/Misc.cs
+++ b/src/kOS/Function/Misc.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using kOS.Execution;
 using kOS.Safe.Compilation;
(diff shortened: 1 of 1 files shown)


pgodd:
Sorry, still learning.
I have to resubmit the pull request with ALL the changes.

PR #558, opened 2015-02-09
This adds HudTxt function

pgodd:
command format hudtxt ( "text"+var, int delay, int style, int size, "color", bool(0|1) mirror in terminal).
e.g.: hudtxt ("Hello Kerbin!, I'm "+round(alt:radar)+"m above you", 2, 3, 25, "red", 0).

in addition you can enter some HTML tags. I haven't explored this one much "Hello world" will produce italics.
Delay = seconds integers only, the amount of time the message will be displayed. New messages within this time will cause message to scroll
Style = choice of styles between 1 and 4 ~ 1=Upper_left 2=Upper_center 3=Upper_right 4=Lower_center The positions of upper right is actually lower right... comes from KSP code. If an improper choice is made i.e.:5 then defaults and prefixes with *.
Size = ? not really sure what this is in terms of how big the text is. 15 will get you started on any. 150 is huge. Effect varies depending on style
Color = (hurts my Canadian eyes, COLOUR!) this is the same as an HTML color tag. e.g. "red" or "#ff0099ff"
Mirror = mirror the on screen display in the terminal 0=no 1=yes (I want to change this to a real boolean when I figure out how)

That's all I got. The command lines are long, but easily remembered.


Dunbaratu:
Actually you were right the first time.  It should have been a pull into develop, not a pull into master.

pgodd:
Story of my life. Second guessing. Do I need to do anything?

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -606,6 +606,9 @@ private void VisitNode(ParseNode node)
                 case TokenType.arglist:
                     VisitArgList(node);
                     break;
+				case TokenType.hudtxt_stmt:
+					VisitHudtxtStatement(node);
+					break;
```
Dunbaratu: There's a common trend where programmers tend to think of the "tab" ascii character as a software re-assignable width that can be whatever you like, and that tends to mess up when different people have it set to different values in their viewers.  And in all cases where it's hardcoded and can't be changed, it's 8 columns, and thus picking a value of 4 makes it come out wrong in those places.

To get around this problem we've made the standard that the source files just shouldn't have any tabs at all.  It makes it simpler to deal with.  Most editors can be made to export the file with tabs turned into runs of spaces.

It's easy enough to fix for this pull request.  Just something to remember if you want to make more later.

pgodd:
Thanks Steve! I really appreciate the extra time you take to explain things. 
I went through a number of text editors before arriving at Sublime.
I read about the concern over tabs and spaces in the style guidelines, but incorrectly assumed I was doing it right since I had downloaded xml files to set up Sublime for KSP.
I'll make sure I make that correction.
Please educate me on any mistakes I make. I'm a quick study.
cartoon6

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -1703,6 +1706,20 @@ private void VisitToggleStatement(ParseNode node)
...
+				VisitNode(node.Nodes[2]);
+				VisitNode(node.Nodes[4]);
+				VisitNode(node.Nodes[6]);
+				VisitNode(node.Nodes[8]);
+				VisitNode(node.Nodes[10]);
+				VisitNode(node.Nodes[12]);
+				AddOpcode(new OpcodeCall("hudtxt()"));
+			}
+		}
+
```
Dunbaratu: This can probably be re-indented by an auto-reformat tool on our end, but I really hate using them because if we run all files through them we preclude the abiliy to use intelligent indents in places where it makes sense to look a bit different.  (i.e. places where people have padded spaces into the code to make columns of stuff line up, for example.)

pgodd:
LOL Rant accepted as filed.

src/kOS.Safe/Screen/ScreenBuffer.cs
```diff
@@ -203,12 +202,17 @@ private void PrintLine(string textToPrint)
             MoveColumn(textToPrint.Length);
         }
 
+		public void HudTxt(string textToHud, int delay, int style, int size, string color, int mirror)
+		{
+		}
+
```
Dunbaratu: As far as I can tell, I don't think this method is actually being called by anywhere.

Was it something you stared thinking you'd use, and then you changed your mind and it's just leftover in the code from an older idea (meaning we should delete it), or is it a stub you plan to use in a future update (meaning we shouldn't delete it) ?

src/kOS/Function/Misc.cs
```diff
@@ -18,6 +18,51 @@ public override void Execute(SharedObjects shared)
...
+				ScreenMessages.PostScreenMessage("<color="+color+"><size="+size+">"+textToHud+"</size></color>", delay, ScreenMessageStyle.UPPER_RIGHT);
+				}
+			else if (style == 4)
+				{
+				ScreenMessages.PostScreenMessage("<color="+color+"><size="+size+">"+textToHud+"</size></color>", delay, ScreenMessageStyle.LOWER_CENTER);
+				}
+			else 
+				{
+				ScreenMessages.PostScreenMessage("*"+textToHud, 3f, ScreenMessageStyle.UPPER_CENTER);
+				}
```
Dunbaratu: Note to self:  kerboscript could really use a named enum type so this doesn't have to be done as an integer.  But for now, it has to be this way.

pgodd:
I was about 24 hours into trying to figure out why my command wasn't being passed. I resorted to tracing the print function and seeing where it was different (aside from the obvious extended capabilities).
Suddenly when I added this line it started working.
But at that point I wasn't clear on what other changes I had made, and I took the fact it was working as an indication it was necessary (correlation bias) Originally, it just had the textToHud variable, but as I added in variables, I extended it. 
I can try nixing it and see if the results are the same. Won't take but a few minutes.

Dunbaratu:
The lines I was referring to were the ones above.  The use of 1 to mean UPPER_LEFT, 2 to mean UPPER_CENTER, and so on.

src/kOS/Function/Misc.cs
pgodd: Was the best of two ways. Originally I planned to convert the input as in named enum, but it would still be a decision tree. 
Looked into switch/case as well, but opinions were it would add overhead.

pgodd:
Lost me. I was just trying to avoid a line wrap in the command line. The method names are nice and self explanatory, but kOS processors have 1970's storage.

Dunbaratu:
Did you make any edits to the kRISC.tpg file?
https://github.com/KSP-KOS/KOS/blob/develop/src/kOS.Safe/Compilation/KS/kRISC.tpg

Because if you did, I don't see it in the pull request, but I assume you must have because you have changes to scanner.cs, parser.cs, and ParseTree.cs. which are the output of the parser generator.

(And if you edited those three files by hand and didn't generate them from kRISC.tpg, then wow.  That would be a lot of work.)

Incidentally, you don't really need to alter the parser at all if you just want people to call hudtxt as a function.

i.e. if you want to support this:
```myfunction(arg1, arg2, arg3, arg4, arg5).```
Then just the change you did to Function/Misc.cs is enough on its own.
The parser changes are to support if you want the syntax to be special and not look like a function call (even though it might turn into one when executed), for example:
```myfunction arg1, arg2, arg3, arg4, arg5.```
or
```myfunction arg1 arg2 arg3  arg4  arg5.```
or
```myfunction { arg1 # arg2 # arg3  # arg4  # arg5 }```
Or whatever other weird syntax you like.

Basically, if the syntax is the normal function call syntax:

* functionName left-paren comma-list-of-args right-paren

then you don't need to bother with the syntax file changes.

pgodd:
Aha! See I was on the right track because my original intent was to have the command line without the need for quotes wrapping the text. But it got messy when I discovered the screen messages would accept HTML tags. It just seemed easier to go back to the default. 
I did edit the parser files by hand at first. But then discovered the TinyPG method.
Incidentally, It should be clear you need a specific branch of TinyPG and not the original author version. 
I'm going to try it on this end, but if you think it's not necessary, by all means ditch it.
This makes adding things a lot easier.
Do you want the krisc file?

pgodd:
So, just checked and it seems all I need is the function call.
This is going to make future add ons a breeze.
I won't call it wasted time, since I learned tons from doing this.
Thanks again!

Dunbaratu:
You are right that there needs to be some sort of "for programmers" documentation too.

With things like "Step 1: Before you make any changes, try to just compile your own copy of the plugin DLL's as they are to make sure you've set up the environment right, and here's the steps you'll need to follow to do so...."

The commentary about our forked branch of TinyPG would go there.

teleksterling:
@Dunbaratu, I know *I'd* find that really helpful.

erendrake:
@pgodd Do you want to put up a new PR with the changes that were discussed?

pgodd:
I'd be happy to.
I'll get to it later today.
The changed code is so much simpler thanks to Dunbaratu's advice.

pgodd:
closed for repairs.
New version in new pull request.

PR #560, opened 2015-02-13
HudTxt function

pgodd:
Adds function to allow text to be displayed on screen via kOS script.

command format hudtxt ( "text"+var, int delay, int style, int size, "color", bool(true/false) echo in terminal).
e.g.: hudtxt ("Hello Kerbin!, I'm "+round(alt:radar)+"m above you", 2, 3, 25, "red", 0).

in addition you can enter some HTML tags in the text field. I haven't explored this one much <i>"Hello world"</i> will produce italics.
Delay = seconds in integers only, the amount of time the message will be displayed. New messages within this time will cause previous message to scroll up.
Style = choice of styles between 1 and 4 ~ 1=Upper_left 2=Upper_center 3=Upper_right 4=Lower_center The positions of upper right is actually lower right... comes from KSP code. If an improper choice is made i.e.:5 then defaults and prefixes with *.
Size = ? not really sure what this is in terms of how big the text is. 15 will get you started on any. 150 is huge. Effect varies depending on style.
Colour = you can enter an rgb colour value in the form rgb(0.002,0.768,0.768) (this is a nice turquoise)
Colour values are floats between 0 and 1, per Unity. Take your normal RGB numbers and divide by 255 to get the float. 
Echo = echo's the on screen display in the terminal false=no true=yes

### DIFF
--- a/src/kOS/Function/Misc.cs
+++ b/src/kOS/Function/Misc.cs
@@ -28,7 +28,7 @@ namespace kOS.Function
         }
     }
 
-[Function("hudtxt")]
+[Function("hudtext")]
     public class FunctionHudTxt : FunctionBase
     {
 
--- a/src/kOS/Function/Misc.cs
+++ b/src/kOS/Function/Misc.cs
@@ -27,52 +27,40 @@ namespace kOS.Function
             shared.Screen.Print(textToPrint);
         }
     }
-
-[Function("hudtext")]
+    
+    [Function("hudtext")]
     public class FunctionHudText : FunctionBase
     {
-
-        public override void Execute(SharedObjects shared)
+        public override void Execute (SharedObjects shared)
 
         {
-            int mirror          = Convert.ToInt32(shared.Cpu.PopValue());
-            string color        = shared.Cpu.PopValue().ToString();
-            int size            = Convert.ToInt32(shared.Cpu.PopValue());   
-            int style           = Convert.ToInt32(shared.Cpu.PopValue());
-            int delay           = Convert.ToInt32(shared.Cpu.PopValue());   
-            string textToHud    = shared.Cpu.PopValue().ToString();
-
-            if (style == 1)
-                {
-                ScreenMessages.PostScreenMessage("<color="+color+"><size="+size+">"+textToHud+"</size></color>", delay, ScreenMessageStyle.UPPER_LEFT);
-                }
-            else if (style == 2)
-                {
-                ScreenMessages.PostScreenMessage("<color="+color+"><size="+size+">"+textToHud+"</size></color>", delay, ScreenMessageStyle.UPPER_CENTER);
-                }
-            else if (style == 3)
-                {
-                ScreenMessages.PostScreenMessage("<color="+color+"><size="+size+">"+textToHud+"</size></color>", delay, ScreenMessageStyle.UPPER_RIGHT);
-                }
-            else if (style == 4)
-                {
-                ScreenMessages.PostScreenMessage("<color="+color+"><size="+size+">"+textToHud+"</size></color>", delay, ScreenMessageStyle.LOWER_CENTER);
-                }
-            else 
-                {
-                ScreenMessages.PostScreenMessage("*"+textToHud, 3f, ScreenMessageStyle.UPPER_CENTER);
-                }
-            if (mirror == 1) 
-                {
+            bool      echo      = Convert.ToBoolean(shared.Cpu.PopValue());
+            RgbaColor rgba      = GetRgba(shared.Cpu.PopValue());
+            int       size      = Convert.ToInt32 (shared.Cpu.PopValue ());    
+            int       style     = Convert.ToInt32 (shared.Cpu.PopValue ());
+            int       delay     = Convert.ToInt32 (shared.Cpu.PopValue ());   
+            string    textToHud = shared.Cpu.PopValue ().ToString ();
+            string   htmlColour = rgba.ToHTMLString();
+            {                            
+            if (style == 1) {
+                ScreenMessages.PostScreenMessage("<color=" + htmlColour + "><size=" + size + ">" + textToHud + "</size></color>",delay,ScreenMessageStyle.UPPER_LEFT);
+            } else if (style == 2) {
+                ScreenMessages.PostScreenMessage("<color=" + htmlColour + "><size=" + size + ">" + textToHud + "</size></color>",delay,ScreenMessageStyle.UPPER_CENTER);
+            } else if (style == 3) {
+                ScreenMessages.PostScreenMessage("<color=" + htmlColour + "><size=" + size + ">" + textToHud + "</size></color>",delay,ScreenMessageStyle.UPPER_RIGHT);
+            } else if (style == 4) {
+                ScreenMessages.PostScreenMessage("<color=" + htmlColour + "><size=" + size + ">" + textToHud + "</size></color>",delay,ScreenMessageStyle.LOWER_CENTER);
+            } else {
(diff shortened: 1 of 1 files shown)


src/kOS/Function/Misc.cs
```diff
@@ -28,6 +28,51 @@ public override void Execute(SharedObjects shared)
...
 
+[Function("hudtxt")]
+    public class FunctionHudTxt : FunctionBase
+    {
+
+        public override void Execute(SharedObjects shared)
+
+        {
+            int mirror          = Convert.ToInt32(shared.Cpu.PopValue());
+            string color        = shared.Cpu.PopValue().ToString();
```
Dunbaratu: I'd prefer this to use the already-existing kOS RGBA color type, for proper coherency in the mod.

i.e.  HUDTXT( "pink text",  6, 1, 2, RGB(1,0.7,0.7), false).
or.  HUDTXT( "pink text",  6, 1, 2, red, false).  // no quotes around 'red' because it's already a built-in term in kOS.

(In which case, you'd be expecting to pop an object of type kOS.Suffixed.RgbaColor from the stack here.)

Dunbaratu: To make this work (since you're using HTML color tags to communicate this), I could provide you with a ToHTMLColor() method from RgbaColor that, for example, if you have RGB(1.0,0.5,0.5) would out put a string like "#ff8080".

Dunbaratu: Note, none of these ideas should prevent the Pull Request.  They're just ideas for a further refinement prior to a release if you like.  (They would have to be done prior to release or not at all, though, otherwise we have backward compatibility problems)

src/kOS/Function/Misc.cs
```diff
@@ -28,6 +28,51 @@ public override void Execute(SharedObjects shared)
         }
     }
 
+[Function("hudtxt")]
```
Dunbaratu: Would you be adverse to calling it "hudtext" (spell out the 'e')?

pgodd:
I like it.
It makes sense and promotes flow in writing scripts.
I might have to lean on your help for this, I'm such pathetic noob.

pgodd:
I prefer hudtext. I changed it early on when I ran into a naming conflict in development. It just stuck.

pgodd:
So, a couple of thoughts.
Firstly, the HTML tag compatibility is inherited from the postScreenMessage function. I know it also responds to the italics tag, but I don't know what else. 
I struggled with the concept of saddling a user with things like the HUD: prefix hardcoded into the mirror to terminal option. I injected the HUD: prefix as a bit of "branding", perhaps self aggrandizing...Just so you know how I understand how some choices are more than simply objective. You guys have really respected other authors with regards to how you extend their plugins etc.
I thought about maybe giving the hudtext function a distinct colour. Originally I thought about limiting it to one screen placement. But as you learn about the extent of the API you start to rationalize more and more freedom. 
What I am getting at is: if we are deciding now and forever what functionality hudtext is going to have, maybe I should invest a few hours to see what's possible. Can we change fonts? Can we underline? Can we change the background cold? Can we change alignment? Can we superscript or subscript? do we really care?
There is a lot to suggest a decision needs to be made to either lock down functions, add the RGBA, or leave it open to whatever <tags> it supports.

src/kOS/Function/Misc.cs
pgodd: I prefer hudtext. I changed it early on when I ran into a naming conflict in development. It just stuck.

src/kOS/Function/Misc.cs
pgodd: I like it.
It makes sense and promotes flow in writing scripts.
I might have to lean on your help for this, I'm such pathetic noob.
So, a couple of thoughts.
Firstly, the HTML tag compatibility is inherited from the postScreenMessage function. I know it also responds to the italics tag, but I don't know what else. 
I struggled with the concept of saddling a user with things like the HUD: prefix hardcoded into the mirror to terminal option. I injected the HUD: prefix as a bit of "branding", perhaps self aggrandizing...Just so you know how I understand how some choices are more than simply objective. You guys have really respected other authors with regards to how you extend their plugins etc.
I thought about maybe giving the hudtext function a distinct colour. Originally I thought about limiting it to one screen placement. But as you learn about the extent of the API you start to rationalize more and more freedom. 
What I am getting at is: if we are deciding now and forever what functionality hudtext is going to have, maybe I should invest a few hours to see what's possible. Can we change fonts? Can we underline? Can we change the background cold? Can we change alignment? Can we superscript or subscript? do we really care?
There is a lot to suggest a decision needs to be made to either lock down functions, add the RGBA, or leave it open to whatever it supports.

Dunbaratu: well it's minor for me to make a RGBAColor.ToHTML() that does what I described, so I'll get on that in just a bit when I finish reading through a thing.  What you choose to do with it is up to you.

Dunbaratu:
I just provided you with the function I was talking about earlier:

RgbaColor.ToHTMLString().

You can get it by taking pull request #562 and bringing it into your own private branch to work with.  It should indirectly end up in the main develop then when we merge your stuff, if you've used it.

src/kOS/Function/Misc.cs
erendrake: @pgodd We dont have to decide anything is always and forever a single way. we can revisit this if we find a need or a better way :)

pgodd: Thanks, I'm figuring this out. ;)

pgodd:
Finally got the time to do this. The Unity colour convention really threw me initially. New version is up.

Dunbaratu:
One of the rules we operate under is that we don't take a pull request until the docs for it are written (so basically the same PR contains both the code and the changes to user docs (if any) that go with it.  That way both the code and its user docs are either rejected or accepted as a set.  (Hopefully that way the docs stay correct.)

For this, I'll add the doc pages to the pull request for you as I merge it in, since they're not well explained (we need docs for the docs - explaining how to change the docs).

src/kOS/Function/Misc.cs
```diff
@@ -27,7 +27,40 @@ public override void Execute(SharedObjects shared)
...
+                ScreenMessages.PostScreenMessage("<color=" + htmlColour + "><size=" + size + ">" + textToHud + "</size></color>",delay,ScreenMessageStyle.UPPER_LEFT);
+            } else if (style == 2) {
+                ScreenMessages.PostScreenMessage("<color=" + htmlColour + "><size=" + size + ">" + textToHud + "</size></color>",delay,ScreenMessageStyle.UPPER_CENTER);
+            } else if (style == 3) {
+                ScreenMessages.PostScreenMessage("<color=" + htmlColour + "><size=" + size + ">" + textToHud + "</size></color>",delay,ScreenMessageStyle.UPPER_RIGHT);
+            } else if (style == 4) {
+                ScreenMessages.PostScreenMessage("<color=" + htmlColour + "><size=" + size + ">" + textToHud + "</size></color>",delay,ScreenMessageStyle.LOWER_CENTER);
+            } else {
+                ScreenMessages.PostScreenMessage("*" + textToHud, 3f, ScreenMessageStyle.UPPER_CENTER);
+            }
```
Dunbaratu: As I merge this in, I'm changing the above to a switch statement instead of an if-else ladder.

I originally started changing it to the indentation style the rest of the code uses (which I really hate, by the way.  I prefer the type you use for exactly this reason - it makes if/else ladders waste less screen real estate).

But then I realized as I was doing that that a simple switch with case 1: case 2 and s on works here too.

PR #561, opened 2015-02-14
Telnet server

Dunbaratu:
Here's my checklist as I was working on this.  You can use it as a test check list when you try things out.

Telnet features work-list:

- [done] bridge between (threaded) socket code and (unthreaded) Unity, via putting chars into input and output queues.
- [done] Welcome menu that is a Unity Monobehavior and proves the above works.
- [done] Get attached to a kOC CPU from the Welcome menu.
- [done] Able to detach from kOS CPU and return to Welcome menu.
- [done] Resize from client to server. 
- [done] Resize from server to client.
- [done] Input: simple ascii.
- [done] Terminal Make & Model abstraction infrastructure, input side.
- [done-ish] Terminal Make & Model abstraction infrastructure, output side.
- [done] Input fancy chars, VT100 / XTERM only for now.
- [done] Display telnet connection status in main GUI Terminal window's lower area (icon, count of how many).
- Disconnect telnet on these activities - try to give different messages for why:
    - [done] CPU poweroff. (effect was interesting - terminal stays connected but nothing happens when you type.  I think that's actually appropriate.)
    - [done] CPU part destructed because of scene switch.
    - [done] Camera moves too far from CPU so the CPU unloads from memory.
    - [done] CPU part destructed because of lithobreaking or thermal exhuberance.
- [goodenough] output "diffing" system. (still could be made more efficient but it works)
- [not my fault] PageUp and PageDn seem to leave the terminal messed up.  Fix it.  (update: I now think this problem always was there.  If you page-up and then page-down a few times, it ends up putting the cursor in the middle of your scrollback buffer when you type instead of teleporting you down to the bottom where the cursor belongs.  This happens in GUI mode too and should be a separate issue.
- [done] Telnet status visible on applauncher control panel.
- [done] Allow non-loopback connection (with opt-in only)
- [done] Get it working with a passive-mode telnet client. (all my early testing was with putty's default active mode.  The difference is in which side of the connection initiates setting negotiations and which side responds to them.).
- [done] Use C# "lock" keyword on inQueue/outQueue to assure multichar control sequences aren't enqueued until the whole string is there.
- [done] More than one telnet client both attached to the same kOS CPU - suppress the resize message looping.
- [done] trap control-C and send it right.
- [ah forget it - it's not important enough] fix the terminal type detection (currently commented out in ContinuousChecks())
- [done] Opt-in warning message about enabling the feature the first time you turn it on.
- [done] Even bigger warning about non-loopback telnet.
- [done] Implement ctrl-A and ctrl-E as alternate ways to "HOME" and "END", following ancient traditions.
- [goodenough for PR] Documentation!!  : control-D for example, new terminal suffixes, etc.
------ Minimum for release, things above this line -------

----- Future ideas?  Pie in the sky? ------
- [] Revamp the display listers (i.e. LIST FILES. LIST CONFIG. etc) to read terminal width and change output accordingly.
- [] suffixes for TERMINAL to detect sessions - harder than it seemed at first - have to think about it more.  The way multiple sessions get cloned to one terminal makes it hard for the system to tell which terminal typed the command.  (i.e. if you PRINT TERMINAL:MODEL, which terminal issued that command when all of the typing is being funneled to the same point?  And if it's a script which of the many equal-stutus terminals connected to it is the "runner" of it?)
- [] Some kind of equivalent to doing the unix command "cat > filename.ks", so the telnet session can write out a program file on the host.
- [] Support color codes, and other atribute codes (underline, reverse, etc)
- [] Support terminals that report as ANSI.
- [] Support artificial RemoteTech delay on the terminal's processing of input or output.
- [] Support line-at-a-time buffered communication mode.  (for sending lines of text at once when there's a RemoteTech delay).


Dunbaratu:
When testing the mandatory opt-in notification messages, you may find that you want to reset the conditions back to what they would be for a new user again (bascially, to undo the "don't ask me again" selection so you can try other combinations).

If you want to un-set the "don't ask me again' option on the mod, you can do so by editing:

GameData/kOS/Plugins/PluginData/config.xml

And removing the lines:
```
    <bool name="PermanentListenPermission">1</bool>
    <bool name="PermanentRealIPPermission">1</bool>
```

Resources/GameData/kOS/Parts/kOSMachine1m/part.cfg
```diff
@@ -13,8 +13,8 @@ iconCenter = 0, 3, 0
 
 // --- node definitions ---
 node_attach = 0.0, 0.0, 0.0, 0.0, -1.0, 0.0
-node_stack_bottom = 0.0, -0.14, 0.0, 0.0, -1.0, 0.0
-node_stack_top = 0.0, 0.14, 0.0, 0.0, 1.0, 0.0
+node_stack_bottom = 0.0, -0.173, 0.0, 0.0, -1.0, 0.0
+node_stack_top = 0.0, 0.173, 0.0, 0.0, 1.0, 0.0
```
erendrake: this looks like a dupe from the other PR?

src/kOS/Screen/TermWindow.cs
```diff
@@ -184,13 +200,9 @@ void OnGUI()
 
             // Should probably make "gui screen name for my CPU part" into some sort of utility method:
             KOSNameTag partTag = shared.KSPPart.Modules.OfType<KOSNameTag>().FirstOrDefault();
```
erendrake: we arent using this partTag anymore?

src/kOS/UserIO/TelnetMainServer.cs
```diff
@@ -0,0 +1,449 @@
...
+            catch (Exception ex)
+            {
+                UnityEngine.Debug.LogError("kOS: Exception Loading TelnetMainServer.xml (maybe the first time you ran and its not there yet): " + ex.Message);
+            }
+            return; // In principle it should never reach here.
+        }
+        
+        public void SetConfigEnable(bool newVal)
+        {
+            bool isLoopback = (bindAddr == IPAddress.Loopback);
```
erendrake: this will always be a reference comparison rather than value. is that what you wanted?

src/kOS/UserIO/TelnetSingletonServer.cs
```diff
@@ -0,0 +1,1064 @@
...
+        /// </summary>
+        private bool flushPendingInput;
+                        
+        /// <summary>
+        /// What order was I launched in?  Was I the 1st TelnetSingletonServer? The 2nd?  The 423rd?
+        /// </summary>
+        private int MySpawnOrder { get; set; }
+        
+        private bool isLineAtATime;
+        private bool allowResize;
```
erendrake: we assign to this but never use it?

src/kOS/UserIO/TelnetSingletonServer.cs
```diff
@@ -0,0 +1,1064 @@
...
+        }
+
+        
+        /// <summary>
+        /// Whenever an IAC WONT command is seen (see RFC 854), this reads through it and understands some parts of it.
+        /// </summary>
+        /// <param name="remainingBuff"> the buffer of raw stuff</param>
+        /// <param name="index">the offset into the buffer where the RFC654_WILL byte started</param>
+        /// <returns>how many bytes of the buffer should get skipped over because I dealt with them.</returns>
+        private int TelnetConsumeWont( byte[] remainingBuff, int index)
```
erendrake: never used?

Resources/GameData/kOS/Parts/kOSMachine1m/part.cfg
Dunbaratu: except that if it shows as a diff that means its not in the develop, doesn't it?  did the other pr not get merged right?  at any rate 0.173 is the right number.  Ive been using it a while now.

src/kOS/Screen/TermWindow.cs
Dunbaratu: Corrent . it can be removed.  Cut and Paste error.

src/kOS/UserIO/TelnetMainServer.cs
Dunbaratu: IPAddress.Loopback is a static.  when using the loopback address, I assign bindAddr to IPAddress.Loopback.  In principle I think that should work.  Then again... if its a property then who knows what it does under the hood.  I *assumed* it returned a reference to a static global thing because that's how I'd have designed it.  But yeah to be safe it can be changed.  Get the ToString() of each and compare them as strings and that should give a value equals.

src/kOS/UserIO/TelnetSingletonServer.cs
Dunbaratu: I had an idea of making one window slaved to the other so as to avoid the resize arguments between them but it turns out you cannot enforce that over telnet anyway.  If the client wants to resize itself then tough, you have to let it.  This can be removed.

src/kOS/UserIO/TelnetSingletonServer.cs
Dunbaratu: Not yet but please don't delete it.  it's part of the protocol that I want to be in place to use later if and when we expand it.  The DO/DONT and WILL/WONT are parts of the RFC's for telnet and I just happen to have not made use of any of the options that have the server side expect a "wont" message from the client.  But if we expand this for things line line-at-a-time mode, then I'll need this and I wanted it here when I that part of the code and how to write it was still fresh in my head.

Dunbaratu: good catch.  It was meant to be called from above.  (I added a line note above).

src/kOS/UserIO/TelnetSingletonServer.cs
```diff
@@ -0,0 +1,1064 @@
...
+                        byte commandByte = inRawBuff[++rawIndex];
+                        switch (commandByte)
+                        {
+                            case RFC854_DO:
+                                rawIndex += TelnetConsumeDo(inRawBuff, rawIndex);
+                                break;
+                            case RFC854_DONT:
+                                rawIndex += TelnetConsumeDont(inRawBuff, rawIndex);
+                                break;
+                            case RFC854_WILL:
```
Dunbaratu: Should have also added this case:
```
case RGC854_WONT:
   rawIndex += TelnetConsumeWont(inWarBuff, rawIndex);
   break;
```

PR #572, opened 2015-02-16
fixes #72 and resolves #548

### LINKED ISSUE #72: (regression) Event handler fails to stage twice
This event handler successfully separates the spent stage but fails to ignite the 2nd stage. I can ignite the 2nd stage by pressing the space key. Works with "uploads_2014_01_KOS.zip".
```
// event handler for staging
when stage:liquidfuel < 0.1 then { 
    print "T+" + round(missiontime) + " Stage 1 separation.". 
    stage. 
    wait 1. 
    print "T+" + round(missiontime) + " Stage 2 ignition.". 
    stage. 
}
```

### LINKED ISSUE #548: If SHIP is not the same as ActiveVessel, then executing STAGE stages the wrong vessel.
User Motokid600 in the KSP forums was having trouble getting a script to stage properly when trying to observe the vessel from the vantage point of another vessel nearby on the ground.

By eyeballing the code I think this is definitely a bug, and the culprit is here, in Function/Misc.cs:

```
    [Function("stage")]
    public class FunctionStage : FunctionBase
    {
        public override void Execute(SharedObjects shared)
        {
            Staging.ActivateNextStage();
        }
    }
```

```Staging.ActivateNextStage()``` comes from the KSP stock API itself, and assumes you're trying to operate on FlightGlobals.ActiveVessel. 

I think this line of code is very old and dates all the way back to the original Nivekk code (from before kOS was being careful to separate the notions of active vessel from cpu vessel).

I won't be able to fix this myself for a while, as I'm knee-deep in telnet for the time being, but I'm noting it and moving on.  If anyone else wants to take a stab at fixing it, be my guest.

### PR DESCRIPTION
**Pull request by erendrake**
this PR gives us two new suffixes

* STAGE:READY
* STAGE:NUMBER

this will let someone who wants to know when they can stage and what stage number they are on the ability to do so.

This also resolves #548 by not letting you stage and having STAGE:READY always be false on craft other than the active craft. This is a Kerbal Space Program restriction :(

### DIFF
--- a/doc/source/structures/vessels/stage.rst
+++ b/doc/source/structures/vessels/stage.rst
@@ -1,0 +1,98 @@
+.. _stage:
+
+Stage
+=============
+
+*Contents*
+
+    - :global:`EXAMPLE`
+    - :global:`READY`
+    - :global:`RESOURCES`
+    - :struct:`Stage`
+
+A planned velocity change along an orbit. These are the nodes that you can set in the KSP user interface. Setting one through kOS will make it appear on the in-game map view, and creating one manually on the in-game map view will cause it to be visible to kOS.
+
+Access
+--------
+
+You access the current stage for the vessel the kOS core is attached to with the STAGE: command.
+
+.. global::EXAMPLE	 
+	
+	
+    A very simple auto-stager using :READY
+	
+	LIST ENGINES IN elist.
+
+	UNTIL false {
+	    PRINT "Stage: " + STAGE:NUMBER AT (0,0).
+		FOR e IN elist {
+			IF e:FLAMEOUT {
+				STAGE.
+				PRINT "STAGING!" AT (0,0).
+				
+				UNTIL STAGE:READY {	} 
+				
+				LIST ENGINES IN elist.
+				CLEARSCREEN.
+				BREAK.    
+			}
+		}
+	}
+
+.. global::NUMBER
+
+	Every craft has a current stage, and that stage is represented by a number, this is it!
+	
+.. global::RESOURCES
+    
+	
+	
+Structure
+---------
+
+.. structure:: Stage
+
+    .. list-table:: Members
+        :header-rows: 1
+        :widths: 1 1 1 2
+
+        * - Suffix
+          - Type (units)
+          - Access
+          - Description
+
+        * - :attr:`READY`
+          - bool
+          - Get only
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: doc/source/structures/vessels/stage.rst [hunk 1]
```diff
@@ -0,0 +1,91 @@
...
+	
+    Every craft has a current stage, and that stage is represented by a number, this is it!
+
+.. attribute:: Stage:Resources
+
+    :access: Get
+    :type: :struct:`List`
+
+    This is a collection of the available :struct:`Resource` for the current stage.
+
```
Dunbaratu: Somewhere in the above document file, it should probably mention how stage:ready is always false for the non-active vessel, and how the problem is imposed on us from stock KSP.

Dunbaratu: Maybe also mention it in the main command stuff where the "STAGE." command is mentioned.

Dunbaratu: Also, are we seting ourselves up for ambiguous parse confusion since STAGE is now both a command to run (STAGE.) and a built-in variable at the same time?

Dunbaratu:
I can pull this but I'd like a response on the question about the naming conventions.  Do you think it's okay to use the word "stage" to both be the name of a bound variable AND the name of a command?

```
STAGE.
PRINT STAGE:NUMBER.
```

doc/source/structures/vessels/stage.rst
erendrake: I agree that it should be mentioned somewhere other than the structure list. Not sure why i thought this was enough :P

The two meanings of STAGE can currently co-exist but you are right in saying that we could have a big problem later. Maybe we could add STAGE:ACTIVATENEXT or simply STAGE:NEXT and deprecate STAGE?

Dunbaratu: I dislike that because STAGE is so fundamental that deprecating it breaks everyone's scripts, not just a few people's.   But i see no way around it since we already have the problem now that there's a STAGE: object and a STAGE command.  I forgot that you didn't  add STAGE: - it was already there.

Dunbaratu:
Okay merging in.  I realized that the objection I had wasn't really caused by you anyway.  It was already present before this change because of things like STAGE:LIQUIDFUEL.  I still don't like it, but.... it's not a new problem, its an existing one.

PR #579, opened 2015-02-19
Fixes #555 by altering the new vessel check logic.

### LINKED ISSUE #555: Error detaching a kOS-enabled part from a vessel
Full details in [forum thread](http://forum.kerbalspaceprogram.com/threads/68089-0-90-kOS-Scriptable-Autopilot-System-v0-15-5-2015-1-3?p=1711892&viewfull=1#post1711892).

Basically, anytime something is decoupled and contains a kOS module it's impossible to run a script on it until you reload the craft

Tested on 0.90 w/ latest 0.15.5 release

Dunbaratu:
The old code had some wonky behavior where as long as the new
vessel being switched to was valid... it refused to do the switch.

Here's a screenshot of the previous wonky behavior:

The vessel before decoupling
[!http://i.imgur.com/nQILanS.png]

The vessel after decoupling:  The terminal with the errors on it is the one attached to the lower half of the ship.
[!http://i.imgur.com/sXtBcyj.png]

The same test works fine now with the new change, and I tested the that controls do in fact try to control the new ship( not the old ship), by adding a bunch of overpowered torque wheel units and proving that I can yaw and pitch the new detached ship with SHIP:CONTROL:YAW or SHIP:CONTROL:PITCH and it won't try to yaw and pitch the old ship like it used to.

### DIFF
--- a/src/kOS/Binding/FlightControlManager.cs
+++ b/src/kOS/Binding/FlightControlManager.cs
@@ -71,18 +71,33 @@ namespace kOS.Binding
         {
             UnbindUnloaded();
 
-            if (currentVessel.id == Shared.Vessel.id) return;
+            // Why the "currentVessel != null checks?
+            //   Because of a timing issue where it can still be set to null during the span of one single
+            //   update if the new vessel isn't valid and set up yet when the old vessel connection got
+            //   broken off.
+            //
+            if (currentVessel != null && currentVessel.id == Shared.Vessel.id) return;
 
-            // Try to re-establish connection to vessel
-            if (VesselIsValid(currentVessel))
+            // If it gets this far, that means the part the kOSProcessor module is inside of
+            // got disconnected from its original vessel and became a member
+            // of a new child vessel, either do to undocking, decoupling, or breakage.
+
+            // currentVessel is now a stale reference to the vessel this manager used to be a member of,
+            // while Shared.Vessel is the new vessel it is now contained in.
+
+            // Before updating currentVessel, use it to break connection from the old vessel,
+            // so this this stops trying to pilot the vessel it's not attached to anymore:
+            if (currentVessel != null && VesselIsValid(currentVessel))
             {
                 currentVessel.OnPreAutopilotUpdate -= OnFlyByWire;
                 currentVessel = null;
             }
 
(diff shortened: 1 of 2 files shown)


src/kOS/Binding/FlightControlManager.cs
```diff
@@ -71,18 +71,33 @@ public override void Update()
...
+            // If it gets this far, that means the part the kOSProcessor module is inside of
+            // got disconnected from its original vessel and became a member
+            // of a new child vessel, either do to undocking, decoupling, or breakage.
+
+            // currentVessel is now a stale reference to the vessel this manager used to be a member of,
+            // while Shared.Vessel is the new vessel it is now contained in.
+
+            // Before updating currentVessel, use it to break connection from the old vessel,
+            // so this this stops trying to pilot the vessel it's not attached to anymore:
+            if (currentVessel != null && VesselIsValid(currentVessel))
```
Dunbaratu: I added the verbose comment because I was trying to explain to *myself* what was really going on here.

src/kOS/Binding/FlightControlManager.cs
```diff
@@ -71,18 +71,33 @@ public override void Update()
...
             {
                 currentVessel.OnPreAutopilotUpdate -= OnFlyByWire;
                 currentVessel = null;
             }
 
-            if (VesselIsValid(Shared.Vessel)) return;
-
-            currentVessel = Shared.Vessel;
+            // If the new vessel isn't ready for it, then don't attach to it yet (wait for a future update):
+            if (! VesselIsValid(Shared.Vessel)) return;
```
Dunbaratu: The flipping of the Boolean meaning with '!' was really the only major edit I did.  Before it would only assign to the new vessel when the new vessel was NOT valid, which seemed backward.

This cascaded into the possiblity of the above "currentVessel = null;" being executed even though the new assignment to currentVessel = Shared.Vessel gets skipped.  I added the extra 'currentVessel != null' checks above just in case that case ever crops up.  In principle I don't think it can (Because Shared.Vessel would be invalid only if it doesn't exist or has no root part.  If it has no root part, how on earth can it contain this part inside it?)

Issue #568, opened 2015-02-16
Reboot command not clean out old states before boot


hvacengi:
When you issue a reboot command, it translates to the Cpu.Boot method, and nothing is done to clean out old variable states besides what is done in the normal boot initialization.  This can cause some issues, but the most noticeable one I've experienced is that `lock steering to` commands that were still active when the reboot command was issues will persist.  If steering is locked to a dynamic direction (like prograde) the direction does not continue to update, but the vessel steers towards where the direction pointed when reboot was issued (so for the example of prograde, it will slowly pitch up from the horizon as you move around the planet).  Manually pointing to another direction works temporarily, but as soon as the controls are released, the vessel snaps back to the direction.

I was thinking that we might want to add a shutdown and/or reboot method to the cpu which is dedicated to disposing of the bindings and managers, and any thing else that ought to be cleaned out for a truly fresh cpu boot.

Issue #362, opened 2014-11-11
Unlock throttle not working?


erendrake:
We have a report on the forums that i havent reproduced yet, I have a feeling that it might be related to the lock dictionary being case sensitive and i havent spotted any place we force case in the LockCollection class. 

That is of course if this is real at all.

Issue #359, opened 2014-11-11
kOS refuses to let go on CPU poweroff.


Dunbaratu:
To recreate:

Make a program that does this at any point during a flight:
```
LOCK STEERING TO UP.
wait 1000.
```

Then run it, let the steering settle to its new orientation, then power off the CPU itself via the new applauncher panel, and turn it back on again.

Now when you try to manually steer the ship you can do it a little bit but you're fighting a force that wants to recenter you.  When you let go of the controls, the ship rotates back to the steering setting it had when kOS module was still on.  There appears to be NO way to undo this.  ```lock steering``` and ```unlock steering``` don't help.  Powering it off and on again does not help.

I discovered it via the applauncher, but I suspect the problem is connected more to the new changes to the steering system and would also happen if you rightclicked the part and powered it down.

austinjames314:
Aha! This bug has absolutely doing my head in. I thought I was doing something wrong, but it turns out it's a bug all along.
 Please fix this! I basically can't get anything to work, because as soon as the kOS module loses power on the night side, it's basically broken from then on. Then because steering wont respond, I can't point the solar panels at the sun, and things go downhill from there.

Dunbaratu:
Is this still happening in 0.15.5?  I know piloting went through a bit of a redesign but if this is still a problem then this really needs to be dealt with.

austinjames314:
I just triple checked: Yes it's 0.15.5. Downloaded the new version a few hours ago.

I'm 99% sure it's the same thing. The difference to the steps in the first post above are, that I'm powering off either due to running out of electric charge, or via the right click menu of the part. (If it matters, it's a probe core that I added the kOS Module to, due to the kOS parts not staying attached).

I've tried 'UNLOCK STEERING.' a few times to no effect. Saving and re-loading seems to fix the problem up until the next power-off.

austinjames314:
While futzing around trying to get my probe to rotate 90 degrees, I think I just managed to reproduce this without losing power or shutting down (Oh yeah, I think the 'SHUTDOWN' command also triggers this bug. I don't know at what point it started though, so I'm sorry to say I can't say what I did to cause it. I'll post more if I notice anything specific.

go1dfish:
I've been running into this sort of behavior on a script that lock's steering to a variable at startup and changes that variable throughout the flight to new direction instances created from HEADING(..) or NODE:DELTAV:VEC:NORMALIZED:DIRECTION

[...]

Issue #172, opened 2014-06-29
Lock states persist through power cycling unit, become default for unlocked state


mileshatem:
For example, if you lock the throttle to 1, then toggle the kOS unit power off the throttle will remain on full (which by it self is fine in my opinion). However, then if you toggle the power back on the engine still remains on full (again, still fine), but the lock state seems to be reset to thinking it's 0, but not 0....

If you do an unlock all in that state, the engine is still on. If you lock throttle to 0 it shuts the engine off, but then if you unlock all the engine goes back to full on. 

Only way I've found to fix it is to lock it to 0, then cycle the power again.

PR #583, opened 2015-02-19
when we are changing power states to off or starved, we should unbind from steering

erendrake:
This looks like it might fix a slew of bugs in https://github.com/KSP-KOS/KOS/milestones/0.16%20punchlist

#172 
#359 
#362?

### DIFF
--- a/src/kOS/Function/Misc.cs
+++ b/src/kOS/Function/Misc.cs
@@ -278,7 +278,11 @@ namespace kOS.Function
     {
         public override void Execute(SharedObjects shared)
         {
-            if (shared.Cpu != null) shared.Cpu.Boot();
+            if (shared.Processor != null)
+            {
+                shared.Processor.SetMode(ProcessorModes.OFF);
+                shared.Processor.SetMode(ProcessorModes.READY);
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


erendrake:
this will need some testing because i did this on my lunch break and I didnt want to open KSP at my desk :)

hvacengi:
I will download and test this patch to see how it works.  If we can implement it so it works with the reboot command as well, I think that it would fix #568 as well.  Possibly add the same line before line 260 in kOS/Function/Misc.cs.  I'll copy it there too and see if it works.  It's a small enough addition that I don't think I want to spend the time figuring out how to make a pull request to your pull request.

erendrake:
@hvacengi its a bit of work to set me up as a remote, once you do that it isnt very hard to setup a pull request.

I actually think that the change to the reboot command i would like to see is two "setmode" commands

something like 

      [Function("reboot")]
    public class FunctionReboot : FunctionBase
    {
        public override void Execute(SharedObjects shared)
        {
            if (shared.Processor != null)
            {
                shared.Processor.SetMode(ProcessorModes.OFF);
                shared.Processor.SetMode(ProcessorModes.READY);
            }
        }
    }

Dunbaratu:
I'm testing this now.  I have to launch a new vessel to test it because all mine are currently too good at being recharged.  They have solar panels sticking out everywhere and thus never lose power.

hvacengi:
@Dunbaratu just disable the resource transfer and retract the panels ;-)

It worked for both starving and toggling power in my testing.  But it appears to have broken the reboot command.  I'm looking into that now.

Also, it revealed an issue with the boot scripts that I didn't account for when I made my previous modifications.  Apparently the boot script is still registered when you toggle the power even after it is deleted.  I'll address that in a separate pull request though.

Dunbaratu:
I usually use the un-retractable single-panel solar panels all over the craft facing every which direction specificially so there is no way to make it lose power.

PR #587, opened 2015-02-20
Rt terminal access for local controlled craft

erendrake:
Also, changed everything from RT2 to RT, this will force the default setting on people because honestly i dont really want to figure out how to fix it for the one guy im about to piss off :)

### DIFF
--- a/src/kOS.Safe/Encapsulation/IConfig.cs
+++ b/src/kOS.Safe/Encapsulation/IConfig.cs
@@ -5,7 +5,7 @@ namespace kOS.Safe.Encapsulation
         int InstructionsPerUpdate { get; set; }
         bool UseCompressedPersistence { get; set; }
         bool ShowStatistics { get; set; }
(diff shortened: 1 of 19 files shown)


hvacengi:
Do you just mean that it will force the default setting onto some one because it won't fall back to the old config key for EnableRT2Integration, or that it will be impossible to disable the RT integration?  It looks to me like it should just fall back to the default, which should be perfectly acceptable.  Just make a note to include it in the change log so that one guy has a chance to figure it out :-)

Dunbaratu:
It just means it will "forget" the setting from your config.xml file when you update and you'll have to re-set it.  (because it's under a different name now).  Not a big deal -just remember to mention it in the changelog.)

erendrake:
forgot to change all of the instances of 'RemoteTech2' along with the change from 'RT2' to 'RT'

erendrake:
sorry i meant reassert, not force. steven has it right

PR #589, opened 2015-02-20
Add support for switching to default volume on mode change

hvacengi:
kOSProcessor.cs - added logic to switch to the default volume if the cpu
is being turned on.  Also removed the check for STARVED or OFF modes
before calling Cpu.Boot, since one of those two conditions must be true.

### DIFF
--- a/src/kOS/Module/kOSProcessor.cs
+++ b/src/kOS/Module/kOSProcessor.cs
@@ -528,8 +528,15 @@ namespace kOS.Module
                 switch (newProcessorMode)
                 {
                     case ProcessorModes.READY:
-                        if ((ProcessorMode == ProcessorModes.STARVED || ProcessorMode == ProcessorModes.OFF) && shared.Cpu != null)
-                            shared.Cpu.Boot();
+                        if (Config.Instance.StartOnArchive)
+                        {
+                            shared.VolumeMgr.SwitchTo(shared.VolumeMgr.GetVolume(0));
+                        }
+                        else
+                        {
+                            shared.VolumeMgr.SwitchTo(HardDisk);
+                        }
+                        if (shared.Cpu != null) shared.Cpu.Boot();
                         if (shared.Interpreter != null) shared.Interpreter.SetInputLock(false);
                         if (shared.Window != null) shared.Window.IsPowered = true;
                         break;


src/kOS/Module/kOSProcessor.cs
```diff
@@ -528,8 +528,15 @@ public void SetMode(ProcessorModes newProcessorMode)
                 switch (newProcessorMode)
                 {
                     case ProcessorModes.READY:
-                        if ((ProcessorMode == ProcessorModes.STARVED || ProcessorMode == ProcessorModes.OFF) && shared.Cpu != null)
-                            shared.Cpu.Boot();
+                        if (Config.Instance.StartOnArchive)
```
Dunbaratu: The removal of :
```
if ((ProcessorMode == ProcessorModes.STARVED || ProcessorMode == ProcessorModes.OFF) && shared.Cpu != null)
```
Is very wrong.  If one part of the code calls SetMode(ProcessorModes.READY) and then another part of the code calls it again unnecessarily, you don't want that to cause a reboot.  It should only reboot when it wasn't already in the READY setting to begin with.

You could change it to check for ```if ((ProcessorMode != ProcessorModes.READY)...etc``` instead, so it really does work into the future if more states are added, but I don't like the idea that setting it to READY when it was already READY causes a reboot.

erendrake: We actually already check to make sure it has changed with https://github.com/hvacengi/KOS/blob/update-setmode/src/kOS/Module/kOSProcessor.cs#L526 if we get a READY call followed by another READY call it will just be skipped.

hvacengi: Unless I'm reading the code wrong, the explanation that @erendrake provides is why I removed it.  We only reach the switch if the ProcessorMode != newProcessorMode, and therefore newProcessorMode must be either STARVED or OFF by definition.  Realistically, the extra check won't have a significant effect on performance, but I just thought I would clear out a redundancy while I was adding the volume switch.

hvacengi: @erendrake looks like I may have missed an edit to your comment, or I failed to read it all the way through.  I still think that `if (ProcessorMode != ProcessorModes.READY)` is redundant.

The question I have is how do we want to future proof the code?  If it is more likely that a mode is added which does not require running the boot method (for instance a sleep mode) then the code should check against known transitions that do require it (thus reverting the line I edited).  If it is more likely that a mode is added which will still require the boot method to be called (like some kind of docked/undocked mode), then I think it would be best to leave it as it is in my commit.

I'm not sure if that makes any sense. I can re-word it tomorrow if needed.  I'm kind of tired right now.

I guess that means my vote is either
`if ((ProcessorMode == ProcessorModes.STARVED || ProcessorMode == ProcessorModes.OFF) && shared.Cpu != null)`
or
`if (shared.Cpu != null)`
Let me know which way you want to go with it, and I'll update the commit to whatever your choice is.  The part I thought was important was to switch back to the default volume (it's pretty important with boot scripts, since booting with the archive selected will launch boot.ks if it is stored in the script folder).

Dunbaratu: You're right . I missed the line above saying :
```if (ProcessorMode != newProcessorMode)```
It was in the bit of the github diff output that was folded away and hidden.

PR #592, opened 2015-02-20
Fixes #427 - easy one.

### LINKED ISSUE #427: Stack doesn't clear on exception termination
When an exception is thrown, causing execution to stop, the stack remains as it is..  When another command is issued after the exception happened, it lays its stack information on top of the leftover garbage was still there when the exception happened.

You don't notice much of an effect when running a program with an exception, because running a new program will clear the stack.  But you do notice when running commands interactively.  The effect.the user sees is that the second time an exception happens, when printing the stack trace to the screen it will claim, quite incorrectly, that the second command was called from the first one.  If the user keeps typing more commands at the interactive prompt and gets a third command with an exception, then the stack trace will claim the third command was called from the second which was called from the first, and so on. This is because the call stack information contained in the stack is left behind from each previous time there was an exception.

Dunbaratu:
I was expecting more problems but actually the one-line change
seems to be fine.

### DIFF
--- a/src/kOS/AddOns/RemoteTech/RemoteTechHook.cs
+++ b/src/kOS/AddOns/RemoteTech/RemoteTechHook.cs
@@ -44,7 +44,7 @@ namespace kOS.AddOns.RemoteTech
 
         public static bool IsAvailable()
         {
-            var integrationEnabled = Config.Instance.EnableRT2Integration;
+            var integrationEnabled = Config.Instance.EnableRTIntegration;
             if (!integrationEnabled)
             {
                 return false;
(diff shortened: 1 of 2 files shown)


src/kOS/AddOns/RemoteTech/RemoteTechHook.cs
```diff
@@ -44,7 +44,7 @@ public static bool IsAvailable(Guid vesselId)
 
         public static bool IsAvailable()
         {
-            var integrationEnabled = Config.Instance.EnableRT2Integration;
+            var integrationEnabled = Config.Instance.EnableRTIntegration;
```
Dunbaratu: Did this because it wasn't compiling.

erendrake: my bad :P

PR #606, opened 2015-02-22
fixes #605

### LINKED ISSUE #605: translation vector is wrong
SHIP:CONTROL:PILOTTRANSLATION and SHIP:CONTROL:TRANSLATION are in an inconsistent order and they are also out of step with the documentation. We need to reconcile them.

erendrake:
from: http://www.reddit.com/r/Kos/comments/2wq9aw/shipcontroltranslation_vector_order/


erendrake:
you are correct, this was a knee jerk thing and the docs do need to change instead.

PR #608, opened 2015-02-22
first draft of elements and highlighting

erendrake:
This is a double PR because its a bit of a try it out PR, we could merge it except there are no docs. 

    LIST ELEMENTS IN elist.

will give you a new list with all of the docked sections of your craft. pretty sure this will always have one element in it. From each of these you can get the parts contained in it, the vessel name, and the uid. More to come

    SET FOO TO HIGHLIGHT(elist[0], red). 

will highlight that element red

    SET FOO:ENABLED TO false. 

will disable the highlight

    SET FOO:COLOR TO green.

will change the color

    SET BAR TO HIGHLIGHT(elist[1]:PARTS[0], blue).

will work too

* * * 

there is more work to do I just wanted to give a preview and look for feedback. I can see a lot of use for this just like there is a ton for vecdraw.

### DIFF
--- a/src/kOS.Safe/Encapsulation/ListValue.cs
+++ b/src/kOS.Safe/Encapsulation/ListValue.cs
@@ -297,7 +297,7 @@ namespace kOS.Safe.Encapsulation
             AddSuffix("COPY", new NoArgsSuffix<ListValue>(() => new ListValue(this)));
         }
 
-        public static ListValue CreateList<T>(List<T> toCopy)
+        public static ListValue CreateList<T>(IList<T> toCopy)
         {
             return new ListValue(toCopy.Cast<object>());
         }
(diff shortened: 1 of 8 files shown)


Dunbaratu:
Isn't highlighting a part and grouping the vessel's parts into "elements" two separate concepts that can be independent of each other?

i.e. you could do SET GLOWER TO HIGHLIGHT(SHIP:PARTS[0], blue), without using elist at all.

erendrake:
they are two different concepts, I ended up wanting the highlighter for testing if the element stuff was working correctly so they ended up together, ill separate them back out when Im done I just wanted to discuss the direction I took 

    SET GLOWER TO HIGHLIGHT(SHIP:PARTS[0], blue).

Will surely work, any part, list of parts or element should work fine, so getting tagged parts or one of our other prebuilts should work fine too :)

PR #617, opened 2015-02-24
Added Fixedupdate Handler

erendrake:
This is ready to be merged

### DIFF
--- a/src/kOS.Safe/IUpdateObserver.cs
+++ b/src/kOS.Safe/IUpdateObserver.cs
@@ -1,6 +1,8 @@
+using System;
+
 namespace kOS.Safe
 {
-    public interface IUpdateObserver
+    public interface IUpdateObserver : IDisposable
     {
         void Update(double deltaTime);
     }
--- a/src/kOS.Safe/UpdateHandler.cs
+++ b/src/kOS.Safe/UpdateHandler.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Collections.Generic;
 
 namespace kOS.Safe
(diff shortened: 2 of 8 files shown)


src/kOS.Safe/UpdateHandler.cs
```diff
@@ -9,27 +8,22 @@ public class UpdateHandler
         // insert itself more than once into the observer list, it still only gets in the list
         // once and therefore only gets its Update() called once per update.
         // The value of the KeyValuePair, the int, is unused.
-        private readonly Dictionary<IUpdateObserver, int> observers = new Dictionary<IUpdateObserver, int>();
-        
+        private readonly HashSet<IUpdateObserver> observers = new HashSet<IUpdateObserver>();
+        private readonly HashSet<IUpdateObserver> fixedObservers = new HashSet<IUpdateObserver>();
```
erendrake: I am assuming you were using a dictionary before to Assert a unique constraint? A hashset should do the same without throwing an exception :)

src/kOS.Safe/IUpdateObserver.cs
```diff
@@ -1,7 +1,9 @@
+using System;
+
 namespace kOS.Safe
 {
-    public interface IUpdateObserver
+    public interface IUpdateObserver : IDisposable
```
erendrake: I added this interface to give anyone who implements it a chance to unregister.

src/kOS.Safe/UpdateHandler.cs
Dunbaratu: I was searching the .net docs for something like a hash and could't find it and gave up and used Dictionary.

Dunbaratu:
Looks good.  Now that this is here we can seriously consider what happens if we move CPU update to fixedupdate (or just try it and see what happens).

Once we do that step, we'll have to dig through a lot of the docs and make sure they stop talking about stuff happening during Update.

PR #625, opened 2015-02-28
Fixes 624

erendrake:


src/kOS.Safe/Encapsulation/Suffixes/SetSuffix.cs
```diff
@@ -14,7 +14,15 @@ public SetSuffix(SuffixGetDlg<TValue> getter, SuffixSetDlg<TValue> setter, strin
 
         public virtual void Set(object value)
         {
-            var toSet = (TValue)Convert.ChangeType(value, typeof(TValue));
```
erendrake: before we attempt to force the type change, We attempt a simple cast and if that works we dont have to go to the trouble of the convert.

PR #626, opened 2015-03-01
Hsva color

### LINKED ISSUE #619: Alternative color scheme HSV
with the docking vecdraw and the new highlights I have feel a need for a different color scale. I have experience with HSV and so i think ill implement it. its a cylindrical projection of the RGB scale 

image

http://en.wikipedia.org/wiki/HSL_and_HSV

sample algorithm 

http://www.splinter.com.au/converting-hsv-to-rgb-colour-using-c/

erendrake:
Resolves #619

### DIFF
--- a/doc/source/structures/misc/colors.rst
+++ b/doc/source/structures/misc/colors.rst
@@ -94,3 +94,60 @@ Examples::
 
     PRINT GREEN:HTML. // prints #00ff00
 
+.. function:: HSV(h,s,v)
+
+    This global function creates a color from hue, saturation and value::
+
+        SET myColor TO HSV(h,s,v).
+		
+	`More Information about HSV <http://en.wikipedia.org/wiki/HSL_and_HSV>`_,
+
+    where:
+
+    ``h``
+        A floating point number from 0.0 to 360.0 for the hue component.
+    ``s``
+        A floating point number from 0.0 to 1.0 for the saturation component.
+    ``v``
+        A floating point number from 0.0 to 1.0 for the value component.
+
+.. function:: HSVA(h,s,v,a)
+
+    Same as :func:`HSV()` but with an alpha (transparency) channel::
+
+        SET myColor TO HSVA(h,s,v,a).
+
+    ``h, s, v`` are the same as above.
+
+    ``a``
+        A floating point number from 0.0 to 1.0 for the alpha component. (1.0 means opaque, 0.0 means invisibly transparent).
+
+.. structure:: HSVA
+
+	The HSVA structure contains all of the suffixes from the RGBA structure in addition to these
+    .. list-table:: Members
+        :header-rows: 1
+        :widths: 2 1 4
+
+        * - Suffix
+          - Type
+          - Description
+
+        * - :H or :HUE
+          - scalar
+          - the hue component of the color. It is a value from 0.0 to 360.0
+        * - :S or :SATURATION
+          - scalar
+          - the saturation component of the color. It has a value from 0.0 to 1.0
+        * - :V or :VALUE
+          - scalar
+          - the value component of the color. It has a value from 0.0 to 1.0
+		  
+
+Examples::
+
+    SET myarrow TO VECDRAW.
+    SET myarrow:VEC to V(10,10,10).
+    SET myarrow:COLOR to HSV(60,1,1). // Yellow
+    SET myarrow:COLOR:S to 0.5. // Light yellow
+    SET myarrow:COLOR:H to 0. // pink
--- a/src/kOS/Function/Misc.cs
+++ b/src/kOS/Function/Misc.cs
@@ -44,7 +44,7 @@ namespace kOS.Function
             int       style     = Convert.ToInt32 (shared.Cpu.PopValue ());
             int       delay     = Convert.ToInt32 (shared.Cpu.PopValue ());   
             string    textToHud = shared.Cpu.PopValue ().ToString ();
-            string   htmlColour = rgba.ToHTMLString();
+            string   htmlColour = rgba.ToHexNotation();
             switch (style)
             {
                 case 1:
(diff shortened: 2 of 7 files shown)


Dunbaratu:
Given that the mapping is two-way, why do HSV's get to output RGB's but RGB's don't get to output HSV's?

That's my point about needing a common unified color class that is always used regardless of whether a color was created as RGB or HSV.  If you insist on making HSV a separate class, then at least make it an *equal* citizen of RGB by making them BOTH derive from a common shared base class or interface.  Then stuff like VectorRenderer can just say it expects a thing of that base type, and know that it will be possible to obtain some RGB values out of it, **or some HSV values out of it** regardless of exactly what kind of color it was made as.

i.e the base color type would have **6** suffixes, not 3.  It would have R,G,B, *AND* H,S.V.  It will be implemented one of two ways - either the R,G, and B are native and the H,S, and V are calculated, or the H,S, and V are native and the R, G, and B are calculated.  There can be two varieties of derived class - one that does it one way and the other that does it the opposite way.  But either way we'd have a common base type that we know does BOTH types in the same class so everything can be written to expect that type and it's all universal.

For exactly the same reason that a method that expects to work in the RGB world should be able to be passed in an HSV as a starting point instead, a method that expects to work in the HSV world should be able to be passed in an RGB as a starting point instead..

erendrake:
I dont believe there is a unified color class that will make all suffixes available all the time. 

There is a shared base class for all of the color projections, its the RGBAColor class. it has the raw coordinates on the Cartesian cube that we pass to the display. All of the other schemes are  projections i dont think that dumping all of them we decide to implement into one structure will lead to an easy to use color system. When you pick a projection you do so for a reason and i think it makes sense for it to be a deliberate choice.

I considered hiding the RGB in the HSV class but that didnt seem nice :)

Dunbaratu:
But what about making both RGB and HSV be derived classes of the base class?  You haven't commented on that one.

The red herring about "when you pick a projection you do so for a reason" has nothing to do with what I said.

I thought it would be obvious what I meant but apparently since you responded to a strawman of what I said, it wasn't.

**Base class**: - it's abstract.  You MUST derive from it to make something useful.  It would be what every other part of kOS expects to be given.  It has abstract properties R, G, B, H, S, V, and A.

**HSV Class** - It's derived from base class.  in this variant, H,S, and V are native numbers, and R,G, and B are calculated.

**RGB class** -It's derived from base class.  In this variant, R,G, and B are native numbers and H,S, and V are calculated.

Both derived classes can share the same A, alpha.

erendrake:
I am trying to tell you that i dont want to have both structures have all 6 members. I WANT there to be a raw color type in RGBA that is the base class and have other color schemes project onto that raw cartesian cube.

Dunbaratu:
Can I have a guarantee that under absolutely no circumstances will you ever create any method anywhere in Kos that insists upon being passed an HSV, and instead every single color-expecting method anywhere will use RGB and RGB only?  That guarantee is absolutely necessary in order for your claim that your way of doing it is just as good to be correct.  Your design hardcodes the notion that absolutely everything must in reality be an RGB no matter what it may look like on the surface

Dunbaratu:
Can I have a guarantee that under absolutely no circumstances will you ever create any method anywhere in Kos that insists upon being passed an HSV, and instead every single color-expecting method anywhere will use RGB and RGB only?  That guarantee is absolutely necessary in order for your claim that your way of doing it is just as good to be correct.  Your design hardcodes the notion that absolutely everything must in reality be an RGB no matter what it may look like on the surface

Dunbaratu:
Can I have a guarantee that under absolutely no circumstances will you ever create any method anywhere in Kos that insists upon being passed an HSV, and instead every single color-expecting method anywhere will use RGB and RGB only?  That guarantee is absolutely necessary in order for your claim that your way of doing it is just as good to be correct.  Your design hardcodes the notion that absolutely everything must in reality be an RGB no matter what it may look like on the surface while my design least more open ended and more open for future expansion

Dunbaratu:
Can I have a guarantee that under absolutely no circumstances will you ever create any method anywhere in Kos that insists upon being passed an HSV, and instead every single color-expecting method anywhere will use RGB and RGB only?  That guarantee is absolutely necessary in order for your claim that your way of doing it is just as good to be correct.  Your design hardcodes the notion that absolutely everything must in reality be an RGB no matter what it may look like on the surface while my design least more open ended and more open for future expansion

Dunbaratu:
Can I have a guarantee that under absolutely no circumstances will you ever create any method anywhere in Kos that insists upon being passed an HSV, and instead every single color-expecting method anywhere will use RGB and RGB only?  That guarantee is absolutely necessary in order for your claim that your way of doing it is just as good to be correct.  Your design hardcodes the notion that absolutely everything must in reality be an RGB no matter what it may look like on the surface while my design least more open ended and more open for future expansion

Dunbaratu:
What I want is for all kOS methods to say "give me a color and I don't care what kind".  When you make HSV derive from RGB instead of making them both derive from something common in a dependecy inversion sort of way, you make it so that only works in one direction and not the other. 

With your design you can use an HSV

Dunbaratu:
(Damn Github mobile web page interface wont let me edit my post)
With your design you can use an HSV where a RGB is expected but not the other way around.

erendrake:
You have it, i consider it a matter of theme that you would always be communicating to the hardware in terms it 'understands' and that raw color would always be the RGBAColor.

Dunbaratu:
I still dont think your way is as good because it would be better for that to be inherent in the design rather than just a convention but I tire of trying to get you to see that.  So fine when I get home ill pull it as is.

Dunbaratu:
Part of the reason I give up is that HSV isn't going to be used much.  Its harder to use unless you're in a scenario where you *start* already knowing the wavelength or frequency of light and want to show it to the user, like showing spectral data.

PR #627, opened 2015-03-01
added grey

erendrake:
there is no issue for this, this class talks about GRAY as an alternative spelling but we didnt actually have the alternative?


Dunbaratu:
it may be a good idea to alias both spellings - GREY and GRAY.
by the way this doesn't have the 'ready to merge' label, which is why I'm leaving it - not because of the alias thing.

erendrake:
I should have marked it with discussion. I am ambivalent about adding both spellings i just wanted to bring it up. It is ready to merge :)

Issue #635, opened 2015-03-03
Suggestion: replace current control for boot file selection.


ZiwKerman:
As discussed with erendrake, I am raising this issue and will follow it up with a PR later today.

Current boot.ks file selection UI is quite broken and unintuitive. By introducing the KSPAPIExtensions.dll dependancy I suggest bringing in UI_ChooseOption control for boot file selection.
Example:
![Example](http://i.imgur.com/05kp7Sy.jpg)

I will also bring in some bits of documentation of how boot scripts work in kOS right now.

erendrake:
The boot feature today is pretty arcane. I welcome any effort to make it better.

ZiwKerman:
Please advise where it would be better to put the documentation part of this feature - to `general/volumes.rst` or to `commands/files.rst`?

gillonba:
Not to digress too much, but is there currently documentation anywhere on how to use the boot file feature?  I would really like to use this feature but cannot find anyone explaining how

ZiwKerman:
I am planning on posting a pull request with some basic documentation in 1-2 hours from now, then it's up to devs to merge it into main release, but you'll be able to read a preview of it on my fork.

ZiwKerman:
By the way, if we release a version of kOS with KSPAPIExtensions as a dependency we should let the creator/maintainers of the mod know by posting in mod's thread on forums.

Issue #618, opened 2015-02-24
OS X system file .DS_Store messes with boot file select


pgodd:
If OS X hidden system files .DS_Store or other hidden system files are present in the scripts directory they throw off the index variable for filenames with boot prefix and create crazy results when user attempts to select boot file.
I traced the problem through and it looks like the cure (aside from Apple removing this nutty file) is to either delete it or filter it in the file manager overrides.
File management sees the file for the purposes of a count, but seemingly cannot delete the file by name.

Dunbaratu:
We could just make the general rule that any filename starting with a dot doesn't count and should be treated as if it wasn't there.  That would "theme" better for all UNIXes than just making the exception for the one hidden file Apple is a jerk about constantly re-adding after you delete it.  To keep the theme uniform across OSes, we could say that all files marked as 'hidden' don't  count either, for windows.

pgodd:
It looks like the place to filter these files out is archive.cs ksp API has a function to distinguish hidden files. I'll take a crack at this.

PR #637, opened 2015-03-03
Boot file selection UI control update + documentation on boot scripts feature.

ZiwKerman:
As discussed in issue #635 

Introduces dependency to KSPAPIExtensions.dll, uses UI_ChooseOption control to select boot script in editor.

For more information on KSPAPIExtensions see http://forum.kerbalspaceprogram.com/threads/81496-0-90-KSPAPIExtensions-V1-7-2-Utilities-for-shared-mod-use-16-Dec

I put the paragraph on boot scripts in `commands/files.rst` as it is related to files, but it could also be mentioned in `general/volumes.rst`


src/kOS/Module/kOSProcessor.cs
```diff
@@ -132,6 +130,26 @@ public Screen.TermWindow GetWindow()
 
         public override void OnStart(StartState state)
         {
+			//Populate selector for boot scripts
```
erendrake: Would we want to do this if we are not in the editor? Also would you mind moving it into its own method?:

erendrake:
Thank you very much for the PR :)

src/kOS/Module/kOSProcessor.cs
ZiwKerman: I'll add a check for StartSate and move all the logic into a separate Method for a cleaner look.

Dunbaratu:
To be honest I've never ever had the boot feature work and I never spoke up about how it's implemented because I don't have any examples of it working to compare against.  After this gets all merged in I might give it a try again.  Perhaps the problem I had is that since I dual boot back and forth between OSX and Windows 7, it might be the case that while under OSX I viewed the script folder with Mac's Finder program, which causes that hidden .DS_STORE file to appear (EVEN when you use Finder to view a non-UNIX filesystem where dotfiles aren't really hidden files like they are on UNIX filesystems, which is a infuriating), which apparently messed everything up.

ZiwKerman:
I also have 2 macs - Mac Pro with Win10 installed for gaming and OSX for everything else, and a Macbook with OSX and Ubuntu installed. And boot script feature only worked for me on clean install of KSP + KOS + IR on my Mac Pro in OSX. All other installs ether refused to run boot scripts completely, or I wasn't able to get anything apart from default "boot.ks" to work, which was not enough for my purposes, as I wanted different boot scripts for several cores on my vessel (I want to automate RT network creation).

erendrake:
@ZiwKerman have you looked at https://github.com/KSP-KOS/KOS/issues/618 ?

ZiwKerman:
@erendrake I did, but it does not explain why my pure Win10 install wasn't working, so I decided to fix the problem myself :)

erendrake:
One thing about this change @ZiwKerman. you sent us a PR against the master branch. We normally want requests to be for the develop branch.

this request also didnt add a reference to the csproj file for the new dependency 

Also.... tabs!!! 

this will not show as merged until i the next release because it is against master so ill just close it.

ZiwKerman:
@erendrake 

Hmm, that was a mistake. I'm new to github, should've asked before. Also I could not compile develop branch maybe due to some changes by @Dunbaratu

erendrake:
They were my changes that he merged in. I messed up and made some of the PRs dependent on one of the others. If you pull again it should be fixed.

ZiwKerman:
And by "tabs!!!" you mean that I forgot to set autoconversion of tabs to spaces?

ZiwKerman:
I pulled the update of develop branch, now it compiles. Should I resubmit a PR, for develop branch?

erendrake:
Nope, i merged your changes across and resolved the issues. you are in :)

ZiwKerman:
@pgodd 
At the first load on the launchpad after VAB/SPH selected boot file is copied into probe's local hard disk and named boot.ks

KOS.CPU.Boot(), which is called whenever probe comes into physics range from rails or when it is powered on from off state, checks whether there is a file "boot.ks" on starting volume and executes it.

You have to remember that boot file is only copied to probe's disk at the PRELAUNCH status, so any changes you make to the file afterwards would not be applied to probe's local copy, but you may manually update the file on probe's disk whenever you like.

I've checked 4 different scenarios:
 * switching to vessel from map view from outside of physics range - boot.ks is run
 * turning off the core and turning it back on - boot.ks is run
 * coming out of warp - boot is *not* run
 * switching between vessels in physics range - boot is *not* run

PR #640, opened 2015-03-04
Fixes #618 Boot selection, hidden and system files

### LINKED ISSUE #618: OS X system file .DS_Store messes with boot file select

pgodd:
Hidden and system files on PC/Windows and "dot" files on Mac disrupt file enumeration for boot file selection.
This filters those out.
Strange but true... Mac has a "hidden" file attribute as does Linux/Unix, that is different from PC/Windows attribute.
Chased this special hidden attribute down for a while, but this is so obscure it's not likely to ever come up.

### DIFF
--- a/src/kOS.Safe/Persistence/Archive.cs
+++ b/src/kOS.Safe/Persistence/Archive.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Xml.Linq;
 using kOS.Safe.Utilities;
 using FileInfo = kOS.Safe.Encapsulation.FileInfo;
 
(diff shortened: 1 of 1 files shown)


ZiwKerman:
I believe @erendrake merged my changes to boot UI selection yesterday into develop branch, please give it a try, don't forget that it introduces KSPAPIExtensions.dll dependency so you have to put it alongside KOS.dll in plugins folder.

pgodd:
That’s so much better now.

PR #641, opened 2015-03-04
Fix the inaccuracy/vagueness in Vessel:ROOTPART documentation.

ZiwKerman:


doc/source/structures/vessels/vessel.rst
```diff
@@ -217,7 +217,7 @@ All vessels share a structure. To get a variable referring to any vessel you can
     :type: :struct:`Part`
     :access: Get only
 
-    The first :struct:`Part` that was used to begin the ship design - the command core. Vessels in KSP are built in a tree-structure, and the first part that was placed is the root of that tree.
+    The ROOTPART is usually the first :struct:`Part` that was used to begin the ship design - the command core. Vessels in KSP are built in a tree-structure, and the first part that was placed is the root of that tree. It is possible to change the root part in VAB/SPH by using Root tool, so ROOTPART does not always point to command core or command pod. Also keep in mind that when you detach a part of a vessel or sub-craft - for example a ComSat subassembly - it becomes a new Vessel with it's own ROOTPART.
```
Dunbaratu: Unless you can launch a subassembly without adding it to another craft, that second change doesn't make any difference.  The subassembly won't be the root part of whatever vessel you attach it to.

ZiwKerman: I think you misread. A detached sub-assembly becomes a new Vessel, so if your core is attached to this subassembly the ROOTPART will change on detachment.

ZiwKerman: Maybe I need to clarify  - When you detach a part of a vessel *in flight*

Dunbaratu:
In that case you need to drop the "for example a comsat subassembly", which is a misleading part of the description.  You're apparently not talking about the subassembly feature at all.  How it was built is irrelevant to your point about decoupling or undocking.

ZiwKerman:
Hmm, yes. Maybe completely replace last sentence with something like this:

Vessel:ROOTPART may change in flight as a result of docking/undocking or detaching of some part of a Vessel.

It is a bit more vague, but still conveys the message.

Dunbaratu:
I don't agree that it's "a bit more vague".  It's more accurate.  The term "subassembly" has a very specific meaning in KSP and it's not the meaning you meant to say.

PR #645, opened 2015-03-05
user functions and local vars mega-change

Dunbaratu:
These are the 'specs' to which I'll be implementing the
actual code.

Pull it!  Give it a try!
================

To-Do list
------------

To make this manageable, I'm trying to split this up into smaller tasks, and make them contain
the steps in what I *think* is the order it will be possible to do them in.

- [x] Add initializer ```TO``` clause to ```DECLARE```, and make it mandatory.
- [x] Block Scope Local vars (test with if/else and loops since functions don't exist yet)
    - [x] Add local variable dictionaries to the stack.
    - [x] DECLARE to make new variables in localmost dictionary, which are at least limited to file scope.
    - [-] DECLARE GLOBAL to make globals explicitly just like SET does implicitly (so that you can still do this when in NOLAZYGLOBAL where implicitly doing it with SET is forbidden).  (DECIDED TO SCRAP THIS IDEA AND SAY "NO"
    - [x] Alter behavior of CPU.GetVariable to scan up through that stack before giving up.
    - [x] Alter behavior of CPU.GetOrCreateVariable to do the same scan, then create only at global level.
    - NOTE, if the above are all done, it is possible to do preliminary tests and examples with that much.  Therefore this point should be the first **pull target** (point at which people can pull what I have and play with it.)
    - [x] Ensure that all early-breaking escape-from-brace statements, like BREAK, pop all this stuff off properly.
    - [x] Ensure that exceptions that kill the program don't leave the stack behind all messed up.
    - [x] Implement NOLAZYGLOBAL to suppress parts of the above behavior.
- [x] Basic Function, do one pass to bottom of block and stop.
  - [x] Change scoping structure to deal with how function calls work
    - [x] handle jumps that skip to sibling nesting levels, by letting getvariable skip intervening stack scope frames that are "not for me".  DO this by storing lexical nesting level in the scope stack frame.
    - [x] handle jumps that recurse, probably with the same skipping stack frame technique.  (if A calls B calls C calls C calls C calls C, then C should see B's and A's vars, but not all the other copies of C's vars on the stack from the other calls to itself.)
- [x] RETURN keyword from middle of function (Ensure an early-break pops the variable contexts properly).
- [x] Fix FOR LOOP iterator case.  (Something in what I did messed up the scope of the secret iterator variable.)
- [x] Refactor: Rename everything in the code so it no longer calls things "locks" that actually refer to both locks and user functions now.

At this point, in *principle* everything should "magically" work on the following list (yeah right), but the testing should check all the cases:

### Thorny test cases:

- [x] Prove local scoping by nesting sections which all have the same variable name in them with different values in the different variables with the same name.
- [x] Make sure one file can be RUN with functions in it that another file can then call - crucial for the entire concept of users making library routines for themselves.
- [x] Check identifier clashes - two functions with the same global name.
- [x] Recursion, plain case:  Test with simple one-list case with no recursive branching (i.e. print values in a list backward via a dumb recursion algorithm).
- [x] Recursion, more complex case: Using Rootpart and part children, print the parts tree with indentation levels to show the tree structure.
- [x] Pass a function's output as a parameter to another function: Foo( bar(), 3, 5).
- [x] Deliberately make infinite recursion - see if it explodes KSP or if the user gets a useful error message.
- [x] Test binary ML files:  Check that the KSM compiled files behave just like their KS script files.


doc/source/language/variables.rst
```diff
@@ -113,3 +165,171 @@ Sets a variable to ``FALSE``. This is useful for the ``RCS`` and ``SAS`` binding
...
+variable.
+
+.. _scope:
+
+Scoping rules
+-------------
+
+.. note::
+    .. versionadded:: 0.17
+        In prior versions of kerboscript, all identifiers other than
```
erendrake: I believe this is true except for the FOR temporary identifier. So after a FOR BAR IN FOO {} BAR gets removed. at least it was this way before kos risc

doc/source/language/user_functions.rst
```diff
@@ -0,0 +1,339 @@
...
+new copies stacked up of all the local variables made with DECLARE,
+but not of the variables implicitly made global without DECLARE.
+
+An example of using ``declare`` for a local variable can be seen in
+the example above, where it is used for the ``row`` and ``col`` variables.
+
+A more in-depth explanation of kerboscript's scoping rules and how they
+work is found :ref:`on another page <scope>`
+
+``DECLARE TO``
```
erendrake: declare to has the benefit of being consistent with SET .. TO 

If the grammar bothers you, we could do DECLARE .. WITH?

eg DECLARE foo with 'bar'.

doc/source/language/user_functions.rst
```diff
@@ -0,0 +1,339 @@
...
+
+    // An example of calling it:
+    
+    print_corner(4,"That's me in the corner").
+
+``Declare function`` can appear anywhere in a kerboscript program,
+and once its been "parsed" by the compiler, the function can be called
+from anywhere in the program.  
+
+The best design pattern is probably to create your library of function
```
erendrake: what is the rule for duplicate functions? error?

erendrake:
I swear i am not trying to be a pill, I just need to ask about the term 'function'. There are a few alternatives that i like such as 'procedure'. 

I dont think function is a poor choice. I just wanted to take a moment and make sure that is the one we want.

doc/source/language/variables.rst
Dunbaratu: good pooint.  I menat all identifeirs created by 'set' or 'declare'.

doc/source/language/user_functions.rst
Dunbaratu: I thought it was bad looking at first until I started writing it out in examples and after about 5 examples or so it no longer looked weird to me.

Dunbaratu:
classically, function means "must have return value" and procedure means "must not".  So neither term will work 100% of the time.

I'm going with the term function used by C, where because originally ALL functions actually return a value (no void functions), it really was correct.

And behind the scenes, that's what we're doing to.

erendrake:
Sound reasoning. I say lets do this!

doc/source/language/user_functions.rst
Dunbaratu: I'd say we either REPLACE the earlier function, or make an error.

an error would be my preference but that's maybe harder.

On a related note, we have a longstanding bug where a user can make a homemade identifier that masks a bound variable.  i.e. someone can do DECLARE SHIP, and thus prevent SHIP from working in its normal way.

doc/source/language/user_functions.rst
erendrake: It isnt that strange so i think that the consistency benefit is appealing. Declare to it is

doc/source/language/user_functions.rst
erendrake: I think it should error as well. i cannot think of a good reason actually want to replace a function or a kOS binding

erendrake:
I just found a reference in https://github.com/KSP-KOS/KOS/blob/develop/doc/source/commands/parts.rst to the lack of user functions. we need to go back and clean these that kind of thing as part of this.

Dunbaratu:
thanks for catching that.  I'll have it in the next commit.

doc/source/language/variables.rst
Dunbaratu: I broke FOR loops when I changed the scoping rules, it turns out, but that's okay because I just changed the FOR's secret iterator to now be a local scope variable anyway, as it should be, and that fixed it.

Dunbaratu:
GO!!!  Give it a try.  I want other devs to beat on it and try it with old scripts.

I'm especially worried about how I had to run through all 77 built-in functions and change them all slightly.  There's opportunity in there for me to have screwed one or two of them up.

doc/source/commands/parts.rst
```diff
@@ -72,12 +72,11 @@ current part in the tree:
 You could keep walking down the tree this way, or go upward with PARENT
 and HASPARENT:
 
+TODO - **NEED TO MAKE A GOOD EXAMPLE OF WALKING THE PARTS TREE HERE WITH RECURSION ONCE THE SYNTAX IS NAILED DOWN FOR THAT.**
+
```
Dunbaratu: We could pull the code I have in functest3.ks into here.

src/kOS.Safe/Compilation/KS/UserFunctionCollection.cs
```diff
@@ -0,0 +1,75 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace kOS.Safe.Compilation.KS
+{
+    public class UserFunctionCollection
```
Dunbaratu: This file isn't *actually* new.  It's just the renamed LockCollection.cs.

src/kOS.Safe/Compilation/KS/kRISC.tpg
```diff
@@ -141,7 +152,8 @@ remove_stmt    -> REMOVE expr EOI;
 log_stmt       -> LOG expr TO expr EOI;
 break_stmt     -> BREAK EOI;
 preserve_stmt  -> PRESERVE EOI;
-declare_stmt   -> DECLARE PARAMETER? IDENTIFIER (COMMA IDENTIFIER)* EOI;
+declare_stmt   -> DECLARE ( (PARAMETER IDENTIFIER (COMMA IDENTIFIER)* EOI) | (FUNCTION IDENTIFIER instruction_block EOI?) | (IDENTIFIER TO expr) EOI);
```
Dunbaratu: If people decide they hate the ```DECLARE foo TO val ``` syntax, we can always add an optional token "INIT" here.  That would be easy.  Thus it would support saying ```DECLARE foo INIT TO val.```.

Dunbaratu:
@erendrake, @ZiwKerman , Please re-pull this branch!!

I discovered a bug last night (and Ziw hit the same bug) where when a script runs another script with arguments, it does so with the args shifted off by one (arg2 is in arg1, arg3 is in arg1, etc).

It was quick to fix because I already knew the fix but had somehow lost it in my major refactor work.

doc/source/language/variables.rst
```diff
@@ -113,3 +205,207 @@ Sets a variable to ``FALSE``. This is useful for the ``RCS`` and ``SAS`` binding
...
+Do not work predictably when you use local variables in the <expression>
+part of them.  They need to be designed to use global variables only,
+because they outlive the duration of any particular scoping braces.
+You can declare local variables within their <statements> in their bodies,
+just don't use local variables in the trigger conditions.
+
+.. _nolazyglobal:
+
+``NOLAZYGLOBAL``
+::::::::::::::::
```
Dunbaratu: Whoops.  This syntax changed and the docs didn't get updated.  It's now:
```@LAZYGLOBAL OFF.```, and it must be at the top, and ignores scope.  This section needs to be re-written.

File: doc/source/language/variables.rst [hunk 8]
```diff
@@ -1,27 +1,77 @@
...
+section it was created in.  After that code block is finished, the variable
+will no longer exist.
+
+.. note::
+    There is an implicit outer scope block around a whole kerboscript
+    program, such that if you use a DECLARE .. TO statement inside
+    a script, it will be one nesting level "higher" than the implicit
+    globals you make with SET.  In other words, using DECLARE always
+    makes a variable that is at least limited to the program script
+    it is inside of.
```
Dunbaratu: Arrrg.  This will have to be edited again also.  This isn't true anymore.  Now file scope and global scope are the same.

doc/source/language/variables.rst
```diff
@@ -70,9 +120,39 @@ The ``DECLARE PARAMETER`` statements can appear anywhere in a program as long as
...
+variable that already exists, if it can find one, and it only 
+creates a new variable if it *has* to because there isn't one that
+already exists.  *(That's the first difference)*.  Because ``SET``
+doesn't make a new variable until it has exhausted the attempts to
+find an existing one by looking up the "scope stack", ``SET`` only
+is capable of creating **global** variables.  *(That's the second
+difference.)*
+
+Also, be aware that DECLARE, in effect, is actually *incapable* of
+creating global variables.  There is an implicit scope block
```
Dunbaratu: Again, not true anymore.

doc/source/language/variables.rst
```diff
@@ -113,3 +205,207 @@ Sets a variable to ``FALSE``. This is useful for the ``RCS`` and ``SAS`` binding
...
+    of statements is, say, the body of an IF check, or the body of
+    an UNTIL loop.
+
+    Be aware that whenever you use the DECLARE..TO statement, you are
+    making a variable that is local to the scope in which it appears.
+    If you use DECLARE in the live interpreter, it makes a variable
+    that doesn't exist from inside a program.  If you use DECLARE in
+    a program script at the outermost nesting level of that script, it
+    still makes a variable that can only be seen from inside THAT 
+    program script.  If you have gotten used to the easy 'sloppy'
```
Dunbaratu: Need rewrite here too, same reason - not true anymore.

erendrake:
With all of these comments about documentation, would you still like me to merge it?

Dunbaratu:
I'll have to do a run through all the docs anyway just before release.  I'll just make a note to revisit these then.

PR #649, opened 2015-03-05
Fixes a typo in Quickstart tutorial

MaStr:


Dunbaratu:
thanks for the fix.  we'll have to pull it into develop branch, not gh-pages branch (gh-pages is purely output-only.  we never do source file changes there.)

MaStr:
argh.. I see now where the difference is... derp.

I'll make a new one. Sorry, for bothering.

Issue #653, opened 2015-03-06
New Feature: Make diskSpace parameter of KOSProcessor module tweakable


ZiwKerman:

hvacengi:
Looks like this may also eliminate the need for #573 which is good since I haven't spent any time working on my suggestion anyways.  I'm compiling the source to check a fix for another issue anyways, I'll check out this new option at the same time.  I think I'd recommend closing #573 at the same time as this issue.  At the very least the multipliers can always be changed if the need arises in the future.

ZiwKerman:
Solved and merged

PR #654, opened 2015-03-06
Tweakable diskSpace

ZiwKerman:
Introduces ability to upgrade KOSProcessor diskSpace to 2x and 4x times baseDiskSpace (as defined in diskSpace field of part.cfg or any ModuleManager config).
This upgrade comes at a cost of both funds and increased part mass. Current ratios are approximately 20kg of mass and 100 funds per additional 4096 bytes of storage.

As a side effect comes some UI code cleanup.


ZiwKerman:
Related to #653

Issue #656, opened 2015-03-08
Resource Transfer 


erendrake:
There is no way right now to achieve a resource transfer in kOS. that sucks really bad and i want to fix it :). such a system needs to be able to 

* Let you start a transfer of a specified amount from one set of parts to another
* Let you start a transfer of a all possible resource from one set of parts to another
* Lets you pause/restart a transfer
* Lets you monitor a transfer's progress

constraints on a transfer include:

* all parts must be on the same craft. the first revision would not be compatible with KAS attachment 
* if you specified an amount, there must be enough available space to accept the resource among the destination parts
* if you specified an amount, there must be enough available resource to supply the resource among the source parts

Failing any of these will stop the transfer.

ZiwKerman:
For clarity on constraints when amount is specified incorrectly: will the transfer stop after all available amount is transferred/all available space filled or immediately upon receiving incorrect instruction?

erendrake:
My first thought is to have it stop immediately.

erendrake:
one thing i would like to discuss is the current behavior which is to instantiate in an inactive state. So you must set :ACTIVE to true before the transfer would start. I am thinking about reversing this decision because if the user didnt want to transfer right away they could just defer creating the transfer in the first place.

ZiwKerman:
Well, yes I agree. Why would you create a TRANSFER in advance?

Dunbaratu:
I disagree.
You can control timing more accurately for multiple transfers if you set up all the transfers first without them moving yet, and then, so to speak, go "on the count of 3 ready.... one two three... GO."

ZiwKerman:
But you can already set up multiple-part to multiple part transfer in one go, no? or you think it may be more complex than that?

Dunbaratu:
Also, there's the issue of giving the user a choice.

If the transfer doesn't go until the user says so, then the user has the choice and the control whether or not to get it started or not.

If the transfer automatically goes until the user says to stop, then the user does NOT have the choice to set up a transfer that doesn't start going yet.

Starting with it stopped is the way that gives the user the most choice.

ZiwKerman:
Maybe then add a parameter on creation of Transfer whether you want to start it immediately or not? IMO it would feel more intuitive
a lot of users will forget to set TRANSFER:ACTIVE, I would be the first

Dunbaratu:
A choice in the constructor is fine by me.
[...]

PR #657, opened 2015-03-09
Resource transfer

erendrake:
Resolves https://github.com/KSP-KOS/KOS/issues/656


src/kOS.Safe/Encapsulation/ListValue.cs
```diff
@@ -297,7 +297,7 @@ private void InitializeSuffixes()
             AddSuffix("COPY", new NoArgsSuffix<ListValue>(() => new ListValue(this)));
         }
 
-        public static ListValue CreateList<T>(IList<T> toCopy)
+        public new static ListValue CreateList<T>(IEnumerable<T> toCopy)
```
erendrake: The two list value classes had different signatures for this method. this reconciles them.

src/kOS/Binding/FlightControlManager.cs
```diff
@@ -292,7 +292,7 @@ public bool Enabled
                 get { return enabled; }
                 set
                 {
-                    SafeHouse.Logger.Log(string.Format("kOS: FlightCtrlParam: Enabled: {0} {1} => {2}", name, enabled, value));
+                    SafeHouse.Logger.Log(string.Format("FlightCtrlParam: Enabled: {0} {1} => {2}", name, enabled, value));
```
erendrake: these were makiing log entries that started with "kOS: kOS:..."

src/kOS/Function/FunctionBase.cs
```diff
@@ -4,11 +4,9 @@
 
 namespace kOS.Function
 {
-    public class FunctionBase
+    public abstract class FunctionBase
     {
-        public virtual void Execute(SharedObjects shared)
-        {
-        }
+        public abstract void Execute(SharedObjects shared);
```
erendrake: function base has always been abstract in practice. when i noticed it wasnt i fixed it. It should have been moved out of this PR :(

src/kOS/Suffixed/ElementValue.cs
```diff
@@ -154,7 +152,7 @@ private static Queue<ElementPair> InitQueue(Vessel vessel, Dictionary<uint, Elem
 
         public override string ToString()
         {
-            return "ELEMENT( " + dockedVesselInfo.name + ", " + parts.Count + " )" ;
+            return string.Format("ELEMENT: {0} {1} Parts", dockedVesselInfo.name, parts.Count);
```
erendrake: in my testing i realized that the element to string had this number that was never explained. so i labeled it

doc/source/commands/resource_transfer.rst
```diff
@@ -0,0 +1,64 @@
...
+	
+TRANSFER will move the specified amount of a resource, where TRANSFERALL will move
+the resource until the source is empty or the destination is full. Both functions
+return a new :struct:`ResourceTransfer`. 
+
+Resource
+--------
+
+in all transfers, you must specify a resource by name 
+(eg "oxidixer", "LIQUIDFUEL", ect). this term is not case sensitive.
```
ZiwKerman: typo in oxidizer

doc/source/commands/resource_transfer.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Finding two parts by index, then transferring all of the oxidizer from 
+part 1 to part 2.
+
+::
+
+	SET foo TO TRANSFERALL("OXIDIZER", SHIP:PARTS[0], SHIP:PARTS[1]).
+	SET foo:ACTIVE to TRUE.
+	
+Finding two lists of parts by tag, then transferring 100 units of liquidfuel.
```
ZiwKerman: How exactly this would work? listOut[0] to listIn[0] until either is complete and them move to either listOut[1] or listIN[1]?

src/kOS.Safe/Exceptions/KOSInvalidArgumentException.cs
```diff
@@ -0,0 +1,11 @@
+namespace kOS.Safe.Exceptions
+{
+    public class KOSInvalidArgumentException : KOSException
+    {
+        private const string MSG = "Wile Invoking function {0}, argument {1} was invalid for the reason: {2}";
```
ZiwKerman: Typo in "wile" - > "while"

doc/source/commands/resource_transfer.rst
erendrake: this transfer would take a proportional share of the available resources from the parts in sourceParts  and transfer it to the parts in  destinationParts. the transfer is throttled to 1/200th of the maxamount in each source part per fixed update.

Dunbaratu:
I'll have a chance to give this a try this evening, and hopefully merge it.

doc/source/commands/resource_transfer.rst
Dunbaratu: This file has hardcoded tabs in it.  I'm altering them as I do the PR pull

doc/source/structures/misc/resource_transfer.rst
```diff
@@ -0,0 +1,100 @@
+.. _resource_transfer:
+
+Resource Transfer 
+=================
+
+Usage is covered elsewhere <../commands/resource_transfer.html>`__
```
Dunbaratu: I'm fairly sure this link syntax is messed up but I'll get to it later when I'm on the unix boot where I have the stuff to parse the RST files installed.

erendrake: I am bad, i never installed it so i have been just guessing, i need to just pull the trigger on it :(

Dunbaratu: When you are ready for a release later let me know and I'll take one last pass at parsing the docs across all the recent PRs rather than looking at them one PR at a time.

src/kOS.Safe/Exceptions/KOSInvalidArgumentException.cs
```diff
@@ -0,0 +1,11 @@
...
+{
+    public class KOSInvalidArgumentException : KOSException
+    {
+        private const string MSG = "While Invoking function {0}, argument {1} was invalid for the reason: {2}";
+
+        public KOSInvalidArgumentException(string functionName, string argumentName, string reason) : base(string.Format(MSG, functionName, argumentName, reason))
+        {
+        }
+    }
+}
```
Dunbaratu: We probably need to go back and find other places where this can be thrown too.  It looks like it could be used all over the place.

src/kOS/Execution/TransferManager.cs
```diff
@@ -0,0 +1,76 @@
...
+        public ResourceTransferValue CreateTransfer(PartResourceDefinition resourceInfo, object transferTo, object transferFrom)
+        {
+            var toReturn = new ResourceTransferValue(this, resourceInfo, transferTo, transferFrom);
+            transfers.Add(toReturn);
+            return toReturn;
+        }
+
+        public void Dispose()
+        {
+            shared.UpdateHandler.RemoveFixedObserver(this);
```
Dunbaratu: If a user stores a transfer object in a global variable, this Dispose will never happen and the object will be getting pointlessly Update()d from then on until they power cycle their CPU.  Maybe ```RemoveFixedObserver(this)``` should also be happening when the transfer completes?

erendrake: This is the transfer manager which is in shared and will only get disposed when shared does. His job is to hold the list of transfers. He could subscribe and unsub when he doesnt have anything. or I could just leave it updating and add this to the first line.

    if (!transfers.Any()) return;

That's a pretty cheap check.

Dunbaratu: So is there exactly one of these per CPU?

erendrake: yeah, the manager there is only one. The reason each transfer doesnt get "disposed" is that the user can reactivate one after it is finished. i imagined this working on a TRANSFERALL command and you just want to keep nudging it.

src/kOS/Suffixed/ResourceTransferValue.cs
```diff
@@ -0,0 +1,387 @@
...
+            {
+                return;
+            }
+            WorkTransfer(fromParts, toParts);
+        }
+
+        public override string ToString()
+        {
+            return string.Format("Transfer( {0}, {1} )", Status, StatusMessage);
+        }
```
Dunbaratu: I'm not sure I like ToString being used to convey status instead of being used in a way that identifies "what is this thing".  Being used to convey status *in addition to* "what is this thing" would be fine by me, but this basically masks what the transfer actually is.  If a user wands to print out the transfer object, just to see what it is, this ToString doesn't mention what the resource is, where its going, etc.  It just says if it's done or not.

erendrake: thats fair, i was using this as a status for most of development which was useful but now that it works thats not such a big deal.

erendrake: so maybe something like 

    if(amount.hasValue){
        return "TRANSFER(oxidizer, transferFromType, transferToType)";
    else {
        return "TRANSFERALL(oxidizer, transferFromType, transferToType)";

Dunbaratu: Sure,  I can add it right now if you like - I have the file open.

erendrake: go for it

src/kOS/Suffixed/ResourceTransferValue.cs
```diff
@@ -0,0 +1,387 @@
...
+            foreach (var part in parts)
+            {
+                var resource = part.Resources.Get(resourceInfo.id);
+                if (resource == null) continue;
+
+                var thisPartsPercentage = resource.amount/availableResources;
+
+                // Throttle the transfer
+                var thisPartsShare = transferGoal*thisPartsPercentage;
+                var thisPartsRate = resource.maxAmount*RESOURCE_SHARE_PER_UPDATE;
```
Dunbaratu: So the transfer rate is that it always takes 200 fixedupdates to fill the tank, for any sized fuel tank, if I'm understanding this right?  That works out to about 8 seconds for most people's Unity settings if they never altered them on the setup screen.

That seems okay.

erendrake: Yeah, i couldnt find the actual rate so i took a guess and it seemed ok. we can of course always change it.

src/kOS/Suffixed/ResourceTransferValue.cs
Dunbaratu: Okay, I went with this:
```
            return string.Format("{0}({1}, {2}, {3}{4})",
                                 amount.HasValue ? "TRANSFER" : "TRANSFERALL",
                                 resourceInfo, transferFromType, transferToType,
                                 amount.HasValue ? (", " + amount) : "" );
```

src/kOS/Suffixed/ResourceTransferValue.cs
```diff
@@ -0,0 +1,387 @@
...
+
+            if (!amount.HasValue)
+            {
+                toReturn = destinationAvailableCapacity;
+            }
+            else
+            {
+                var rawGoal = amount.Value - transferedAmount; 
+                toReturn = Math.Min(destinationAvailableCapacity, rawGoal);
+            }
```
Dunbaratu: As a rule of thumb, I usually don't use the "!" not operator for cases where I have both an IF and an ELSE.  There's no point when you could just swap the two clauses and then not use the "!".  One thing I never liked about the old C syntax this is inherited from is that "!" is an easy character to not notice.  It doesn't stand out like some of the other punctuation marks.  It looks too similar to the letter 'i'.  So I usually either find excuses to not use it, or I violate formatting slightly to force it to stand out better, like ``if ( !(amount.HasValue) )```.  But in this case, can we just swap the clauses of IF and ELSE?

erendrake: I completely agree. Like you said, its pretty easy to miss and they should be swapped. Thanks for catching that

src/kOS/Suffixed/ResourceTransferValue.cs
```diff
@@ -0,0 +1,387 @@
...
+            DetermineTypes();
+            InitializeSuffixes();
+        }
+
+        public void Update(double deltaTime)
+        {
+            if (Status != TransferManager.TransferStatus.Transferring) { return; }
+
+            IList<global::Part> fromParts = GetParts(transferFromType, transferFrom);
+            SafeHouse.Logger.Log("TRANSFER: FromParts Count: " + fromParts.Count);
```
Dunbaratu: There's a lot of these spamming logs and they're not SuperVerbose or protected by a flag of any kind.  Do we want to spew one of these 25 times a second, or is this leftover from debugging and should be removed?

erendrake: because i forgot to mark them as TODO:Remove i just spaced that they were there. I think just one when its starts and another when it fails or succeeds should be fine.

Dunbaratu: Okay I'll remove most of them except for those two.,

Dunbaratu: Okay here's what I'm going with instead, as the only logging that will occur:

```
        public TransferManager.TransferStatus Status
        {
            get { return status; }
            private set
            {
                status = value;
                transferManager.ReregisterTransfer(this);
                SafeHouse.Logger.Log( StatusChangeMessage ); // <--- THE LINE I ADDED
            }
        }
```
Then further down:
```
        private string StatusChangeMessage()
        {
            return string.Format)("{0} is now {1}.", ToString(), Status);
        }
```

for example: it might log:  ```TRANSFERALL("oxidizer", LIST of Parts(), List of Parts()) is now Inactive``` (or "is now Transferring", etc).

Dunbaratu:
I removed the "ready to merge" label so that you can use it again as a flag to tell me when you're done editing it.

erendrake:
Thank you :-)

src/kOS/Function/Suffixed.cs
```diff
@@ -405,4 +407,52 @@ public override void Execute(SharedObjects shared)
...
+            {
+                throw new KOSInvalidArgumentException("TransferAll", "Resource",
+                    resourceName + " was not found in the resource list");
+            }
+
+            object toPush = shared.TransferManager.CreateTransfer(resourceInfo, transferTo, transferFrom);
+            shared.Cpu.PushStack(toPush);
+        }
+
+    }
```
Dunbaratu: This function will need to be updated to use the new user_functions branch's paradigm.  I didn't do it in user_functions because this branch and that branch were awaiting merge simultaneously, and this function didn't exist yet in develop when I went through and changed them all.

src/kOS/Function/Suffixed.cs
```diff
@@ -405,4 +407,52 @@ public override void Execute(SharedObjects shared)
...
+            }
+
+            double parsedAmount;
+            if (Double.TryParse(amount.ToString(), out parsedAmount))
+            {
+                object toPush = shared.TransferManager.CreateTransfer(resourceInfo, transferTo, transferFrom, parsedAmount);
+                shared.Cpu.PushStack(toPush);
+            }
+        }
+    }
```
Dunbaratu: This function will need to be updated to use the new user_functions branch's paradigm.  I didn't do it in user_functions because this branch and that branch were awaiting merge simultaneously, and this function didn't exist yet in develop when I went through and changed them all.

src/kOS/Function/Suffixed.cs
```diff
@@ -405,4 +407,52 @@ public override void Execute(SharedObjects shared)
...
     }    
+
+    [Function("transferall")]
+    public class FunctionTransferAll : FunctionBase
+    {
+        public override void Execute(SharedObjects shared)
+        {
+            var transferTo = shared.Cpu.PopValue();
+            var transferFrom = shared.Cpu.PopValue();
+            var resourceName = shared.Cpu.PopValue().ToString();
```
Dunbaratu: The above three shared.Cpu.PopValue()'s  should be replaced with PopValueAssert(shared), after user_functions is merged in.

Dunbaratu: Then a line added afterward, here, saying:
```AssertArgBottomAndConsume(shared);```

File: src/kOS/Function/Suffixed.cs [hunk 18]
```diff
@@ -405,4 +407,52 @@ public override void Execute(SharedObjects shared)
...
+
+            var resourceInfo = TransferManager.ParseResource(resourceName);
+            if (resourceInfo == null)
+            {
+                throw new KOSInvalidArgumentException("TransferAll", "Resource",
+                    resourceName + " was not found in the resource list");
+            }
+
+            object toPush = shared.TransferManager.CreateTransfer(resourceInfo, transferTo, transferFrom);
+            shared.Cpu.PushStack(toPush);
```
Dunbaratu: And finally, this PushStack line replaced with:
```
    ReturnValue = toPush;
```

Issue #662, opened 2015-03-11
Colon in the string "ALT:RADAR" gets mangled when stored in a string variable.


Dunbaratu:
```
SET FOO TO "??? ALT:RADAR".
PRINT FOO.
??? ALT_RADAR

PRINT "Watch the colon in ALT:RADAR magically become an underscore".
Watch the colon in ALT_RADAR magically become an underscore
PRINT "Other colons Like:This are fine"
Other colons Like:This are fine.
```

There's some code in the mod that tries to make ```ALT:RADAR``` and ```ALT_RADAR`` aliases for the same variable.  More than likely that's getting activated when they are just dumb strings and not identifiers.

There are a few other similar variable bindings like I think ETA:apaopsis and ETA:periapsis that do the same thing.

This might get fixed automatically if we make two different kinds of strings - user  strings and identifier strings.
.

abenkovskii:
#931
> Technically, ALT and ETA are now objects

Close?

PR #665, opened 2015-03-13
added suffixes to make future orbits more like the current orbit

### LINKED ISSUE #550: no ETAs for future orbital patches
Situation: I'm still around Kerbin, bound for the Mun, and would like to place a maneuver node at the munar periapsis on my next orbital patch.

Problem: I can't figure out when/where the interesting points are. ETA:SOMETHING only works on the current patch. 

I can see how many and what kind of future orbits I will have, together with all defining parameters, so it should be possible to piece it together in some way... but I don't know how, and don't think it should be that complicated.

erendrake:
Resolves #550

### DIFF
--- a/src/kOS/Suffixed/OrbitInfo.cs
+++ b/src/kOS/Suffixed/OrbitInfo.cs
@@ -30,6 +30,8 @@ namespace kOS.Suffixed
             AddSuffix("NAME", new Suffix<string>(() => name));
             AddSuffix("APOAPSIS", new Suffix<double>(() => orbit.ApA));
             AddSuffix("PERIAPSIS", new Suffix<double>(() => orbit.PeA));
+            AddSuffix("ETA_APOAPSIS", new Suffix<double>(() => orbit.timeToAp));
+            AddSuffix("ETA_PERIAPSIS", new Suffix<double>(() => orbit.timeToPe));
             AddSuffix("BODY", new Suffix<BodyTarget>(() => new BodyTarget(orbit.referenceBody, shared)));
             AddSuffix("PERIOD", new Suffix<double>(() => orbit.period));
             AddSuffix("INCLINATION", new Suffix<double>(() => orbit.inclination));
--- a/doc/source/structures/orbits/orbit.rst
+++ b/doc/source/structures/orbits/orbit.rst
@@ -36,9 +36,15 @@ Structure
         * - :attr:`APOAPSIS`
           - scalar (m)
           - Maximum altitude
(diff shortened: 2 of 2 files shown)


src/kOS/Suffixed/OrbitInfo.cs
```diff
@@ -30,6 +30,8 @@ private void InitializeSuffixes()
             AddSuffix("NAME", new Suffix<string>(() => name));
             AddSuffix("APOAPSIS", new Suffix<double>(() => orbit.ApA));
             AddSuffix("PERIAPSIS", new Suffix<double>(() => orbit.PeA));
+            AddSuffix("ETA_APOAPSIS", new Suffix<double>(() => orbit.timeToAp));
+            AddSuffix("ETA_PERIAPSIS", new Suffix<double>(() => orbit.timeToPe));
```
Dunbaratu: On a vaguely related note, I've never liked how we have to have those weird exception cases for mapping ETA:Apoapsis to ETA_Apoapsis, and so on.  It parses like it was a suffix, but then the compiler goes out of its way to make a special case for it and change it into the underscore version, which also seems to have caused condition #662.

I dislike it because it's misleading to the script writers.  There is no such object as "ETA".  Therefore ETA:<anything> really shouldn't be a thing.

Unless... there really is an object called ETA so the exception case can be changed into something that is really using proper syntax. 

I'd like to propose that we either deprecate ETA:apoapsis and it's similar brethren, or if we want to keep it for compatibility, we actually make an ETA object so the parsing of it isn't a weird special case.  (i.e. An ETA object that holds the ETA for (or has runtime gettable properties for) apoapsis, periapsis, et all)

erendrake:
@Dunbaratu i think we should make an orbit eta object, people have been asking for more of them anyway like ETA:DN and ETA:AN. I agree with you that the shortcut is lame.

Dunbaratu:
Sorry I didn't notice this.  I'll have a look at it as soon as I get functions in, unless you need it sooner.  (I expect to get functions in today or tomorrow.

Dunbaratu:
When you said "I think we should make an orbit ETA object", do you think that means we should hold off on merging this until that can be done?  Basically if we had that, then it would instantly deprecate the user interface you just made here (it would no longer be OBT:ETA_APOAPSIS and would become instead OBT:ETA:APOAPSIS), OBT would return its ETA object which would return its APOAPSIS).

Should we release a user interface that we know we're going to deprecate soon afterward?

erendrake:
Seems worth it to do it right

PR #679, opened 2015-03-18
Radial mounted computer

pgodd:
Now with meagre solar power!


Resources/GameData/kOS/Parts/kOSMachineRad/part.cfg
```diff
@@ -0,0 +1,71 @@
...
+minimum_drag = 0.0
+angularDrag = 0
+crashTolerance = 6
+maxTemp = 3400
+
+
+MODULE
+{
+	name = kOSProcessor
+	diskSpace = 60000
```
erendrake: 60k is quite a bit bigger than our other processors. It is designed to be an end game part?

pgodd: It’s up for discussion.
I believe it should be treated as a more advanced part.
This is consistent with the gradual progression of technology suggested by the game.
I believe the current parts are at the same level on the tech tree yes?
In the plot laid out for the kOS computer, we are initially emulating 1970’s technology. 
It made sense to me that a more advanced technology (miniaturization, integrated circuits, Nvram etc. ) would get around the  
serious limitations of hard drive space.
Depending on when we expect we might develop global variables, direct data exchange between modules etc. we might want to scale this down.
Another solution might be to add another model for “storage expansion”.

You will also note the weight and drag characteristics of the module.
This is quite consistent with other radial mounted “black boxes” developed for the game.
The problem being we have to mount things on the exterior (with minor exceptions for mods that allow otherwise).\
If drag and weight were significant, that would pose an unrealistic challenge for designing spacecraft, as the engineers would simply locate the 
devices on the interior of the ship.

Additionally, the power consumption is quite high given the hardwired link between the HD size and power draw.
This is only slightly offset by the addition of a weak solar panel.
There are still plenty of challenges to managing this computer through a mission.

erendrake: right now the basic 1.25 part is unlocked in Flight Control and the 0.625 part is unlocked in Precision Engineering. I think this would make most sense in Unmanned Tech or Advanced Unmanned Tech.

With a solar panel i would say we should not get unlocked before the stock solar panels.

As far as the disk limit goes, you are right that it is does draw extra power. lower down in the tech tree i am worried that it would draw so much power that it would be unusable.

erendrake:
Im sorry, i started out with a critical comment. Thank you very much for taking the time to build us a radial part! i am very excited to include this.

pgodd:
Didn’t see it as critical.

;)

Resources/GameData/kOS/Parts/kOSMachineRad/part.cfg
pgodd: I would have to defer to you.
I don’t have much interest in the tech tree.
I agree it seems to be a big advantage for unmanned work. Particularly with the combined solar cell.

It makes a heck of a nice sandbox toy.

erendrake: and for sandbox it would of course be unlocked :)

erendrake: So i guess what i am proposing is either 

* Move the part farther up the tech tree

or

* reduce the available space so people dont add it to their craft and run out of power right away


if i was picking i would leave everything the same and just move it to to unmanned tech.

pgodd: Done, as far as I’m concerned.

BTW I really enjoyed learning how to make parts. 
It’s really quite simple once you get the hang of it.

PR #721, opened 2015-04-03
Cpu fixed update DONT MERGE

### LINKED ISSUE #526: CPU draws power even when game is paused?
Two users have both reported in the forums that the kOS CPU part draws power even when the KSP game itself is paused.  Pause the game with the CX-4181 turned on, go away for a few hours leaving your gaming PC turned on, and come back, and your ship is out of power even though no time has passed on the game clock.

I suspect that Timewarp.fixedDeltaTime is the wrong way to be calculating this, as well, because Updates() happen more (or less if your computer sucks) often than fixed updates.  Draining power by one fixed Delta Time step each Update time means people who's frame rate is 2x physics rate are getting charged for electricity doubly, I think.

So there's really two problems at once, if this is happening:

1 - If Update() gets called when the game is paused, then we have to catch that condition instead of running IPU number of instructions each Update while paused.  Pausing the gameworld should pause the kOS script.
2 - If running Update() is causing power to drain even when the game is paused (and presumably time isn't passing on the world clock) then that's evidence that we shouldn't be consuming power as if Timewarp.fixedDeltaTime was passing every Update, because clearly it's not.  Even without the pausing problem, there's still the framerate problem and something else should be used to calculate power usage.

**Power save mode??**
Perhaps this is a good opportunity to consider implementing some sort of clock speed power save idea while we're at it.  Make the unit drain less power when IPU is lower.  We could do this perhaps very easily by doing the electrical charge as a micropayment.  Find out how much drain typically happens within one Update(), divide that by 150 to get a tiny drain amount, and then charge that tiny amount just before executing each instruction.  This would depend on the floating point numbers being able to operate lots of times on values that small without too much cumulative error, of course.

If this was implemented, then faster computers would take more energy drain, and putting the computer into sleep mode by intentionally reducing IPU to something like 1 (the current gui widget clamps to no lower than 50 but that could be changed), makes it drain less power.  With that you could put your script into a mode where pretty much ALL it does is wait to wake itself up in a tiny simple loop, with all Whens, Ons, locks, etc turned off.  an IPU of 1 would be enough to run in this mode and then it could put the IPU back to 150 when it wakes up.

### LINKED ISSUE #527: Finally move script execution to FixedUpdate instead of Update.
(This will likely fix #526 when its implemented, and in fact issue #526 is what triggered the decision to finally get going on this.)

We've always known it's sort of wrong that the kOS CPU executes instructions based on the speed of Unity Update()s rather than at the fixed speed of the universe.  Update() gets called as many times as the Unity engine can spare, depending on the computer's speed.  However much time is leftover after running the mandatory code in FixedUpdates, the leftover is used for Updates.  Fast gaming computers run many Update()s per physics tic, and slow ones run fewer Update()s than physics tics.  

Therefore CONFIG:IPU should really be Instructions Per **Fixed** Update, not Per *Update*.

----------

Here is an initial design suggestion for how to go about doing it:

What exists now:
--------------------

This is the description of how it works at the moment:

- Various parts of kOS can register themselves as Update listeners, by calling UpdateHandler.AddObserver().  Currently this is done by each instance of CPU, each instance of VectorRenderer, and each instance of RemoteTechInterpreter.  To do this, an object needs to implement Interface IUpdateObserver, which merely means it has an Update() method in it.

- Unity itself calls kOSProcessor's Update() method.

- kOSProcessor's Update() method in turn tells UpdateHandler's UpdateObservers.

- UpdateHandler goes through all the objects that have registered themselves with AddObserver and calls their Update()'s.  This includes the CPU, The RemoteTechInterpreter, and the VectorRenderer.

What it should change to: 
-------------------------------

UpdateHandler should add these methods:

- UpdateHandler.AddFixedObserver() - for a class instance to register itself as a fixed updater.  Must be an implementer of IFixedUpdateObserver, meaning it has a FixedUpdate() method. 

- UpdateHandler.UpdateFixedObservers() - to be called on FixedUpdate to run through all observers and call them.

- kOSProcessor should call UpdateHandler.UpdateFixedObservers() in its own FixedUpdate().

- CPU should no longer call UpdateHandler.AddObserver(this) and instead do UpdateHandler.AddFixedObserver(this), and it should rename its Update() to FixedUpdate()

This *might* be all it takes to change things to work.  Maybe.  It's just my initial suggestion.

It also may mean we should increase or decrease the default config:IPU.

It also means we should edit a few bits of the documentation site which currently do tell users that everything happens in Unity's Update.

### LINKED ISSUE #539: power use fluctuations when kOS module installed in part
if I use Module Manager to add the kOSProcessor module to a part, when I load it into the flight scene the power use according to the resource panel will fluctuate wildly between 0.17 and 0.20 ec/s. The only thing that should be draining is the probe core which uses 0.05 ec/s. When I remove the module and go back to the flight scene the power drain is a steady 0.05 ec/s as it should be.

Using latest 0.90 release 0.15.5

erendrake:
i realized that there was an old branch that i had never merged that renamed the update types. so its in here along with the change to CPU,

im saying dont merge because i havent tested it in game yet.

fixes #527 
fixes #526
fixes #539

### DIFF
--- a/src/kOS.Safe/Execution/ICpu.cs
+++ b/src/kOS.Safe/Execution/ICpu.cs
@@ -3,7 +3,7 @@ using kOS.Safe.Compilation;
 
 namespace kOS.Safe.Execution
 {
-    public interface ICpu : IUpdateObserver
+    public interface ICpu : IFixedUpdateObserver
     {
         void PushStack(object item);
         object PopStack();
--- a/src/kOS.Safe/IUpdateObserver.cs
+++ b/src/kOS.Safe/IUpdateObserver.cs
@@ -4,6 +4,11 @@ namespace kOS.Safe
 {
     public interface IUpdateObserver : IDisposable
     {
-        void Update(double deltaTime);
+        void KOSUpdate(double deltaTime);
+    }
+
+    public interface IFixedUpdateObserver : IDisposable
+    {
+        void KOSFixedUpdate(double deltaTime);
     }
 }
--- a/src/kOS.Safe/UpdateHandler.cs
+++ b/src/kOS.Safe/UpdateHandler.cs
@@ -9,7 +9,7 @@ namespace kOS.Safe
         // once and therefore only gets its Update() called once per update.
         // The value of the KeyValuePair, the int, is unused.
         private readonly HashSet<IUpdateObserver> observers = new HashSet<IUpdateObserver>();
-        private readonly HashSet<IUpdateObserver> fixedObservers = new HashSet<IUpdateObserver>();
+        private readonly HashSet<IFixedUpdateObserver> fixedObservers = new HashSet<IFixedUpdateObserver>();
 
         public double CurrentFixedTime { get; private set; }
         public double LastDeltaFixedTime { get; private set; }
@@ -21,7 +21,7 @@ namespace kOS.Safe
             observers.Add(observer);
         }
 
-        public void AddFixedObserver(IUpdateObserver observer)
+        public void AddFixedObserver(IFixedUpdateObserver observer)
         {
             fixedObservers.Add(observer);
         }
@@ -31,7 +31,7 @@ namespace kOS.Safe
             observers.Remove(observer);
         }
 
-        public void RemoveFixedObserver(IUpdateObserver observer)
+        public void RemoveFixedObserver(IFixedUpdateObserver observer)
         {
             fixedObservers.Remove(observer);
         }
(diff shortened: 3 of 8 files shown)


Dunbaratu:
Do not forget that accompanying this change should be some edits to the doc pages.  There's a few places referencing how we do things in **Update ticks** and describe the weird behavior we have at the moment.

erendrake:
I merged this because i was pretty happy with how it was going last night. I wanted to let everyone play with it too and ill make a new issue for the Documentation piece because i cannot finish that today.

Dunbaratu:
I'll give the docs an overhaul.  The stuff talking about fixed vs framerate updates is mostly my own words so I sort of remember where everything is.

Dunbaratu:
btw: glad this is in.  I definitely want to give it a try later tonight.

Issue #722, opened 2015-04-03
BUG: Inside functions wait until does not work


ZiwKerman:
To reproduce run the following script
```javascript
    declare function bugtest1 {

        print "waiting for 3 seconds".
        wait 3.
        print "done!".

        declare t to time:seconds + 5.

        print "waiting until time:seconds > t = time:seconds + 5".
        wait until time:seconds > t.
        print "done 5!".
    }

    bugtest1().
```
The second wait finishes instantly.

Dunbaratu:
```wait until``` is implemented as a trigger. 

Just to give an idea, here's how complex it is now:

at the top it makes this trigger:
```
push $time     ------.  <-- $wait-0000001 is a variable with this instruction's address location in it.
push "seconds"       |
getmember            |--- from your expression
push $t              |      "time:seconds < t"
gt               ------'
br.zero 4
endwait  // changes the CPU state back to "Running"
push $wait-0000003  // id of this trigger, i.e. the 3rd trigger the compiler made.
removeytrigger
EOF
```
Then down in the body where the wait statement actually existed, it turns that trigger on with:
```
push $wait-0000003 // id of this trigger
addtrigger true  // 'true' means "do turn on the CPU's Wait state".
```

So it sets the CPU's state to ```Waiting``` due to the 'true' parameter to addtrigger, and waits for the trigger to turn it back off again.

What's causing the problem is the fact that the trigger's scope is outside the scope of the function, so it's not allowed to see the functions' variables in its expression check.  (it's probably falling back to a more global $t instead of the local one).

I see no reason at all for this to be implemented on the back of the already beleaguered trigger system, with its upper limit on instructions allowed and so on, when you could just do all the logic right there inline in the code like so:
```
wait 0.0000001
push $time     ------.
push "seconds"       |
getmember            |--- from your expression
push $t              |      "time:seconds < t"
gt               ------'
br.zero -6  // jump negative 6 instructions (up to wait 0.0000001)).
```
So it basically just spins in a loop forever checking the expression *right there* in the inline code, but in a way that only loops once per update rather than busy polling.

So much cleaner.  Doing this will also solve issue #688, as it is caused by accidentally triggering the "trigger" logic when it should have not done so (using a zero wait makes the system think it's a WAIT UNTIL when it isn't).

Dunbaratu:
actually I think I got the gt/lt 's backward in my above examples typed out manually, but you get the gist of it.

Dunbaratu:
[...]

PR #726, opened 2015-04-04
Resolves #688, #722 - WAIT is no longer a trigger.

### LINKED ISSUE #688: WAIT 0. freezes script/terminal
The command `WAIT 0.` if typed directly to the terminal locks you out until you cycle power (never completes and you can't Ctrl+C out.)
When in a script it freezes the script until you abort it. 
This also happens if the value is negative.

Dunbaratu:
In addition to fixing the mentioned issues, this should also be slightly simpler and more efficient.

Removed WAIT from the trigger system and just made it do a simple loop right there in the mainline code where it was defined,  akin to this psuedocode:
```
    do
            wait 1 (fixed)update tick.
            evaluate expression
    until expression is true
```
This is literally exactly the same performance as the old way (because the trigger checked the condition every (fixed)update, and so does this), but without eating into the precious limited resource of trigger instructions per update.

This will also be slightly less ML code *except* in the case where the script uses the same WAIT UNTIL expression repeatedly in more than one place in the source code.  Then this will admittedly result in a slightly bigger ML file than it would have before (because before the expression's instructions were only compiled once regardless of how many times the expression was mentioned in wait statements, and all the additional places the same wait expression existed it just referred back to the existing trigger.)

I would argue that that's actually a "good" thing, though, because users *should* expect that programming by cut-and-paste would result in inefficiently repeated code in the ML file, and not rely on the compiler to optimize away what is essentially a redundancy that exists in the source code.

### DIFF
--- a/doc/source/general/cpu_hardware.rst
+++ b/doc/source/general/cpu_hardware.rst
@@ -43,6 +43,14 @@ Triggers are all of the following:
 -  ON condition { some commands }.
 -  WHEN condition THEN { some commands }.
 
+.. note::
+
+    The :ref:`WAIT <wait>` command only causes mainline code
+    to be suspended.  Trigger code such as WHEN, ON, LOCK STEERING,
+    and LOCK THROTTLE, will continue executing while your program
+    is sitting still on the WAIT command.
+
+
 The way these work is that once per **update tick**, all the LOCK expressions which directly affect flight control are re-executed, and then each conditional trigger's condition is checked, and if true, then the entire body of the trigger is executed all the way to the bottom \*before any more instructions of the main body are executed\*. This means that execution of a trigger never gets interleaved with the main code. Once a trigger happens, the entire trigger occurs all in one go before the rest of the main body continues.
 
 Do Not Loop a Long Time in a Trigger Body!
@@ -81,6 +89,14 @@ But the duration of the next update tick is actually 0.09 seconds, then you will
     WAIT UNTIL TRUE.
 
 Then even though the condition is immediately true, it will still wait one update tick to discover this fact and continue.
+
+.. note::
+
+    The :ref:`WAIT <wait>` command only causes mainline code
+    to be suspended.  Trigger code such as WHEN, ON, LOCK STEERING,
+    and LOCK THROTTLE, will continue executing while your program
+    is sitting still on the WAIT command.
+
 
 The Frozen Universe
 -------------------
@@ -161,7 +177,3 @@ forums. You may put this code up near the top of your script::
       PRESERVE.
     }
 
-An Even Better Solution
-~~~~~~~~~~~~~~~~~~~~~~~
-
-There has been talk of instituting a special command: WAIT UNTIL PHYSICS that will sleep until there has been a physics update, and it's a good idea but it hasn't been implemented yet.
--- a/src/kOS.Safe/Compilation/KS/Compiler.cs
+++ b/src/kOS.Safe/Compilation/KS/Compiler.cs
@@ -245,9 +245,6 @@ namespace kOS.Safe.Compilation.KS
                     PreProcessChildNodes(node);
                     PreProcessWhenStatement(node);
                     break;
-                case TokenType.wait_stmt:
-                    PreProcessWaitStatement(node);
-                    break;
                 case TokenType.declare_stmt:
                     PreProcessProgramParameters(node);
                     break;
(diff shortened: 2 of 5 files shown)


doc/source/general/cpu_hardware.rst
```diff
@@ -43,6 +43,14 @@ Triggers are all of the following:
 -  ON condition { some commands }.
 -  WHEN condition THEN { some commands }.
 
+.. note::
+
+    The :ref:`WAIT <wait>` command only causes mainline code
+    to be suspended.  Trigger code such as WHEN, ON, LOCK STEERING,
+    and LOCK THROTTLE, will continue executing while your program
+    is sitting still on the WAIT command.
+
```
Dunbaratu: NOTE: This documentation notice is NOT referring to a change I made.  This was actually true all along even before this Pull Request.  I only just *noticed* that it was true because I was looking at that code, and decided that it should be clarified in the docs.  It surprised me when I found this out and it would probably not be what others would have assumed either.

I considered making an issue to change it so it *does* freeze everything even the triggers, until I remembered that triggers are used by flybywire's settings and thought that might open a can of worms to freeze it's work too when the script does a WAIT.

doc/source/general/cpu_hardware.rst
```diff
@@ -161,7 +177,3 @@ forums. You may put this code up near the top of your script::
       PRESERVE.
     }
 
-An Even Better Solution
-~~~~~~~~~~~~~~~~~~~~~~~
-
-There has been talk of instituting a special command: WAIT UNTIL PHYSICS that will sleep until there has been a physics update, and it's a good idea but it hasn't been implemented yet.
```
Dunbaratu: I decided that this doesn't belong in the documentation anymore.  I put it there thinking this change was going to happen "any day now" and it's been like that for months so clearly it's not happening real soon.

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -340,28 +337,6 @@ private void PreProcessWhenStatement(ParseNode node)
...
-                Opcode branchOpcode = AddOpcode(new OpcodeBranchIfFalse());
-                AddOpcode(new OpcodeEndWait());
-                AddOpcode(new OpcodePushRelocateLater(null), triggerObject.GetFunctionLabel());
-                AddOpcode(new OpcodeRemoveTrigger());
-                Opcode eofOpcode = AddOpcode(new OpcodeEOF());
-                branchOpcode.DestinationLabel = eofOpcode.Label;
-            }
-        }
-        
-
```
Dunbaratu: It's rare that I get to fix a bug in a way that *reduces* how much code there is.

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -1680,41 +1680,18 @@ public override string ToString()
     
     public class OpcodeAddTrigger : Opcode
     {
-        [MLField(1,false)]
-        private bool ShouldWait { get; set; }
-
```
Dunbaratu: Again, it feels good to fix something by getting to *cut* code and replace it with .. nothing.

The ```ShouldWait``` parameter to the ```addtrigger``` opcode only served one purpose and that was to differentiate the triggers created by WAIT from the ones created by other kinds of trigger statements like WHEN and ON.  It's gone now because there should be no such thing as a "waiting" trigger like this anymore.

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -1739,11 +1716,8 @@ public class OpcodeWait : Opcode
 
         public override void Execute(ICpu cpu)
         {
-            object waitTime = cpu.PopValue();
-            if (waitTime is double)
-                cpu.StartWait((double)waitTime);
-            else if (waitTime is int)
-                cpu.StartWait((int)waitTime);
```
Dunbaratu: cpu.StartWait actually takes a double anyway, so it seemed ridiculous to cast an int value to an int, pass it to StartWait, which will coerce it to a double.  May as well just convert it to a double unconditionally whether it was a double or an int.

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -1739,11 +1716,8 @@ public class OpcodeWait : Opcode
...
+            object arg = cpu.PopValue();
+            cpu.StartWait(Convert.ToDouble(arg));
```
Dunbaratu: We have methods over in FunctionBase that do safe numeric conversions (safe in that they will trap System casting exceptions and transform them into KOS exceptions instead, so users don't see the raw System exceptions when they pass bogus types of values to the functions.) 

It occurs to me that we might want to do a refactor later that adds that sort of thinking to all these kRISC opcode parameters.  Many of them are coming from the compiler directly injecting the ML code with the results of user expressions.

src/kOS/Execution/CPU.cs
```diff
@@ -759,10 +759,7 @@ public void RemoveTrigger(int triggerFunctionPointer)
 
         public void StartWait(double waitTime)
         {
-            if (waitTime > 0)
-            {
-                timeWaitUntil = currentTime + waitTime;
-            }
+            timeWaitUntil = currentTime + waitTime;
             currentStatus = Status.Waiting;
```
Dunbaratu: This was the cause of issue #688 .  The parameter waitTime comes directly from the "N" in the ```wait N.``` command, and to tell the CPU that this was the kind of WAIT that was a ```WAIT UNTIL FOO``` instead of a ```WAIT N```, the OpcodeAddTrigger(true) opcode would call cpu.StartWait(0).  The Cpu would then know it needed to rely on the execution of the triggers to see if the value had become true anywhere during the triggers.   But there was nothing preventing a ```wait N.``` from also passing in a value of N=0 and ending up calling cpu.StartWait(0) as well.

So the Cpu got confused and thought that ```wait 0.``` was really a ```wait until foo.``` but the compiler had never built the trigger for the expression that ```wait until```'s depend on, so it was always stuck on the default value of false.  Permanently.

But now that there are no trigger-based waits anymore, and the only kind of wait is a numeric seconds wait as far as the CPU cares (because the UNTIL logic exists out in user-code now not in the triggers), there's no problem with passing in a zero anymore.  Now it just means "wait until a tick in which the timer is later than right now", or in other words, the same thing as we had been using ```wait 0.0001

PR #739, opened 2015-04-05
Resolves #341 unary vector negation

### LINKED ISSUE #341: unary vector negation doesn't work
This:
``` set oppositeVector to - theVector. ```
doesn't work, but it should.  It should mean  the same thing as:
``` set oppositeVector to (-1) * theVector. ```

### PR DESCRIPTION
**Pull request by Dunbaratu**
Vectors now can be negated and Directions too (for directions, it does an :INVERSE suffix on it).

### CODE REVIEW

File: src/kOS.Safe/Compilation/Opcode.cs [hunk 1]
```diff
@@ -893,7 +892,17 @@ public override void Execute(ICpu cpu)
...
             else
-                throw new KOSUnaryOperandTypeException("negate", value);
+            {
+                // Generic last-ditch to catch any sort of object that has
+                // overloaded the unary negate operator '-'.
+                // (For example, kOS.Suffixed.Vector and kOS.Suffixed.Direction)
+                Type t = value.GetType();
+                MethodInfo negateMe = t.GetMethod("op_UnaryNegation", BindingFlags.Static | BindingFlags.Public); // C#'s alternate name for '-' operator
+                if (negateMe != null)
+                    result = negateMe.Invoke(null, new[]{value}); // value is an arg, not the 'this'.  (Method is static.)
```
Dunbaratu: I did it this way because I didn't want to have a Vector and Direction Interface in kOS.Safe when the public methods of them contain mention of Unity classes like ```Vector3``` and KSP classes like ```Vector3D```.

Therefore I picked a solution in which ANY negatable object works without knowing what its type is at compile time.  In general I think it's a more elegant solution anyway, as it's more future-compatible that way.

Issue #193, opened 2014-07-14
set target returns empty string until next tick


Gaiiden:
this will return a string error:
````
set target to waypointName + currentWaypoint.
print "<" + time:clock + "> driving " + round(target:distance, 2) + "m".
````
unless you place a `wait 0.001.` statement in the middle. Interestingly, this code executes just fine:
````
set target to MyTarget.
lock wheelsteering to target.
````
perhaps because the actual `lock` does not occur until the next tick

PR #742, opened 2015-04-06
Bound Variables set instantly 

erendrake:
#193 turned out to be an issue where the target was not actually being set until the end of the update. I considered making some kind of new bound variable that would interrupt the existing flow. Then i looked at the API calls we were making and I could not understand the reason for continuing to defer.

### DIFF
--- a/src/kOS.Safe/Binding/BoundVariable.cs
+++ b/src/kOS.Safe/Binding/BoundVariable.cs
@@ -33,5 +33,10 @@ namespace kOS.Safe.Binding
                 Set(value);
             }
         }
+
+        public void ClearCache()
+        {
+            currentValue = null;
+        }
     }
 }
--- a/src/kOS.Safe/Binding/BoundVariable.cs
+++ b/src/kOS.Safe/Binding/BoundVariable.cs
@@ -9,7 +9,6 @@ namespace kOS.Safe.Binding
         public BindingGetDlg Get;
 
         private object currentValue;
-        private bool wasUpdated;
 
         public override object Value
         {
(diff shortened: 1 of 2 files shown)


Dunbaratu:
I encountered the following two problems in testing:

### Problem 1 : Boolean variables for action groups permanently stuck at false.

Reproduce with this test, which you can just run directly from the interactive prompt:
```
until false {
  print "AG1 is " + ag1.
  wait 0.5
}
```
Now, with keyboard focus outside the window, press '1' several times to toggle the state.  The terminal still keeps saying "AG1 is False" over and over.
It works on the develop branch, where it toggles the value as expected.

The same problem happens with the 'named' action groups like "gear" and "lights".

The test case that revealed the problem was trying to run my testlibpid hoverbot script, and the buttons 1 and 2 weren't causing the seek altitude to change anymore.

### Problem 2: time:seconds is stuck.

Reproduce with this test:  Just keep doing ```print time:seconds.``` over and over and the same number gets displayed every time.
```
print time:seconds.
366684.023769342.
print time:seconds.
366684.023769342.
print time:seconds.
366684.023769342.
```
The test case that revealed this problem was that my ascend script's countdown display that was supposed to go 10,9,8,7,6... was instead going 10,10,10,10,10 forever so it never started the liftoff.

erendrake:
this is all because i am a dumb, once you 'get' the value it is cached and i removed the clear cache step so you will always get the first value always and forever :P

erendrake:
I have fixed the cache clearing issue. 

    print ag1.

can change when you hit 1 multiple times. toggling its state.

    print time:seconds.

will also advance normally.

PR #749, opened 2015-04-08
Resolves #741 : locks now have closures

### LINKED ISSUE #741: functions need closures for locks to work right
(Not sure if this is "bug" or "enhancement".  I'll call it "enhancement", but will represent its bug-ness by linking it to several existing github issues that are labeled as "bug" already.)

Several recent bugs surfaced around the practice of trying to use LOCK inside a local scope, to do things like this:
```
declare function throt_set {
  declare parameter user_scale.
  declare coefficient to 0.5.

  lock throttle to user_scale* coefficient * ship:maxthrust/ship:mass..
}

throt_set(0.75).
```
The problem being that users expect LOCK THROTTLE to keep working after the function throt_set is finished and its local scope, in which some of the values it depends on, are gone.

The only way to fix this is to implement [closures](http://en.wikipedia.org/wiki/Closure_%28computer_programming%29) for locks.

And since locks are indistinguishable from user functions once compiled, that means basically doing it for all user functions.

Dunbaratu:
I decided to make the issues about allowing local locks into a different issue, for two reasons:

1 - The possible solutions introduce new syntax that affects 'passive' variables too, so it should be address outside of a commentary about locks.

2 - Under the hood, the lock "variable" is really just a structure containing the integer program location of the function and the closure dictionaries for it.  As this variable is just a userland variable with a weird name ($varname*), any changes to address making a lock local would be do-able entirely separately from how the closure is implemented.  (In other words, what to do once you have a variable with a closure is what this PR is about.  How you make such a variable with a closure go away can be orthogonal to that, with different code that never touches the same area.)


src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -1187,14 +1189,23 @@ public override void Execute(ICpu cpu)
...
                 int currentPointer = cpu.InstructionPointer;
                 DeltaInstructionPointer = (int)functionPointer - currentPointer;
                 var contextRecord = new SubroutineContext(currentPointer+1);
                 cpu.PushAboveStack(contextRecord);
+                if (userDelegate != null)
+                {
+                    // Reverse-push the closure's scope record, just after the function return context got put on the stack.
+                    for (int i = userDelegate.Closure.Count - 1 ; i >= 0 ; --i)
+                        cpu.PushAboveStack(userDelegate.Closure[i]);
+                }
```
Dunbaratu: In general I'm starting to get a bit nervous about how long the body of OpcodeCall.Execute() is getting.  Eventually it should be split up into separate sections *somehow*, but every time I look into how I'd do it I just shudder and leave it as-is.  The reason it's all intertwined is that I don't want to cut-n-paste as much as I'd have to of I just had separate cases splitting into their own subroutines (if it's direct, do this, if it's indirect with a C# delegate do this, if it's indirect with a UserDelegate do this, if it's indirect with an integer delegate do this...).  These cases end up being about 75% the same as each other.

But what makes it hard to split up that other 25% into separate routines is that it's not the *same* 25%.  The bits they share in common are not the SAME - so you get an overlapping Venn diagram of which cases call which code, that isn't nice and clear to separate.

So, I'm saying I know the current long length with intermingled conditions is poor form, but fixing it is nontrivial and should probably be its own refactor rather than doing it simultaneously with a PR that is changing functionality.

When it gets refactored, it should be the sort of frozen-functionality PR that allows for regression testing (make sure this didn't change anything....).

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -1674,14 +1699,70 @@ public override void Execute(ICpu cpu)
...
+        {
+        }
 
+        /// <summary>
+        /// This variant of the constructor is just for ML file save/load to use.
+        /// </summary>
+        protected OpcodePushDelegateRelocateLater() : base() { }
+    }
+    
     #endregion
```
erendrake: remind me after 0.17 is out to split up this file :P

erendrake:
@Dunbaratu it looks like we are missing some files here?

* Exe\IStack
* Exe\IUserDelegate

Dunbaratu:
I don't know how I did it but somehow those files got attached to the wrong branch.  They're on the branch I was working on for fixing #756.  Well, since they're interrelated to the same general thing (functions and returns) I'll end up merging that fix in with this one and pushing the new version.  I'm just doing my own testing first before I push it.  if all goes well, expect it in about 10-20 min.

Dunbaratu:
okay ready again.

erendrake:
back on it. :)

Issue #763, opened 2015-04-08
Unable to launch new rockets


abenkovskii:
![screenshot0](https://cloud.githubusercontent.com/assets/11031560/7056438/47b5f184-de54-11e4-823a-9fc1a7a8eb36.png)
log: https://www.dropbox.com/s/bfeyha5qz4ajn6r/output_log-launch_bug.txt?dl=0

erendrake:
is this with the last canary i gave you or a new one?

abenkovskii:
The problem is with the version @ZiwKerman compiled for me. He experiences the same bug.

Tabris83:
i get this most times in relation to [issue #751](https://github.com/KSP-KOS/KOS/issues/751)

ascensionislander:
FWIW, sometimes I'll get this exact same bug with the current version when reverting a flight. I can't confirm that it's kOS causing it, but [others have had it too](http://www.reddit.com/r/KerbalSpaceProgram/comments/31ewu5/chart_i_used_kos_to_log_some_data_from_one_of_my/cq15qa8) and seem confident that it's kOS causing it.

Dunbaratu:
I'm reading through the log, and I notice this suspicious thing right off during the part at the beginning when all the mods are loaded:
```
NullReferenceException
  at (wrapper managed-to-native) UnityEngine.WWW:LoadImageIntoTexture (UnityEngine.Texture2D)
  at kOS.Screen.KOSToolBarWindow.FirstTimeSetup () [0x00000] in <filename unknown>:0 
  at kOS.Screen.KOSToolBarWindow.Start () [0x00000] in <filename unknown>:0 
```
Are all the PNG files for kOS's icons installed properly in your installation?  Or, did anyone recently touch the code that reads the PNG files?  Is there a way we may be failing to read the files?  Because failing that step is what is going to cause this exception.  And, looking at the code, it will abort FirstTimeSetup() before it finishes all its initial work getting the starting values right, which might cause all sorts of other errors down the road.

(For example, if it fails there, it never gets to the line where it assigns the panel's skin (the skin will still be null), or the line where it registers the panel with SQUAD's applauncher properly.)

Further down in the log I see problems all related to the panel, as more exceptions occur.

Just as it tries to enter FLIGHT mode during that last (fatal) scene change, I see this happening:
```
[HighLogic]: =========================== Scene Change : From EDITOR to FLIGHT (Async) ===================== 
[...]

Issue #761, opened 2015-04-08
[bug] can't unset boot file in the VAB


abenkovskii:
This issue was split from #758.

ZiwKerman:
It is in the docs.

Usually you should keep the default boot.ks as neutral as possible as a workaround

I have mine consisting of 1 line
```
   switch to 1.
```

abenkovskii:
@ZiwKerman, but @erendrake says it is a bug.

ZiwKerman:
It's not a bug if it's documented. This is a limitation of the UI tweakable and of how the boot files were programmed

There is a workaround possible, but we had other mattters to attend to.

erendrake:
@ZiwKerman It is documented, but it really sucks, even if we have to add a "clear boot selection" button i think its better than locking someone into a boot file. its not that i think this is a super high priority issue I just want to have it recorded that we should do something if we have time.

ZiwKerman:
@erendrake: I agree that we need to add the ability to set boot script to "None".

erendrake:
could we simply add another option in the slider with some kind of special sentinel "none" value that the processor knows means "i dont want a boot file"?

ZiwKerman:
That is what I intend to do

hvacengi:
As it is, doesn't the first boot file (usually boot.ks) get listed twice?  Or is that the last boot file?  I seem to recall seeing that when I was looking into how to revise the boot file logic (which I never did, and I'm happy @ZiwKerman is handling that).  If one file gets listed twice, one of the instances could be used as a "none" option.

ZiwKerman:
Solved and merged. @erendrake  you can close this one.

PR #769, opened 2015-04-09
Fixes: kOSToolbarWindow, bootFile can be set to None in VAB/SPH

ZiwKerman:
#761 

Fixes some kOSToolbarWindow long standing issues. Rewrote texture handling to use GameDatabase loaded textures and made textures static, so there are no multiple instances of textures.

BTW @Dunbaratu - I removed Instance counting as it is really unnecessary anymore, every instance created gets destroyed properly (left some logging in Start and OnDestroy).


ZiwKerman:
also fixes #763

Dunbaratu:
This looks like a good change.  I wrote that kostoolbar stuff based entirely on a post in the forums by maxmaps that made very little sense to me, and had to add all the instance counting because it kept throwing exceptions when doing it the way his post said to - frustrating sensitive timing exceptions that you can't trace without knowing a lot about Unity itself, which I don't.

I've long known it was bad, but had no idea what I was meant to do instead because the only documentation was that one post that didn't explain it well.

Thanks for the fix.  I'll be testing it out today.

ZiwKerman:
One of the major things I added is this 2 lines in OnDestroy (GoAway)

```
   GameEvents.onGUIApplicationLauncherReady.Remove(RunWhenReady);
   GameEvents.onGUIApplicationLauncherDestroyed.Remove(GoAway);
```
Because they were missing the instances of kOSToolbarWindow did not despawn properly, but now all should be fine

Dunbaratu:
Yeah the original post by maxmaps never mentioned making calls to GameEvents, or what it was.  It just said to set up things this way and it will work... which it didn't.

Dunbaratu:
I will be testing the toolbar stuff.  I am in no position to test the bootfile stuff that is also intermingled into this PR, as I know nothing of that system, and can't compare what I see to how it worked before.

src/kOS/Screen/KOSToolBarWindow.cs
```diff
@@ -80,100 +80,65 @@ public class KOSToolBarWindow : MonoBehaviour
...
-            terminalOpenTelnetIconImage.LoadImageIntoTexture(terminalOpenTelnetIconTexture);
-            terminalClosedTelnetIconImage.LoadImageIntoTexture(terminalClosedTelnetIconTexture);
-
-            windowRect = new Rect(0,0,width,height); // this origin point will move when opened/closed.
+            launcherButtonTexture = GameDatabase.Instance.GetTexture ("kOS/GFX/launcher-button",false);
+            terminalOpenIconTexture = GameDatabase.Instance.GetTexture ("kOS/GFX/terminal-icon-open", false);
+            terminalClosedIconTexture = GameDatabase.Instance.GetTexture ("kOS/GFX/terminal-icon-closed", false);
+            terminalOpenTelnetIconTexture = GameDatabase.Instance.GetTexture ("kOS/GFX/terminal-icon-open-telnet", false);
+            terminalClosedTelnetIconTexture = GameDatabase.Instance.GetTexture ("kOS/GFX/terminal-icon-closed-telnet", false);
+
```
Dunbaratu: At first I didn't like this change away from using constant strings for the names, but now that I think of it, there's nowhere else we'd be using those strings anyway except for right here, so I guess this is okay.

Dunbaratu:
The KOSToolbarWindow portion of this PR has my blessing.  I read the code changes and tested the following cases:

  * revert to launch with kos panel open.
  * revert to launch with kos panel closed.
  * revert to VAB, then relaunch, with kos panel open.
  * revert to VAB, then relaunch, with kos panel closed.
  * new craft in VAB, with kos panel open in VAB, then launch.
  * new craft in VAB, with kos panel closed in VAB, then launch.

The strange behaviors do indeed seem to be gone now.

However, I have not tested (and will not be testing) the associated boot file change that is also in the same PR.  Can someone else test that part and if it's good, merge this?

src/packages/repositories.config
```diff
@@ -1,4 +1,5 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<repositories>
-  <repository path="..\kOS.Safe.Test\packages.config" />
```
erendrake: what OS/filesystem are you using for development?

ZiwKerman: this is maybe from my xamarin/monodevelop

erendrake:
@ZiwKerman Thank you :)

In the future i would prefer to have 1 PR per issue unless multiple issues are very closely related. that way we can test the one change and merge it.

ZiwKerman:
Mea culpa

The changes seemed so small at first :)

PR #771, opened 2015-04-09
Update AGX documentation

SirDiazo:
As per previous discussion, I've gone ahead and revised the AGX integration documentation.

### DIFF
--- a/doc/source/addons/AGX.rst
+++ b/doc/source/addons/AGX.rst
@@ -28,15 +28,15 @@ For Action Groups 11 through 250 there must be an action assigned to the group i
 
 Note that this mod only adds action grousp 11 through 250, it does not change how action groups 1 through 10 behave in any way.
 
-**Known limitations:** 
+**Known limitations (Action groups 11 through 250 only):** 
 
 - For an action group to be useable, it must have an action assigned to it. When installed, AGX adds a "Script Trigger" action to the kOS computer part that serves this purpose if you want an "empty" action group to trigger kOS scripts with. 
-- Be aware that if you query an empty action group, it will always return a state of False and trying to turn an emtpy action group On will do nothing and silently fail without any sort of error message. (Groups AG11 through AG250 only. Groups AG1 through AG10 can be empty and will turn On and Off when commanded to.)
-- At this point, AG11 through AG250 do not support RemoteTech. Triggering those action groups will bypass the signal delay and execute those actions immediately. (On the immediate fix list.)
+- Be aware that if you query an empty action group, it will always return a state of False and trying to turn an emtpy action group On will do nothing and silently fail without any sort of error message. 
+- At this point, AG11 through AG250 do not officially support RemoteTech through kOS. (Support will happen once all three mods involved have updated to KSP version 1.0 and made any internal changes necessary.) All three mods can be installed at the same time without issue, just be aware there may be unexpected behavior when using action groups 11 through 250 from a kOS script in terms of RemoteTech signal delay and connection state.
 
 **Action state monitoring**
 
-Note that the state of action groups is tracked on a per-action basis, rather then on a per-group basis. This results in the group state being handled differently. (AG 11 through AG250 only.)
+Note that the state of action groups is tracked on a per-action basis, rather then on a per-group basis. This results in the group state being handled differently.
 
 - The Script Trigger action found on the kOS computer module is not subject to the below considerations and is the recommended action to use when interacting with a running kOS script.
 - The state of actions are monitored on the part and updated automatically. A closed solar panel will return a state of false for all it's actions. (Extend Panels, Retract Panels, Toggle Panels) When you extend the solar panel with either the Extend Panels or Toggle Panels action, all three actions will change to a state of True. Retract the panels and the state of all three actions will become False. Note that this state will update in any action group that contains that action, not just the action group that was activated.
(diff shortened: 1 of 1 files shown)


erendrake:
@SirDiazo would you consider a change for AGX that changes this behavior based on the presence of KOS?  I admit i have been out of the loop on AGX development for the last few months. It would be ideal to have the API always register the action. even if there are no active parts in it.

Dunbaratu:
If AGX 'grabs' all the actions even the ones that have no parts in them, I think this might be a problem because there's so many AGX actions that they'd be basically stealing the entire keyboard away from the user.  The default keybindings would (I think) get in the way of all the keys all over the keyboard, which is why only the active ones can be enabled.  I'm not a big fan of it, but I see why it has to be that way.

SirDiazo:
@erendrake The issue with that is that the way AGX is setup to work directly with actions, not with action groups, if there is no action assigned to an action group, that group does not exist and so I don't have an "action group" object existing in code to work with.

I've hacked into the action groups 1-10 present in stock KSP to give that status feedback to match a stock kOS installation, but I don't have that available for groups 11-250.

While I have not given it serious thought, at the least I would have to create the "action group" object from scratch in order to monitor it like this. 

Currently, the virtual "action group" presented on screen to the player is simply a list of "AGX Action" objects assigned to that group number so it is nice and straightforward to keep track of everything. I'd rather not be tacking an actual "action group" object onto this and have to keep everything sync'd.

@Dunbaratu The number of keys available for keybinding is actually not an issue. In a default AGX installation, only action groups 1-10 are bound to keys 1-10 the same as stock KSP. Action groups 11-250 have no key bound and can only be activated via mouse-click on the GUI. (Until the player assigns a key to them anyways.)

Which ties back into why there is no "action group" object. As groups 11-250 have no keyboard key assigned and if there is no action assigned to that group (and so does not show in the GUI), the player has no way to interact with the group so why would I create a bunch of empty action group objects that the player has no way of interacting with?

Now, anything is possible in code of course, but for me to do this is not just adding a few lines onto my flight scene GUI, but would require me to stick my fingers into the back-end guts of AGX where the data is stored and I'd really rather avoid that if I can.


Having said that and considering the following:

1) Groups 1-10 do behave as stock so there are those groups available to the player to use as empty action groups to trigger.

2) If the player wants that many empty action groups to use as triggers for scripts, I would expect them to start using the "Script Trigger" action so they could name the action group to remember what activating said group does in their script. As the presence of the Script Trigger object makes the group behave as expected, any work I do to support triggering of empty action groups would not be used in this case.

(Note I have not polled users of kOS to see if they use the Script Trigger in this fashion or not, that is just my personal expectation.)

With those two points in mind, I feel this works well enough as groups 1-10 behave exactly as stock and while groups 11-250 do require adding the Script Trigger action, I'm expecting players to do so anyway as I would think the ability to name your action groups so you can remember what they do in a script is a big plus to the player. (Just a matter of being clear enough in the documentation.)

SirDiazo:
Ugh, wrong button.

Dunbaratu:
I suspect most kOS users don't even know what ScriptTrigger is or what it's for.  The fixes to the docs you're submitting should help in that regard a bit though.

Here's a more elegant solution that I like because forcing users to read walls of text in doc pages is a sure-fire way to make sure most of them won't know about it :

I wonder if it would be possible for the kOS mod's own code to catch the condition of a script program attempting to query the action boolean variable (AG11, AG12, etc) for an AGX action that has nothing assigned to it, and to throw an exception in that case that teaches the user why that can't work and explains things like the ScriptTrigger in its verbose text, and maybe has a URL to the relevant paragraph of the AGX.rst page.

I'm also picturing the case of someone releasing public script code, saying "hey everybody check out my awesome script you can try....".  The users downloading and trying it need to ensure they built their vessels with the same AGX groups assigned or that script won't work.  Such an exception message would help them understand what the problem is.

The only reasons I can see for *not* doing it are:

  * If there are cases where the condition is entirely valid and the script should go on anyway.  (And not be killed by an exception).
  * If it's not easy for kOS to query AGX to find out that there are no assigned actions for that number.  (I assume it is easy but I haven't looked into that code).

If we can query it it might actually be nice for us to provide a query a script can do for "is this action assigned to anything?" as well - and see if we can make that work on the stock groups too (1-10 and 'abort', 'gear', etc).

SirDiazo:
First, for reference, if AGX gets a group state request of an empty action group, it always returns False.
If AGX gets a activate request on an empty action group, it simply does nothing, including not returning an error. This means currently there is no way for kOS to know if the "activate group" command it passed to AGX worked or not.

Note that for groups 1-10, kOS uses stock KSP commands, it only attempts to use AGX for action groups 11+.

Second, AGX does expose a way to check if a group has actions:
|| public static List<AGXAction> AGX2VslGroupActions(uint FlightID, int group); //FlightID of root part of vessel
Which I just realized returns a List<AGXAction> which makes it useless. I fix that when I get a chance.

For now 
||public static List<PartModule> AGX2VslListOfPartModulesInGroup(uint flightID, int i)
will have to do. If the returned list has zero objects, there is no actions in group(int i).


Now, looking at your previous post, I think the biggest thing that needs to be dealt with is the sharing of scripts. 

My biggest objection to just simply adding the "action group" object was the fact I'd have to mess around in the guts of AGX to make sure everything saved correctly. But if it is just a script trigger, does it matter if the group state saves? IE: If everytime a vessel loads, the group state of all empty groups was set to false regardless of their state the last time the vessel was flown, is this a big deal?

If that is acceptable, that is something I would be okay doing as it does not require me to change how AGX handles its data, I can handle it in the flight scene class exclusively.

Dunbaratu:
Hmm.... I think that is fine.  kOS is so clunky right now about how it saves/loads when leaving the scene that people are just in the habit of having to re-run their scripts from scratch when they reload the scene anyway.

The problem is that for kOS to save/load properly, it would have to have a means to serialize the entire state of the computer, memory and all, out to the save file, and bring it back - that's a "pie in the sky" future idea but it's certainly not trivial.  For one thing it means doing deep reference re-assignment all over the place (i.e. this object is a LIST().  It contains 5 thngs.  The third one is a Vessel.  It contains 25 parts.  One of those parts is reference #123456789.... Oh, I mean that is the reference it *used* to have before KSP rebuilt the entire vessel from scratch again as it reloaded the scene.  NOW the part has reference #6916193591... and we have to reassign this reference....)

Therefore, I wouldn't call "it forgets action group states when you reload" to be that big of a deal, not compared to that problem.

SirDiazo:
Hmmm.

In that case, no promises but I'll open AGX and take a look at the code when I get home. (On break at work at the moment.)

Just adding a dummy "action group" object that can be toggled should not be a big deal if I'm not worried about saving.

Issue #746, opened 2015-04-07
on and when in functions does not work


abenkovskii:
this code:
```
declare function test1 {
  when abort then {
    print "test1_trigger".
  }
  wait 10.
}
test1().
```
and this code:
```
declare function test1 {
  on abort {
    print "test1_trigger".
  }
  wait 10.
}
test1().
```
Both in the console and in the file print nothing. No matter how many times I hit the "abort" button
However this code:
```
on abort {
  print "test1_trigger".
}
wait 10.
```
prints "test1_trigger"
and this code:
```
when abort then {
  print "test1_trigger".
}
wait 10.
```
Both in the console and in the file print "test1_trigger" when I press the "abort" button.

erendrake:
this is likely related to the closure work @Dunbaratu is doing?

Dunbaratu:
Not sure - I'll have to have a look at things.

abenkovskii:
code:
```
// file bug.ks
declare function test1 {
  when true then {
    print "test1_trigger".
  }
  wait 10.
}
test1().
set a to 1/0.
```
log:
```
(Filename: C:/BuildAgent/work/d63dfc6385190b60/artifacts/StandalonePlayerGenerated/UnityEngineDebug.cpp Line: 49)

kOS: At bug2 on 1, line 8
set a to 1/0.
          ^

(Filename: C:/BuildAgent/work/d63dfc6385190b60/artifacts/StandalonePlayerGenerated/UnityEngineDebug.cpp Line: 49)

System.DivideByZeroException: Division by zero

  at kOS.Safe.Compilation.CalculatorIntInt.Divide (System.Object argument1, System.Object argument2) [0x00000] in <filename unknown>:0 

  at kOS.Safe.Compilation.OpcodeMathDivide.ExecuteCalculation (kOS.Safe.Compilation.Calculator calc) [0x00000] in <filename unknown>:0 

  at kOS.Safe.Compilation.BinaryOpcode.Execute (ICpu cpu) [0x00000] in <filename unknown>:0 

  at kOS.Execution.CPU.ExecuteInstruction (kOS.Execution.ProgramContext context) [0x00000] in <filename unknown>:0 
 
(Filename: C:/BuildAgent/work/d63dfc6385190b60/artifacts/StandalonePlayerGenerated/UnityEngineDebug.cpp Line: 49)

Code Fragment

File                 Line:Col IP   opcode operand
----                 ----:--- ---- --------------------- 
[...]

PR #772, opened 2015-04-09
Resolves #732 and #746

### LINKED ISSUE #732: runing programs from inside functions
exception
log: http://pastebin.com/Yie6RS6n

Dunbaratu:
These two things were in fact caused by the same exact problem - PreProcessStatements() wasn't recursing down into the children of a DECLARE statement, which meant it never looked inside the bodies of functions for the things that need preprocess work, like the run command, and triggers.

Also, had to split up the order of identifying functions versus identifying triggers, to solve a chicken-and-egg problem trying to nest them inside each other.

### DIFF
--- a/src/kOS.Safe/Compilation/KS/Compiler.cs
+++ b/src/kOS.Safe/Compilation/KS/Compiler.cs
@@ -173,14 +173,9 @@ namespace kOS.Safe.Compilation.KS
         {
             ParseNode rootNode = tree.Nodes[0];
             TraverseScopeBranch(rootNode);
+            IterateUserFunctions(rootNode, IdentifyUserFunctions);
             PreProcessStatements(rootNode);
-            PreProcessUserFunctions(rootNode);
-        }
-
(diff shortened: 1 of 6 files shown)


src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -173,14 +173,9 @@ private void PreProcess(ParseTree tree)
...
             TraverseScopeBranch(rootNode);
-            PreProcessUserFunctions(rootNode);
+            IterateUserFunctions(rootNode, IdentifyUserFunctions);
             PreProcessStatements(rootNode);
-        }
-
-        private void PreProcessUserFunctions(ParseNode node)
-        {
-            IterateUserFunctions(node, IdentifyUserFunctions);
-            IterateUserFunctions(node, PreProcessUserFunctionStatement);
```
Dunbaratu: This was removed because I had to split up these two steps and do one of them BEFORE PreProcessStatements and one of them AFTER PreProcessStatements.

PR #776, opened 2015-04-10
AGX Documentation update, round 2

SirDiazo:
As per conversation earlier, here is how AGX works as of Version 1.31d that I released a minute ago.

https://github.com/SirDiazo/AGExt/releases/tag/1.31d


erendrake:
Thank you very much :-)

Issue #782, opened 2015-04-12
part:uid changes every time I load a craft


abenkovskii:
What the title says. According to [this page](https://github.com/Anatid/XML-Documentation-for-the-KSP-API/blob/master/src/Part.cs#L83|flightID) KSP API exposes a part id that never changes and is stored in the save file. It would be nice to have part:uid that never changes.

erendrake:
The old part UID was removed from the api a few KSP releases ago. When that happened I reimplemented it with a generic hashcode which changes every time because I thought it would be enough to be unique in a mission. 

You could use the part:tag value. As it does not change.

abenkovskii:
Because I'll have to set part:tag using a script and there is no easy way of making part:tag unique i ksp saved game. Also: I sort of want to use part:tag for other purposes than just storing a unique identifier.

erendrake:
You don't have to set it with a script. You can set it with the right click menu in the VAB.

TDW89:
@erendrake This has mostly come up because I was writing a script (with assistance from @abenkovskii ) that tried to link volumes to their respective parts (without going through and doing it all by hand). The script in its current state looks like this:
boot.ks:
```
declare found.
declare fileList.
until found {
 list files in fileList.
 for file in fileList {
  if file = "exec.ks" {
   set found to true.
  }.
 }.
}.
log "idle:add("+mac+")." to idle.ks.
```
exec.ks:
```
declare parameter code.
log "" to exec.tmp.
delete exec.tmp.
log code to exec.tmp.
run exec.tmp.
```
boot_control.ks:
```
wait 0.001. // for physics to load
print "start".
declare controlTag.
set controlTag to "control".
copy exec from 0.
print "copying exec".
declare volList.
declare fileList.
list volumes in volList.
declare vol.
set vol to volList:iterator.
declare cpuList.
set cpuList to ship:modulesnamed("kOSProcessor").
declare cpu.
declare my_mac.
set my_mac to false.
declare mac.
declare found.
declare warning.

print 0.

Switch to 1.
set found to false.
list files in fileList.
for file in filelist {
 if file = "my_uid.ks" {
  set found to true.
 }.
}.
print 1.
if found {
 run my_uid.ks.
 set my_mac to mac.
 set found to false.
 for cpu in cpuList {
  if cpu:part:tag = controlTag and cpu:part:uid = my_mac {
   set found to true.
   break.
  }.
 }.
 if not found {
  print "This isnt your boot script.".
[...]

PR #783, opened 2015-04-12
Should solve #782

abenkovskii:
I tested it:
1. loaded a vessel
2. printed and wrote down `ship:rootpart:uid`
3. leaved flight scene and returned to the same vehicle
4. UID was the same
5. Than I opened persistent.sfs and yes it is the same uid as in the save file

### DIFF
--- a/src/kOS/Suffixed/Part/PartValue.cs
+++ b/src/kOS/Suffixed/Part/PartValue.cs
@@ -34,7 +34,7 @@ namespace kOS.Suffixed.Part
             AddSuffix("FUELCROSSFEED", new Suffix<bool>(() => Part.fuelCrossFeed));
             AddSuffix("TITLE", new Suffix<string>(() => Part.partInfo.title));
             AddSuffix("STAGE", new Suffix<int>(() => Part.inverseStage));
-            AddSuffix("UID", new Suffix<string>(() => Part.uid().ToString()));
+            AddSuffix("UID", new Suffix<string>(() => Part.flightID.ToString));
             AddSuffix("ROTATION", new Suffix<Direction>(() => new Direction( Part.transform.rotation) ));
             AddSuffix("POSITION", new Suffix<Vector>(() => new Vector( Part.transform.position - shared.Vessel.findWorldCenterOfMass() )));
             AddSuffix("TAG", new SetSuffix<string>(GetTagName, SetTagName));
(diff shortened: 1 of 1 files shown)


erendrake:
@abenkovskii this change is fine by me. I have a few comments.

* It doesn't compile .ToString needs to be invoked as a method .ToString(), easy to fix. 
* We have never been explicit about the intended behavior of the UID, it might be nice to make it more clear in the docs, specifically this page.
    * http://ksp-kos.github.io/KOS_DOC/structures/vessels/part.html
* I am a little worried about scripts that depend on a more persistent ID rather than the tagging tool. It might make scripts using it harder to share. If you are needing this to find the part tied to the core i think we should add a binding that lets you get that link more easily. something to the effect of CORE:PART:TAG.

abenkovskii:
@erendrake 
1. oops. There was an accident: for some reason my IDE messed up indention of this line and than another IDE removed trailing white space in empty lines. So I reverted to the original state and edited it with notepad. I forgot to remove `() => `.
2. I'll add the docs.
3.1 Are you talking about scripts hard coded to work only on one specific vessel? Why someone might make such a strange thing? This things should be in the "data files" that program creates when you run it the first time.
3.2 do you mean `volume:part`? Because I don't see any structure named `core`.

erendrake:
3.2, i am talking about a possible new binding that we have talked about for some time. We had not done it because i was not sure how useful it would be. It clearly is :)

abenkovskii:
3.2 `ship:core` would be really handy in some complex situations.
But:
1. It will make the structure `ship` differ form the structure target from `list targets in var.`
2.1 It only solves the problem of linking volume 1 to it's part and for other volumes it still would be quite tricky.
2.2 It will simplify part 2.1 though by giving you an easy check: "Am I about to shutdown myself?"
P.S. Could you explain what did you mean when you said that it will be harder to share scripts using new `part:uid` behavior?

erendrake:
The reason it might be harder to share scripts is if it is abused. Just like any tool. If you hardcode a UID in the middle of the script it might make it impossible to reuse the script. Esp if you couple it with a bad variable name. Just like hardcoding any id in your code. They can be exceedingly brittle. 

I only bring it up to share our goal of having code be portable and easy to share. So that when you come up with a reason you want to rely on the UID, we can talk about what other tools we need to add.

abenkovskii:
@erendrake 
1. I'll add a documentation than.
2 Actually the only use of this feature I can think of is attaching a multiple properties to one part. All other goals can be achieved by other (including not currently existing like `ship:core`) features of kOS.

PR #787, opened 2015-04-13
fixes #786

### LINKED ISSUE #786: ship:control:neutral not giving true/false.
```
print ship:control:neutral. //this gives
kOS.Safe.Utilities.flushable'1[system.Boolean]  //this
```
```
if ship:control:neutral {print false.}. // this gives
cannot cast from source type..... // this
```

erendrake:


src/kOS/Suffixed/FlightControl.cs
```diff
@@ -183,7 +183,7 @@ private void InitializeSuffixes()
 
             //OTHER
             AddSuffix(new[] { "BOUND" }, new SetSuffix<bool>(() => bound, value => bound = value));
-            AddSuffix(new[] { "NEUTRAL" }, new Suffix<Flushable<bool>>(() => neutral));
```
erendrake: im not sure why i thought that would ever work :P

PR #793, opened 2015-04-13
Resolves #750 - the global/local keyword issues.

### LINKED ISSUE #750: Better keywords for global/local control (for locks but also vars)
Proposal: The syntax is getting clunky for differentiating scope, and as I wish to add scoping to LOCK as well, it's getting needlessly verbose.

I propose this instead:

TO and IS synonyms:
----------------------------

We allow "TO" and "IS" to be the same: i.e. *declare x is 10* is the same as *declare x to 10*  This is needed for some of the changes below to keep things making gramattical sense:

We keep using "declare" only in a backward-compatibility sense.  We make the grammar allow it as an alternative extra keyword but prefer the use of "local".

DECLARE, LOCAL, GLOBAL:
----------------------------------------

Saying *declare local* : means you are making a variable local to current scope.  It's the same exact thing as saying just "declare" now does.

Saying *declare global* : means you are explicitly asking for a variable that lives at global scope *instead* of the current local scope.

Saying *local* only : means the same as *declare local*.  The word *declare* is entirely optional when "local" or "global" is used.

Saying *global* only : means the same as *declare global*.  The word *declare* is entirely optional when "local" or "global" is used.

Saying *declare* only : means the same as *declare local*.  The "local" is presumed when left off.

*set* : same as before - it stores in the first variable it can find when starting from localmost and going outward to globalmost, and if it finds none, then and only then does it make a variable, and it's global.

@LAZYGLOBAL:
-----------------------

Using *@LAZYGLOBAL OFF* now has a new meaning:  It still disallows *set* from implicitly making lazy globals, but it also disallows the use of saying *declare* only and having the "local" be presumed.  It now requires the use of either "global" or "local".  You can, however, leave the word "declare" off and just say "local" or just say "global".

PARAMETER:
-------------------

The word "parameter" can be left as-is now, dropping the "declare" if you like, but then like the cases above, an optional "declare" won't break anything.  You can say "declare parameter", "declare local parameter", or "local parameter", or just "parameter" and they are all the same, but the compiler will call it an error to say "global parameter" - parameters cannot be anything but local to where they are declared and saying otherwise is confusing.

FUNCTION:
----------------

The word "function" can be left as-is now, dropping the "declare" if you like, but an optional 'declare' is still allowed and won't break anything. You can say:
```
declare function foo {stuff}
```
or just
```
function foo {stuff}
```
and it means the same thing.

FILE SCOPING:
----------------------

At the moment, file scope and global scope are identical to each other.  With the new GLOBAL keyword, we could make file scope be its own indent level, and thus when declared in file scope, *local x is 1.* would be different from *global x is 1*.

Examples of valid lines of script using the above rules:
----------------------------------------------------------------------

The following would be equivalent and all valid:
```
local x is 0. // this is the version we'd encourage and people would probably prefer anyway.
local x to 0.
declare x to 0.
declare x is 0.
declare local x to 0.
declare local x is 0.
```

The following would be equivalent and all valid:
```
global x is 0. // this is the preferred way and the one people are more likely to use anyway.
global x to 0.
declare global x is 0.
declare global x to 0.
```

The following would be equivalent and all valid:
```
global lock foo to some_expression_here.
declare global lock foo to some_expression_here.
lock foo to some_expression_here. // only equivalent if there is not already a foo lock at some local scope.  Like set, it only makes a global one if there isn't one it finds to override already.
```

The following would be equivalent and all valid:
```
local lock foo to some_expression_here.
declare local lock foo to some_expression_here.
declare lock foo to some_expression_here.
```

The following would be equivalent and all valid:
```
declare parameter foo.
parameter foo.
local parameter foo.
declare local parameter foo.
```

The following would be disallowed:
```
global parameter foo.
declare global parameter foo.
```

### Not as complex to implement as it looks.

Despite how long it takes to explain all the possibilities in every detail, to *implement* this is really just a matter of changing the grammar file to allow the keywords to be optional, and making the compiler set some flags for the statement based on which keywords were actually present.

Why?
------

Issue #741 dealt with giving locks closures but didn't deal with the fact that all locks are global.  The effect is that when you make a lock, it's a lock forever.  Since they now have closures, that means they keep around the entire set of stacked variables in memory somewhere forever.
Consider the following:
```
// make g_list a list of 100 empty vectors:
set g_list to list().
set i to 0. until i >= 100 { g_list:add(v(0,0,0)). set i to i +1}.

declare function foo {
  declare a to 10.

  // make foo_list a local list of 100 empty vectors:
  declare foo_list to list().
  set i to 0. until i >= 100 { foo_list:add(v(0,0,0)). set i to i +1}.

  declare function bar {
    declare b to 20.

    // create a lock indented to be used in bar:
    lock l_bar to a+b.

    // make bar_list a local list of 100 empty vectors:
    declare bar_list to list().
    set i to 0. until i >= 100 { bar_list:add(v(0,0,0)). set i to i +1}.

    // ...Assume there's other code here, that uses the lock bar, and the bar_list and the foo_list .....
  }.
}.

foo().

// assume more code here for the rest of the program...

```
The problem with the
```
    lock l_bar to a+b.
```
being global is that even though it only uses A and B, it still keeps the entire scope alive in memory for both function bar and its containing function, function foo, and because it's global, it keeps that scope alive forever even after the function call foo() is long done.  So the two big lists of vectors (and whatever other locals may exist in foo() and bar() never really go away.  You can't reach them anymore - only l_bar can, but they still take up room in memory that cannot go away.

That's correct behavior for when you *need* that, like when doing a LOCK STEERING and it has to keep being capable of seeing those closure variables indefinitely until you LOCK STEERING to something else.

But there will be cases where you don't need it, and for them we need to provide users with the ability to make their locks local if they like.  Thus the impetus for this issue.

Dunbaratu:
Also found a lot of bugs in how it dealt with local scoping
and when and lock interactions in the compiler.


src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -279,7 +299,7 @@ private void PreProcessOnStatement(ParseNode node)
             PushTriggerRemoveName(triggerRemoveVarName);
             AddOpcode(new OpcodePush(triggerRemoveVarName));
             AddOpcode(new OpcodePush(true));
-            AddOpcode(new OpcodeStore());
+            AddOpcode(new OpcodeStoreGlobal());
```
Dunbaratu: The "remove-trigger-number....." var name can be always global because the number part of the name is the unique hashcode of the instance of the trigger body we're in.  In fact it's better to force it to be global because it shouldn't live in whatever runtime location happened to get interrupted by this trigger.

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -341,7 +361,7 @@ private void PreProcessWhenStatement(ParseNode node)
         /// </summary>
         private string ConcatenateNodes(ParseNode node)
         {
-            return string.Format("{0}{1}", context.NumCompilesSoFar, ConcatenateNodesRecurse(node));
+            return string.Format("{0}{1}{2}", context.NumCompilesSoFar, GetContainingScopeId(node), ConcatenateNodesRecurse(node));
```
Dunbaratu: Added the extra term because basically the attempt to mash together different instances of the same expression into the same function or trigger isn't so reliable when using closures - the same expression in these braces might not REALLY be the same as that expression over there in those other braces becasue they need to have a different scope ID for their locals.  So mostly the entire concept of these hashcodes is nearly pointless now.  We could change the hashcode to just something that simply is unique for "this exact location in the source file of this exact compile instance" and stop paying attention to the contents of the node.

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -442,19 +511,19 @@ private void PreProcessUserFunctionStatement(ParseNode node)
 
                 if (userFuncObject.IsSystemLock())
                 {
-                    AddOpcode(new OpcodePush(userFuncObject.PointerIdentifier));
+                    AddOpcode(new OpcodePush(userFuncObject.ScopelessPointerIdentifier));
```
Dunbaratu: to keep the names unique when looking up the function, the same identifier name gets different values based on the scope ID it was in, so function Foo in global scope gets stored as having identifier "foo`0" and a foo function in, say, the 23rd brace section encountered in the compiler gets stored as identifier "foo`23".  The ScopelessPointerIdentifier strips off that trailing number again to just give you "foo".

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -476,32 +545,35 @@ private void PreProcessUserFunctionStatement(ParseNode node)
...
                 AddOpcode(new OpcodeReturn());
             }
 
             // lock expression's or function body's code
             currentCodeSection = userFuncObject.GetUserFunctionOpcodes(expressionHash);
-            if (isLock) // locks need to behave as if they had braces even though they don't - so they get lexical scope ids for closure reasons:
-                BeginScope(node);
-            if (isDefFunc)
-                nextBraceIsFunction = true;
+            bool secondInstanceSameLock = currentCodeSection.Count > 0;
```
Dunbaratu: The first time the same exact lock expression is found, it fills the code with the expression contents.  The second time the same expression is used, it previously used to pointlessly append the same code again, that would never get reached because it would hit a "return" before encountering it.  This catches that case by seeing if the lock already got code compiled into it.  (the hashcode ensures that it's the same code body from the same scope level).  This works better than the old way, which assumed the lock would be unintialized the first time you encounter it.  That's not true anymore because we have to initialize the locks up front for them to be acessible from trigger bodies.

src/kOS.Safe/Compilation/KS/kRISC.tpg
```diff
@@ -152,7 +153,31 @@ remove_stmt    -> REMOVE expr EOI;
...
+                              (LOCAL|GLOBAL)
+                          )
+                          (
+                              declare_parameter_clause |
+                              declare_function_clause |
+                              declare_identifier_clause |
+                              declare_lock_clause
+                          )
+                      )
+                  );
```
Dunbaratu: Many of the changes to the Compiler.cs file are a consequence of this.  Functions and locks and parameters were moved down a step in the parse tree to put them under a parent generic "declare" rule.  Thus the parent rule deals with the 'global', 'local' and 'declare' keywords (if you look at the above rules, you can see it's still called a declare_stmt even when the 'declare' keyword is omitted and it just says "local" or "global" or none of the above.).

src/kOS.Safe/Utilities/Debug.cs
```diff
@@ -63,5 +64,33 @@ public static List<string> GetPendingNags()
...
+                                               codes[index].SourceColumn ,
+                                               index,
+                                               codes[index].Label ?? "null",
+                                               codes[index] ?? new OpcodeBogus(),
+                                               "DEST: " + (codes[index].DestinationLabel ?? "null" ),
+                                               "" ) );
+            }
+
+            return codeFragment.Aggregate(string.Empty, (current, s) => current + (s + "\n"));
+        }
```
Dunbaratu: I moved the above function out of kOS Utilities into kOS.Safe Utilities so I could call it from places in the compiler during my checking.  It seemed like a good move so I left it there in my PR.

Issue #652, opened 2015-03-05
Direction documentation doesn't explain :inverse suffix


Dunbaratu:
On:
http://ksp-kos.github.io/KOS_DOC/math/direction.html

The :inverse suffix for a Direction is mentioned in the summary list, but then it doesn't link down into anything in the detail section below.  The only documentation for it that people can see is the one-line summary and that's it.

PR #794, opened 2015-04-13
Added :inverse detailed section

TDW89:
I think I have done this right. Let me know if I have set this up wrong.
Documentation for issue #652

### DIFF
--- a/doc/source/math/direction.rst
+++ b/doc/source/math/direction.rst
@@ -190,7 +190,7 @@ Structure
     :type: :struct:`Direction`
     :access: Get only
     
-    :struct: The reverse of a `Direction`. So `PROGRADE:INVERSE` is the same as `RETROGRADE`.
+    :struct: Gives a `Direction` with the opposite rotation around its axes. So `r(30,15,5)` becomes `r(-30,-15,-5)`.
     
 
 .. note:: **The difference between a :struct:`Direction` and a ``Vector``**
--- a/doc/source/math/direction.rst
+++ b/doc/source/math/direction.rst
@@ -185,6 +185,13 @@ Structure
 
     :struct:`Vector` of length 1 that is in the same direction as the "starboard side" of this Direction.  Note that it is the same meaning as "what the X axis of the universe would be rotated to if this rotation was applied to the basis axes of the universe". When you LOCK STEERING to a direction, that direction's STARVECTOR is the vector the right wing of the ship will orient to.  SHIP:FACING:STARVECTOR is the way the ship's right wing is aimed right now.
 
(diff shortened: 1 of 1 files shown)


doc/source/math/direction.rst
```diff
@@ -185,6 +185,13 @@ Structure
 
     :struct:`Vector` of length 1 that is in the same direction as the "starboard side" of this Direction.  Note that it is the same meaning as "what the X axis of the universe would be rotated to if this rotation was applied to the basis axes of the universe". When you LOCK STEERING to a direction, that direction's STARVECTOR is the vector the right wing of the ship will orient to.  SHIP:FACING:STARVECTOR is the way the ship's right wing is aimed right now.
 
+.. attribute:: Direction:INVERSE
+
+    :type: :struct:`Direction`
+    :access: Get only
+    
+    :struct: Gives a `Direction` with the opposite rotation around its axes. So `r(30,15,5)` becomes `r(-30,-15,-5)`.
+
```
Dunbaratu: Is this true?  I was unsure.  (remembering of course, that if it says, say, 330 that does mean -30 - I think Rots get printed clamped into the 0-360 range so you never see negative numbers when you print them (i.e. -10 prints as +350).)

but I don't remember if this is really true - has this claim been tested?

TDW89: I didn't test whether Rots ever get printed as negative but i did test it accepting those rotations. Have just done more testing on them and using `vang` there is a 0.5deg (might be floating point error - might not) between `r(30,15,5):inverse:vector` and `r(-30,-15,5):vector`. There is also a similar difference between them and `r(330,345,355):vector`.
Also `print r(30,15,5):inverse. gives `R(329.756,345.538,2.968)` the first 2 are close enough that i just assumed floating point but the `2.968` makes me uncertain.

Dunbaratu: The best test to be sure is to print them from a ship in orbit, time warp about 1/4 of the way around the planet, and try again.  That will ensure whether or not the axes rotating will mess up the result or not.

TDW89: just warped round kerbin running this script:
```
Until false {
 set dir1 to vecdrawargs(v(0,0,0),r(30,15,5):inverse:vector,red,"Dir1",5,true).
 set dir2 to vecdrawargs(v(0,0,0),r(-30,-15,-5):vector,green,"Dir2",5,true).
}.
```
Both arrows stayed over lapping the whole way round.

Dunbaratu: I think no example is better than a misleading one.  I'd say remove it.

Issue #685, opened 2015-03-24
Minor, Documentation: There is no explanation of SHIP:STATUS


ZiwKerman:
In "Bound Variables" page STATUS is referred to SHIP:STATUS, but on Vessel page of the documentation there is no mention of it. In fact there is no mention of SHIP:STATUS anywhere else and it is never explained.

TDW89:
Was about to create documentation for this but noticed that `ship:status` doesn't work. I can `print status.` but I can't `print ship:status.` or `print target:status.`.

Dunbaratu:
Added "bug" tag to reflect the fact that ship:status is missing.

TDW89:
Thought I would add this here as its on the same doc page.
The table of suffixes is missing `:TYPE` it is however included in the detailed explanations below.

PR #798, opened 2015-04-14
Fix for #685

TDW89:
Looks like i don't understand github that well (cancelled the last PR after updating my fork got added to the PR)


erendrake:
the PR is always looking at the HEAD of the branch that you specify. so you cant share a branch for two PRs

PR #800, opened 2015-04-14
update fork

TDW89:


TDW89:
sorry got the fields the wrong way round

PR #810, opened 2015-04-14
Fixes #779 lock steering in triggers

### LINKED ISSUE #779: lock steering and lock throttle broke in triggers
It looks like something I pulled from upstream today broke using `lock steering` and `lock throttle` from inside a `when ... then ...` trigger.  The version of kOS that I compiled yesterday morning still works fine.  This morning I compiled from the latest "develop" source, and now my launch script stalls goes straight up and doesn't change orientation (I use triggers to change the lock based on altitude and orbital parameters).  No errors are thrown, it just doesn't turn or adjust the throttle while the script continues to process.  When it gets to the point where it executes a node to circulize, it can steer the ship and adjust throttle just fine (this section does not use triggers).  The issue persisted through recompiling the code again.

I pulled the git source yesterday around 8:30AM (EDT) and today around 10:00AM (EDT), so the change happened some where between those times.  I'm guessing it has something to do with the preparsing changes, but I can't be sure.  I don't see any errors in the debug log surrounding the lock commands.

Dunbaratu:
Moved the compilation around such that it will
set up the triggers properly even when the
lock is only mentioned in another trigger.

### DIFF
--- a/src/kOS.Safe/Compilation/KS/Compiler.cs
+++ b/src/kOS.Safe/Compilation/KS/Compiler.cs
@@ -397,12 +397,14 @@ namespace kOS.Safe.Compilation.KS
                 bodyNode = lastSubNode.Nodes[2];
             }
             else
-                return; // not one of the types of statement we're really meant to run IdentifyLocks on.
+                return; // not one of the types of statement we're really meant to run IdentifyUserFunctions on.
             
             UserFunction userFuncObject =
                 context.UserFunctions.GetUserFunction(funcIdentifier, storageType == StorageModifier.GLOBAL ? (Int16)0 : GetContainingScopeId(node), node);
             int expressionHash = ConcatenateNodes(bodyNode).GetHashCode();
             userFuncObject.GetUserFunctionOpcodes(expressionHash);
+            if (userFuncObject.IsSystemLock())
+                BuildSystemTrigger(userFuncObject);
         }
         
         /// <summary>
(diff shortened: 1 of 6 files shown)


src/kOS/Suffixed/VesselTarget.cs
```diff
@@ -422,7 +422,7 @@ private void InitializeSuffixes()
                 () => System.Math.Min(Vessel.distanceLandedPackThreshold, Vessel.distancePackThreshold), 
                 value => { Vessel.distanceLandedPackThreshold = Vessel.distancePackThreshold = value; }));
             AddSuffix("ISDEAD", new NoArgsSuffix<bool>(() => (Vessel.state == Vessel.State.DEAD) ));
-            AddSuffix("STATUS", new Suffix<String>(() => shared.Vessel.situation.ToString()));
+            AddSuffix("STATUS", new Suffix<String>(() => Vessel.situation.ToString()));
```
Dunbaratu: I thought this fix had already been pulled into develop?  I find it strange that this is showing up as a diff.

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -1762,7 +1763,8 @@ public override void PopulateFromMLFields(List<object> fields)
 
         public override void Execute(ICpu cpu)
         {
-            cpu.PushStack(cpu.MakeUserDelegate(EntryPoint));
+            IUserDelegate pushMe = cpu.MakeUserDelegate(EntryPoint);
+            cpu.PushStack(pushMe);
```
Dunbaratu: This got split into two lines because I was doing a debug print of the pushMe value.  It doesn't hurt to leave it like this, but I thought I'd mention why just so it wasn't confusing.

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -1165,6 +1165,8 @@ public override void Execute(ICpu cpu)
             if (Direct)
             {
                 functionPointer = cpu.GetValue(Destination);
+                if (functionPointer == null)
+                    throw new KOSException("Attempt to call function failed - Value of function pointer for " + Destination + " is null.");
```
Dunbaratu: This is one of those "the user should never see this message" types of errors.  I added it to help me force a crash right at the spot where the problem first occurred, so I could get the dump of what the stack was like at that point.  But I decided to leave it in, because if this ever DOES happen, well, its really not possible without some kind of kOS bug and I'd like to know about it.

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -575,10 +561,37 @@ private void PreProcessUserFunctionStatement(ParseNode node)
...
         }
         
+        
+        /// <summary>
+        /// Build the system trigger to go with a user function (lock)
+        /// such as LOCK STEERING or LOCK THROTTLE
+        /// </summary>
+        /// <param name="func">Represents the lock object, which might not be fully populated yet.</param>
+        private void BuildSystemTrigger(UserFunction func)
+        {
```
Dunbaratu: This logic used to be occurring when preparsing the lock statement, the very first time it encountered this particular system variable being bound.  But now that it's possible to visit things out of order (where the triggers get compiled before the locks do), it's important to get this code ready up front at the start of the compile regardless of what order things happen in.  This is the trigger that sets boundvariable $STEERING and $THROTTLE once per FixedUpdate by executing the function in the lock $steering* and $throttle*.  Without it being in place, flybywire cannot work.

hvacengi:
I'll download the source tonight and test out on my machine as well.  At this point, I think my script for automatic launch and dock may test every case of locks and user functions ;-)

erendrake:
@hvacengi if you have a second would you mind sending me an email? I would like to ask you a question :)

PR #827, opened 2015-04-15
upped config:ipu default for new users to 200.

Dunbaratu:


erendrake:
do we think that 200 is enough?

Dunbaratu:
No clue.  It's what I needed to make my previously running program keep working and stay within the trigger limit.  Chances are it's a bit more than is necessary.  You can make it bigger if you like.  I have no clue.  Any any benchmarking I did would only answer the question "how fast is it on MY computer", rather than give a general case answer.

PR #832, opened 2015-04-15
Fixes #676 no more second error on syntax errors.

### LINKED ISSUE #676: Weird additional error after syntax error - tries to re-call FunctionCall.Execute().
I keep getting this extra second error message after a syntax error in a program I run:
Whatever the real syntax error is always get followed by a second exception - a bogus phantom nullref exceptiom.  Here's an example:
```
switch to 0.
run functest1.
Syntax: Unexpected token '.' found. Expected TO at line 6
declare local_x.
               ^
(Cannot Show kOS Error Location - error might really be internal. See kOS devs.)
At interpreter history, line 3
run functest1.
    ^
```
The "unexpected token '.' found. is a correct error message.  But the "nullref" that follows is bogus.

After sprinkling the code with some debug printouts to narrow down where the problem is, I've discovered that after the syntax error, something in the system is attempting to RE-execute the FunctionRun.Execute() method a second time, and it's that second call that throws the nullref.  I don't know what cause it to try to run a second time, but it shouldn't be doing that.

It's unrelated to what I'm working on now so I'm just mentioning it in passing and moving on.

Dunbaratu:
When I examined it, I saw that this is how it's
meant to work in normal cases:

There are two instruction pointers - one for where
you are in the interpreter context and another for
where you are in the program context.

When you run a program, it does this:

  1. Interpreter context has an OpcodeCall that calls built-in function run().
  2. Function run() tells CPU it should switch to program context on the
     next instruction.
  3. Function run() calls compiler, and populates the CPU's program context
     with the resulting program.
  4. Function run() finishes.
  5. CPU, like it does for all opcode executions, updates the instruction
     pointer of the current context to the next opcode.
  6. CPU finishes loop iteration, and begins new loop.
  7. CPU now is in the new program context, and is running the program.
  8. When program finishes or errors out, CPU returns to interpreter context and picks up where it left the instruction pointer for the interpreter.

The problem is when step 3 barfs and throws an exception, the CPU was
*already* told in step 2 that it is no longer going to be in interpreter
context, so when it handles the error it's handling it as if it had come
from a program instead of from the interpreter, and thus is does NOT advance
the interpreter's instruction pointer, thinking that had already been
done before in step 5 of the previous iteration just before it had
jumped into the "program" it thought it was running.

Thus the interpreter tries to run the same Opcode a second time, and gets
the new error because the stack was already popped in the first attempt
to call the run() function.  The second time it tries to call run(), the
args it needs aren't there on the stack anymore, thus the null ref error.

The fix was to detect when the compiler failed in step 3, and when it did,
to undo the "you are now in program context" setup it did in step 2, so
the CPU knows it's still in interpreter context and it does therefore need
to advance the interpreter's instruction pointer.

### DIFF
--- a/src/kOS/Execution/CPU.cs
+++ b/src/kOS/Execution/CPU.cs
@@ -112,7 +112,7 @@ namespace kOS.Execution
             else {
                 shared.ScriptHandler.ClearContext("program");
 
-                var programContext = ((CPU)shared.Cpu).GetProgramContext();
+                var programContext = ((CPU)shared.Cpu).SwitchToProgramContext();
                 programContext.Silent = true;
                 var options = new CompilerOptions { LoadProgramsInSameAddressSpace = true };
                 string filePath = shared.VolumeMgr.GetVolumeRawIdentifier(shared.VolumeMgr.CurrentVolume) + "/" + "boot";
@@ -238,7 +238,7 @@ namespace kOS.Execution
             return contexts[0];
         }
         
-        public ProgramContext GetProgramContext()
+        public ProgramContext SwitchToProgramContext()
         {
             if (contexts.Count == 1)
             {
--- a/src/kOS/Function/Misc.cs
+++ b/src/kOS/Function/Misc.cs
@@ -179,7 +179,7 @@ namespace kOS.Function
                 shared.ScriptHandler.ClearContext("program");
                 string filePath = shared.VolumeMgr.GetVolumeRawIdentifier(shared.VolumeMgr.CurrentVolume) + "/" + fileName ;
                 var options = new CompilerOptions {LoadProgramsInSameAddressSpace = true, FuncManager = shared.FunctionManager};
-                var programContext = ((CPU)shared.Cpu).GetProgramContext();
+                var programContext = ((CPU)shared.Cpu).SwitchToProgramContext();
 
                 List<CodePart> codeParts;
                 if (file.Category == FileCategory.KSM)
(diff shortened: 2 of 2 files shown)


src/kOS/Execution/CPU.cs
```diff
@@ -238,7 +238,7 @@ public ProgramContext GetInterpreterContext()
             return contexts[0];
         }
         
-        public ProgramContext GetProgramContext()
+        public ProgramContext SwitchToProgramContext()
```
Dunbaratu: I renamed this because it took me a while to find the problem, and part of the reason it took so long was that the original naming convention of "Get" ProgramContext hid the fact that actually there's a major side effect hiding inside this "get" method: Not only does it GET the program context, but it also tells the cpu to switch to starting to USE that program context as its next instruction it's going to execute after the current one is done.

That is crucial to how run() works.  I kept running around in circles trying to find "yes I see that run() loads the program, but where on earth does it tell the CPU to jump to start running it?  I can't find it anywhere!"

And the reason I couldn't find it is that it was hidden inside this "Get" method as a secret side-effect.  Thus I figured we really should rename it to make it clear it's not just doing a "Get".

erendrake: or we should break it into two methods that each have one job

Dunbaratu: Maybe.  I wanted to hurry up and get the bug fixed.  The rename was a good first step that doesn't feel like doing a refactor but at least "un-hides" the effect the method has.

PR #838, opened 2015-04-16
Set sasmode documentation - issue #837

### LINKED ISSUE #837: document setting the stock sasmode
I am working on documentation for the stock sasmode control to be included with the 0.17 release.

hvacengi:
Updated the SAS warning in flight.rst according to new SAS behavior.
Added the explaination for sasmode to systems.rst.  Listed right below
the section for toggling sas itself to keep connected thoughts together.

This should close #837 on merge.


space-is-hard:
Quick note:

SAS direction hold will vary depending on whether the navball is in surface or orbit mode. Can we make a way to get/set the navball mode if that feature doesn't currently exist?

hvacengi:
That feature does not currently exist.  However I knew the question was going to come up.  I honestly don't know if it is possible.  I'll look into it.  Maybe I can find it before release, otherwise I think it can go in 0.17.1.

PR #846, opened 2015-04-17
coerced some more degrees into reasonable range.

Dunbaratu:
Fixes the case when longitude is reported wrapped around
the world several times, but Latitude can still be
wonky and record things on the
wrong side of the world i.e. report 60 north as
instead 120 north from a longitude on the other
side of the world, with a latitude so big that it
is wrapping up and over the north pole to the
other side.  That's a nontrivial fix so it's still
being left that way so at least kOS won't report
false data - just ... report true data in a stupid
way like the main game does.

### DIFF
--- a/src/kOS/Suffixed/GeoCoordinates.cs
+++ b/src/kOS/Suffixed/GeoCoordinates.cs
@@ -7,8 +7,30 @@ namespace kOS.Suffixed
 {
     public class GeoCoordinates : Structure
     {
-        public double Lat { get; private set; }
-        public double Lng { get; private set; }
+        private double lat;
+        private double lng;
+        public double Lat
+        {
+            get
+            {
+                return Utils.DegreeFix(lat,-180);
+            }
+            private set
+            {
+                lat = value;
+            }
+        }
+        public double Lng
+        {
+            get
+            {
+                return Utils.DegreeFix(lng,-180);
+            }
+            private set
+            {
+                lng = value;
+            }
+        }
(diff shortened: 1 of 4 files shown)


src/kOS/Suffixed/GeoCoordinates.cs
```diff
@@ -51,10 +73,10 @@ public GeoCoordinates(CelestialBody body, SharedObjects sharedObj, double lat, d
         ///   Build a GeoCoordinates from any arbitrary lat/long pair of floats.
         /// </summary>
         /// <param name="sharedObj">to know the current CPU's running vessel</param>
-        /// <param name="lat">latitude</param>
-        /// <param name="lng">longitude</param>
-        public GeoCoordinates(SharedObjects sharedObj, float lat, float lng) :
-            this(sharedObj.Vessel.GetOrbit().referenceBody, sharedObj, lat, lng)
+        /// <param name="latitude">latitude</param>
+        /// <param name="lnongitude">longitude</param>
```
erendrake: spelling

src/kOS/Suffixed/GeoCoordinates.cs
```diff
@@ -7,8 +7,30 @@ namespace kOS.Suffixed
 {
     public class GeoCoordinates : Structure
     {
-        public double Lat { get; private set; }
-        public double Lng { get; private set; }
+        private double lat;
+        private double lng;
+        public double Lat
```
erendrake: could we just rename these to their full names?

PR #850, opened 2015-04-18
Set sasmode documentation (continues #837)

hvacengi:
Add RemoteTech Warning

Continues #837


Dunbaratu:
I confirmed that these changes are already present in develop due to my other merge and github is just being too boneheaded to detect that when I merged this into my branch and then my branch into develop, that this means this is merged already.

PR #880, opened 2015-04-20
removed the print message from GetPosition

Dunbaratu:
ziw - can you test this?  I'm not in a position to be able to run the KSP game right now.


ZiwKerman:
Spam is gone.

PR #901, opened 2015-04-27
Fix keyword lexing.  Resolves #900

### LINKED ISSUE #900: [bug] Variables with names starting with "not" interpreted as "NOT <rest of var>"
SET note TO "some text".
    PRINT note.
    Undefined Variable Name 'e'.
    PRINT notasdasd.
    Undefined Variable Name 'asdasd'.

Relevant sections of stack trace: 

    1/planeautopilot.ks    91:9   0753 push $<argstart>
    1/planeautopilot.ks    91:18  0754 push $e 
    1/planeautopilot.ks    91:18  0755 not <<--INSTRUCTION POINTER--
    1/planeautopilot.ks    91:18  0756 call print() 

    interpreter history    70:1   0877 push $<argstart> 
    interpreter history    70:10  0878 push $asdsdsd 
    interpreter history    70:10  0879 not <<--INSTRUCTION POINTER--
    interpreter history    70:10  0880 call print()

This will occur with any variable name starting with NOT. Does not occur with variable names starting with OR or AND.

Dunbaratu:
I don't know why TinyPG doesn't take the longest possible regex match and instead picks a shorter token when available, but at any rate I fixed it by using the regex special term "\b" to indicate a word boundary, putting it around all the keywords like 'not'.  

So, for example, ``laforge`` won't be taken as the keyword ``for`` because it's now defined as ``\bfor\b``.


erendrake:
Should we be changing this in TinyPG? is this just a stopgap?

Dunbaratu:
I don't want to touch tinyPG code if I don't have to.  This is too complex a feature to quickly fix.   It would be a rabbit hole I'd be stuck in a long time to try to work out the logic of what its lexxer is doing.  I think this is fine.

PR #907, opened 2015-04-27
1.0 release

erendrake:
This is the WIP for getting kOS released with 1.0

PR #909, opened 2015-04-27
Core binding

erendrake:
This is an addendum on top of #907


abenkovskii:
Reminder: `print core.` prints "structure"

PR #910, opened 2015-04-27
first draft of navball mode

erendrake:
This allows you to set the current navball mode, this was started to allow you to correctly select the correct autopilot mode.

### DIFF
--- a/src/kOS/Binding/MissionSettings.cs
+++ b/src/kOS/Binding/MissionSettings.cs
@@ -1,4 +1,5 @@
 using kOS.Safe.Binding;
+using kOS.Safe.Exceptions;
 using kOS.Suffixed;
 using kOS.Suffixed.Part;
 using kOS.Utilities;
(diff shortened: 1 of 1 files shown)


hvacengi:
Two issues when I checked this code:
1) the string in the setter is case sensitive,  So "orbit" returns an argument is out of range exception, while "ORBIT" does not.
2) it didn't actually toggle the mode on my nav ball, and the gettter returns the value that corresponds with what the navball shows.
screenshot34

erendrake:
i guess i was changing the mode, just not the text. Thank you for trying it out. it should work better now :)

Dunbaratu:
I'm holding off on merging this because of the label "Pending Documentation".

erendrake:
yeah, i just havent gotten back to this. i should have time soon.

erendrake:
im going to have to work on this some more.

PR #915, opened 2015-04-27
aggregate resource doesnt cache anymore

### LINKED ISSUE #914: AggregateResources don't update amount on fuel transfer
If you set up a loop to watch the amount value on each AggregateResource as you transfer fuel from one element to another, the amount never reads differently from the amount that was in the container when the AR was instantiated.

I suspect that this is because the struct is created by reading the values of child parts, without updating them after each FixedUpdate.

If this proves to be too large of a processor overhead, it might be worth not fixing it or providing a direct query method which allows a user to opt in to the overhead only when they need to update the value.

erendrake:
fixes #914


hvacengi:
I can confirm that this is working in my refueling script.  If no one else has testing issue, this should close #914

Issue #919, opened 2015-04-28
Part not attachable.


zanderzhng:
KR-2042 b scriptable control system's bottom surface is not attachable.
using v0.17.2 and ksp1.0

ZiwKerman:
This should be a simple part.cfg fix. I'll try to do it right now.

PR #920, opened 2015-04-28
Fix part attach node

ZiwKerman:
#919


erendrake:
this was closed by removing the `1.0_release` branch, the changes did make it into develop so it is resolved.

PR #924, opened 2015-04-29
Fixes #923 by making return capable of popscope.

### LINKED ISSUE #923: Locks lose scope before returning, breaking the return val and breaking closure.
A typical lock, like ```lock steering to mysteer.```, if declared in a nested scope context such as in a function call, retains a closure of the variables of that scope, that gets pushed back on the stack as the function is called, and popped back off again afterward, like so:

lock steering to mysteer does this 'function':
```
pushscope 21 20
push $mysteer
popscope 1
return
```
The closure trick is that the return is going to try to eval the "$mysteer" atop the stack, to read its value and push the resulting R() back on top, so the caller can read it without needing local var $mysteer in scope.  That's fine except for a problem: it's doing that AFTER the popscope already occurred, so $mysteer has already gone out of scope by then. The closure is already invalid, and the return statement barfs, claiming identifier $mysteer doesn't exist.

To fix, we either have to perform an eval prior to the popscope, or roll the popscope into the return opcode so a return statement can also popscope in one statement - that way ``return`` can eval the top of the stack first, and THEN popscope implicitly internally.

@ZiwKerman had a script that first triggered the problem.

Dunbaratu:
Now it is possible to have an OpcodeReturn that does its own
popscope, AFTER it evals the thing atop the stack.  This is
now used in:

   returning from locks.
   explicit return statements from user functions like so:
```
    function foo {
      local val is 3.
      if true {
        return val.
      }
      return 0.
    }
```

When you call that, now, the return val will perform its own internal
popscope 2 (because it's jumping outside of both the if and the
function's braces), and the return 0 will perform its own internal
popscope 1 (because its outside the if but still inside the function's
braces).

This is as opposed to before, where it would do it in two separate
steps, first doing a popscope, and then doing a return.

So for example, the above code's return val will now do this:

```
    push $val
    return 2 depth
```

Where what it used to do was this:
```
    push $val
    popscope 2
    return
```

When looking at an opcode dump, the format for the new return statement
in the "assembly language" will be:
``return`` number ``depth``.

Under the hood, that means it was called with constructor:

  new OpcodeReturn(number).

### DIFF
--- a/src/kOS.Safe/Compilation/Opcode.cs
+++ b/src/kOS.Safe/Compilation/Opcode.cs
@@ -1401,32 +1401,64 @@ namespace kOS.Safe.Compilation
         }
     }
     
+    /// <summary>
+    /// Returns from an OpcodeCall, popping a number of scope depths off
+    /// the stack as it does so.  It evals the topmost thing on the stack.
+    /// to remove any local variable references and replace them with their
+    /// current values, and then performs the equivalent of a popscope, then
+    /// jumps back to where the routine was called from.
+    /// It also checks to ensure that the argument stack contains the arg
+    /// bottom marker.  If it does not, that proves the number of parameters
+    /// consumed did not match the number of arguments passed and it throws
+    /// an exception (to avoid stack misalignment that would happen if it
+    /// tried to continue).
+    /// </summary>
     public class OpcodeReturn : Opcode
     {
         protected override string Name { get { return "return"; } }
         public override ByteCode Code { get { return ByteCode.RETURN; } }
+        
+        [MLField(0,true)]
+        public Int16 Depth { get; private set; } // Determines how many levels to popscope.
 
+        public override void PopulateFromMLFields(List<object> fields)
+        {
+            // Expect fields in the same order as the [MLField] properties of this class:
+            if (fields == null || fields.Count<1)
+                throw new Exception("Saved field in ML file for OpcodeCall seems to be missing.  Version mismatch?");
+            Depth = (Int16)fields[0];
+        }
+        
+        // Default constructor is needed for PopulateFromMLFields but shouldn't be used outside the KSM file handler:
+        private OpcodeReturn()
+        {
+        }
+        
+        /// <summary>
+        /// Make a return, telling it how many levels of the scope stack it should
+        /// be popping as it does so.  It compbines the behavior of a PopScope inside
+        /// itself, AFTER it reads and evaluates the thing atop the stack for return
+        /// purposes (that way it evals the top thing BEFORE it pops the scope and forgets
+        /// what variables exist).<br/
+        /// <br/>
+        /// Doing this:<br/>
+        ///   push $val<br/>
+        ///   return 2 deep<br/>
+        /// is the same as this:<br/>
+        ///   push $val<br/>
+        ///   eval<br/>
+        ///   popscope 2<br/>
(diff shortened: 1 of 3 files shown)


src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -1401,32 +1401,64 @@ public override string ToString()
...
+    /// the stack as it does so.  It evals the topmost thing on the stack.
+    /// to remove any local variable references and replace them with their
+    /// current values, and then performs the equivalent of a popscope, then
+    /// jumps back to where the routine was called from.
+    /// It also checks to ensure that the argument stack contains the arg
+    /// bottom marker.  If it does not, that proves the number of parameters
+    /// consumed did not match the number of arguments passed and it throws
+    /// an exception (to avoid stack misalignment that would happen if it
+    /// tried to continue).
+    /// </summary>
```
Dunbaratu: Half of what this is talking about was already true before, but as OpcodeReturn was getting complex now I decided it was best to document all of what it's doing.

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -1401,32 +1401,64 @@ public override string ToString()
...
-            {
-                cpu.PopAboveStack(1);
-                peeked = cpu.PeekRaw(-1, out okay) as VariableScope;
-            }
-            object shouldBeContextRecord = cpu.PopAboveStack(1);
-            if ( !(shouldBeContextRecord is SubroutineContext) )
-            {
-                // This should never happen with any user code:
-                throw new Exception( "kOS internal error: Stack misalignment detected when returning from routine.");
-            }
```
Dunbaratu: This had to be moved further down because the SubroutineContext on the abovestack still has the scope frames in the way if the popscope is no longer happening before the return.  It has to wait until the internal "popscope" inside Return has happened before it tries this check, else it produces false positives.

src/kOS.Safe/Compilation/ProgramBuilder.cs
```diff
@@ -99,7 +99,7 @@ protected virtual void AddEndOfProgram(CodePart linkedObject, bool isMainProgram
             else
             {
                 linkedObject.MainCode.Add(new OpcodePush(0)); // all Returns now need a dummy return value on them.
-                linkedObject.MainCode.Add(new OpcodeReturn());
+                linkedObject.MainCode.Add(new OpcodeReturn(0));
```
erendrake: we are now not using the parameterless constructor for `OpcodeReturn` i would remove it but there is a comment to not do so.

Dunbaratu: That's correct.  The KSM builder used for COMPILE needs all Opcodes to have a default constructor, so it can make a "dummy" of the opcode during load time to query its attributes (Because C# can't do proper overrides of things without an actual instance present, none of the derived classes of Opcode can have their attributes quered in a static state.  An instance needs to exist in order to check its attributes.  So it has to be able to make a dummy of each just long enough to use Reflection on it and find them, then it throws it away.  That's why they all need a default constructor.

If the default constructor isn't supposed to be used "for real", then make it private so that ONLY the KSM reflection code can use it (which I think is the case for OpcodeReturn.  I made it private so the compiler would help find for me all the places it was being used (like the one above) so I can edit them ).

PR #929, opened 2015-04-29
added version:build

TDW89:
#928

### DIFF
--- a/doc/source/bindings.rst
+++ b/doc/source/bindings.rst
@@ -239,6 +239,7 @@ Returns values about kOS and hardware
     PRINT VERSION.            // Returns operating system version number. i.e. 0.8.6
     PRINT VERSION:MAJOR.      // Returns major version number. e.g. 0
     PRINT VERSION:MINOR.      // Returns minor version number. e.g. 8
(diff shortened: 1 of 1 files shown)


doc/source/bindings.rst
```diff
@@ -239,6 +239,7 @@ Returns values about kOS and hardware
     PRINT VERSION.            // Returns operating system version number. i.e. 0.8.6
     PRINT VERSION:MAJOR.      // Returns major version number. e.g. 0
     PRINT VERSION:MINOR.      // Returns minor version number. e.g. 8
+    PRINT VERSION:BUILD.      // Returns build version number. e.g. 2
```
erendrake: we have an example above `i.e. 0.8.6` then here we use `2` as the build number when we used `0` and `8` for major and minor. It would be nice to have them the same to make it clear that we are talking about part of the above concept.

doc/source/bindings.rst
```diff
@@ -239,6 +239,7 @@ Returns values about kOS and hardware
     PRINT VERSION.            // Returns operating system version number. i.e. 0.8.6
```
erendrake: this i.e. should be an e.g.

PR #931, opened 2015-04-30
Fixes #925 and changes how lowercasing gets done.

### LINKED ISSUE #925: List with a single comma string value sometimes turns uppercase strings to lower
This came originally from a more complex example encountered by @tdw89.  I've managed to trim it down to this as a minimal example that causes it:

The following program operates as you'd expect:
```
local L1 is list("a","b","c","d", "e", "f").
local L2 is list("A","B","C","D", "E", "F").
print L1[1].
print L2[1].
```
It prints:
```
b
B
```
Like it should.

But now, edit the first line so that the 'f' is a comma instead, like so:
```
local L1 is list("a","b","c","d", "e", ",").
local L2 is list("A","B","C","D", "E", "F").
print L1[1].
print L2[1].
```
Now it prints this instead:
```
b
b
```

Behind the scenes, it really is genuinely doing a `push b` instead of a `push B` when building the args to the LIST() constructor.  Something about that single comma confuses everything so it starts treating the uppercase strings as lowercase after that.

But weirdly, it is very touchy and specific.  If you alter that example too much, removing terms from it, or making it not be a list anymore, the problem does not surface.

Dunbaratu:
Also required moving how ALT:APOAPSIS and ETA:APOAPSIS
type strings work.

Technically, ALT and ETA are now objects which should mean
they get documented, but as their purpose was to refactor the old
code to keep working... that documentation isn't vital and can
be a separate issue for later.


erendrake:
Did you want to include a test script for this?

Dunbaratu:
Too weirdly specific tor a test script, and it's just not going to come up again unless we decided to go back to he old way of doing it.  It's not a problem that would accidentally pop back up again somehow, and any test for it would end up testing JUST the cases that are no longer relevant.  I tested the case we had here, and it was fixed, but there's no point in archiving the test for re-use later.

src/kOS/Binding/FlightStats.cs
```diff
@@ -12,15 +12,17 @@ public class FlightStats : Binding
     {
         public override void AddTo(SharedObjects shared)
         {
-            shared.BindingMgr.AddGetter("ALT_APOAPSIS", () => shared.Vessel.orbit.ApA);
-            shared.BindingMgr.AddGetter("ALT_PERIAPSIS", () => shared.Vessel.orbit.PeA);
-            shared.BindingMgr.AddGetter("ALT_RADAR", () => Convert.ToDouble(shared.Vessel.heightFromTerrain > 0 ? Mathf.Min(shared.Vessel.heightFromTerrain, (float)shared.Vessel.altitude) : (float)shared.Vessel.altitude));
+            shared.BindingMgr.AddGetter("ALT", () => new Alt(shared));
+            shared.BindingMgr.AddGetter("ALT_APOAPSIS", () => (new Alt(shared)).GetApoapais());
```
erendrake: i thought these _ suffixes would be gone when you were finished?

Dunbaratu: And break scripts?

src/kOS.Safe/Compilation/KS/KSScript.cs
```diff
@@ -21,10 +21,6 @@ public override List<CodePart> Compile(string filePath, int startLineNum, string
         {
             List<CodePart> parts = null;
             
-            // make the code lowercase
-            scriptText = MakeLowerCase(scriptText);
-            scriptText = ReplaceIdentifiers(scriptText);
-            
             //if (contextId != "interpreter") parts = _cache.GetFromCache(scriptText);
 
             // if parts is null means the code doesn't exists in the cache
```
erendrake: parts will always be null now. we could remove this check.

Dunbaratu: Entirely unrelated to this commit.  If you look, that was already the case before I touched anything.

src/kOS/Binding/FlightStats.cs
erendrake: `ALT:APOAPSIS` still works, and `ALT_APOAPSIS` was, to my knowledge, never part of the documented binding list. we can leave it if the _ version is in wide use.

Dunbaratu: I was operating under the mandate to just refactor to mimic the existing behavior in a new way.  So I did.  If you don't mind a 'breaking' message, then yeah we can get rid of them.

src/kOS/Suffixed/Alt.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            return Convert.ToDouble(
+                shared.Vessel.heightFromTerrain > 0 ?
+                    Mathf.Min(shared.Vessel.heightFromTerrain, (float)shared.Vessel.altitude) :
+                    (float)shared.Vessel.altitude);
+        }
+
+        public override string ToString()
+        {
+            return string.Format("ALT: Apoapsis={0} Periapsis={0} Radar={0}", GetApoapais(), GetPeriapsis(), GetRadar());
```
erendrake: this is going to print `GetApoapsis` 3 times

src/kOS/Suffixed/Alt.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        private void InitializeSuffixAlt()
+        {
+            AddSuffix("APOAPSIS", new NoArgsSuffix<double>(GetApoapais));
+            AddSuffix("PERIAPSIS", new NoArgsSuffix<double>(GetPeriapsis));
+            AddSuffix("RADAR", new NoArgsSuffix<double>(GetRadar));
+        }
+        
+        public double GetApoapais()
```
erendrake: spelling?

src/kOS/Suffixed/Eta.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+        
+        public double GetTransition()
+        {
+            return shared.Vessel.orbit.EndUT - Planetarium.GetUniversalTime();
+        }
+        
+        public override string ToString()
+        {
+            return string.Format("ETA: Apoapsis={0} Periapsis={0} Transition={0}", GetApoapsis(), GetPeriapsis(), GetTransition());
```
erendrake: string format never uses 1 or 2

Dunbaratu: true.  this was never really part of what is mentioned in the docs and was added only because it's a structure now and they all have one.  change it.

src/kOS/Suffixed/Alt.cs
Dunbaratu: I hate that word.

src/kOS/Binding/FlightStats.cs
erendrake: alright! im down with breaking these

PR #943, opened 2015-05-01
[breaking] Fix thrust calculation for 1.0 (ready)

### LINKED ISSUE #940: MAXTHRUST and AVAILABLETHRUST only account for sea level thrust

hvacengi:
When complete, this will fix #940.

I have updated the VesselUtils to calculate maxthrust based on isp, atmospheric density, and velocity.  This works for traditional rocket engines and jet engines alike.

TODO:
- [x] Move calculation to within EngineModuleAdapter
- [x] Port to individual engine calculations
- [x] Port to availablethrust
- [x] Test against all stock engines
- [x] Test against non-stock if possible
- [x] Create "at pressure" variants
- [x] Implement available thrust on EngineValue
- [x] Documentation

### DIFF
--- a/src/kOS/Utilities/VesselUtils.cs
+++ b/src/kOS/Utilities/VesselUtils.cs
@@ -365,7 +365,7 @@ namespace kOS.Utilities
                 {
                     if (pm.isEnabled && pm is ModuleEngines)
                     {
-                        thrust += ModuleEngineAdapter.GetEngineThrust((ModuleEngines)pm);
+                        thrust += ModuleEngineAdapter.GetEngineAvailableThrust((ModuleEngines)pm);
                     }
                 }
             }
--- a/src/kOS/Utilities/VesselUtils.cs
+++ b/src/kOS/Utilities/VesselUtils.cs
@@ -113,43 +113,9 @@ namespace kOS.Utilities
                 foreach (PartModule pm in p.Modules)
                 {
                     if (!pm.isEnabled) continue;
-                    if (!(pm is ModuleEngines || pm is ModuleEnginesFX)) continue;
-
-                    var engine = pm as ModuleEngines;
-                    var enginefx = pm as ModuleEnginesFX;
-
-                    if (enginefx != null)
+                    if (pm is ModuleEngines)
                     {
-                        if (!enginefx.isOperational) continue;
-                        float flowMod = (float)(engine.part.atmDensity / 1.225f);
-                        float velMod = 1.0f;
-                        if (enginefx.atmChangeFlow && enginefx.atmCurve != null)
-                        {
-                            flowMod = enginefx.atmCurve.Evaluate((float)(enginefx.part.atmDensity / 1.225));
-                        }
-                        if (enginefx.velCurve != null)
-                        {
-                            velMod = velMod * enginefx.velCurve.Evaluate((float)vessel.mach);
-                        }
-                        // thrust is fuel flow rate times isp time g times the velocity modifier for jet engines (as of KSP 1.0)
-                        thrust += enginefx.maxFuelFlow * flowMod * enginefx.atmosphereCurve.Evaluate((float)enginefx.part.staticPressureAtm) * enginefx.g * velMod;
-                    }
-                    else if (engine != null)
-                    {
-                        if (!engine.isOperational) continue;
-                        float flowMod = (float)(engine.part.atmDensity / 1.225f);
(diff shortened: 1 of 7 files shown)


src/kOS/Utilities/VesselUtils.cs
```diff
@@ -121,13 +121,35 @@ public static double GetMaxThrust(Vessel vessel)
...
+                        if (enginefx.velCurve != null)
+                        {
+                            velMod = velMod * enginefx.velCurve.Evaluate((float)vessel.mach);
+                        }
+                        // thrust is fuel flow rate times isp time g times the velocity modifier for jet engines (as of KSP 1.0)
+                        thrust += enginefx.maxFuelFlow * flowMod * enginefx.atmosphereCurve.Evaluate((float)enginefx.part.staticPressureAtm) * enginefx.g * velMod;
                     }
-
-                    if (engine != null)
+                    else if (engine != null)
```
hvacengi: When I first found all of the equations, I was reading twice the thrust shown on the right click menu (as near to exactly double as I could see).  It looks like it was the result of counting the thrust twice, once as EngineModulesFx and again as EngineModules.  EngineModulesFx is now a subclass of EngineModules, and it must not have been before.  My first solution was to insert an `else` in the 2nd `if` statement so that it doesn't count it if the 2nd time.  In final practice, we should see if there is any issue just using the EngineModules base class to get this information.  To that same end, we could write it as an extension method of the EngineModules class for MaxThrust and AvailableThrust, which can then be used anywhere else that the thrust needs to be calculated (and also means we only need to edit once if the math ever changes).

src/kOS/Utilities/VesselUtils.cs
```diff
@@ -121,13 +121,35 @@ public static double GetMaxThrust(Vessel vessel)
                     if (enginefx != null)
                     {
                         if (!enginefx.isOperational) continue;
-                        thrust += enginefx.maxThrust;
+                        float flowMod = (float)(engine.part.atmDensity / 1.225f);
```
hvacengi: I never bothered to change the variable name for `flowMod`, but really it should be called `atmMod` or some such for better readability in the future.

space-is-hard:
Just FYI, @dunbaratu was discussing splitting the max/availablethrust up into six different suffixes:

* MAXTHRUST - Current max thrust
* MAXTHRUSTSL - Max thrust were the vessel at sea level
* MAXTHRUSTVAC - Max thrust were the vessel in vaccuum
* AVAILABLETHRUST - Current available thrust (takes into account thrust limiters)
* AVAILABLETHRUSTSL - Available thrust were the vessel at sea level
* AVAILABLETHRUSTVAC - Available thrust were the vessel in vaccuum

I'm pulling this from memory from the slack chat; @dunbaratu: correct me if I'm wrong on any of it

hvacengi:
I've added sea level and vacuum to my todo list.  The only question I would have there is: how do we account for velocity in get engines at that point?  We could theoretically use the current mach number for everything, but I don't even know what the system reports for mach when in orbit under a vacuum (but we can find out).  Velocity itself will be well above sea level mach 1 at that point, but the local mach would be undefined (or maybe it gets termed as 0m/s).  I guess I don't even know if the system changes the mach value according to density.  I can expose that mach number easily enough though (if only for testing).

While I'll read and respond comments on github and slack, I won't be making edits to this until tonight at the earliest and more likely tomorrow.  If you (or any one else) want to test or make further modifications before I get a chance, feel free.

erendrake:
@space-is-hard I would advice restoring the API we had and then enhancing it with whatever new fields we decide to add based on the new Aero.

hvacengi:
OK, I think I've updated the thrust to be accurate for all engines.  If any one else might test it with whatever random engines they have in their library (as well as double checking the stock engines) I would appreciate it.

Dunbaratu:
Added '[breaking]' tag, as no matter what this looks like in the end, the release announcement will need a breaking warning about how to use it, since the main game made it inevitable that it needs to break.

hvacengi:
I think this just needs to be documented and then we should be good to go.

hvacengi:
OK.  Documentation should be complete.  Some one else should review it (possibly compile it) to make sure it renders correctly.  Also, the more people who test the code itself, the more comfortable I'll be.  This is a kind of complex modification, and I'm just sure there is going to be some fringe case that I haven't accounted for.

Issue #947, opened 2015-05-01
strings returned by :allevents and :allactions are unusable.


TDW89:
would it be possible to have kOS return a list of just the event or action names in a module.
At the moment :allevents gives a list of string where each string is in the format `(callable) <event name> , is KSPEvent`

As far as I can tell this means that you cant write a script that is un-awair of the event name.
In some situations it would be useful to be able to do this eg. running all experiments with:
```
for experiment in ship:modulesnamed("ModuleScienceExperiment") {
 for event in experiment:allevents {
  experiment:doevent(event).
 }
}
```
possible solutions: allevents could either return a list of structures (so it maintained it's printed appearance in the terminal)  but have a suffix that was just the event name, or a list of  strings that were just the event names.

erendrake:
because you are right that `:ALLFIELDS` has less utility when you are trying to call a field directly doing something like 

    SET bar TO SHIP:PARTS[0]:GETMODULE("FOO").
    bar:DOEVENT(bar:ALLEVENTS[0]).

Would it be better if you could do 

    SET bar TO SHIP:PARTS[0]:GETMODULE("FOO").
    bar:DOEVENT(bar:ALLEVENTS[0]*:NAME*).

Dunbaratu:
I thought of making an event into an object when I first implemented this.  The reason I didn't go for it was because of the annoying way the menus work, where the GUI changes constantly and replaces an event with the opposite event for some toggleable things.  That makes it so that an event is only valid the moment you see it - you can't save a handle to the event and use it later because it can go away by then from the UI, and we're trying to adhere to the rule that we don't second-guess why other mods have removed a thing from the GUI - once they have, it's immediately disallowed to be called.  Allowing handles to live on so they could be called after the event is removed from the GUI menu would violate that.

TDW89:
Would adding a `:visible` suffix as well as a `:name` suffix work? That way kos could throw an exception if the suffix can't be used and the user could write in protection for it.

Dunbaratu:
didn't auto-close when PR was merged.

PR #948, opened 2015-05-02
PartModule suffixes 

TDW89:
Quick fix for #947 - this was my quick fix for running script that didn't know what the event names were going to be for a part module. I have compiled and run it and it works. However you will probably want to change the suffix names or maybe implement this a different way (if at all).

If you use this I will add docs once the suffix names are decided upon.
Suffixes added:
`:ALLFIELDNAMES` - provides a list of just the field names
`:ALLEVENTNAMES` - returns a list of event names (not formated with `, is KSPEvent` )
`:ALLACTIONNAMES` - returns a list of action names

### DIFF
--- a/doc/source/structures/vessels/partmodule.rst
+++ b/doc/source/structures/vessels/partmodule.rst
@@ -29,12 +29,21 @@ Once you have a :struct:`PartModule`, you can use it to invoke the behaviors tha
         * - :attr:`ALLFIELDS`
           - :struct:`List` of strings
           - Accessible fields
+        * - :attr:`ALLFIELDNAMES`
+          - :struct:`List` of strings
+          - Accessible fields (name only)
         * - :attr:`ALLEVENTS`
           - :struct:`List` of strings
           - Triggerable events
+        * - :attr:`ALLEVENTNAMES`
+          - :struct:`List` of strings
+          - Triggerable event names
         * - :attr:`ALLACTIONS`
           - :struct:`List` of strings
           - Triggerable actions
+        * - :attr:`ALLACTIONNAMES`
+          - :struct:`List` of strings
+          - Triggerable event names
         * - :meth:`GETFIELD(name)`
           -
           - Get value of a field by name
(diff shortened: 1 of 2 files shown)


erendrake:
@TDW89 how would you feel about having what gets returned change from a list of strings to a list of structures. we would have to build structures for them and it might be a little tough to avoid an API break but i think it could be very useful to be able to store actions from many parts.

TDW89:
I would be fine if it worked that way its probably a better approach (and was the first thing i put under possible solutions in the issue) I just don’t know enough to write it hence the quick fix. Which was mostly just copy and past of the old suffixes with the formatting bits removed.

Dunbaratu:
Not pulling because there are no changes to the sphinx documentation mentioning the 3 new suffixes.

Dunbaratu:
@erendrake - I considered doing that very thing when I first wrote it.  the reason I didn't is the following example problem:

A docking port is docked right now.  It's rightclick menu has an action called "undock".  You do:
```
   set evt to part[30]:event("undock").
```
and now your variable evt is a handle that lets you access that KSPEvent.
Then the docking port is undocked by the script, or by some other means  This causes the "undock" KSPEvent to become hidden in the gui, and no longer usable by the manual pilot.

But your variable evt is still referencing the "undock" event that a manual pilot would be unable to call through the normal gameplay interface.

If you try to then invoke evt:execute() (or whatever we call it), then what happens?  It's now going to cause the docking port code to execute an event at a time the manual pilot wouldn't be allowed to.  Does that code have a check for this?  Well, it never needed one before so it might not.

Allowing you to hold a handle on an event or action that can live longer than the event or action is visible in the gui panel would violate the principle that kOS shouldn't be able to access anything that the PartModule never meant a pilot to have access to.

TDW89:
@Dunbaratu 
should I keep those suffix names or would you prefer them called something else?

Dunbaratu:
@tdw89 your suffix names are fine.

TDW89:
documentation added.

PR #950, opened 2015-05-02
Enable kOS icon in KSC view

ZiwKerman:


Dunbaratu:
Agreed this is a good idea - at least when it appears it can be turned off this way.

PR #957, opened 2015-05-04
Fixes #834. protect for null patchedConicSolver

### LINKED ISSUE #834: POSITIONAT() only works on other vessels that have previously been selected as target.
I have three vessels in orbit: kCom 1, kCom 2, and kCom 3.
Attempting to use `POSITIONAT` to get the location of any non-active vessel with any value for time returns a null reference exception.

However, if I first `SET TARGET TO vessel("kCom 1")`, I can suddenly read it's values for the rest of the flight scene even if I un-target it later. kCom 2 will continue to fail until it is targeted as well. Upon scene reload the problem returns until I perform the sacred target-dance again.

This is all being done from the interpreter. I get the same results from a script, other than that I need to make sure I wait a brief moment each time I set the target to make sure KSP actually registers the target change.

I have reproduced this on a clean install with only kOS. I have checked against vessels that are both inside and outside physics loading range, both exhibit the same behavior.

[KSP.log](http://pastebin.com/F5d2NJVr)

Steps to reproduce:
1. Have two or more vessels.
2. `print POSITIONAT(vessel("name"),time).` where the vessel chosen is not the active vessel.
3. `Object reference not set to an instance of an object` 
4. `set target to vessel("name").` then repeat step 2.
5. Receive an actual position vector.

Dunbaratu:
vessel.patchedConicSolver appears to be null whenever the vessel in question is not the ActiveVessel.  There's a number of places in our code where we just presume it will exist for every vessel and blindly try to use it without checking if it's null.  This patch fixes that.

### DIFF
--- a/src/kOS/Utilities/VesselUtils.cs
+++ b/src/kOS/Utilities/VesselUtils.cs
@@ -227,6 +227,11 @@ namespace kOS.Utilities
 
         public static object TryGetEncounter(Vessel vessel, SharedObjects sharedObj)
         {
+            // If not the active vessel, it will be on rails, and therefore won't
+            // be able to have "encounters" via its patchedConicSolver.
+            if (vessel.patchedConicSolver == null)
+                return "None";
+
             foreach (var patch in vessel.patchedConicSolver.flightPlan)
             {
                 if (patch.patchStartTransition == Orbit.PatchTransitionType.ENCOUNTER)
--- a/doc/source/commands/prediction.rst
+++ b/doc/source/commands/prediction.rst
@@ -13,6 +13,23 @@ Predictions of Flight Path
 
     Using the Add and Remove commands as described on that page, you may alter the flight plan of the CPU\_vessel, however kOS does not automatically execute the nodes. You still have to write the code to decide how to successfully execute a planned maneuver node.
 
+.. warning::
+
+    Be aware that a limitation of KSP makes it so that some vessels'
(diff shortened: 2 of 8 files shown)


src/kOS/Utilities/VesselUtils.cs
```diff
@@ -227,6 +227,11 @@ public static Vector3d GetNorthVector(Vessel vessel)
 
         public static object TryGetEncounter(Vessel vessel, SharedObjects sharedObj)
         {
+            // If not the active vessel, it will be on rails, and therefore won't
+            // be able to have "encounters" via its patchedConicSolver.
+            if (vessel.patchedConicSolver == null)
+                return "None";
+
```
Dunbaratu: I never liked the fact that the ENCOUNTER suffix only finds *maneuver node* encounters and can't find just natural encounters that you'll hit by momentum from where you are now.  (the way you'd do that is to follow the vessel's orbit object, which has a next orbit linked list structure to it to represent future non-manuever node encounters).

But as the "remit" for this PR is JUST to fix the crashes from patchedConicSolver being null, that fell outside the remit of this change.

It might make a good separate issue.

Dunbaratu:
Added pending documentation because even though this isn't a change, I'm going to write up a bit warning users not to try to use maneuver nodes on the non-active vessel.  That was not mentioned before (because we didn't know).

PR #964, opened 2015-05-06
Fixes #858 making empty programs have a NOP.

### LINKED ISSUE #858: running an empty file restarts the main program
Only works when you run an empty file from inside another program:
```
// main.ks
print "program started".
log "" to empty_file.
delete empty_file.
log "" to empty_file.
print "next line will implicitly restart your program".
run empty_file.
```
The code above prints:
```
program started
next line will implicitly restart your program
program started
next line will implicitly restart your program
program started
next line will implicitly restart your program
program started
next line will implicitly restart your program
program started
next line will implicitly restart your program
...
```
Possible cause: load() returns 0 when parsing an empty file.

Dunbaratu:
Now all empty program files will get at least 1 instruction
in them - a NOP, which (importantly) is labeled.  Empty programs
did have instructions before, but they didn't come from the
compiler.  The compiler made a zero-length program, and the
opcodes that existed were added afterward by
ProgramBuilder and they just consisted of a push 0, return.
The reason they didn't work is that since they were built outside
the compiler, they had no Label field filled in.  The jump for
a CALL is taken by reading the label of the entrypoint of
the program, which is its zeroth opcode.  When the zeroth
opcode had a missing .Label because it was added ad-hoc after
the compiler was done, the lookup system broke and the result was
returned as zero.

Now the compile simply performs a check.  If the list of opcodes
is length zero after the compile is done, then a dummy NOP is
added by the compiler before it finishes, so there's always at
least one opcode the compiler put there, with all its proper labelling
work that it knows how to do via AddOpcode().

### DIFF
--- a/src/kOS.Safe/Compilation/KS/Compiler.cs
+++ b/src/kOS.Safe/Compilation/KS/Compiler.cs
@@ -123,7 +123,7 @@ namespace kOS.Safe.Compilation.KS
             PushReversedParameters();
             VisitNode(tree.Nodes[0]);
             
-            if (addBranchDestination)
+            if (addBranchDestination || currentCodeSection.Count == 0)
             {
                 AddOpcode(new OpcodeNOP());
             }
@@ -663,7 +663,7 @@ namespace kOS.Safe.Compilation.KS
                         currentCodeSection = subprogramObject.FunctionCode;
                         // verify if the program has been loaded
                         Opcode functionStart = AddOpcode(new OpcodePush(subprogramObject.PointerIdentifier));
-                        AddOpcode(new OpcodePush(0));
+                        AddOpcode(new OpcodePush(-1));
                         AddOpcode(new OpcodeCompareEqual());
                         OpcodeBranchIfFalse branchOpcode = new OpcodeBranchIfFalse();
                         AddOpcode(branchOpcode);
@@ -700,7 +700,7 @@ namespace kOS.Safe.Compilation.KS
                         currentCodeSection = subprogramObject.InitializationCode;
                         // initialize the pointer to zero
                         AddOpcode(new OpcodePush(subprogramObject.PointerIdentifier));
-                        AddOpcode(new OpcodePush(0));
+                        AddOpcode(new OpcodePush(-1));
                         AddOpcode(new OpcodeStore());
                     }
                 }
--- a/src/kOS.Safe/Exceptions/KOSBadJumpException.cs
+++ b/src/kOS.Safe/Exceptions/KOSBadJumpException.cs
@@ -1,0 +1,25 @@
+﻿using System;
+
+namespace kOS.Safe.Exceptions
+{
+    public class KOSBadJumpException : Exception
+    {
+        public KOSBadJumpException(int destination, string message):
+            base( String.Format("Can't jump to instruction {0}.  No opcode there: {1}: {2} ",
+                                destination,
+                                message,
+                                "If you see this message, something is broken about how this program got compiled."))
+        {
+        }
+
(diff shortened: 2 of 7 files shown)


src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -663,7 +663,7 @@ private void PreProcessRunStatement(ParseNode node)
                         currentCodeSection = subprogramObject.FunctionCode;
                         // verify if the program has been loaded
                         Opcode functionStart = AddOpcode(new OpcodePush(subprogramObject.PointerIdentifier));
-                        AddOpcode(new OpcodePush(0));
+                        AddOpcode(new OpcodePush(-1));
```
Dunbaratu: I changed the dummy value to one that deliberately fails instead of 0, to help avoid the danger of jumping to the start of the program again like it did before.  If it is having the error of calling an uninitialized program address, I'd rather it crashed the program than recursively start itself over because it happens to jump to zero.

src/kOS.Safe/Exceptions/KOSBadJumpException.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            get { return base.Message; }
+        }
+
+        public virtual string HelpURL
+        {
+            get { return string.Empty; }
+        }
+    }
+}
```
Dunbaratu: This new exception is just here so that if the problem, or a similar one appears again, the attempt to jump to instruction -1 should throw this error instead of re-running the program.

src/kOS/Execution/ProgramContext.cs
```diff
@@ -111,13 +111,21 @@ public List<string> GetCodeFragment(int start, int stop)
         {
             var codeFragment = new List<string>();
 
-            const string FORMAT_STR = "{0,-20} {1,4}:{2,-3} {3:0000} {4} {5}";
-            codeFragment.Add(string.Format(FORMAT_STR, "File", "Line", "Col", "IP  ", "opcode", "operand"));
-            codeFragment.Add(string.Format(FORMAT_STR, "----", "----", "---", "----", "---------------------", ""));
+            const string FORMAT_STR = "{0,-20} {1,4}:{2,-3} {3:0000} {4} {5} {6}";
+            codeFragment.Add(string.Format(FORMAT_STR, "File", "Line", "Col", "IP  ", "label", "opcode", "operand"));
+            codeFragment.Add(string.Format(FORMAT_STR, "----", "----", "---", "----", "-------------------------------", "", ""));
```
Dunbaratu: Changed the format because this isn't the first time I've had trouble diagnosing a problem because the debug dump didn't show the string label that the opcode was using.

src/kOS/Execution/ProgramContext.cs
```diff
@@ -111,13 +111,21 @@ public List<string> GetCodeFragment(int start, int stop)
...
-                    codeFragment.Add(string.Format(FORMAT_STR, Program[index].SourceName, Program[index].SourceLine, Program[index].SourceColumn, index, Program[index], (index == InstructionPointer ? "<<--INSTRUCTION POINTER--" : "")));
+                    codeFragment.Add(string.Format(
+                        FORMAT_STR,
+                        Program[index].SourceName,
+                        Program[index].SourceLine,
+                        Program[index].SourceColumn,
+                        index,
+                        Program[index].Label,
+                        Program[index],
+                        (index == InstructionPointer ? "<<--INSTRUCTION POINTER--" : "")));
```
Dunbaratu: After adding yet another parameter I decided this was getting absurdly long for one line.

PR #965, opened 2015-05-06
Fixes #927 by making mean anomaly into degrees.

### LINKED ISSUE #927: mean anomaly is in radians
maybe this is intended and there's a reason i'm not aware of. docs mention all attempts were made to give everything in degrees...this might be one that was missed.

obt:meananomalyatepoch is 0 at periapsis -> ~2pi where i'd expect 0 at periapsis -> ~360

i'm learning as i go and that had me banging my head on the wall for a while!

Dunbaratu:


src/kOS/Suffixed/OrbitInfo.cs
```diff
@@ -38,8 +39,8 @@ private void InitializeSuffixes()
             AddSuffix("SEMIMINORAXIS", new Suffix<double>(() => orbit.semiMinorAxis));
             AddSuffix(new[]{"LAN", "LONGITUDEOFASCENDINGNODE"}, new Suffix<double>(() => orbit.LAN));
             AddSuffix("ARGUMENTOFPERIAPSIS", new Suffix<double>(() => orbit.argumentOfPeriapsis));
-            AddSuffix("TRUEANOMALY", new Suffix<double>(() => orbit.trueAnomaly));
-            AddSuffix("MEANANOMALYATEPOCH", new Suffix<double>(() => orbit.meanAnomalyAtEpoch));
+            AddSuffix("TRUEANOMALY", new Suffix<double>(() => Utilities.Utils.DegreeFix(orbit.trueAnomaly,0.0)));
```
Dunbaratu: I figured it didn't hurt to add this while I was at it, just in case.

Dunbaratu:
**breaking**: Technically, although unlikely because I don't know how many people were using this, the previous behavior might have been noticed by someone and they never said anything and just wrote their script to expect radians here.  If they did so, this fix would break their script.

PR #976, opened 2015-05-08
Fixes #956 - MLFields for OpcodeDelegateRelocateLater was wrong.

### LINKED ISSUE #956: error using functions inside .ksm
cant find any mention of it in the docs but .ksm files containing functions fail to run.
to reproduce

test.ks:
```
run test2.
print_one().
compile test2.
```
test2.ks:
```
function print_one {
print 1.
}.
```
run test twice from the terminal the second time it will preferentially try to run the .ksm file and fail.
error is:
```
Cannot cast from source type to destination type.
```

Dunbaratu:


src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -1875,11 +1875,12 @@ public OpcodePushDelegateRelocateLater(string destLabel, bool withClosure) : bas
...
         
         public override void PopulateFromMLFields(List<object> fields)
-        {
+        {            
             // Expect fields in the same order as the [MLField] properties of this class:
             if (fields == null || fields.Count<1)
                 throw new Exception("Saved field in ML file for OpcodePushDelegateRelocatelater seems to be missing.  Version mismatch?");
-            WithClosure = (bool)fields[0];
+            DestinationLabel = (string)( fields[0] ); // this is really from the base class.
+            WithClosure = (bool)fields[1];
```
Dunbaratu: The problem, in a nutshell, is that once you derive a class from another Opcode, and the base class used MLFields, and filled them with its own PopulateFromMLFields() method, and your derived class also adds MLFields, you now have an opcode that has to fill SOME of its MLFields from the derived class, and SOME from the base class, but the system wasn't designed to take that.  It presumes you can hardcode where to expect the fields in the fields array.  i.e. you KNOW this field will be the 0th, this one will be the 1st, that one will be the 2nd, and so on.  That means putting knowledge of the base class into the derived class where it doesn't belong. 

In the example above, note my derived class has to be aware that its base class has an MLField called DestinationLabel that is in need of filling, and will come first in the list.  It can't just call the base PopulateFromMLFields() either, because it would then still need the knowledge of the fact that the base class consumed exactly 1 field, so it knows to start at index 1 not 0.

This system may need a refactor, or we may need to declare these things as final to disallow derived classes.

For now this patch fixes this ONE problem with this ONE and only place where the problem has appeared so far (this is the only such derived class situation we have where the base class has ML fields - I checked).

A more general solution would be to change it so that PopulateFromMLFields is given a malleable list of fields rather than an array of fields, so the base method can consume and erase from the list as it eats through the fields, letting the derived class start at the "front" of the part of the list that remains.

abenkovskii:
I compiled and tested akrOS: https://github.com/akrasuski1/akrOS/issues/26. It works with this fix.

src/kOS.Safe/Compilation/Opcode.cs
erendrake: do you expect to get a lot of use out of derived classes in opcodes? if not i think we should seal/final the classes to prevent confusion.

Dunbaratu: I think the original reason this was a derived class is that there's a few places where I try to catch both kinds of relocatable opcode of them with an 'is' check.  (i.e. catch both the base and the derived opcode in the same check)

That may be changed to perhaps an Opcode base class field for things like "isRelocatable", and then what you say could work.

Issue #631, opened 2015-03-01
Edit terminal out of focus transparency.


TDW89:
This stems from #630 but is more of an enhancement idea based off using multiple resized terminals as readouts.

Would it be possible to allow in script, and or control panel, modification of the out of focus terminal window transparency.
If using kOS as a custom instrument panel during the day it can be hard to read when out of focus.

Dunbaratu:
perhaps we can leave it opaque at all times, and just change the border color.

abenkovskii:
It seems to be an easy to do. I'll try changing it. Expect a PR soon.

PR #984, opened 2015-05-09
Now only the terminal border changes when terminal is out on focus

abenkovskii:
Issue:   #631
screenshot0


abenkovskii:
@akrasuski1 Check this out.
Does it need some tweaking?

akrasuski1:
Judging from the screenshot, it looks nice. This would enable user input via action groups etc. without having to squint at the window due to its transparency.

kaufman-bryce:
If possible, could this be a config setting? Sometimes it is useful to be able to see through the kOS window.

abenkovskii:
@kaufman-bryce It's an interesting idea

abenkovskii:
[do not merge] found a new idea to implement

abenkovskii:
@erendrake What do you think about adding a new config setting for it?
P.S. I know how to do it
P.P.S. I'm going to remove redundant config settings in the next PR

erendrake:
@abenkovskii because we have the `TERMINAL:` object, i would say we could leave the default as it has been and allow the scripter control the overall opacity on blur.

`SET TERMINAL:OPACITY:ONBLUR TO 0.5`
`SET TERMINAL:OPACITY:ONFOCUS TO 1`

then if you want we could additionally control the two separate elements of the wrapper and the terminal screen like you have in this PR.

`SET TERMINAL:BEZELOPACITY:ONBLUR TO 0.5`
`SET TERMINAL:BEZELOPACITY:ONFOCUS TO 1`

`SET TERMINAL:SCREENOPACITY:ONBLUR TO 0.5`
`SET TERMINAL:SCREENOPACITY:ONFOCUS TO 1`

this would allow us to have more control, in script, over this behavior.

abenkovskii:
@erendrake :+1: Nice idea. I'll try implementing it. May be we can give scripter a full access to colour of those two terminal parts (rgba)? Or it will be excessive?

erendrake:
@abenkovskii lets just do that opacity in this pass and then we can regroup and see about adding color. I really liked what @Nivekk was showing in his last classic video.


....


@abenkovskii @Dunbaratu  how would you feel about building a new in game terminal that would get unlocked at some point that was a lot more modern looking? maybe these advanced opacity controls would make more sense for a more advanced terminal/

erendrake:
@abenkovskii i think there is much more work to do before this PR will be ready. im going to close it for now and we can reopen it.

abenkovskii:
@erendrake 
1. I like the idea of upgradable terminal
2. Technical difficulty: can TELNET support a tabbed interface?
3. In the Nivekk's video the background is transparent but the text is visible. If we make something like this there will be no need in opacity controls

Dunbaratu:
I think the problem with what we have now is that the transparency is applied to BOTH the background and the text letters.  Also, the choice of green letters, while thematically right for some old terminals, looks just like Kerbin's ground and that helps make it hard to see when the background goes transparent.  We could also go with white.

I dislike the notion that terminal opacity is an upgrade option, because the problem of terminal opacity is entirely on the wrong side of the "fourth wall" for an in-game solution to make sense.  The Kerbals IN the universe don't have to look through a terminal screen to see the landscape behind it.  Presumably their terminals are opaque CRT's in the dashboard somewhere, or on a ground controller's desk.

abenkovskii:
What about syntax highlighting, tabbed interface and resizable window (@Nivekk's suggestion)? They seem to be on the right side of the fourth wall.

Dunbaratu:
Syntax highlighting isn't caused by the terminal.  Being able to do color at all is caused by the terminal.  Once you have that color, then using it for highlighting is caused by whatever editor you're running inside it.

abenkovskii:
It looks like there are a lot of questions to discuss about changing the terminal. I'll create a new issue for the discussion.

Issue #897, opened 2015-04-25
Round number floats not treated as ints for volume number. i.e. switch to 0.00. fails.


abenkovskii:
![screenshot1](https://cloud.githubusercontent.com/assets/11031560/7332354/413e538a-eb44-11e4-812f-b7df0bcb3ae2.png)
P.S. encountered it when using `mod()` on the volume number.

abenkovskii:
@erendrake Switch does accept variables (integers and probably strings). It doesn't accept floats.
![screenshot2](https://cloud.githubusercontent.com/assets/11031560/7333672/052cfad0-eb81-11e4-820a-e85ffd55e1cf.png)

erendrake:
Ah that should be much simpler to fix. Thank you for the clarification

abenkovskii:
Not at all.

Dunbaratu:
Why on earth is it a bug that it can't switch to volume number 0.00001 ?   That looks pretty correct to me.

abenkovskii:
Because you can't switch to the volume `mod(7, vol_list:length)`

Dunbaratu:
``switch to round(i,0).``
There is no such thing as volume 0.00001.  That is correct that you can't switch to it.  It's only a problem if it won't let you switch to, say, 0.00000 or 1.00000 - THEN it would be a problem because floats and ints are meant to be interchangeable WHEN and only when the float has no fractional component.

abenkovskii:
@Dunbaratu Need a "real" example? Here you are:
![screenshot3](https://cloud.githubusercontent.com/assets/11031560/7333876/75747f56-eb87-11e4-904e-612e451d52ca.png)
P.S. some_list[0.2] works

Dunbaratu:
Then it is what I said with:

> It's only a problem if it won't let you switch to, say, 0.00000 or 1.00000

Using a weird example of 0.00001, where it's *correct* that it doesn't allow it, doesn't illustrate the problem.  The problem is that every time you push a float to the stack it is *supposed* to be transforming it into an int if its a round number.  That is clearly not happening here.

What should NOT be happening is that we sprinkle special case code everywhere in the mod to sanitize the integer-expecting inputs in those places where integers are expected.  The transformation of round number floats into integers should be happening universally in one place.

some_list[0.2] shouldn't work.  If it does then that's a bug.  What *should* work is some_list[0.0000] or some_list[2.0000].  You can use a float in place of an integer, but only if the float really IS a round number.

[...]

PR #987, opened 2015-05-09
Round floats are now automatically converted to integers

abenkovskii:
screenshot4
issue: #897


src/kOS/Execution/Stack.cs
```diff
@@ -48,9 +48,13 @@ private void ThrowIfInvalid(object item)
...
         private object ProcessItem(object item)
         {
-            if (item is float)
+            if (item is float) {
                 // promote floats to doubles
-                return Convert.ToDouble(item);
+                item = Convert.ToDouble (item);
+            }
+            if (item is double && (double)item == Math.Truncate((double)item)) {
+                return Convert.ToInt32 (item); // or should it just be (int)?
```
Dunbaratu: suggest: 
```
if (item is double)
{
    double rounded = Math.Truncate((double)item);
    if ((double)item == rounded)
        return Convert.ToInt32(rounded);
}
```
Same thing, but it only performs the rounding once, instead of doing it in both Truncate and again in Convert.

abenkovskii: > only performs the rounding once

1. Does `Convert.ToInt32()` work faster if it receives round number?
2. `return Convert.ToInt32 (item);` in original code executes if and only if `item` is exactly equal it `Math.Truncate((double)item)` so in this particular situation it shouldn't matter.

abenkovskii: But I can try this:
```
if (item is double)
{
    int intPart = Convert.ToInt32(item);
    if ((double)item == (double)intPart)
        return intPart;
}
```

abenkovskii: Or this if you say that it's faster:
```
if (item is double)
{
    double truncated = Math.Truncate((double)item);
    int intPart = Convert.ToInt32(truncated);
    if ((double)item == (double)intPart)
        return intPart;
}
```

abenkovskii:
[don't merge] found some bugs

abenkovskii:
I fixed all the bugs I found.

abenkovskii:
There might be some bugs :beetle: caused by my misunderstanding of C#. Could someone look through the code?

src/kOS/Execution/Stack.cs
```diff
@@ -48,9 +52,19 @@ private void ThrowIfInvalid(object item)
...
+            }
+            if (item is double)
+            {
+                if (!Double.IsNaN((double)item) && Int32.MinValue < (double)item && (double)item < Int32.MaxValue)
+                {
+                    int intPart = Convert.ToInt32(item);
+                    if ((double)item == (double)intPart)
+                        item = intPart;
+                }
+            }
```
Dunbaratu: This looks correct to me.
@erendrake will probably complain a bit about the formatting, but the logic is correct.

abenkovskii:
I found this: https://github.com/KSP-KOS/KOS/blob/develop/src/kOS.Safe/Compilation/Calculator.cs#L171
So `switch to 2.0.` doesn't work without my fix but `switch to 2.0 + 0.0.` works fine in the current kOS version.

src/kOS/Execution/Stack.cs
erendrake: i am actually more concerned about the fact that we have do so much before we push every single double onto the stack.

1. Parse the double to see if its NaN
2. Convert to Int
3. strict equality of the double value which will fail unless this is right after some kind of truncating event. (eg round)

@abenkovskii  What is the reason for not doing this in the round function itself?

Dunbaratu: This is part of the logic that says in kOS, int and double numbers are interchangeable.  Rather than having each and every place that expects a number to be an int do the check, just make it so that when a round double is pushed, it becomes an int.

64-bit IEEE doubles can store accurate integers up to 2^53 before you hit the first integer that cannot be stored correctly without roundoff.  So the strict equality check works fine until you have integers bigger than a 32-bit integer anyway.

erendrake: @Dunbaratu i was more worried about floating point error making many of the ints not match if is has been a part of any math operation.

it still seems pretty heavy for something we do each time we put something on the stack

it seems like either of these would be cheaper to use for the check

    double d = 5.0;
    bool isInt = d == (int)d;

    double d = 5.0;
    bool isInt = d % 1 == 0;

but if you guys are ok with it maybe im just worried over nothing :)

Dunbaratu: It's not a matter of expense but of when.  The idea is to fix all the spots where an int is presumed by putting the fix in one place instead of in 20 places.  But I'm still unsure if that's the right thing to do, because then what about all the places that expect a double and now will be getting an int?  I think it's inevitable that you'll end up having to perform that cast on a per-instance basis in the C# code, because there are cases where an int is right and cases where a double is right, so maybe trying to solve it in one universal place isn't really practical.

abenkovskii: @Dunbaratu Show me at least one place that expects double and will break if receives int.

abenkovskii: The current version of kOS is doing the same check and conversion of round doubles to int in the Calculator class: https://github.com/KSP-KOS/KOS/blob/develop/src/kOS.Safe/Compilation/Calculator.cs#L171. I tested it: `switch to 2.0.` doesn't work without my fix but `switch to 2.0 + 0.0.` does.

Dunbaratu: @abenkovskii You are asking the wrong question.  The question should be "show me one place where a cast to a double exists just in case the value is an int", rather than trying to find a place where a bug exists because that cast has failed to be done and therefore the code is currently already broken.

There are places that expect a double, but they won't break when given an int precisely because that casting I'm referring to is in the code.  My point is that regardless of whether you pick a standard that the stack will always turn things into integers or the pick a standard that it will always turn things into doubles, either way there will have to be places where that standard isn't what's needed and they'll need to do a cast from that back to what they need..  Places like that already exist for making ints into floating point (for example, when you ``lock throttle to 1`` instead of ``lock throttle to 1.0``, kOS's flybywire logic will convert that integer 1 into a floating point 1.0 for you.  The stack isn't doing it - it's an exception that had to be hardcoded for that one case.  The point I'm making is that either you default to using ints for all values, in which case you cast the ints to floating point where needed in C#'s code, or you default to using doubles for all values, in which case you cast the doubles to ints where needed in kOS's C# code.  Either way the standard doesn't solve all your casting needs on its own because whichever standard you pick, there will be parts of the code that have to convert back out of it again - so maybe the right answer is to assume neither and have the casting everywhere that it matters to the application at hand, which is pretty much how it works right now except that this bug is that the casting was missing in one spot where it was needed.  That's no more or less a hole in the design than the fact that it's possible in the future to forget a cast to a double where it is needed.

That's why I'm beginning to think doing this at a low level of the stack isn't the right place for it - unless we make a Number wrapper class like we're planning on making a String wrapper class.
Then the Number wrapper class would have a GetAsInt or GetAsDouble or GetAsFloat, depending on what the underlying kOS code needs to do with it, and you'd make it so that it always has to pick one of those (so that you can't forget to cast).

abenkovskii: @Dunbaratu I like the idea of number wrapper. Eventually we will wrap up everything and get rid of Calculator.cs that puts operation logic outside of class.

Dunbaratu:
I'm saying the error, at it's core, is that any code that absolutely needs the value to be double, and any code that absolutely needs it to be integer, should be doing that safety check itself instead of making the fragile assumption that it was done for it by some other level.  The reason for the standard of trying to cast to int in the math if possible is purely for the sake of preventing turning integers into doubles for speed (so that, for example, ``2^3`` doesn't get calculated as the more expensive ``2.0^3.0``)  It shouldn't be used as the means of fixing the fact that some places that expect integers are failing on doubles.  The code still has to deal with what happens when given a value like 0.9999999999999 instead of 1.  If it is a case where the API needs integers, relying on the stack push as the solution is fragile.

There is an important difference between "this is what the low level parts of the system happen to be doing at the moment" versus "this is what the low level parts of the system have advertised that they are guaranteed to always do".  The fact that 2.0 + 0.0 happens to return an integer belongs squarely in that "happens to do" category, not the "guaranteed to always do by contract" category.

src/kOS/Execution/Stack.cs
Dunbaratu: Yes, if we had a Number wrapper, one of its responsibilities would be to perform all the Calculator.cs logic itself.

PR #992, opened 2015-05-10
kOS toolbar fix

### LINKED ISSUE #951: [bug][able to reproduce] kOS toolbar window appearing in Space Center scene
Steps to reproduce:
* Go to space center scene
* Open any one of the following: Astronaut Center, R&D Center, Mission Control, Admin Building
* Exit back to space center scene
* Open any one of the above buildings again (doesn't have to be same building)
* Exit back to space center scene again
* Toolbar window should now be open despite not having its icon in the toolbar

abenkovskii:
fixes #951
Also fixed this:
screenshot5
P.S. the second bug happens when you first hover over the kOS button.
P.P.S. My IDE changed .csproj, .sln and .userprefs should I revert this changes?


abenkovskii:
@erendrake Check this out. Can it be added to the next release?

ZiwKerman:
@abenkovskii 

I suggest we bring a more resuable and unified code, we can borrow from here I believe:
https://github.com/MagicSmokeIndustries/InfernalRobotics/blob/develop/InfernalRobotics/InfernalRobotics/Extension/RectExtensions.cs

ZiwKerman:
From what I see in your fix you want to clamp the window to the screen so it is always visible, if it is enabled, and it is what `ClampToScreenEdge(Rect)` does

abenkovskii:
The code you are talking about is not generic enough. It can clamps a rect only to the screen. But kOS needs to clamp it's window to a smaller rect:
screenshot21

ZiwKerman:
What I'm trying to say is we should probably follow the same approach, create some Utility code that will be more reusable, so we can use it not only for Toolbar but for all other windows in kOS

abenkovskii:
I see now. Yes generic approach is probably better.

ZiwKerman:
For example create ClampToRect() and make ClampToScreen() using it

ZiwKerman:
Also it seems your branch is a bit outdated- you should try pulling latest code. We've added kOS Icon to SpaceCenter view some time ago.

abenkovskii:
I know. And I also know that it was added as a "quick and dirty" solution to the first bug this PR solves so I deleted it.

abenkovskii:
[don't merge yet] ok. I'll change it to the more generic variant.

abenkovskii:
[ready] I believe it's ready again

ZiwKerman:
I will try it now
What should I do to test it out?

abenkovskii:
1. Try reproducing this bug: #951
2. Open the VAB, mouseover kOS app launcher button but **don't** click it. kOS window should appear in the right place.
3. Optional: test other situations where kOS toolbar window should appear

erendrake:
I am waiting on @ZiwKerman to test this before i merge. Thank you both for working on this.

src/kOS/Screen/KOSToolBarWindow.cs
```diff
@@ -56,7 +54,7 @@ public class KOSToolBarWindow : MonoBehaviour
         // ReSharper restore RedundantDefaultFieldInitializer
         private Vector2 scrollPos = new Vector2(200,350);
 
-        private static Rect windowRect;
+        private static Rect windowRect; // does anybody know why this is ststic?
```
ZiwKerman: Because we only have one window?

ZiwKerman:
Toolbar window now behaves properly, at least I could not reproduce previous bugs.

src/kOS/Screen/KOSToolBarWindow.cs
abenkovskii: But some per-window fields (e.g. isOpen, clickedOn, width, height) are not static.

PR #999, opened 2015-05-11
Develop

pgodd:
Parts used to sink into launch pad.
Modified 1m part to fix collision mesh and problem corrected.
Modified 0m part and found that parts below a certain thickness will
always fall through launchpad.
Modified light animation in 0m part.
Source files created for all parts.


pgodd:
Grrrr. Not what I wanted to do.

PR #1000, opened 2015-05-11
Fixes #980 parts sinking into pad

### LINKED ISSUE #980: kOS processor sinks into the launchpad:
I'm not sure if it's a kOS problem or a KSP bug but when I launch this:
screenshot9
The kOS processor sinks into the launchpad:
screenshot10
It only happens with stack mounted parts, radial part doesn't sink.
@pgodd Is something wrong in the model or in the .cfg? Or is it supposed to work like this?

pgodd:
Includes new 1m and 0.625m parts


abenkovskii:
:fireworks: 1000

Dunbaratu:
It looks as if the reason this has over 1000 files is because it contains all of SQUAD's PartTools as well.

pgodd:
It's down to 44 files I think

Issue #342, opened 2014-11-07
BODY needs a sphere of influence radius suffix


Dunbaratu:
Something like BODY:SOIRADIUS.

Returns: the radius of the body's sphere of influence.  If your orbit goes higher than that number you will escape the body EVEN IF your math still shows an elliptical rather than hyperbolic orbit.  This is a place where you have to break from roleplaying the reality a bit and look at the simulation artifact to get the autopilot to do the correct thing because doing what should work in the real world will NOT work in the simulated world of KSP.

Alternatively: BODY:SOIALTITUDE.  (The difference being whether it's measured from the body center or the body surface.  Periapsis and apoapsis numbers are reported in terms of altitude not radius so this might be the better way to show it to the user.)

Either number will be sufficient, because you can add or subtract BODY:RADIUS to get the other one.  It's just a matter of which we want to show.

ndevenish:
This would be very useful. I'm writing a periapsis kicker and need to enter this information manually at the moment.

abenkovskii:
Looks easy enough. Can I do it?

erendrake:
@abenkovskii go for it!

PR #1011, opened 2015-05-13
Adds SOIRADIUS suffix to the BODY structure

abenkovskii:
Issue: #342
screenshot13


src/kOS/Suffixed/BodyTarget.cs
```diff
@@ -96,6 +96,7 @@ private void BodyInitializeSuffixes()
             AddSuffix("ROTATIONPERIOD", new Suffix<double>(() => Body.rotationPeriod));
             AddSuffix("ATM", new Suffix<BodyAtmosphere>(() => new BodyAtmosphere(Body)));
             AddSuffix("ANGULARVEL", new Suffix<Vector>(() => RawAngularVelFromRelative(Body.angularVelocity)));
+            AddSuffix("SOIRADIUS", new Suffix<double>(() => Body.sphereOfInfluence));
```
erendrake: @abenkovskii have you tested this in game? i thought i remembered this not being a super reliable property of the body class but squad may have fixed it.

abenkovskii: Well.
1. It complies
2. it runs. 
3. I can even print SOI radius on the screen. (see screenshot) 
4. There are no warnings about potential problems in the [KSP API documentation](http://wiki.kerbalspaceprogram.com/wiki/API:CelestialBody)
P.S. what exactly do you mean by "not super reliable"?

erendrake:
@abenkovskii the reason i didnt add this feature before was that way back when i first looked at it apparently this property would not line up with the patched conics solution for when you left the SOI.

This would mean that kOS would report that you would leave the SOI at X altitude but the solver and therefore the rest of the game might have you leaving at a much different altitude. we had the maneuver node system that would tell you when a node would have you leaving the SOI and it was much more accurate so i left this for later. Thinking i would have to do something crazy to make it work.

 Harv made quite a few changes to the way the solver works in 0.90 and i just havent been back in here to see if this property is ok to use raw now.

abenkovskii:
Looks fairly accurate:
screenshot10
Could you forget to subtract bodie's radius?

abenkovskii:
It matches the value from [KSP WIKI](http://wiki.kerbalspaceprogram.com/wiki/Kerbin#cite_note-centered-1)

abenkovskii:
screenshot13
Works fine with Kerbin's SOI.

erendrake:
Could you forget to subtract bodie's radius?

this is the second time today that you have assumed that we have no clue what we are doing and apparently didnt read what i wrote. The issue was not a limitation of kOS or our understanding of the wiki. this property has always matched the wiki as far as i know. 

The issue was that the value from the wiki was not always correct when you were flying the craft. many times you could escape at a lower altitude or couldnt escape at a higher altitude so the property was not very useful because it lied. Back then i decided to nix this suffix because the maneuver node system was more accurate at showing SOI transitions. 

It looks like the work harvester did a few releases ago fixed it, which is good because it is exactly the type of bug that he mentioned back then. Thank you for the PR :)

abenkovskii:
> this is the second time today that you have assumed that we have no clue what we are doing

Sorry. I just assume that everyone sometimes does stupid errors.

erendrake:
And lord knows i make stupid errors :) thank you for doing the leg work to show that this property of the body is safe to use and implementing the new suffix.

I just realized that we didnt have the documentation for this new suffix in the PR. Would you mind adding it, even just to this branch and ill pull it in

PR #1019, opened 2015-05-16
Resolves #40 with new FROM loop syntax.

### LINKED ISSUE #40: (enhancement) versatile loop akin to C's FOR loop
C's FOR loop is quite versatile.  A lot of other languages have loops that sound nicer in English when read out ("for x from 1 to 10 step 2" for example), but what they gain in looks and legibility, they lose in versatility.  They constrain the type of loop you can have.  It *must* have exactly one loop counter variable, it *must* increment by a value known ahead of time, with an endpoint based upon a greater-than or less-than check which isn't appropriate for some algorithms where the counter isn't a number, and so on.

What I'm considering is a loop with the versatility of the C FOR loop , with the same 3 parts to it (do this when starting, check this to see when to stop, do this to increment), but done in a way more consistent with kosscript's "special" style of syntax.

I propose this syntax:
```FROM``` *init* ```UNTIL``` *end-check* ```GOING``` *incrimenter* ```{``` … ```}```.
Or the same thing by overloading the FOR syntax to also accept:
```FOR``` *init* ```UNTIL``` *end-check* ```GOING``` *incrimenter* ```{``` … ```}```.

(I'm not sure which I prefer.  "FOR" is more traditional for programming, and also used by the list iterator loop already, but "FROM" flows better in the way it sounds in English and they way I've written it.)

examples:
```
FOR x=0 UNTIL x>10 GOING x=x+1
{
  print x.
}.
```
I wold propose that the expressions *init*, *end-check*, and *incrimenter* actually be designed to be any block of code, so they could include curly braced complex sections, like so:
```
// Iterate with x counting up while y counts down and they meet in the middle:
FOR {x=0. y=100.} UNTIL x >= y GOING {x=x+1. y=y-1}
{
  … stuff ...
}.
```
It should be instantly familiar to experienced programmers, while at the same time being easy to understand for first-timers because it describes the clauses and what they mean.

The implementation of parsing it is very simple, I think.  It's just that you take anything of the form:
FOR _A_ UNTIL _B_ GOING _C_ { _body_ }.
And make it behave like so:
_A_.
UNTIL _B_ {
  _body_.
  _C_.
}.
Most of the routines to do this are already there.  It's just a matter of linking them together.

### PR DESCRIPTION
**Pull request by Dunbaratu**
After much discussion in the slack chat, I settled on this
syntax:

FROM {statements} UNTIL expression STEP {statements} DO {body}

The braces in the FROM and STEP clauses are mandatory, to
help make them stand out, and to avoid the ugly look
of having naked periods in the middle of the statement.  Periods are
necessary because of how statements are defined all over the
place to mandate the period INSIDE the statement rather than
outside it.

(It's too hard now to go back in time and change
that rule to a proper one where the period is defined in a wrapper
rule around statements, and can therefore be left off in cases
like this.  To make that change now requires a major compiler
refactor, as it adds an extra nesting depth level that none
of the VisitBlahNode() methods in the compiler are currently
expecting, and changes their counts of how many nodes to
expect in node.Nodes.Count for certain alternations.)

Therefore since removing the periods is hard at this stage,
leaving the braces optional would have allowed this:

FROM stmt. UNTIL expr STEP stmt. DO {body}

Which looks really messed up with those naked periods in there.

That is why the braces are being made mandatory for the FROM and STEP clauses.

The rule is that the statements in the FROM braces are not REALLY in their own sseparate scope local to those braces (if they were, it would be impossible to use them inside the loop body), but that they will be unrolled and wrapped AROUND the loop body, so the body can see the scope.

The actual scoping turns from this:

FROM {AAAA.} UNTIL BBBB STEP {CCCC.} DO {DDDD.}

into this:

    {
      AAAA.
      UNTIL BBBB {
        DDDD.
        CCCC.
      }
    }

The initializer variables in the FROM are still local to the loop, but it's scope is wrapped *around* the loop so it can be seen inside the loop.

### CODE REVIEW

**Comment by Dunbaratu** [comment]
This was... suspiciously easy to implement once I came up with "compiling" a FROM loop by just rearranging its parse tree parts into a normal UNTIL loop and then letting VisitUntilNode() do all the heavy lifting later.  It meant I didn't have to do the error-prone work of repeating everything VisitUntilNode() was doing in regards to tracking return nesting levels and the like.

Still needs documentation.  And I'm nervous because, well, things aren't supposed to be this easy.  I must have missed something.

Please beat on it and try to break it.

**Comment by abenkovskii** [comment]
I asked it quite a few times here and there but I no one answered: why we have only `until` but not `while` in the kerboscript?

**Comment by Dunbaratu** [comment]
Once you have one there just isn't any point in having the other.

**Comment by abenkovskii** [comment]
`until not` looks strange to me. May be because in the russian for "until" is "пока не" and therefore for "until not" it is "пока не не".
> Once you have one there just isn't any point in having the other.

I know at least one place where `while` will be better:
```
set it to list:iterator().
until not it:next {
  // do stuff
}
```
```
set it to list:iterator().
while it:next {
  // do stuff
}
```
3. From my programming experience: `while` is more commonly used than `until`. Some languages even don't have the `until` loop at all but I can't remember other languages that don't have `while`.
4. For me using `while` to test loop's invariant looks more natural than using `until` to iets it's inversion.
5. Every unusual choice in the language design (e.g. having `while` *instead* of `until`) should have strong reasons. What are the reasons to prefer `until` to `while`?
P.S.
> Once you have one there just isn't any point in having the other.

Having both kinds of loops is a syntactic sugar. A very sweet piece of syntactic sugar if you ask me.

PR #1025, opened 2015-05-17
Add post-build events to copy the dlls into the plugin directory

yump:
This makes the edit/build/test cycle quite fast, if Resources/GameData/kOS is symlinked from your KSP GameData directory.  The manual copy is also removed from the release checklist.  Works on Linux and Windows.  Hasn't been tested on OSX, but the event for Linux should do the same thing there.


Dunbaratu:
Symlinking KSP/GameData/kOS to the build's GameData is problematic when you want to test the manufactured ZIP file like a normal user.  You don't quite get the same experience a non-developer would when installing it.

I'd much rather the symlink go in the opposite direction - that the dev environment links to the installed KSP rather than the other way around (so the KSP install contains no links and looks "normal").  And there would be a separate directory, lets call it, say, ``LinkToInstalled`` alongside of the build's GameData.

Then the idea is that the post-build install step would be "copy some stuff from dev's GameData to dev's LinkToInstalled".  Devs would then have to link FROM LinkToInstalled TO KSP/GameData/kOS, not the other way around.

If we did it that way, then we could delete our KSP/GameData/kOS folder, make a new one by doing the unzip of our package ZIP file like a normal user, and see everything working like a normal user.  We could do that without ever clobbering our dev links setups, and without accidentally getting a failed ZIP to appear falsely to be working because we're picking up our developer files intermingled with the installed ZIP files.

erendrake:
so symlink discussion aside. It doesnt look like there are any outstanding issues on this. I like the change. the pre/post build steps can get stupid complicated but this seems pretty good to me.

erendrake:
@yump how will this patch behave on OSX?

yump:
@erendrake [According to the documentation](http://www.mono-project.com/archived/porting_msbuild_projects_to_xbuild/#prepostbuildevents), it should do the copy as for Linux.  `cp` is a standard Unix command, so it should work on OSX.  I don't have an OSX machine to test it, though.

@Dunbaratu If KSP/GameData/kOS is a symlink to this_repo/Resources/GameData/kOS, your testing copy of KSP is automatically working with everything that goes into the zip, including the part configs and textures and whatnot.  I did notice, however, that kOS writes its persistent configuration into GameData/kOS/Plugins/PluginData, which is not something that should be included in the release zips.

Your outbound symlink idea does solve this problem and strictly enforce unidirectional flow of data from the git repo directory to the KSP install.  It does, however, complicate the build process.  The copying would likely have to be factored out into a shell script and a batch script that do nothing if the outbound symlink doesn't exist, rather than causing the build to fail. Plus, there's the question of whether building should just copy the .dlls to your KSP directory, or clobber all of KSP/GameData/kOS with the current contents of this_repo/Resources/GameData/kOS. For now, I've added a direction to the release checklist to delete the PluginData directory if it exists.

Dunbaratu:
@yump, The question wasn't "does the cp command exist on OSX". Yes of course it does.  What I was asking is whether or not the msbuild system "squashes" all the types of Unix down into one common string and just generically returns the string "Unix" when asked for the value of '$OS', regardless of whether that Unix is Linux, OSX, or whatever.  This is relevant because the test is '$OS'=='Unix' rather than something like '$OS' == 'Unix' || '$OS' == 'OSX' || '$OS' == 'Linux'. 

The doc page linked to didn't really clearly answer that very well.  It mentions that different Unixes will have a 'cp' command, but doesn't really explicitly say that msbuild gives all Unixes the same value for '$OS'.

I see nothing in my suggestion about the outbound symlink that requires any sort of change to the build system itself.  It's just a change to which way around you set it up before running the build system.  You either just mkdir the directory if you don't want to auto-install on build, or symlink it to point at KSP/GameData/kOS if you do want to auto-install on build.

One important note about making it auto-install on build that might make me not want to set it up that way anyway is this:  Sometimes I'm running the previous build of kOS in KSP to gather some data while I'm compiling the new change I just made just to test that it builds (without wanting to install yet).  Because of how Windows hates copying over a DLL that's being run right now, it will fail any build while the KSP program is running, if you make the copying of the DLL a part of the build process.

yump:
I should have linked to the other anchor, "Platform Specific Items":

>The default environment variable ‘OS’ is set to “Windows_NT” on all currently supported versions on Windows. xbuild sets this to “Unix” on non-Windows platforms, even on MacOS.

Issue #993, opened 2015-05-10
Can we now have one Function("round")?


abenkovskii:
0. Now functions can take variable number of arguments
1. Function("roundnearest") was never documented
2. I assume it will require some refactoring of complier too

hvacengi:
Technically, don't we just some how alias `round(d).` to be `round(d, 0).`?  It looks like there is a functionsOverloads dictionary in Compiler.cs, which only holds the round function.  It looks like we can just remove the round functions from this dictionary and proceed as normal.

@Dunbaratu Is there any reason to leave this legacy overload capacity in place?  This is literally the only reference I could find to the overload dictionary.

abenkovskii:
@hvacengi 
> Technically, don't we just some how alias round(d). to be round(d, 0).?

Currently they are separate functions.

> It looks like we can just remove the round functions from this dictionary and proceed as normal.

I already did this in my fork.

hvacengi:
Sorry, I wrote that first part before reviewing the code.  We do use the two functions, but it is still a matter of aliasing (overloading).  I meant it more to mean that it is documented, it's just documented using the overload not the base function.  But I agree that it would be cleaned up to be in one function, if for no better reason than to have a single function to maintain if we need to make a change.

If you have it running in your fork, submit a pull request.  The refactor itself can follow after you've updated round.  You might add "TODO" comment tags at the spots you expect to need a refactor, just so it's easy to find when you/we/they get around to it :wink:

Dunbaratu:
@hvacengi : 

> Is there any reason to leave this legacy overload capacity in place?

No.  It can be removed after changing round() as you suggest.

abenkovskii:
It's merged. I'm closing the issue.

PR #1030, opened 2015-05-19
Changing `round` to use the new arg count feature

abenkovskii:
solves #993 
P.S. In my view removing an old `functionsOverloads` should be a separate refactoring pass

### DIFF
--- a/src/kOS.Safe/Compilation/KS/Compiler.cs
+++ b/src/kOS.Safe/Compilation/KS/Compiler.cs
@@ -36,7 +36,7 @@ namespace kOS.Safe.Compilation.KS
         private CompilerOptions options;
         private const bool TRACE_PARSE = false; // set to true to Debug Log each ParseNode as it's visited.
 
-        private readonly Dictionary<string, string> functionsOverloads = new Dictionary<string, string>
+        private readonly Dictionary<string, string> functionsOverloads = new Dictionary<string, string> // TODO: remove this
         { 
         };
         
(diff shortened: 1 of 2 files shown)


src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -36,10 +36,8 @@ class Compiler
         private CompilerOptions options;
         private const bool TRACE_PARSE = false; // set to true to Debug Log each ParseNode as it's visited.
 
-        private readonly Dictionary<string, string> functionsOverloads = new Dictionary<string, string>
+        private readonly Dictionary<string, string> functionsOverloads = new Dictionary<string, string> // TODO: remove this
```
erendrake: if we dont need it, remove it.

erendrake:
I am fine with removing unused features like this with the same PR, but you are welcome to do it in two of course.

Dunbaratu:
In this case the feature didn't BECOME unused until this PR removed the need for it, so it makes sense for its removal to be part of this PR.  Prior to this PR's changes to the round() function, that piece of code was still needed.

erendrake:
@Dunbaratu yeah, that would be my preference too. @abenkovskii if you pull out the feature in this PR or another ill merge them at the same time.

abenkovskii:
OK. I'll try to figure out what parts of the compiler are unused now and can safely be removed and add commits to this PR. Should I test all the `kerboscript_tests/` every time I modify the compiler?

erendrake:
@abenkovskii i just sent you a PR that makes a few changes, including the compiler changes. i didnt have a chance to test it yet.

Issue #1029, opened 2015-05-19
Implement stock "Warp To" function


hvacengi:
Implement access to the stock "Warp To" function.

This is pretty easy to do, but has one significant limitation: you cannot cancel "Warp To" without keyboard interaction.  As long as we make that limitation explicitly clear to the user, that should be acceptable.

Currently, I'm implementing this as a function (in Misc.cs) which accepts 1 to 3 parameters (required number of parameters is shown in parenthesis): 
ut: the time to warp to (1-2-3)
maxwarp: the maximum warp setting to use (2-3)
minwarp: the minimum warp setting to use (3)

I don't really know what the minwarp setting does, but it is a stock option so I wanted to expose it.

evg-zhabotinsky:
Well, I have just written my own timedwarp.ks and run it whenever I need:
```
parameter warp_end.
local lock tm to warp_end - time:seconds.
print "Warping " + tm + " seconds forward.".
set warp to 0.
if tm > 5 {
	set warp to 1.
	if tm > 10 {
		set warp to 2.
		// And so on, nested further for higher warp speeds
		wait until tm < 10.
		set warp to 1.
	}
	wait until tm < 5.
	set warp to 0.
}
wait until tm <= 0.
print "Timed warp done.".
```

It takes 1 parameter (moment to warp to) and decreases warp speed 1 second (real time) before destination (because it takes some time to stop warp).
Only problem: it fails when you run out of power :)

------------

**On subject:**
To overcome "can't stop" problem of KSP's "Warp to" function, you can implement it yourself.
Basically the above logic, invoked with a function call and running in background, but also stop on writes to WARP variable.

PR #1032, opened 2015-05-19
Implement warpto

hvacengi:
Implement warpto function per #1029

* Added a new function `warpto`
* Accepts from 1 to 3 parameters
* 1 parameter: just the ut time to warp to
* 2 parameters: ut time and maximum warp rate
* 3 parameters: ut time, maximum and minimum warp rate

Todo:
- [x] Implement warpto
- [x] Figure out minimum warp parameter *(doesn't do anything I can see)*
- [x] <del>Maximum warp ignored below 3.0, use 3.0 as a minimum value for maximum?</del>
- [x] Documentation

### DIFF
--- a/src/kOS/Function/Misc.cs
+++ b/src/kOS/Function/Misc.cs
@@ -383,32 +383,35 @@ namespace kOS.Function
     {
         public override void Execute(SharedObjects shared)
         {
+            // TODO: As of KSP v1.0.2, the maxTimeWarping and minTimeWarping parameters behave unpredictably.  Disabling for now, we should revisit it in a later version.
             int args = CountRemainingArgs(shared);
-            double maxWarp = 8.0;
-            double minWarp = 2.5;
+            //double maxWarp = 8.0;
+            //double minWarp = 2.5;
             double ut = 0.0;
             switch (args)
             {
-                case 3:
-                    minWarp = GetDouble(PopValueAssert(shared));
-                    maxWarp = GetDouble(PopValueAssert(shared));
-                    ut = GetDouble(PopValueAssert(shared));
-                    break;
+                //case 3:
+                //    minWarp = GetDouble(PopValueAssert(shared));
+                //    maxWarp = GetDouble(PopValueAssert(shared));
+                //    ut = GetDouble(PopValueAssert(shared));
+                //    break;
 
-                case 2:
-                    maxWarp = GetDouble(PopValueAssert(shared));
-                    ut = GetDouble(PopValueAssert(shared));
-                    break;
+                //case 2:
+                //    maxWarp = GetDouble(PopValueAssert(shared));
+                //    ut = GetDouble(PopValueAssert(shared));
+                //    break;
 
                 case 1:
                     ut = GetDouble(PopValueAssert(shared));
                     break;
 
                 default:
-                    throw new KOSArgumentMismatchException(new[] { 1, 2, 3 }, args);
+                    throw new KOSArgumentMismatchException(new[] { 1 }, args);
+                    //throw new KOSArgumentMismatchException(new[] { 1, 2, 3 }, args);
             }
             AssertArgBottomAndConsume(shared);
-            TimeWarp.fetch.WarpTo(ut, maxTimeWarping: maxWarp, minTimeWarping: minWarp);
+            TimeWarp.fetch.WarpTo(ut);
+            //TimeWarp.fetch.WarpTo(ut, maxTimeWarping: maxWarp, minTimeWarping: minWarp);
         }
     }
 }
--- a/src/kOS/Function/Misc.cs
+++ b/src/kOS/Function/Misc.cs
@@ -1,13 +1,13 @@
-﻿using System;
-using System.Collections.Generic;
-using kOS.Execution;
+﻿using kOS.Execution;
 using kOS.Safe.Compilation;
 using kOS.Safe.Exceptions;
 using kOS.Safe.Function;
 using kOS.Safe.Module;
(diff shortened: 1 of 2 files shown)


src/kOS/Function/Misc.cs
```diff
@@ -373,4 +373,35 @@ public override void Execute(SharedObjects shared)
...
+                    break;
+                case 2:
+                    maxWarp = GetDouble(PopValueAssert(shared));
+                    ut = GetDouble(PopValueAssert(shared));
+                    break;
+                case 1:
+                    ut = GetDouble(PopValueAssert(shared));
+                    break;
+                default:
+                    break;
```
erendrake: I think that we should get into the practice of throwing here. we have an exception just for this that i recently changed in another PR to take multiple expected args. @Dunbaratu what do you think?

hvacengi: I forgot to move the case 1 logic into the default switch.  That's why I made sure I didn't set the timewarp until after we assert the bottom args.  In which case, passing zero args will throw an exception when trying to pop, and passing too many args will throw an exception with the assert before triggering the timewarp.  Unless you would prefer an explicit throw in the switch.

Dunbaratu: I think throwing an exception here when number of args is unexpected is a good idea.

erendrake:
@hvacengi you have some checkboxes here. Is that because its not ready to merge?

hvacengi:
@erendrake more to remind me to do it.  It is not ready to merge, if for no other reason than I haven't added documentation yet.  I can do that tonight.  The other things are just to try and keep me organized.

hvacengi:
I just did some more testing.  It looks like the effect of the maximum value (and possibly the minimum) is... unpredictable.  I think it depends at least partially on the maximum available time warp.  If I'm on the launchpad, it seems to ignore the value altogether and maxes out at mode 5.  If I'm in low orbit, it will respect a max value of 3, but anything lower than 3 defaults back to the maximum possible (5).  If my orbit is above 250km, it will respect max values down to 4, but below 4 defaults back to the maximum possible for that altitude again.

I'm starting to think that we should just ignore the maximum and minimum values for the time being, and then if @erendrake could put in a request for both the ability to cancel "WarpTo" and for better handling of the "maxTimeWarping" and "minTimeWarping" parameters.

I'm going to comment these cases and add a TODO: comment to remind us to revisit it in a later version, unless someone else has a better idea.

erendrake:
@hvacengi because this is a convenience method for doing something that we can already do with the existing WARP: command. By all means do the happy path first and we can do more with subsequent PRs

hvacengi:
Then it should be operational and documented.  Ready to merge once some one else has tested and reviewed.

evg-zhabotinsky:
@hvacengi , Do i understand correctly that this (KSP) function takes *floating point arguments* for minWarp and maxWarp? If so, than it is a bit... strange. As far as I can see, KSP only uses the same predefined 5x, 10x, 50x and so on when warping to specific time moment (with [click on orbit] > [warp here]).
I expected this function to take the same kind of arguments as TimeWarp.SetRate (integer from 0 to 7, or from 0 to 3 for physics warp). If not, then I don't understand what the hell these arguments mean.

evg-zhabotinsky:
I got it!
These `minTimeWarping` and `maxTimeWarping` are limits of **how long that warp should take in real world seconds!**

erendrake:
@evg-zhabotinsky have you tested this branch? I would like to have at least one person test it before we merge it.

evg-zhabotinsky:
Yes, kind of tested it.
I just uncommented lines related to these extra 2 arguments and launched KSP to find out what they actually do.
It also works as-is (with single argument only, obviously, but as I said, I don't think they are really important).

BTW, [this](https://github.com/hvacengi/KOS/blob/cc2a95660697d243649c146bdd4185951d362cbe/src/kOS/Function/Misc.cs#L386) comment is no longer correct.
Alternative suggestions:
* Change code to accept all 3 arguments and document their meaning.
* Remove code related to these arguments and add comment that they are not useful.
* Just fix the comment to reflect actual meaning of these arguments, leave actual code unchanged.

src/kOS/Function/Misc.cs
```diff
@@ -373,4 +377,41 @@ public override void Execute(SharedObjects shared)
...
+                    ut = GetDouble(PopValueAssert(shared));
+                    break;
+
+                default:
+                    throw new KOSArgumentMismatchException(new[] { 1 }, args);
+                    //throw new KOSArgumentMismatchException(new[] { 1, 2, 3 }, args);
+            }
+            AssertArgBottomAndConsume(shared);
+            TimeWarp.fetch.WarpTo(ut);
+            //TimeWarp.fetch.WarpTo(ut, maxTimeWarping: maxWarp, minTimeWarping: minWarp);
```
erendrake: @hvacengi i am going to remove the comments in this code because i dont think that they are accurate anymore?

hvacengi:
I'll be happy to remove the comments from the code now that we have an idea of how it works.  I'm not really sure how to document that behavior though.  Honestly, I'm having a hard time figuring out when you would want to use these parameters as units of time.  The only reason I limit my maximum warp mode is to prevent EC consumption bugs.  I can push an updated version with the 3 parameters enabled again, but I may need some one to help with documentation if nothing inspires me.  It may still make sense to just use the single parameter option for simplicity sake though.

Thanks @evg-zhabotinsky for figuring out the strange implementation. (looking at the parameter name, it actually makes sense now but it was not obvious to me initially).

Issue #968, opened 2015-05-06
[documentation] Arithmetic and logic operators are bad documented


abenkovskii:
1. There is no information about the order of operations
2. `True` and `false` are listed as logic operators
3. `(` and `)` are listed as  arithmetic operators
4. Some operators (`^` and `e`) probably should be explained

P.S. I know what `^` does but I don't know what `e` is for.

Dunbaratu:
`e` is for scientific notation.  1.432e32, for example.  But it's incorrect to call that an operator.  It's actually part of the parser's definition of what makes a number literal.

Dunbaratu:
merged PR, but the issue didn't close - dunno why.

PR #1035, opened 2015-05-20
Fixes #1034 Boolean AND and OR operations now are short-circuiting.

### LINKED ISSUE #1034: Short-circuit Boolean AND and OR operators.
At the moment, Kerboscript's AND and OR operators don't do short-circuit Booleans.  It was never formally defined whether they do or don't - its just that they coincidentally happen not to.

They really should.

(After this is implemented, then we can revisit old issue #18.)

This is my plan of attack:
---------------------------------

### AND case:

``expr1 AND expr2`` currently does this:

- Execute expr1, which pushes its result on the stack.
- Execute expr2, which pushes its result on the stack on top of expr1's result.
- Perform an OpcodeAnd - which pops the top two things and and's them, pushing the Boolean result of the ``and`` on the stack top

Changing it to short-circuiting means making it do this instead:

- Execute expr1, which pushes its result on the stack.
- br.false to "push false" opcode below.
- Execute expr2, which pushes its result on the stack.
- br.false to "push false" opcode below.
- push true // if it got this far, both expressions were true.
- branch to NOP Opcode below. // skip the false case.
- push false [target branched to from the br.false cases above]
- NOP.

### OR case:

``expr1 OR expr2`` currently does this:

- Execute expr1, which pushes its result on the stack.
- Execute expr2, which pushes its result on the stack on top of expr1's result.
- Perform an OpcodeOr - which pops the top two things and or's them, pushing the Boolean result of the ``or`` on the stack top

Changing it to short-circuiting means making it do this instead:

- Execute expr1, which pushes its result on the stack.
- br.true to "push true" opcode below. // (br.true doesn't exist yet- may make a new opcode or else do a 'not' followed by a br.false.)
- Execute expr2, which pushes its result on the stack.
- br.true to "push true" opcode below. // (br.true doesn't exist yet- may make a new opcode or else do a 'not' followed by a br.false.)
- push false // if it got this far, both expressions were false.
- branch to NOP Opcode below. // skip the true case.
- push true [target branched to from the br.true cases above]
- NOP.

BREAKING: This makes some simple small boolean expressions take more opcodes to execute, but makes more complex ones take less opcodes to execute.  Increasing number of opcodes executed means breaking some people's trigger code that was nearing the limit, or increasing some people's program sizes.

This will have the side effect of rendering OpcodeOr and OpcodeAnd rather useless, as short circuiting accomplishes the same effect by jumping past sections of code rather than by comparing two operands.  It's unclear whether they should remain in the Opcodes.  They probably should for the sake of other future language bindings, or for running KSM direct files, even though I'm nervous about leaving in a feature that doesn't get exercised (and therefore can become incorrect after future changes, without anyone noticing that it broke.)

Dunbaratu:
Breaking:  Not much.  Just if you used side-effects before, or were
near the upper limit of IPUS or KSM code size.

Note, this means the OpcodeOr and OpcodeAnd are in fact not being
used at all, but I didn't remove them.  I'm reluctant to cripple
the virtual machine's capacity for things just because kerboscript
doesn't use them.

### DIFF
--- a/src/kOS.Safe/Compilation/KS/Compiler.cs
+++ b/src/kOS.Safe/Compilation/KS/Compiler.cs
@@ -1130,13 +1130,17 @@ namespace kOS.Safe.Compilation.KS
                 case TokenType.arglist:
                     VisitArgList(node);
                     break;
-                case TokenType.expr:
                 case TokenType.compare_expr: // for issue #20
-                case TokenType.and_expr:
                 case TokenType.arith_expr:
                 case TokenType.multdiv_expr:
                 case TokenType.factor:
-                    VisitExpression(node);
+                    VisitExpressionChain(node);
+                    break;
+                case TokenType.expr: // expr is really the rule for doing an expression with optional trailing or-expression.
+                                     // 'or' just happens to have the lowest operator precedence level so it ends up being
+                                     // the outermost expression rule, thus getting the generic name 'expr'.
+                case TokenType.and_expr:
+                    VisitShortCircuitBoolean(node);
                     break;
                 case TokenType.suffix:
                     VisitSuffix(node);
@@ -1233,7 +1237,22 @@ namespace kOS.Safe.Compilation.KS
             identifierIsVariable = false;
         }
 
-        private void VisitExpression(ParseNode node)
+        /// <summary>
+        /// Performs the work for a number of different expressions that all
+        /// share the following universal basic properties:<br/>
+        /// - They contain optional binary operators.<br/>
+        /// - The terms are all at the same precedence level.<br/>
+        /// - Because of the tie of precedence level, the terms are to be evaluated left-to-right.<br/>
+        /// - No special extra work is needed, such that simply doing "push expr1, push expr2, then do operator" is all that's needed.<br/>
+        /// <br/>
+        /// Examples:<br/>
+        ///   5 + 4 - x + 2 // because + and - are in the same parse rule, these all get the same flat precedence.<br/>
+        ///   x * y * z<br/>
+        /// In cases like that where all the operators "tie", the entire chain of terms lives in the same ParseNode,<br/>
+        /// and we have to unroll those terms and presume left-to-right precedence.  That is what this method does.<br/>
+        /// </summary>
+        /// <param name="node"></param>
+        private void VisitExpressionChain(ParseNode node)
         {
             NodeStartHousekeeping(node);
             if (node.Nodes.Count > 1)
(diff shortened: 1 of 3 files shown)


src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -1233,7 +1237,22 @@ private void VisitVariableNode(ParseNode node)
...
+        /// - No special extra work is needed, such that simply doing "push expr1, push expr2, then do operator" is all that's needed.<br/>
+        /// <br/>
+        /// Examples:<br/>
+        ///   5 + 4 - x + 2 // because + and - are in the same parse rule, these all get the same flat precedence.<br/>
+        ///   x * y * z<br/>
+        /// In cases like that where all the operators "tie", the entire chain of terms lives in the same ParseNode,<br/>
+        /// and we have to unroll those terms and presume left-to-right precedence.  That is what this method does.<br/>
+        /// </summary>
+        /// <param name="node"></param>
+        private void VisitExpressionChain(ParseNode node)
```
Dunbaratu: Note, I didn't actually change anything in this method at all, other than to add the documentation comment.  This is the method that *previously* had been used for AND and OR expressions, along with all the binary operators for math - they were handled by the same method.  As I looked at this and started to edit it, I came to the conclusion that the code was getting too messy when I intermingled it with this method, so I split it out into a new method, VisitShortCircuitBoolean().  But I figured it didn't hurt to leave the explanatory comments behind because previously the method wasn't really explaining itself well.

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -79,6 +79,7 @@ public enum ByteCode :byte
         POPSCOPE       = 0x5b,
         STOREEXIST     = 0x5c,
         PUSHDELEGATE   = 0x5d,
+        BRANCHTRUE     = 0x5e,
```
Dunbaratu: I chose to add this to the end instead of next to BRANCHFALSE because if I moved all the enum values down one, it would break KSM.

Dunbaratu:
I've removed the undocumented math-ish boolean operators from Calculator.cs.  It was also doing really weird things like allowing ``false < true``, which sort of makes sense but really who would use that?  It's just easier to remove that feature.

I originally flagged this as "pending documentation" but I've changed my mind on that and removed the label because as I looked at the docs for where to put it, I realized that we don't really have a place that describes the operators in any detail anyway.  Therefore to add the documentation requires a larger issue first - to just document the math operators better in the first place so there's even a location to add the comment about short circuiting.

abenkovskii:
> Therefore to add the documentation requires a larger issue first - to just document the math operators better in the first place so there's even a location to add the comment about short circuiting.

I believe we already have one: #968

src/kOS.Safe/Compilation/Opcode.cs
erendrake: good call.

abenkovskii: Next update is already a ksm breaking. Isn't it?

Issue #989, opened 2015-05-09
[bug] Setting Warp in PHYSICS warpmode doesn't work.


Feinbube:
SET WARP TO 2. Will result in "Timewarp x1" in game. Both, in the terminal (I'm using putty) and in scripts. Works fine in RAILS warp mode though..

abenkovskii:
Confirmed. Strange. Suggest [bug] label.

abenkovskii:
I can confirm: it's solved now.

PR #1038, opened 2015-05-21
Fixed TimeWarp as per #989

evg-zhabotinsky:
*Offtop: I've finally managed to build this thing on Linux with Mono!*

Cleaned up that mess.
How did it work even for RAILS warp at all?!

Not closing the issue because switching warp modes should probably set warp to zero beforehand.
(Not sure about that, but switching warp modes without stopping warp beforehand is a bit glitchy right now.)


evg-zhabotinsky:
Emmmm....
Why automatic build system works THAT strangely and is broken so badly?

erendrake:
@evg-zhabotinsky Just imagine that the CI box looks like this 



We just started working on the CI setup last week and we have never worked with travis-ci before

PR #1040, opened 2015-05-22
Fixes #1036 with gui terminal beeping and more.

### LINKED ISSUE #1036: Add beep sound to in-game terminal.
EDIT: Changed description.

It is already possible to do beep via telnet interface, but the in-game tutorial should also respond to ASCII 7th character (BEL).

Dunbaratu:
NOTE: The question of how eventually we will actually
support making special characters in kOS strings is
outside the scope of this PR, and is expected to be
addressed as part of a future larger string functions
update.

Testing it:
-----------
In the meantime, to test this feature you'll have to
use the "spec_char.ksm" program from the KSLib github project.
After you do a:

```
   run spec_char.
```

Then you will now have a global variable called `beep` that
contains an ascii 7 BEL char, which you can now print:

```
   print "one beep" + beep.
   print beep + "two beep" + beep.
```

Note, when testing on telnet you may notice multiple beeps get mashed
into one beep if they occur fast enough that the first beep isn't done
yet when the second one starts.  This is not a bug in kOS.  That's just
how the terminals being emulated by most telnet clients actually behaved.

While this was being worked on, further suggested improvements
came up in chat discussion and were added to this PR commit:

- a visual beep setting so beeping can be just a flash.
- a reverse screen setting (as long as we have to reverse the
screen for the sake of the visual beep, it may as well support
just leaving it in reverse mode.


src/kOS/Screen/TermWindow.cs
```diff
@@ -612,18 +724,18 @@ protected void CheckResizeDrag()
...
+            ShowCharacterByXY(x, y, tx, ty, charTextColor, reversingScreen);
         }
 
-        void ShowCharacterByXY(int x, int y, int tx, int ty, Color charTextColor)
+        void ShowCharacterByXY(int x, int y, int tx, int ty, Color charTextColor, bool reversingScreen)
         {
             GUI.BeginGroup(new Rect((x * CHARSIZE), (y * CHARSIZE), CHARSIZE, CHARSIZE));
-            GUI.color = charTextColor;
+            GUI.color = (reversingScreen ? bgColor : charTextColor);
             GUI.DrawTexture(new Rect(tx * -CHARSIZE, ty * -CHARSIZE, fontImage.width, fontImage.height), fontImage);
```
Dunbaratu: This line is very inefficient looking and I'd like to address fixing it - but given then KSP is moving to Unity 5 next, I'm reluctant to mess with it now when everything about the GUI system is getting an overhaul anyway.

What this line does is paint one character out of the fontImage by positioning the font image's upper corner at a negative coordinate relative to the origin of the tiny 8x8 rectangle this BeginGroup is cropping to, such that the little 8x8 rectangle of the BeginGroup will only allow one of the characters of the fontImage to "bleed through" to be seen, cropping off the rest of it.  It still draws the whole fontImage, and relys on the BeginGroup to crop off the unseen parts of it.  It then repeats this over and over for each letter it prints.  It seems to me that it would be more efficient to pull a subrectangle for one of the characters out of the fontImage and paint just that, rather than painting the whole fontImage and relying on the cropping of the output rectangle to hide the parts of the fontimage outside of the desired letter.

abenkovskii:
1. It works.
2. reversed terminal looks weird:
screenshot9

Dunbaratu:
The reversed terminal looks pretty similar to what I remember reversed mode looking like on old terminals I've used in the past.  I allowed the mode because once I had temporary reverse working for the visual beep, it was no work at all to also allow that mode to stay on steady all the time,as an option.

The only hard work was fighting against the annoying Unity GUISkin system to get those little radio buttons on the terminal to show up like I wanted them to, compactly.  Now that I solved that, I may want to go back and revisit the Applauncher panel window and fix that annoying misalignment of those radio buttons.

hvacengi:
I can't test it yet because I've got a really busy day here at work (checking in during lunch).  But I'm curious, what does the inverse look like when the screen is not in focus?  Alpha blending may make that easier to read than the traditional mode, but it may also obfuscate too much more of the screen.

abenkovskii:
@Dunbaratu Let me clarify: a rectangle inside a rounded rectangle looks odd for me. In my view the whole background should change, not only part of it. I.e. it should look like this:
screenshot8
P.S. I'm bad in photoshop but you get the idea.

Dunbaratu:
Sigh.  Young People.  Just because the glass on a CRT tube is curved doesn't mean the electron gun paints the whole glass all the way to the edge of it.

Proof that the squared off sides are correct:  This is my IBM3151 pulled out of storage, set to reverse video mode:

ibm3151_in_reverse_mode

abenkovskii:
@Dunbaratu 
I see what you are trying to emulate. But the in-game version still looks unfinished for me. May be colores are off. Or may be it's because I treat the terminal as a part of GUI and I *know* it *can* print in those areas.

abenkovskii:
Or may be it's because the unprintable area at the bottom is bigger than unprintable area at other three sides.

Dunbaratu:
I might add one or two pixels around the edges of the letters to ensure that you can see the last characters correctly.  (So, for example, the cursor doesn't look like a void space biting out of the screen).

hvacengi:
Wait... we used to shoot electrons at a phosphor coated piece of glass to get images?  Can we implement a degauss feature too? :wink:

abenkovskii:
I personally dislike the idea of simulating the old monitor. As @Nivekk said in his video this doesn't make sense if the terminal is resizable.

Dunbaratu:
I like the way the terminal looks.  The resizing is an exception to the look, but it's a necessary evil.  I'd rather it resized to a few discrete sizes, representing different terminal models (one of which will be using ANSI color down the road), but that just doesn't deal with the telnet client well when we can't stop someone from resizing an xterm in which the telnet is running, and having the GUI terminal and the telnet disagree about the screen dimensions doesn't work (they share a sceenbuffer).  So as long as the telnet terminal can be resized, the GUI one has to be too.

PR #1045, opened 2015-05-23
Fixes #724 fixes #985

### LINKED ISSUE #724: log command does not respect disc size
how to reproduce:
1. let this program run or quite a while
```
set i to 0.
until abort {
  print i.
  log i to log.
  set i to i + 1.
}
list files.
```
2. press abort button
3. now you will see a file on your disk bigger than the disk itself.
P.S.
because it is possible to log kOS code this trick could be used to exceed available disk space.

### LINKED ISSUE #985: The compile command fails silently if there is not enough disk space
In my view it should raise an exception.
[Errors should never pass silently. Unless explicitly silenced.](https://www.python.org/dev/peps/pep-0020/)

abenkovskii:
Fixes #724 fixes #985 - for some reason GitHub doesn't recognize #number in the title.
screenshot10
screenshot11

### DIFF
--- a/src/kOS.Safe/Persistence/Archive.cs
+++ b/src/kOS.Safe/Persistence/Archive.cs
@@ -241,7 +241,7 @@ namespace kOS.Safe.Persistence
             }
         }
 
-        public override void AppendToFile(string name, string textToAppend)
+        public override bool AppendToFile(string name, string textToAppend)
         {
             SafeHouse.Logger.SuperVerbose("Archive: AppendToFile: " + name);
             System.IO.FileInfo info = FileSearch(name);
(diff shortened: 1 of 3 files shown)


src/kOS.Safe/Persistence/Archive.cs
```diff
@@ -255,9 +255,10 @@ public override void AppendToFile(string name, string textToAppend)
                 byte[] binaryLine = System.Text.Encoding.UTF8.GetBytes((textToAppend + "\n").ToCharArray());
                 outfile.Write(binaryLine);
             }
+            return true;  // is this correect?
```
erendrake: yes :)

Dunbaratu:
**breaking** - added breaking to the title to reflect the fact that fixing #724 means people can't "get away with" a thing they used to be able to get away with.

Issue #963, opened 2015-05-05
Add An Associative Array! AAAAAAAA


erendrake:
I have noticed that without user structures people are having a fun time passing lists around and having to worry about index. 

So i have decided to make a new associative array structure it should have the following features by the time i am done.

- [x] Be able to use any user object as a key
- [x] String keys are case preserving by default (for access, replace, removal)
- [x] case sensitivity can be enabled with the `:casesensitivity` or `:case` suffixes.
- [x] CRUD actions against each key
- [x] Replacement without error 
- [x] pretty print dumping like LIST currently has
- [x] dumping any collection type should be able to dump its inner collections.
- [ ] user documentation

because this gets to live in kOS.Safe there will be unit tests!

EDIT: i have been working on this for days and i just realized there was no issue :P
EDIT2: WIP can be found here https://github.com/erendrake/KOS/tree/assoc_array. WARNING i will be rewriting the history on this branch so dont get too attached to it :)

abenkovskii:
> String keys are case insensitive

problem:
```
set s1 to "abc".
set s2 to "ABC".
print s1 = s2. // false
//but my_arr[s1] and my_arr[s2] are the same index
```
It might cause cold fusion.

erendrake:
@abenkovskii when we make string a user object, i hope to change the string equality behavior. my hope being that everything in kOS is case insensitive.

abenkovskii:
@erendrake What about one-line array creation like: `list("abc", 42, 3.1415, my_var)` but for associative arrays?

Dunbaratu:
It is incorrect design to make it impossible for the user to pick case sensitivity vs insensitivity.
If you build insensitivity in at too low a level, you make it impossible for the programmer to choose.
The reason most languages don't do what you're talking about is that case insensitivity is a loss of information, which you don't want to enforce at too low of a level of the system.

Or to put it more clearly what the problem is:
You CAN implement case-insensitivity on top of case sensitivity.
You CANNOT implement case sensitivity on top of case-insensitivity.

Whether the kerboscript language is case sensitive is an entirely separate question from whether or not the values of its variables are.

Dunbaratu:
The solution is to provide users with the option to do case insensitive vs sensitive checking, not to make it impossible to escape from it.

[...]

PR #1049, opened 2015-05-25
 Assoc array

erendrake:
This PR is a big one :P which is too bad it contains the following major parts which if i had thought about ahead of time, i would have done with multiple PRs

* changed the existing `LIST:DUMP` implementation to be interface based. so we could have multiple collection types, this also ditched the reflection strategy that we have been using. This has some risk but with the unit tests @Dunbaratu added previously it was pretty painless.

* Adding a new associative array collection that is talked about over here #963

there are pending tasks to merging this, chief among these are user docs and in game test scripts. I just wanted to get some feedback from the group and maybe some testing :)


Dunbaratu:
If you got rid of the callstack reflection used in DUMP to detect if it was nested inside a verbose versus a terse dump, I'm glad.  I was never too happy with having to do that but it was the only way I could think of without a major refactor that was just outside the scope of what I was doing at the time.

erendrake:
Man, i have been so busy over the last few weeks but i am going to get some of this stuff cleared up today :)

@Dunbaratu  I dont even remember what the `Part 1` was about

cgruber:
Curious what the process is from here.  This work was last touched (in terms of commits on this branch) back in May.  I'm curious where we go from here.  I'm very keen to use associative arrays.  They fill a huge gap in the kscript language.  Valuetypes/user-objects would be better, but I can make do with maps.

hvacengi:
Honestly, I think @erendrake just got super busy and hasn't had a chance to do anything with it.  I've been meaning to check out the PR myself, but got tied up with updating the steering code.  I agree though, it will be a very nice feature to get included.  We're going to try and get some milestones figured out for next release, and should have a better idea of the eta at that point.

doc/source/structures/misc/lexicon.rst
```diff
@@ -0,0 +1,234 @@
...
+    // As a reminder, at this point your lexicon, if you did all the above
+    // steps in order, would look like this now:
+    //
+    //  FOO["ALTITUDE"] = 99999. // or whatever your altitude was when you added it.
+    //  FOO["ETA:] = 99. // or whatever your ETA:APOAPSIS was when you added it.
+
+    PRINT FOO:LENGTH.        // Prints 2
+    PRINT FOO:LENGTH().      // Also prints 2.  LENGTH is a method that, because it takes zero arguments, can omit the parentheses.
+    SET x TO "ALTITUDE". PRINT FOO[x].  // Prints the same thing as FOO["ALTITUDE"].
+
```
erendrake: I only made it this far adapting the list docs for the lexicon. this is where i will start from

doc/source/structures/misc/lexicon.rst
```diff
@@ -0,0 +1,234 @@
...
+    SET FOO TO LEXICON().           // Creates a new empty lexicon in FOO variable
+    FOO:ADD("ALTITUDE", ALTITUDE).  // Adds current altitude number to the lexicon
+    FOO:ADD("ETA", ETA:APOAPSIS).   // Adds current seconds to apoapsis to the lexicon at the index "ETA"
+
+    // As a reminder, at this point your lexicon, if you did all the above
+    // steps in order, would look like this now:
+    //
+    //  FOO["ALTITUDE"] = 99999. // or whatever your altitude was when you added it.
+    //  FOO["ETA:] = 99. // or whatever your ETA:APOAPSIS was when you added it.
```
Dunbaratu: `"ETA:` ? typo?

erendrake: yes

doc/source/structures/misc/lexicon.rst
Dunbaratu: I assume you did a cut-n-paste from array docs and that's why everything from here on down looks a bit wonky is because you haven't got this far yet in re-editing what was cut-n-pasted?  Is that what you mean?

erendrake: yes

src/kOS.Safe/Encapsulation/Lexicon.cs
```diff
@@ -0,0 +1,305 @@
...
+        public object GetIndex(object key)
+        {
+            T castKey;
+            if (key is T)
+            {
+                castKey = (T)key;
+            }
+            else
+            {
+                throw new KOSInvalidArgumentException("LexiconIndexer", "Index", key + " was invalid");
```
Dunbaratu: I think there's actually a KOS casting exception you could throw here.

src/kOS.Safe/Encapsulation/ListValue.cs
```diff
@@ -141,141 +141,78 @@ public override bool SetSuffix(string suffixName, object value)
...
+                    }
+                }
+                else
+                {
+                    var entry = string.Empty.PadLeft(depth * INDENT_SPACES);
+                    entry += string.Format("  [{0,2}]= {1}", index, item);
+                    toReturn.Add(entry); 
+                }
+            }
+            return toReturn.ToArray();
```
Dunbaratu: Wow, Github's diff'er really made a mess of that stuff above this point.  I can't follow at all what happened.  I'm going to have to wait until you're done and I pull this and look at it locally with smarter diffing tools to decipher what changed here.

src/kOS.Safe/Exceptions/KOSKeyNotFoundException.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace kOS.Safe.Exceptions
+{
+    public class KOSKeyNotFoundException : KOSException
+    {
+        private const string MSG = "The given key {0} was not present, use CONTAINS for checking safety. This is a case {1} collection";
+
+        public KOSKeyNotFoundException(string key, bool caseSensitive) : base(string.Format(MSG, key, caseSensitive ? "insensitive" : "sensitive"))
```
Dunbaratu: I'm slightly wary that stating "This is a case [in]sensitive collection" during the error message may trick users into thinking that kOS is trying to tell them this is the reason for their error.  It's printing it unconditionally regardless of whether or not that had anything to do with it.  It makes sense to print it when they looked for "cat" and the key was really "Cat", but not when they looked for "cat" and the key was really "dog".

At the very least, can it only be printed when it's sensitive?  It cannot possibly be the cause of error when it's insensitive.  (And the reverse for the duplicate key exception above, where case can only cause a duplication error when it's insensitive.)

Or maybe just not make it a separate sentence but just make it part of the description of the thing.  That will make it look like less of a claim that it's the cause of the error, while still hinting people that it might be.

i.e. "The given key "{0}" was not present in the case-{sensitive/insensitive} collection.  Use CONTAINS for checking safety".

erendrake: that sounds good to me. done

Dunbaratu: Can you also do it on the other exception above - the one for duplicate key?

erendrake: done

erendrake:
I just rebased this to the top of `develop` if you have checked it out already you might want to reset to the head of this branch rather than trying to pull it.

src/kOS/Function/Suffixed.cs
```diff
@@ -197,6 +197,17 @@ public override void Execute(SharedObjects shared)
...
+    public class FunctionLexicon : FunctionBase
+    {
+        public override void Execute(SharedObjects shared)
+        {
+            AssertArgBottomAndConsume(shared);
+            var listValue = new Lexicon<object,object>();
+            ReturnValue = listValue;
+        }
+    }
+
```
Dunbaratu: I never did like calling it "lexicon", but it's too late to change it now.  (It implies a text-based lookup (lexicographical), which implies the keys are strings, which they don't have to be).

doc/source/structures/misc/lexicon.rst
```diff
@@ -0,0 +1,160 @@
+.. _lexicon:
+
+Lexicon
+====
+
+A :struct:`Lexicon` is an associative array where the keys and values can be of any type in kOS. You can create your own :struct:`Lexicon` variables and kOS itself can return them to you.
```
Dunbaratu: About this -> "and kOS itself can return them to you" :
I had a read through the code and couldn't find where we return one of these from any method or function call.  Is this sentence here for the sake of future expansion?  (i.e. That you anticipate that we'll start returning associative arrays from the system later, but that's not in this particular PR?)

PR #1050, opened 2015-05-27
Changed the sounds, and added their sources to git.

Dunbaratu:


src/.nuget/packages.config
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="NuGet.CommandLine" version="2.8.2" />
```
Dunbaratu: This is not a deliberate change.  Something got automatically altered without me telling it to here.

PR #1054, opened 2015-06-03
Enable boot files on archive

### LINKED ISSUE #1046: Boot files don't work with CONFIG:ARCH
I play with CONFIG:ARCH set to true at all times. Now I tried to use boot files for the first time, and they don't work for me. As @abenkovskii explained, it is the connection of those two things (boot and CONFIG:ARCH) that cause the issue. <sup><sub>connected issue: erendrake is grumpy :)</sub></sup>

hvacengi:
Fixes #1046 
Update boot logic to maintain the original boot file name when copying to the local hard drive.  This allows for uniform file name access either on the archive or local drive.  At the same time I updated some of the underlying logic to make this easier to do, and to allow kOS scripts to be aware of the current volume (at least by name for now).

Todo:
- [x] Documentation

### DIFF
--- a/src/kOS.Safe/Module/IProcessor.cs
+++ b/src/kOS.Safe/Module/IProcessor.cs
@@ -3,6 +3,8 @@ namespace kOS.Safe.Module
     public interface IProcessor
     {
         void SetMode(ProcessorModes newProcessorMode);
+        string GetBootFileName();
+        void SetBootFileName(string name);
     }
     public enum ProcessorModes { READY, STARVED, OFF };
 }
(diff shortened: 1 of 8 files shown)


hvacengi:
Documentation has been updated to go along with the code changes.  If no one finds any bugs, this should be able to be pulled and we can close #1046

src/kOS.Safe/Module/IProcessor.cs
```diff
@@ -3,6 +3,8 @@ namespace kOS.Safe.Module
     public interface IProcessor
     {
         void SetMode(ProcessorModes newProcessorMode);
+        string GetBootFileName();
```
erendrake: @hvacengi why do this as two separate methods and not just a a property?

PR #1060, opened 2015-06-06
Support RemoteTech Local Control Check

### LINKED ISSUE #1005: "Signal lost" when using RT but the vessel is manned.
When using kOS with RemoteTech, If the vessel has an astronaut (*) in it, then you shouldn't need an antenna connection to use the kOS terminal, but the way the check is incorrectly implemented now, you do.

The check always incorrectly presumes it's ground control at KSC that has to issue all autopilot commands.

() - by 'astronaut' I mean a kerbal capable of taking the controls.  Tourists don't count as they are disallowed from touching anything in the cockpit.

hvacengi:
Fixes #1005.

Enable terminal input when the vessel has local control.  The archive is still inaccessible without a connection to the KSC.  This does assume of course that Jeb is able to type commands into the terminal correctly.

Long term I would like to add a delay in program execution whenever the archive needs to be accessed.

TODO:
- [x] Documentation

### DIFF
--- a/src/kOS/AddOns/RemoteTech/Addon.cs
+++ b/src/kOS/AddOns/RemoteTech/Addon.cs
@@ -20,7 +20,7 @@ namespace kOS.AddOns.RemoteTech
 
             AddSuffix("HASKSCCONNECTION", new OneArgsSuffix<bool, VesselTarget>(RTHasKSCConnection, "True if ship has connection to KSC"));
 
-            AddSuffix("HASLOCALCONTROL", new OneArgsSuffix<bool, VesselTarget>(RTHasLocalControl, "True if ship has connection to KSC"));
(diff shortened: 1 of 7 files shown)


src/kOS/AddOns/RemoteTech/Addon.cs
```diff
@@ -20,6 +20,8 @@ private void InitializeSuffixes()
 
             AddSuffix("HASKSCCONNECTION", new OneArgsSuffix<bool, VesselTarget>(RTHasKSCConnection, "True if ship has connection to KSC"));
 
+            AddSuffix("HASLOCALCONTROL", new OneArgsSuffix<bool, VesselTarget>(RTHasLocalControl, "True if ship has connection to KSC"));
```
erendrake: local control is not the same as "has connection to ksc"

hvacengi: Oops. Helps to fully edit when you copy/paste. I'll get it fixed tomorrow morning.

hvacengi:
I just pushed an update that should fix that error.

PR #1065, opened 2015-06-08
Allow to set RemoteTech antenna target

tomekpiotrowski:
Added a `target` field to every `ModuleRTAntenna`. The original `RTAntennaTarget` field is not editable and deals with Guids. This adds a 'virtual' `target` field which converts values to/from Guids on the fly.

### DIFF
--- a/doc/source/addons/RemoteTech.rst
+++ b/doc/source/addons/RemoteTech.rst
@@ -23,8 +23,9 @@ It is possible to activate/deactivate RT antennas, as well as set their targets 
   m:SETFIELD("target", "mission-control").
   // or
   m:SETFIELD("target", mun).
+  m:SETFIELD("target", "minmus").
 
-Acceptable values for `target` are: `no-target`, `active-vessel`, `mission-control`, a :struct:`Body` or a :struct:`Vessel`.
+Acceptable values for `"target"` are: `"no-target"`, `"active-vessel"`, `"mission-control"`, a :struct:`Body`, a :struct:`Vessel`, or a string containing the name of a body or vessel.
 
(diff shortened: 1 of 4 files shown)


hvacengi:
If this works... I will be so grateful to you.  I've wanted to do this for a while, but always figured I was going to need to modify the API from RemoteTech.  I will definitely download tonight and test.

hvacengi:
I've been running this code for about a week now.  It seems to work well.

It looks like the only issue I ran into was that the returned vessel and body are the actual ksp types of Vessel and Body, rather than kOS VesselTarget and BodyTarget types.  Which means the returned values have no suffixes on them, and equality cannot be checked.  I'll get you a PR to your fork with the fixes later today.

tomekpiotrowski:
Thanks!

That was one of the last things I was fixing before the pull request so it is possible that I've messed it up :)

Please note that this is the first piece of c# code I've ever written so I might have done some silly mistakes. Feel free to fix those as well.

hvacengi:
Actually, you did a very good job with the C# itself.  It's very clean.  By the time I'm done with it, it probably won't look as clean! :wink: 

I'm also adding the ability to set a target based on a string name i.e. `"Shuttle-1"` instead of `Vessel("Shuttle-1")` (you can ignore that part if you want).  My main goal there was to allow some one to set the kOS name tag of the antenna to a value (like "Mun" or "Shuttle-1") and then we can easily change the target based on the tag.  By allowing setting by name, you can more easily set the antennae to target different types (one may be the active vessel, another the mun, and another would then be a relay satellite).

tomekpiotrowski:
Great! I was thinking about that functionality as well. If we also had an API like the one mentioned in RemoteTechnologiesGroup/RemoteTech#365 we could make some really interesting things :)

PR #1070, opened 2015-06-16
Fixes #289 making triggers interruptable

### LINKED ISSUE #289: Make triggers able to span multiple updates doing away with all the annoying behaviors.
The fact that triggers must continue to the end and thus must fit inside one update is rather annoying to me and to lots of users who find out about the problem the hard way.  We've made stopgap solutions like throwing the exception on long triggers but there's a much better solution - just make triggers able to span multiple updates, like so:

Current Cpu.Update() behavior is this:

Step 1 - Run flybywire-based lock expressions to completion (not all the locks - just the ones used by LOCK STEERING or LOCK THROTTLE.  Most locks just happen on the fly when they're referenced in an expression and the rest of the time when not being referenced they don't do anything.)

Step 2 - For each trigger:
  If the trigger condition is true
    Run the trigger to the end and (maybe) delete it.

Step 3 - Mainline code:  Until number of instructions executed > CONFING:IPU, run another instruction.  When CONFIG:IPU is reached, then remember the instruction pointer to pick up from there next time, and quit the update.

My proposed change is to do this instead:

Use the same logic as step 3 in Step 2 as well.

Maintain a queue of "what needs to run yet" which is a list of triggers or mainline code contexts.  When a trigger becomes true, add the trigger to the end of the queue UNLESS it's already in the queue then don't.  

Execute items in the queue round-robin, as follows:

When executing a "thing" (trigger or mainline), it will either end on is own before CONFIG:IPU or it will end because it hit CONFIG:IPU limits and there's still more code in the body to go.  When the execution of a "thing" is done, do one of the following:

If it completed and isn't being PRESERVEd,  then remove it from the queue.

If it completed and IS being PRESERVEd, then reset it's execution state (current Instruction pointer) and move it to the back end of the queue.

If it did not complete because it hit CONFIG:IPU first, then remember its execution state (current instruction pointer) in part of the queue's data structure for that item in the queue, and move it to the back of the queue.

In either case where it completed (being PRESERVEd or not) then that means there are still some instructions left in CONFIG:IPU.  Keep going, using them on the next item in the queue.  (i.e. the count of instructions so far this update is global to the CPU across all items in the queue.

What this does is turns triggers into "equal" citizens with mainline code, being swapped in and out of execution along with the mainline code.  In fact, it turns mainline code into essentially a trigger that's always true, in a sense.

I think this would make things a LOT clearer.  Perhaps it can be added to the 0.16 list.

Dunbaratu:
Now triggers can span multiple FixedUpdates, because
the system memorizes where it was within the ProcessTriggers()
when it hit the IPU limit, and it picks up from there next
time the KOSFixedUpdate happens.

If you create a new trigger (i.e. by one program running another
so it compiled new triggers into the program context) during an
update, the system won't start checking for it to fire until after
it finishes going over the list of triggers that existed the last
time it started on the list.  In other words if you have so many
triggers that they're taking, say, 5 FixedUpdates to finish each
pass through the triggers list, then it will be 5 FixedUpdates
before a newly added trigger starts being checked for.


Dunbaratu:
Due to some recent discussions on slackchat #ideas channel, there may be a waaay better way to implement this.  Added the Not Ready label for now.

erendrake:
@Dunbaratu what is the status of this PR?

Dunbaratu:
Do not pull.  This was superceeded by @hvacengi's work on threads, which in turn was superceeded by finding out that KS 1.1 was going to be fully threaded anyway making a lot of the complication we worked out probably unnecessary.  We have to see what Unity 5 does to threading, and most likely revisit all of the work that was threading - based.

erendrake:
Closing this so we can rework a new solution to the problem

Dunbaratu:
Could we use a new tag label for github, "wontmerge" - like "wontfix" but for PR's?

PR #1080, opened 2015-06-28
Cleaning up errors detected while generating docs.

Dunbaratu:
Also, re-instituted the part of changes.rst that referred
to the prev version of kOS.  The intent I had for that page
was for it to have a table of contents for each version you
could click on, not to delete old versions from it.

### DIFF
--- a/doc/source/about.rst
+++ b/doc/source/about.rst
@@ -4,6 +4,6 @@ About **kOS** and **KerboScript**
 =================================
 
 kOS was originally created by Nivekk.
-It is under `active development`_ by the
+It is under active development by the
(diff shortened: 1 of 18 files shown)


Dunbaratu:
don't pull - found an error with the link for CORE struct giving a 404 not found.

Dunbaratu:
Okay false alarm.  The doc sources are fine - I just forgot to do a ``git add *`` after replacing my gh-pages working area with the output from the build, so core.html, being a new file that didn't exist before, didn't make it into the gh-pages branch to be hosted on github.io.  Fixed now.

Dunbaratu:
Dang it - I hate that there's a close button that means close-issue when I keep thinking it means close the user interface box.

Dunbaratu:
okay removed the "not ready" - pull it.

PR #1081, opened 2015-06-30
Quickstart Tutorial update for new aero

### LINKED ISSUE #1072: Quickstart tutorial is obsolete and fails in KSP 1.0's aero model
The documentation's quckstart tutorial is confusing for new players I think because it describes being able to perform a crude launch by going straight up to 10km, then turning 45 degrees, etc - the sloppy way that used to work prior to KSP 1.0.

There's no way that a new user, unaware of the way KSP used to behave in its early access days, would find that tutorial believable anymore.

We need a new quickstart for the new areo model.  Its got to be really simple but still not flip in new stock areodynamics.

Suggestion: instead of claiming it will work with "any" rocket, we instead specify the exact rocket build the tutorial is designed for.  Then we can still hardcode the steering angles but guarantee a successful launch with the example craft.  (Then the tutorial can even go on to try the same program on a second rocket design, known to fail with those steering angles - maybe it has a smaller TWR and thus doesn't have the same flight curve so hardcoding the same steering points will point it off prograde and cause a flip.  It could do this as a teaching moment explaining why it's better to write the script more generically, as a lead-in for a more smooth launch script as a future part 2 to the tutorial.)

space-is-hard:
Resolves #1072


doc/source/tutorials/quickstart.rst
```diff
@@ -146,23 +146,24 @@ But in order to give you an example that you can start with from scratch, that's
 Step 1: Make a vessel
 ^^^^^^^^^^^^^^^^^^^^^
 
-Make any sort of rocket that can lift you to orbit that fills the following pattern:
+Please make the vessel you see here:
```
Dunbaratu: I'd be a bit more explicit here and add something to the effect "This tutorial doesn't work well for all rocket designs, and is tailored to this specific craft.  You need to build this exact craft to work with this example."  Rather than "please" make the vessel you see here.  Imply that it won't work at all if you make a different one (not entirely true, but it gets the point across).

doc/source/tutorials/quickstart.rst
```diff
@@ -357,84 +374,239 @@ Step 6: Now to make it turn
...
-Note that these lack the command PRESERVE like the previous WHEN example had. This is because we want these to trigger just once and then never again. There's no point in constantly telling **kOS** to reset the steering to the same thing over and over as the script runs.
+Instead of using WAIT UNTIL to pause the script and keep it from exiting, we can use an UNTIL loop to constantly perform actions until a certain condition is met. For example::
 
-Now, if you insert this new section to the script, we have a nice nifty example of a start of a launching script. Note that it works even if you insert it at the top of the script, because it sets up the triggers to occur LATER when the condition becomes true. They don't execute right away::
+    UNTIL APOAPSIS > 100000 {
+        LOCK STEERING TO HEADING(90,90). //90 degrees east and pitched up 90 degrees (straight up)
+        PRINT SHIP:APOAPSIS AT (0,15).
+        //We use the PRINT AT() command here to keep from printing the same thing over and
+        //over on a new line every time the loop iterates. Instead, this will always print 
+        //the apoapsis at the same point on the screen.
```
Dunbaratu: Hmm.... maybe two commands back to back:
```
    PRINT "            " at (0,15).
    PRINT ROUND(SHIP:APOAPSIS,0) at (0,15).
````
Just because one of the frequent problems new users encounter when using this print technique is how to deal with a number that gets smaller over time.  In this case it doesn't, but if it does, then you get leftover digits on the screen to the right that falsify the display and make the number look bigger than it is.  This isn't *too* complicated and it lets us introduce the concept.

File: doc/source/tutorials/quickstart.rst [hunk 3]
```diff
@@ -357,84 +374,239 @@ Step 6: Now to make it turn
...
 
-- You could change the steering logic to make a more smooth gravity turn by constantly adjusting the pitch in the HEADING according to some math formula. The example shown here tends to create a "too high" launch that's a bit inefficient.
-- You could complete the launching script by making sure once the vessel breaks the atmosphere it actually makes a circular orbit rather than just stopping after 70000m and coasting.
-- This script just stupidly leaves the throttle at max the whole way. You could make it more sophisticated by adjusting the throttle as necessary to avoid too much wasted energy fighting air friction. (The way **KSP**'s stock areodynamic model works, the optimal speed is terminal velocity, by the way). This is partly addressed in the :ref:`PID Loop Tutorial <pidloops>`.
-- With more sophisticated staging checks, the script could be made to work with solid fuel engines as well.
-- With even more sophisticated checks, the script could be made to work with fancy staging methods like asaparagus.
```
Dunbaratu: I'd put this bullet point back in.  The MAXTHRUST = 0 check does not do asparagus staging properly, so that is still an improvement users can make.  (With asparagus staging, maxthrust doesn't go all the way down to zero when it's time to stage - it drops, but it doesn't drop all the way to zero, because you have some engines burned out but other engines still going.)

hvacengi:
```
set stagemaxthrust to ship:maxthrustat(0).
when (ship:maxthrustat(0)<stagemaxthrust or ship:maxthrustat(0)<1) then {
	if stage:ready {
		stage.
		set stagemaxthrust to ship:maxthrust.
	}
	if ( stage:number > 0 ) { preserve. }
}
```
This is (essentially) my staging script which works for asparagus staging and traditional staging alike.  It doesn't work for staging jets, but I think that's a special enough case.  You could introduce this topic in the code itself, or you might edit this bullet point to suggest comparing maxthrust to a stored historical value, and/or checking the engines for flameout.  Maybe just add "... like asparagus, by checking for changes in `maxthrustat(0)` or checking for engines with the `flameout` suffix returning true.

Issue #188, opened 2014-07-07
With FAR installed SHIP:TERMVELOCITY puts infinity to the stack.


nuclearphoenixegg:
While the ship is in motion and in the atmosphere SHIP:TERMVELOCITY returns infinity. It doesn't do this without FAR installed.

dcrockwell:
+1 for FAR support in kOS. It is extremely popular to the point that it warrants special support.

Perhaps if there were just a way to get atmospheric density at sea level from KerboScript? Then there would be no special support needed and we could do our own calculations based upon this, scale height, and atmospheric height.

Dunbaratu:
-1 to FAR support in kOS.
In addition to my usual reasons that FAR is being a bad citizen when it breaks the API and demands that other mods change to do things its way, there's also the fact that we have no idea yet what to expect when KSP 1.0 changes the stock aero system.  Whatever that does is going to require a change anyway and maybe invalidate anything we do this close to its release.

gillonba:
Also -1 to FAR support.  I would be inclined to agree with dcrockwell based on FAR's popularity, with Squad reworking the aero model there is a good chance that FAR will soon become much less relevant in the near future.  If that is the case, effort spent supporting it will have been wasted effort that could have been better spent elsewhere

dcrockwell:
>  with Squad reworking the aero model there is a good chance that FAR will soon become much less relevant in the near future.

I was unaware. I think you're right. -1 FAR support.

futrtrubl:
I don't think FAR is going anywhere. While the stock aero system is becoming more intuitive it is still not nearly as realistic as FAR. For one I don't believe it changes drag due to mach effects. Also I don't think it sets a precedent as FAR is a huge change to the system so it really had to change the underlying behavior. So I think special support is warranted and I believe just a matter of getting the terminal velocity straight from FAR.

erendrake:
@futrtrubl I dont think FAR is going anywhere either but i would advise letting the smoke clear from the 1.0 release before we continue with this.

ferram4:
I fail to see how FAR is being a bad citizen by breaking a non-existent API; it's not like there's a "vesselDrag" parameter that I can set.  In fact, it doesn't falsify the data at all; you're calculating the terminal velocity based on the drag of the stock massdrag system and getting the expected result when that is disabled with FAR.

[...]

PR #1085, opened 2015-07-01
added binding for dynamic pressure

erendrake:
Progress towards #188

This may not actually deal with the FAR integration problem.

### DIFF
--- a/src/kOS.Safe/Encapsulation/ConstantValue.cs
+++ b/src/kOS.Safe/Encapsulation/ConstantValue.cs
@@ -5,24 +5,25 @@ namespace kOS.Safe.Encapsulation
 {
     public class ConstantValue : Structure
     {
+        /// <summary>
+        /// kiloPascals to atmospheres
+        /// </summary>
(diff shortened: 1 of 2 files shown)


hvacengi:
We should (and I will later today) double check any other values returned in kOS for proper units.  You convert to Pa and most KSP methods return either kPA (as the underlying code here does) or Atm.  I think the units returned should remain consistent unless there is a major reason not to.

erendrake:
Excellent point. What do we think makes the most sense? Atm?

Dunbaratu:
Although ATM is the most human-readable, for most math formulas you end up having to convert ATM into something more universal anyway, and a kOS script is going to be doing math on it, I presume.

erendrake:
So im coming back to this. My thoughts are no matter which we choose we should provide some kind of constant that will let you take the kpa or atm and convert it to the other. 

Right now it seems that we have a bias for display units in the API. Degrees come to mind. Using ATM would fit that pattern. Is that a pattern we like?

Dunbaratu:
edit this to return ATM's -  that was the decision we made in video chat.

PR #1101, opened 2015-07-19
Scriptable science

### LINKED ISSUE #1075: Fully automated science
If I'm not mistaken currently there is no way to fully automate the science gathering process using kOS. After you run for example `module:doevent("observe materials bay")` a window pops up and a user intervention is required. Specifically there is no way to Transfer Data back to Kerbin programmatically. 

Has anyone ever tried to tackle that problem? I saw that it was mentioned briefly in #60. Or does clicking few buttons simply does not bother anyone enough? :)

tomekpiotrowski:
Fixes #1075.

Adds a couple of suffixes to ModuleScienceExperiment that allow to do science without clicking anything. I haven't had time to test it much, so any feedback would be appreciated.

Does not work with SCANsat as their scanners do not implement ModuleScienceExperiment. I imagine there are other mods out there that do this as well.
Does not implement science lab related stuff, but I don't think that's needed.

### DIFF
--- a/doc/source/structures/vessels/scienceexperiment.rst
+++ b/doc/source/structures/vessels/scienceexperiment.rst
@@ -1,0 +1,114 @@
+.. _scienceexperimentmodule:
+
+ScienceExperimentModule
+======
+
+The type of structures returned by kOS when querying a module that contains a science experiment.
+
+Some of the science-related tasks are normally not available to kOS scripts. It is for
+example possible to deploy a science experiment::
+
+    SET P TO SHIP:PARTSNAMED("GooExperiment")[1].
+    SET M TO P:GETMODULE("ModuleScienceExperiment").
+    M:DOEVENT("observe mystery goo").
+
+Hovewer, this results in a dialog being shown to the user. Only from that dialog it is possible
+to reset the experiment or transmit the experiment results back to Kerbin.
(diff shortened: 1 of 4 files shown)


hvacengi:
I'll check this out tomorrow night.  Does it still display the UI, or did you find a way to just save the science directly?

tomekpiotrowski:
The gatherData method from ModuleScienceExperiment actually accepts a boolean parameter, which specifies whether to display the UI or not. The only problem was that it was private, so I'm using reflection to call it.

hvacengi:
That's too bad.  But since it isn't exactly the kind of thing that the user will iterate over 100 times, it shouldn't be a noticeable performance hit (unless they try to run 100 experiments at once...)

Dunbaratu:
The reason using reflection to make the call is potentially dangerous isn't the performance hit.  It's the fact that it violates the privacy settings such that you're using an API that isn't meant to be used.  It could break in a future version very easily, or could currently be bypassing some important step you didn't realize.

tomekpiotrowski:
I realize all that and I agree. However, AFAIK there is no other way of doing this. So it's either reflection and some risk of future incompatibilities or we'll have to skip that feature for now.

Dunbaratu:
merged... but still open?

PR #1108, opened 2015-07-23
Resolves #1096 vecdraw atmo effects.

### LINKED ISSUE #1096: The drawn vectors have atmospheric effects.
only a minor graphical bug
The vecdraw arrows catch fire during re-entry and get Mach effects at high speed.

Dunbaratu:
This was simple - just change the layer to draw the lines on.


Dunbaratu:
Here's an example of screenshots:

Before the fix:

screenshot39

After the fix:

screenshot41

erendrake:
I almost dont want to fix this ;)

Issue #1113, opened 2015-07-29
Trim improvements


ItMustBeACamel:
One bug, and one proposal.

The bug: setting trim via the raw controls like "SHIP:CONTROL:PITCHTRIM" does actually do nothing if the craft is currently controlled by the player. This is due to FlightInputManager overwriting the  control state of the craft with its own on every frame.

The proposal: The way kOS currently handles raw controls makes it necessary to neutralize a control input before the player can control it him- or herself again. For normal controls this makes sense, but not for trim. Trim is normally used to maintain a stable state like level flight/ climb/descend. Real world autopilots often pitch a plane by setting the trim, rather then pulling on the yoke. This way the plane does not suddenly drop into a dive after deactivating the flight assistance. So my proposal is to make all pitch related pilot controls settable, like "PILOTMAINTHRUST" already is. That way it is possible to either lock away the trim from the player (PITCHTRIM) or set the trim permanently (PILOTPITCHTRIM).

Dunbaratu:
Closing as 'wontfix' given that the PR it was attached to was closed as "wontfix".  If I'm wrong to do this, re-open it, @erendrake.

erendrake:
Keep it closed, I don't want kOS setting to to the input handler any more often than absolutely necessary

PR #1114, opened 2015-07-29
trim bug

ItMustBeACamel:
FlightInputManager overwrites the trim setting each frame, so we have to set it in FlightInputManager itself if vessel is controlled by the player. See issue  #1113


erendrake:
First off, thank you very much for your PR! I am sorry we have been so slow in reviewing it. 

I am a little reticent to change the state of the flightinputhandler because up to now it has been reserved for user. Wouldnt this change disable the user from overriding kOS control over trim in flight?

erendrake:
after review i realized that part of our design is to always let the user override all of the settings.

Issue #1107, opened 2015-07-22
Place parameter check up top.  Prevents varargs, but is more robust.


Dunbaratu:
@gisikw posted an issue, #1100 which is closed, but I'd still like to address an issue connected to it.

The problem is that since we don't check number of arguments until the bottom of the function, users typically end up seeing complaints about wrong argument types before they see a complaint about mismatched arguments, because passing the wrong number of args causes stack misalignment.  (Which IS detected eventually, but not until after it tried using the values that were shifted into the wrong variables, which often causes its own errors that stop it before it gets to the check for missing or too many args).

The entire reason for not checking until the bottom was to allow for clever varying args situations where you execute the ``declare parameter`` commands conditionally depending on what you see.

But I doubt anyone is doing this, and it's preventing us from giving proper error messages out for mismatched argument counts.

I'd like to switch it to a system that requires the declare parameter commands to be "flat" in the outermost nesting level (not contained inside loops or if's), and can throw an error right away when they don't have the right number of arguments to match.  I might provide people with an alternate technique where they can get a LIST() of parameters instead if they like to try some vararg logic.

gisikw:
Just curious here on the latter part. Are you referring to setting up an external mechanism to make passing LIST()s to functions easier? Or adding something akin to the JavaScript `arguments` pseudo-array?

Can definitely think of some uses for the latter (particularly passing "flags" to functions, which couldn't be done using conditional declares). Either way, excited for this :)

If I can feature creep a tiny bit...would there be any possibility of allowing functions to be passed as arguments (either directly or through a `&myFunction` syntax or something). Right now, they're automatically executed whenever they're referenced. But if we had higher-order functions, we'd be able to create user pseudostructs in KerboScript!

Dunbaratu:
I was considering a means of varargs that make a list for you, not something where you just pass a list, which you can do now.  The idea is that it would get the remaining args up to the end, like so:

```
function foo {
  parameter x,y.
  parameterlist z.
  // .. body here
}

foo( 1,2,3,4,5,6,7).
```
That would fill `x` with `1`, `y` with `2`, and `z` with `list(3,4,5,6,7)`.

A parameterlist would have to come last, only allowing varying args at the end.


As for function pointers, it's definitely a doable thing if we just add the syntax for it.  Right now we don't just because it wasn't a high priority.

gisikw:
Oh nice, so just a trailing splat. I like it!

hvacengi:
[...]

Issue #1031, opened 2015-05-19
Variable parameter count?


evg-zhabotinsky:
I have found that reaction to passing wrong number of parameters **when running a program** is inconsistent. (Calling functions works as expected.)

Passing too many arguments usually ignores first arguments and everything works.
Passing too few is more funny.

It often shows the proper error message (BTW it is quite misleading, stating things like "Too few arguments were passed **to print**").

It sometines complains about unbound instances.
I get something that prints as "$argpart" (not sure, could not get it while writing this) moderately often.

Running program with 1 parameter declared from terminal with no parameters consistently results in that only parameter being zero. I even thought it is the way all unspecified parameters are marked.

------

Alternative fix/suggestion: report missing parameters as some predefined value (null?) and allow testing for it (both for programs and functions). That way it will be possible to implement functions with variable parameter count. Not sure if it is possible with your bytecode.

abenkovskii:
1. Yes. The the wrong number of arguments passed to a function or to a program is an undefined behaviour because it causes stack misalignment.
2. Your idea is nice but I'd prefer `parameter param_name default expression.` syntax.
3. We can change our bytecode if we want. And we are doing it from time to time.

evg-zhabotinsky:
Ignore the "alternative fix" words i've written.
You still need to figure out if the parameter was specified or not, and this is broken for programs.
Maybe there should be two separate issues?

abenkovskii:
> Maybe there should be two separate issues?

No. In my view it's two sides of the same problem.

Dunbaratu:
Calling a program and calling a function use slightly different techniques, under the hood.  Once upon a time there was no checking for program parameters at all.  It just let the stack get misaligned without mentioning anything and then whatever other error that happened to cause later is the error you saw - usually involving something not being the expected type (because the stack misalignment made it pop the wrong item from the stack into the variable in question).

This problem was dealt with for functions when functions were added, by adding an extra padding stack parameter, a string called "$<argstart>" that gets put on the stack to mark where the boundary between other stuff and the passed arguments are, and then popping it back off when returning from the function.  But not all of that approach got back-ported into the run program code.  The run program code pops the "$<argstart>" to avoid stack mismatch, but it doesn't have all the work to check the number of args.


One very ugly problem is that ages ago, Kerboscript got defined to allow (and I find this a very bad idea, but it was how the language was defined before I got my hands on it) a mishmosh of DECLARE PARAMETER statements mingled with actual code.

[...]

PR #1133, opened 2015-08-26
Fixes 1107 by enforcing arg/param matching.

Dunbaratu:
### The design changes in a nutshell:

1 - All program calls now push an arg bottom mark to the stack under the arguments (this was standard practice for calling user functions, but the practice hadn't been extended to RUN PROG(arg,arg,arg) - now it has been).

2 - For both the outermost list of statements in a program, and the braces of a function, at the moment where the lastmost PARAMETER statement is encountered is when the check to assert the arg bottom mark is atop the stack, ~~and pop it off~~ will happen.  (In the past this check was performed only at the bottom of a function, at the return statement, by which point the misalignment probably would have already caused another fatal error anyway.  Now it's performed as early as possible instead, when there are no more PARAMETER statements remaining, to ensure the params consumed match the args sent up at the top, probably before any of the args get used.) [edit: as described in comments below, it does NOT pop it off at this point - that tactic changed.  This check is performed with a stack *peek* rather than a stack *pop* - it still waits for the actual RETURN or end-of-file to do the popping-off of the argbottom mark.  This is so that it guarantees premature returns that read a shorter number of args, and use them to choose to quit before the lastmost ``parameter`` statement don't end up failing to pop the argbottom.  The RETURN causes the pop, rather than the act of visiting the lastmost `parameter` statement, which the premature return will end up skipping.]

3 - But the checking is still happening at runtime, not compile time, which still leaves open the chance of doing varying args if you're careful.  The new test program `functest31.ks` shows an example of how it can be done.  Have a look at it.  It is not mentioned in any docs yet because I am unsure if we want to expose it, but the ability is present in the code.  (varying args are a good way for people to shoot themselves in the foot).

4 - To enforce the check for there being fewer args on the stack than parameters being pulled from the stack, it now checks every time you try to do an OpcodeStore, OpcodeStoreLocal, or OpcodeStoreGlobal to ensure that the value you're storing isn't allowed to be the arg marker.  If it is, it means you must have dug down too deep into the stack to obtain that value and thus there were too few args passed.

5 - In the process of all the above, I finally did the refactor I'd wanted to do for a while, of changing the arg bottom marker from a magic string into a special type of its own (`KOSArgMarkerType`).  This makes it a faster check than a string compare, and also makes it impossible for a kerboscript to make its own string value that clashes with it.  The reason this step was finally done was because I now need to perform the check much more often and the expense of the string compare was starting to feel icky.


kerboscript_tests/functest30.ks
```diff
@@ -0,0 +1,8 @@
+print "Testing the case of a RUN in a loop.".
```
Dunbaratu: I added this test case because I was messing with the area of the code where terminal RUN differs from in-program RUN, and I wanted to make sure that when doing a re-run of the same subprogram, where it skips the compile step and just uses the previously compiled version, that I hadn't broken that case.  I noticed that none of the existing functests covered this case properly.

kerboscript_tests/functest31.ks
```diff
@@ -0,0 +1,38 @@
...
+// the first arg is a number: how many more args to expect.
+// the rest of the args are just printed for testing.
+function print_var_args {
+  parameter how_many.
+  
+  // read parameters, filling a local string.
+  local my_string is "".
+  local i is 1.
+  until i > how_many {
+    parameter next_arg.
```
Dunbaratu: This is the key part - note the use of a `parameter` statement in a conditional way - where it executes a number of times that varies.

src/kOS.Safe/Compilation/CompiledObject.cs
```diff
@@ -57,6 +58,7 @@ public static void InitTypeData()
             AddTypeData(5, typeof(float));
             AddTypeData(6, typeof(double));
             AddTypeData(7, typeof(string));
+            AddTypeData(8, typeof(KOSArgMarkerType));
```
Dunbaratu: KSM files need to store arguments of the type KOSArgMarkerType because of all the places where we had code that does:
```
   push $<argstart>
```
which used to be pushing a STRING, are now pushing a new type:
```
    push _ARGMARKER_
```
And that opcode argument needs to be storable in KSM.

src/kOS.Safe/Compilation/CompilerOptions.cs
```diff
@@ -4,6 +4,15 @@ namespace kOS.Safe.Compilation
     public class CompilerOptions
     {
         public bool LoadProgramsInSameAddressSpace { get; set; }
+        /// <summary>
+        /// IsCalledFromRun is true when this compile is being made from
+        /// a context in which it was called from a RUN command such that there
+        /// is a chance of there being arguments passed as parameters.  If it is
+        /// the interpreter context, then this should be false to instruct the
+        /// compiler NOT to attempt to look for an arg bottom marker because there
+        /// won't be one.
```
Dunbaratu: The compiler is used to compile interpreter statements one statement at a time.  In that context, there is no arg bottom marker because the interpreter wasn't called with a RUN command.

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -2141,13 +2160,76 @@ private void VisitInstructionBlock(ParseNode node)
                 PushReturnList();
             
             AddFunctionJumpVars(node);
-            VisitChildNodes(node); // nextBraceIsFunction state would get incorrectly inherited by my children here if it wasn't turned off up above.
```
Dunbaratu: Had to explode this out so I could insert the OpcodeArgBottom at the correct index within the list where the last `DECLARE PARAMETER` statement had happened.

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -1237,7 +1246,17 @@ private void VisitNode(ParseNode node)
         private void VisitStartStatement(ParseNode node)
         {
             AddFunctionJumpVars(null);
-            VisitChildNodes(node);
```
Dunbaratu: Had to explode this out so I could insert the OpcodeArgBottom at the correct index within the list where the last DECLARE PARAMETER statement had happened. The exploded-out version is in the lines that follow below.

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -2141,13 +2160,76 @@ private void VisitInstructionBlock(ParseNode node)
...
+                }
+            }
+                
+            return false; // default if nothing found in the above search.
+        }
+        
+        /// <summary>
+        /// When parsing a function, we need to mark the spot where the lastmost PARAMETER
+        /// statement occurred, so it knows that that's the point during runtime where it
+        /// should assert all the arguments passed in have been consumed by the function.
```
Dunbaratu: Note, when encountering a `PARAMETER` statement inside further nested braces, as in this:
```
function foo {
    parameter a.
    if a = 1 {
        parameter b.
    }
    // more code here
}
```
The place where it should count as the end of the parameters is after it pops back out to the outer level again, meaning here:
```
function foo {
    parameter a.
    if a = 1 {
        parameter b.
    }
    // <---- put argbottom check here
    // more code here
}
```
As opposed to here:
```
function foo {
    parameter a.
    if a = 1 {
        parameter b.
        // <---- put argbottom check here
    }
    // more code here
}
```
That's important so in the case where it skips the if body, it doesn't end up also skipping the arg bottom consume.  This is designed to ensure that the argbottom consumption always occurs at the outer nest level of the function call itself, no matter how deeply nested inside the function the `PARAMETER` statement ended up being.

Because this delays the actual check for arg matching until the bottom of the brace, it is still possible for people to write function bodies that don't check parameter mismatch until too late, like so:
````
function foo {
  if true {
    parameter a,b,c.
    // rest of function body here.
  }
}
```
In that example ,the check is still deferred until getting out of the IF, because it won't insert the arg bottom check until popped back out to the nesting level of the function itself.  This doesn't break - it just devolves into working the same way it did before this pull request in that case - it can't check that you made the args match until the bottom of the function if you do this.

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -1387,7 +1401,7 @@ public void ReverseStackArgs(ICpu cpu)
         {
             List<object> args = new List<object>();
             object arg = cpu.PopValue();
-            while (arg != null && (!(arg.ToString().Equals(ARG_MARKER_STRING))))
+            while (arg == null || arg.GetType() != OpcodeCall.ARG_MARKER_TYPE)
```
Dunbaratu: This is slightly different logic than before.  That's not a mistake - I realized the logic it had before in regards to null protection was a bit off- it prevented the existence of nulls at all in the arguments by treating them as if they were the arg bottom marker when they aren't.  So that change was deliberate - just in case we start supporting nulls in the future we don't want to mistake nulls for argbottoms.

src/kOS.Safe/Execution/KOSArgMarkerType.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// ArgMarkerType literally serves no purpose whatsoever other
+    /// than to just be a mark of where on the stack is the bottom of
+    /// the aruments to something.  If an object is of this type, then
+    /// that means it's the argument botttom marker.
+    /// </summary>
+    public class KOSArgMarkerType
+    {
+        public override string ToString()
+        {
+            return "_KOSArgMarker_";
```
Dunbaratu: Just there so it shows up in stack dumps and is easier to read.

src/kOS/Execution/Stack.cs
```diff
@@ -8,7 +8,7 @@ namespace kOS.Execution
 {
     public class Stack : IStack
     {
-        private const int MAX_STACK_SIZE = 1000;
+        private const int MAX_STACK_SIZE = 3000;
```
Dunbaratu: I noticed it wasn't counting the "above stack" in this before, so when I fixed that it also effectively reduced stack capacity, so I increased it to compensate.

src/kOS/Execution/Stack.cs
```diff
@@ -17,7 +17,7 @@ public void Push(object item)
             ThrowIfInvalid(item);
 
             stackPointer++;
-            if (stackPointer < MAX_STACK_SIZE)
+            if (stack.Count < MAX_STACK_SIZE)
```
Dunbaratu: This was an error all along that was being hidden by another error.  This check was wrong because it only checked if the "below stack" overflowed, and allowed the "above stack" to get as big as you like.  In the case of recursion without arguments, it would let you recurse forever, since the return context stack frames on the above stack didn't count in the check, and they were the only thing accumulating.

This error wasn't noticed before because we had a bug where we were in fact leaving an argmarker on the stack as orphaned garbage, and it was the accumulation of this orphaned garbage that was actually causing the stack overflow condition to be detected, rather than the actual thing it was supposed to be detecting.

Dunbaratu:
Adding tag to #1107 because I forgot the '#" in the original title and that made it fail to auto-associate.

kerboscript_tests/functest31.ks
```diff
@@ -0,0 +1,38 @@
...
+    set my_string to my_string + " " + next_arg.
+    set i to i + 1.
+  }
```
Dunbaratu: Note, given the way the compiler walks the instruction_block body of the function, THIS is the point at which it will insert the opcode to pop the argbottom and assert that it was atop the stack, because it only inserts that check at the outermost level, after the last parameter statement it found.  Because the parameter statement was found inside an `until` loop's braces, the check doesn't happen until those braces end, down here.

This is not a bug.  It's deliberate design, such that if you put a `parameter` statement inside a conditional if block, and the conditional if block doesn't happen, that doesn't cause it to skip over the OpcodeArgBottom.  The OpcodeArgBottom has to occur outside of anything conditional, so it always is guaranteed to happen.

Dunbaratu: OOPS - in writing up the above description, a problem case just occurred to me!

Imagine this code:
```
function foo
{
    parameter p1.
    if p1 = 0 { return. }
    parameter p2.
    // other code here.
}
```

I think that will break, big time.  Because when the ``return`` statement kicks in, it circumvents the OpcodeArgBottom that happens after the ``parameter p2.``, and thus, leaving junk atop the stack, causes stack misalignment.

I'll have to think on the problem a bit longer.

Dunbaratu:
Added the "not ready" because I just thought of a problem case I didn't cover correctly, as described here:

https://github.com/KSP-KOS/KOS/pull/1133/files#r38284523

It's too late tonight for me to start working on this, but I'll come back to it.

Dunbaratu:
This is also addressing issue #1031 - just making the comment here to tag it.

Dunbaratu:
Okay, problem solved with the new commit I just pushed.  Ready to merge now.

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -80,6 +80,7 @@ public enum ByteCode :byte
         STOREEXIST     = 0x5c,
         PUSHDELEGATE   = 0x5d,
         BRANCHTRUE     = 0x5e,
+        ARGBOTTOM      = 0x5f,
```
Dunbaratu: Warning: This will merge-conflict with branch "fixes_1092_defined_keyword".  This is because they both added a new Opcode, and both gave it code number ``0x5f``.

That was because I had to make them as separate independant pull requests from the same starting develop branch, and in each one I added just the one opcode.

@erendrake, when merging them, just give one of them the code of ``0x5f`` while you give the other one code ``0x60`` - it doesn't matter which one you do which with.

kerboscript_tests/functest31.ks
Dunbaratu: The problem mentioned here was fixed by switching from doing a POP to check the argbottom, to doing a PEEK to check the argbottom, and leaving it up to OpcodeReturn later to pop the argbottom just like it did before this PR.  Thus a premature return will still pop the argbottom even if it happens before the actual argbottom check that occurs after the lastmost parameter.

Issue #1090, opened 2015-07-08
Suffixes do not operate on bodies in lists


gisikw:
Many of the functions I'm looking to write would be useful for both celestial bodies and vessels. As such, I'd love to be able to write something like:

```
FUNCTION IS_BODY {
  PARAMETER name.

  LIST BODIES IN allBodies.
  FOR b IN allBodies {
    if b:NAME = name {
      RETURN TRUE.
    }
  }

  RETURN FALSE.
}

// e.g.
IF IS_BODY(foo) {
  RETURN BODY(foo):LONGITUDE.
} ELSE {
  RETURN VESSEL(foo):LONGITUDE.
}
```

Unfortunately, I get an error `CelestialBodies cannot have suffixes`. Referring to a single body (e.g. `BODY('Mun'):NAME`) operates just fine.

gisikw:
If I had to guess, `Body`s when dumped as a list aren't being wrapped as `Orbitable`s?

hvacengi:
From buildlist.cs
```
switch (listType)
            {
                case "bodies":
                    list = ListValue.CreateList(FlightGlobals.fetch.bodies);
                    break;
...
```
Yep, it returns the underlying ksp body object.  We'll have to get that fixed in the next update.

gisikw:
Cheers :-) It looks like for my case, I can invert the logic and `LIST TARGETS` instead (which do support suffixes), but I'll look forward to the update!

PR #1134, opened 2015-08-26
Fixes 1090 by listing bodies as BodyTargets.

Dunbaratu:
This one was simple.

### DIFF
--- a/src/kOS/Function/BuildList.cs
+++ b/src/kOS/Function/BuildList.cs
@@ -18,7 +18,10 @@ namespace kOS.Function
             switch (listType)
             {
                 case "bodies":
-                    list = ListValue.CreateList(FlightGlobals.fetch.bodies);
+                    foreach (CelestialBody cBody in FlightGlobals.fetch.bodies)
+                    {                        
+                        list.Add(new BodyTarget(cBody, shared));
+                    }
                     break;
                 case "targets":
                     foreach (var vessel in FlightGlobals.Vessels)


Dunbaratu:
Saying #1090 so this gets a link to the issue.

hvacengi:
Is there a reason this was done previously using the `ListValue.CreateList` method?  It looks like there is an underpinning to allow you to cast the list items to another type, but I don't know if we have enabled the necessary casting on the `BodyTarget` class.  I prefer your solution, but wanted to make sure we weren't breaking something else by moving away from that convention.

Dunbaratu:
CreatList just turns any generic collection into our ListValue collection, but as far as I know it doesn't alter the individual items within the list, which is what's needed here.

hvacengi:
From ListValue.cs:
```
        public static ListValue<T> CreateList<TU>(IEnumerable<TU> list)
        {
            return new ListValue<T>(list.Cast<T>());
        }
```
If I read the generics right there, you could in theory call `ListValue<BodyTarget>.CreateList<Body>(FlightGlobals.fetch.bodies)` But I haven't tried it.  At for what we're doing, I think it just adds a layer of complication for someone who comes back to look at it later.

I say leave it the way you did it.  Just trying to understand the underpinnings myself having come it late.

Dunbaratu:
There exists a constructor for BodyTarget that goes like so:

BodyTarget(CelestialBody cb, SharedOjbects so);

I don't think we have implemented an automatic casing conversion that will create a BodyTarget from JUST a CelestialBody without being given that SharedObjects information, so the cast you're talking about doesn't seem to me like it would do the job.

PR #1135, opened 2015-08-26
Fixes #1092 by implementing a DEFINED keyword.

### LINKED ISSUE #1092: Implement an IDENT_FOUND() or equivalently named language feature.
(This was discussed briefly before in passing here and there as part of other issues but I really am getting tired of this feature not existing in the language so I'll be implementing it myself.  I'll assign it to me so I don't forget.)

Synopsis:
```
    print "A: Does myvar Exist? " + IDENT_FOUND(myvar).
    set myvar to 1.
    print "B: Does myvar Exist? " + IDENT_FOUND(myvar).
```
prints
```
A: Does myvar Exist? False
B: Does myvar Exist? True
```
### scoping.

Note that it would check for any identifier that is IN SCOPE at the moment the call is made.  It would treat identifiers that are out of scope as being identical to as if they had never existed (it would return false for them).  It's a check for "can I refer to this identifier right now without crashing the script when I do so?".

### Why not call it just EXISTS or DECLARED?

Because that may confuse new programmers who think we're talking about objects in the gameworld, as in "does this vessel exist?", when we're actually talking about the program's identifiers themselves.  Basically this differentiates it from something more like an ISNULL check, which we don't have and is still a bit controversial to implement.

### Why not call it VAREXIST?  Why use the word IDENT?

Because what I'm picturing would also work just as well for checking if a function with a given name has been declared.

### Any alternate names?

I'm not entirely happy with the name IDENT_EXISTS, and would entertain suggestions for alternate names for it, but keep the above two points in mind when coming up with alternate names.

Dunbaratu:
Went with a syntax like so:

    if defined foo { blag. }

rather than:

    if isdefined(foo) { blag. }

becuase it's messy to make it work with a function call when foo is an
udefined identifier that's going to bomb out with error if you try to
eval it first to push it onto the stack for a function argument.


cgruber:
+1.  Please.  This is pretty crucial functionality to breaking code into libraries that support configurable behavior without forcing up-front declaration of all of their symbols.

PR #1138, opened 2015-08-28
Fixes #1137 - KSM can now store label-less jumps.

### LINKED ISSUE #1137: Short-circuit boolean logic broken ONLY in KSM files, not KS files.
Here's the reddit thread about it: 
https://www.reddit.com/r/Kos/comments/3im7tm/compiled_script_does_not_seem_to_execute/

Basically, these two examples run fine in uncompiled KS files, but hang forever when run from their compiled KSM form:

```
if true and false {
  print "this should never print".
}
print "done".
```
```
if false or true {
  print "this should always print".
}
print "done".
```

The key problem is that this only triggers when the righthand side of the 'and' or the 'or' **cannot** be short-circuited away and must be tested to determine the expression, and it never happens in KS files, only KSM files.

If the righthand side can be short-circuited and ignored, then the bug does not appear in either KS or KSM (for example ``if true or false`` has no problem, as it always short-circuits, whereas ``if false or true`` never short-circuits, and does have the problem.)

Dunbaratu:
I started to fix it by removing the relative label-less jump
from the compiler, but changed my mind.  If the label-less
jump is a thing you can do in the virtual machine, then KSM files need
to be able to store it correctly.  So instead I worked on changing
it so the KSM file can in fact store the case where a jump or
branch doesn't bother using the label and instead uses just a
relative jump distance only.

Note, if you populate BOTH the label AND the distance, they'd
better agree!! If they don't it's probably wrong in the first place
even before KSM gets involved.  In any case, if you DO populate
both, the KSM file will take the label as official and forget
the relative distance (re-populating it from the label later
when loading, like it does for all the other labeled jumps).

### DIFF
--- a/src/kOS.Safe/Compilation/Opcode.cs
+++ b/src/kOS.Safe/Compilation/Opcode.cs
@@ -730,10 +730,20 @@ namespace kOS.Safe.Compilation
     
     public abstract class BranchOpcode : Opcode
     {
-        // This is identical to the base DestinationLabel, except that it has
-        // the MLFIeld attached to it:
+        // This stores EITHER the label OR the relative distance,
+        // depending, in the KSM packed file.  Only if the label is
+        // an empty string does it store the integer distance instead.
         [MLField(1,true)]
-        public override string DestinationLabel {get;set;}
+        public object KSMLabelOrDistance
+        {
+            get
+            {
+                if (DestinationLabel == String.Empty)
+                    return Distance;
+                else
+                    return DestinationLabel;
+            }
+        }
         
         public int Distance { get; set; }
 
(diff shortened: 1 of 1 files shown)


src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -742,12 +752,19 @@ public override void PopulateFromMLFields(List<object> fields)
...
+            else
+                Distance = (int)fields[0];
         }
 
         public override string ToString()
         {
-            return string.Format("{0} {1}", Name, Distance);
+            // Format string forces printing of '+/-' sign always, even for positive numbers.
+            // The intent is to make it more clear that this is a relative, not absolute jump:
+            return string.Format("{0} {1:+#;-#;+0}", Name, Distance);
```
Dunbaratu: This change was not strictly needed for the fix, but as long as my fingers were in this code, I changed it because I think this communicates the "relativeness" of the jump better to the user.

PR #1142, opened 2015-08-30
Fixes #1141 allows wheelsteering by integer.

### LINKED ISSUE #1141: Wheelsteering slightly broken, does not process heading instructions correctly.
lock wheelsteering to 180.  Does not work.
lock wheelsteering to 180.1.  Does work.

Dunbaratu:
Also fixed the "wraparound at north" bug when
doing LOCK WHEELSTEERING TO 0. (i.e. it would be
1 degree off to the left, and behave as if it was
359 degrees off to the right, and perform a full
loop all the way around, because it doesn't
realize a bearing of -359 really means +1.)

Also added a few more robust conditions for some of the
other steeering checks, although they are unlikely
to trigger.

### DIFF
--- a/src/kOS.Safe/Exceptions/KOSWrongControlValueType.cs
+++ b/src/kOS.Safe/Exceptions/KOSWrongControlValueType.cs
@@ -1,0 +1,36 @@
+﻿using System;
+
+namespace kOS.Safe.Exceptions
+{
+    /// <summary>
+    /// To be thrown whenever the script attempted to perform a
+    /// cooked control "LOCK", or a raw control "SET", to a value<br/>
+    /// that's of an unusable type for that particular control bound variable.
+    /// <br/>
+    /// Examples:<br/>
+    /// - attempting to LOCK THROTTLE TO a string, or a LIST.<br/>
+    /// - attempting to SET SHIP:CONTROL:PITCH to a string, or a RGBA color.<br/>
+    /// </summary>
+    public class KOSWrongControlValueType : KOSException
+    {
+        public KOSWrongControlValueType(string controlType, string didUse, string shouldUse) :
+            base( string.Format("Cannot use a {0} as the value for the {1}.  Should use {2} instead.",
+                                didUse, controlType, shouldUse))
+        {
+        }
+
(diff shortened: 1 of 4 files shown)


src/kOS.Safe/Exceptions/KOSWrongControlValueType.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// <summary>
+    /// To be thrown whenever the script attempted to perform a
+    /// cooked control "LOCK", or a raw control "SET", to a value<br/>
+    /// that's of an unusable type for that particular control bound variable.
+    /// <br/>
+    /// Examples:<br/>
+    /// - attempting to LOCK THROTTLE TO a string, or a LIST.<br/>
+    /// - attempting to SET SHIP:CONTROL:PITCH to a string, or a RGBA color.<br/>
+    /// </summary>
+    public class KOSWrongControlValueType : KOSException
```
erendrake: Name should be `KOSWrongControlValueTypeException`

src/kOS.Safe/Exceptions/KOSWrongControlValueType.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        public KOSWrongControlValueType(string controlType, string didUse, string shouldUse) :
+            base( string.Format("Cannot use a {0} as the value for the {1}.  Should use {2} instead.",
+                                didUse, controlType, shouldUse))
+        {
+        }
+
+        public virtual string VerboseMessage
```
erendrake: `virtual` should be `override`

src/kOS.Safe/Exceptions/KOSWrongControlValueType.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            get { return base.Message + "\n" +
+                         "When setting a special control variable\n" +
+                         "the type of the value has to match what\n" +
+                         "the kOS computer expects that value to be."; }
+        }
+
+        public virtual string HelpURL
```
erendrake: `virtual` should be `override`

PR #1146, opened 2015-08-31
fixes #1106 implemented binding synomyms

erendrake:


Dunbaratu:
I hope you don't mind - when I merge this I'm also going to look for all the other places we have aliased suffixes and move them to the new system too.

erendrake:
By all means lets do it :)

PR #1152, opened 2015-09-06
fixes #1145

### LINKED ISSUE #1145: Vessel:SENSORS uses a set of fixed values that become wrong as time passes.
If you do this:

```
print vessel:sensors:acc.
wait 1.
print vessel:sensors:acc.
wait 1.
print vessel:sensors:acc.
```
You will get slightly different results as the vessel situation changes.

But if you do this instead:

```
set mysensors to vessel:sensors.
print mysensors:acc.
wait 1.
print mysensors:acc.
wait 1.
print mysensors:acc.
```
Then you'll end up with 3 duplicates of the exact same thing printed, because the constructor for VesselSensors creates a frozen snapshot in time of the values as they were at that moment.

I think that to any user reading the docs it would look like vessel:sensors is just a handle giving you access to the vessel's LIVE sensors, not a handle on a snapshot in time of the sensors.

hIGH-aND-mIGHTY:
Been watching this mod and wanted to contribute. Let me know what you think. #1145

### DIFF
--- a/src/kOS/Suffixed/VesselSensors.cs
+++ b/src/kOS/Suffixed/VesselSensors.cs
@@ -1,34 +1,32 @@
 using kOS.Safe.Encapsulation;
 using kOS.Safe.Encapsulation.Suffixes;
+using kOS.Safe.Exceptions;
 
 namespace kOS.Suffixed
 {
     public class VesselSensors : Structure
     {
-        private Vector acceleration = new Vector(0, 0, 0);
-        private Vector geeForce = new Vector(0, 0, 0);
-        private double kerbolExposure;
-        private double temperature;
-        private double pressure;
+        private Vessel vessel;
+
 
         public VesselSensors(Vessel target)
         {
-            FindSensors(target);
+            vessel = target;
             InitializeSuffixes();
         }
 
         private void InitializeSuffixes()
         {
-            AddSuffix("ACC", new Suffix<Vector>(() => acceleration));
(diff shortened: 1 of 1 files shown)


src/kOS/Suffixed/VesselSensors.cs
```diff
@@ -36,29 +34,65 @@ private void FindSensors(Vessel target)
...
-                    foreach (var c in part.FindModulesImplementing<ModuleDeployableSolarPanel>())
+                }
+            }
+            throw new KOSException("Cannot find sensor for " + sensorType);
+        }
+
+        private double GetSensorDoubleValue(string sensorType)
+        {
+            foreach (var part in vessel.Parts)
+            {
```
Dunbaratu: This oarts walk is going to be a bit expensive if someone is trying to query the value over and over in a loop to see how it changes each physics tick.

Could you do the parts walk during the constructor, and save the set of PartModules of each type then?  That would be similar to what it did before, where it stored the values in private variables, but now it's just storing the PartModules rather than the actual values, deferring the query of the partmodules for later when you request a value.

That way you only pay the expense of this parts walk once, during the constructor, and after that you already have the relevant PartModules to query each time.

It does raise an issue with having stale parts that become wrong after a decouple or docking operation changes the parts list, but you can guard against that by checking if the PartModules's Part's Vessel is equal to the vessel you remembered in the constructor.  If it's not, then that means the PartModule is now on a different vessel, and that means you need to re-do the parts walk to re-obtain the PartModule list again.

PR #1153, opened 2015-09-06
Fixes 1151 unlock in trigger

Dunbaratu:
See full comments in the first commit's commit log.

### DIFF
--- a/src/kOS.Safe/Compilation/KS/Compiler.cs
+++ b/src/kOS.Safe/Compilation/KS/Compiler.cs
@@ -522,13 +522,11 @@ namespace kOS.Safe.Compilation.KS
             if (IsLockStatement(node))
             {
                 funcIdentifier = lastSubNode.Nodes[1].Token.Text;
-                Console.WriteLine("eraseme: Identified lock with identifier " + funcIdentifier);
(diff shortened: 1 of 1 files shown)


src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -2263,37 +2263,34 @@ private void VisitUnlockStatement(ParseNode node)
 
         private void UnlockIdentifier(UserFunction lockObject)
         {
-            if (lockObject.IsInitialized())
+            if (lockObject.IsSystemLock())
             {
```
Dunbaratu: Because github's differ doesn't ignore whitespace-only diffs, this section looks like it was massively changed, but if you look at it in detail, it's really not.

All I really did was delete this:
```
If (lockObject.IsInitialized())
{
```
and the associated closing line:
```
}
```
The fact that this meant out-denting the body one level is what's creating all these bogus 'changes' in the diff highlighting here.

For example, the above line pretends I *replaced* the IsInitialized() check with the IsSystemLock() check, when actually the IsSystemLock check was there all along, nested one level deep, and I was just out-denting it, making it falsely assume it was a replacement of the original if check.

Issue #270, opened 2014-10-13
String operations (suggest making a kOS string wrapping structure object)


Dunbaratu:
Lots of people have asked for string manipulator operations.  My suggested solution is to make a new subclass of kos.Safe.Encapsulation.Structure, which contains inside it one single C# string and NO other data (having no other data is important for a reason I'll get to below).  It then wraps all the string manipulator methods we feel like exposing to the users, and might have a lot of very short code routines like so:

```
    protected string data;

    public string ToString() { return data; }
   
    public string ToUpper() { return data.ToUpper();}
    public string ToLower() { return data.ToLower();}
    public bool StartsWith() { return data.StartsWith(); }
    // etc...

    // And then these would be wrapped with AddSuffix calls to set up
    // TOUPPER, TOLOWER, STARTSWITH, and so on for suffixes.
```
This sounds very simple and straightforward. (I'd rather just subclass C#'s string, but it's sealed so you can't.)  But it does have a few consequences elsewhere through the code - one example is that it means now the ML code will need to know to peek inside the wrapper class to get the actual string to store in the ML file (which only stores primitives, not structures, which is the reason it's important that this wrapper class only contain the inner string data and nothing else - so it can be recreated from a vanilla string stored in the ML file).  Another example is that the compiler needs to know to create an object of this new type when it sees a quoted string in kOS script.

It may help to also fix issue #253, as it would present a way for the system to distinguish user strings and internally made identifier strings that doesn't require using a leading "$" character.  User strings would all be inside this wrapper and identifier strings wouldn't.

abenkovskii:
>
Embedding chars inside quotes:
>
// ---------- 3 different ideas for how to do it: -----------
>
//embedded, with backslash or other commonly used encoding method:
SET newString to "This string has \"quoted text\" in it.".
SET newString to "This string has &#42;quoted text&#42; in it.".
>
//provide some constant built-in variables for some special chars like DQUOTE:
SET newString to "This string has " + DQUOTE + "quoted text" + DQUOTE + "in it.".
>
//provide function to convert number code to character:
SET newString to "This string has " + CHR(42) + "quoted text" + CHR(42) + "in it.".

I prefer escape sequences (\n, \\, \" etc):
1. There is no need to remember codes of characters. (magic numbers are the least mnemonic way to represent something)
2. Escape sequences are commonly used. Implementing \n, \\ \" etc will make transition to and from kerboscript easier.

abenkovskii:
[...]

PR #1159, opened 2015-09-16
String manipulation suffixes (addresses #270)

errantsignal:
I added suffixes for basic string manipulation. As described in #270, I added a subclass of Structure to wrap C# strings which defines suffixes for most of the basic string operations in C#. I stuck with the C# function names in general. The wrapper class also adds indexing for individual character access.

The operations are as their C# equivalents for the most part, hence they are case sensitive, strings are immutable, etc. Of course, case insensitivity is easily obtained using ToUpper/ToLower.

The actual wrapping of the string is only done when suffixes are actually resolved; it doesn't keep the wrapped strings around permanently. I decided to do the wrapping here because it required the
simplest change and wouldn't break anything. Certainly may be a better place for it, or maybe there is some advantage to having the wrapped versions hang around on the stack. 

I mostly just needed string manipulation for what I was doing; if you'd rather do it a different way, feel free to ignore my PR, or let me know what you think.

Example code (with everything except Split):
```
                                                                // CORRECT OUTPUTS
SET s TO "Hello, Strings!".                                     // ---------------
PRINT "Original String:               " + s.                    // Hello, Strings!
PRINT "string[7]:                     " + s[7].                 // S
PRINT "LENGTH:                        " + s:LENGTH.             // 15
PRINT "SUBSTRING(7, 6):               " + s:SUBSTRING(7, 6).    // String
PRINT "CONTAINS(''ring''):            " + s:CONTAINS("ring").   // True
PRINT "CONTAINS(''bling''):           " + s:CONTAINS("bling").  // False
PRINT "ENDSWITH(''ings!''):           " + s:ENDSWITH("ings!").  // True
PRINT "ENDSWITH(''outs!''):           " + s:ENDSWITH("outs").   // False
PRINT "FIND(''l''):                   " + s:FIND("l").          // 2
PRINT "FINDLAST(''l''):               " + s:FINDLAST("l").      // 3
PRINT "FINDAT(''l'', 0):              " + s:FINDAT("l", 0).     // 2
PRINT "FINDAT(''l'', 3):              " + s:FINDAT("l", 3).     // 3
PRINT "FINDLASTAT(''l'', 9):          " + s:FINDLASTAT("l", 9). // 3
PRINT "FINDLASTAT(''l'', 2):          " + s:FINDLASTAT("l", 2). // 2
PRINT "INSERT(7, ''Big ''):           " + s:INSERT(7, "Big ").  // Hello, Big Strings!

PRINT " ".
PRINT "                               |------ 18 ------|".
PRINT "PADLEFT(18):                   " + s:PADLEFT(18).        //    Hello, Strings!
PRINT "PADRIGHT(18):                  " + s:PADRIGHT(18).       // Hello, Strings!   
PRINT " ".

PRINT "REMOVE(1, 3):                  " + s:REMOVE(1, 3).               // Ho, Strings!
PRINT "REPLACE(''Hell'', ''Heaven''): " + s:REPLACE("Hell", "Heaven").  // Heaveno, Strings!
PRINT "STARTSWITH(''Hell''):          " + s:STARTSWITH("Hell").         // True
PRINT "STARTSWITH(''Heaven''):        " + s:STARTSWITH("Heaven").       // False
PRINT "TOUPPER:                       " + s:TOUPPER().                  // HELLO, STRINGS!
PRINT "TOLOWER:                       " + s:TOLOWER().                  // hello, strings!

PRINT " ".
PRINT "''  Hello!  '':TRIM():         " + "  Hello!  ":TRIM().          // Hello!
PRINT "''  Hello!  '':TRIMSTART():    " + "  Hello!  ":TRIMSTART().     // Hello!  
PRINT "''  Hello!  '':TRIMEND():      " + "  Hello!  ":TRIMEND().       //   Hello!

PRINT " ".
PRINT "Chained: " + "Hello!":SUBSTRING(0, 4):TOUPPER():REPLACE("ELL", "ELEPHANT").  // HELEPHANT
```
The following operations were added:
[Character Indexing]
CONTAINS
ENDSWITH
FIND/INDEXOF
FINDAT
FINDLAST/LASTINDEXOF
FINDLASTAT
INSERT
LENGTH
PADLEFT
PADRIGHT
REMOVE
REPLACE
SPLIT (Returning a KOS List)
STARTSWITH
SUBSTRING
TOLOWER
TOUPPER
TRIM
TRIMEND
TRIMSTART

No attempt was made to tackle escape characters in strings, which was also briefly mentioned in #270.

### DIFF
--- a/doc/source/structures/misc/string.rst
+++ b/doc/source/structures/misc/string.rst
@@ -1,0 +1,311 @@
+.. _string:
+
+String
+====
+
+A :struct:`String` is an immutable sequence of characters in kOS.
+
+Creating strings
+-------------------
+
+Unlike other structures, strings are created with a special syntax::
+
+    // Create a new string
+    SET s TO "Hello, Strings!".
+    
+
+Strings are immutable. This means, once a string has been created, it 
+can not be directly modified. However, new strings can be created out
+of existing strings. For example::
+
+    // Create a new string with "Hello" replaced with "Goodbye"
+    SET s TO "Hello, Strings!".
+    SET t TO s:REPLACE("Hello", "Goodbye").
+
+Structure
+---------
+
+.. structure:: String
+
+    .. list-table:: Members
+        :header-rows: 1
+        :widths: 2 1 4
+
+        * - Suffix
+          - Type
+          - Description
+
+        * - :meth:`CONTAINS(string)`
+          - boolean
+          - True if the given string is contained within this string (case sensitive) 
+        * - :meth:`ENDSWITH(string)`
+          - boolean
+          - True if this string ends with the given string (case sensitive)
+        * - :meth:`FIND(string)`
+          - integer
+          - Returns the index of the first occurrence of the given string in this string (starting from 0)
+        * - :meth:`FINDAT(string, startAt)`
+          - integer
+          - Returns the index of the first occurrence of the given string in this string (starting from startAt)
+        * - :meth:`FINDLAST(string)`
+          - integer
+          - Returns the index of the last occurrence of the given string in this string (starting from 0)
+        * - :meth:`FINDLASTAT(string, startAt)`
+          - integer
+          - Returns the index of the last occurrence of the given string in this string (starting from startAt)
+        * - :meth:`INDEXOF(string)`
+          - integer
+          - Alias for FIND(string)
+        * - :meth:`INSERT(index, string)`
+          - :struct:`String`
+          - Returns a new string with the given string inserted at the given index into this string
+        * - :meth:`LASTINDEXOF(string)`
+          - integer
+          - Alias for FINDLAST(string)
+        * - :attr:`LENGTH`
+          - integer
+          - Number of characters in the string
+        * - :meth:`PADLEFT(width)`
+          - :struct:`String`
+          - Returns a new right-aligned version of this string padded to the given width by spaces
+        * - :meth:`PADRIGHT(width)`
+          - :struct:`String`
+          - Returns a new left-aligned version of this string padded to the given width by spaces
+        * - :meth:`REMOVE(index,count)`
+          - :struct:`String`
+          - Returns a new string out of this string with the given count of characters removed starting at the given index
+        * - :meth:`REPLACE(oldString, newString)`
+          - :struct:`String`
+          - Returns a new string out of this string with any occurrences of oldString replaced with newString
+        * - :meth:`SPLIT(separator)`
+          - :struct:`String`
+          - Breaks this string up into a list of smaller strings on each occurrence of the given separator
+        * - :meth:`STARTSWITH(string)`
+          - boolean
+          - True if this string starts with the given string (case sensitive)
+        * - :meth:`SUBSTRING(start, count)`
+          - :struct:`String`
+          - Returns a new string with the given count of characters from this string starting from the given start position
+        * - :attr:`TOLOWER`
+          - :struct:`String`
+          - Returns a new string with all characters in this string replaced with their lower case versions
+        * - :attr:`TOUPPER`
+          - :struct:`String`
+          - Returns a new string with all characters in this string replaced with their upper case versions
+        * - :attr:`TRIM`
+          - :struct:`String`
+          - returns a new string with no leading or trailing whitespace
+        * - :attr:`TRIMEND`
+          - :struct:`String`
+          - returns a new string with no trailing whitespace
+        * - :attr:`TRIMSTART`
+          - :struct:`String`
+          - returns a new string with no leading whitespace
+
+
+.. method:: String:CONTAINS(string)
+
+    :parameter string: :struct:`String` to look for
+    :type: boolean
+    
+    True if the given string is contained within this string. This is a case-sensitive search.
+
+.. method:: String:ENDSWITH(string)
+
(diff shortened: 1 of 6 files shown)


erendrake:
Thank you for your pull request! We will review it soon

Dunbaratu:
> Certainly may be a better place for it, or maybe there is some advantage to having the wrapped versions hang around on the stack.

Actually that was the entire point of the wrapper.  To make it so that there are no such things as having a kerboscript string that isn't wrapped.  The whole point is to always have an abstraction between native types and script types so we can make our own type-aware things.  (as part of this, we would also make a wrapper around int, bool, double, etc).

errantsignal:
Yeah, I've just done it this way because it was the simplest way to get this up and running without much knowledge of the code - it's less than 20 lines outside of the string operations themselves, and I only needed to modify one class. If you have grander dreams than I, I wont be offended if you ignore my PR and do it yourselves (I just needed the string operations for my own scripts).

Alternatively, if you're happy with the operations themselves, it's easy enough to just change where the strings are getting wrapped now or at some later point if having all data types wrapped on the stack is necessary for whatever else you're doing. It should be transparent to the users. (I don't really have the familiarity with the code to know what all would break if I tried implementing it.)

Dunbaratu:
No problem.  I'll definitely be taking a closer look at the code in the PR when I can.  I've just been a bit sick the last few days and only able to participate marginally - definitely not up to a task like digging through someone else's code and deciding how to integrate it.

Dunbaratu:
I had a read through it.  It looks good.  I think I'd add the behavior that strings act as indexables, so you can use the '[]' operator to get one character out of the middle of the string, which also would allow you to do a "FOR ch in str {" loop.

Dunbaratu:
If I get this PR in, can we have it in pre1.1, @erendrake ?

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -656,6 +672,12 @@ public override void Execute(ICpu cpu)
             }
             object list = cpu.PopValue();
 
+            // Box strings if necessary to allow them to be indexed
+            if (list is String)
+            {
+                list = new StringValue((String)list);
+            }
+
```
Dunbaratu: I think this sort of conversion needs to be in all things pushed/popped to the stack, rather than here in Get/SetIndex where it will miss a lot of the cases.

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -679,7 +701,9 @@ public override void Execute(ICpu cpu)
             {
                 index = Convert.ToInt32(index);  // allow expressions like (1.0) to be indexes
             }
-            if (!(list is IIndexable)) throw new Exception(string.Format("Can't iterate on an object of type {0}", list.GetType()));
+
+            // Adjusted error message to reflect that there are now read-only indexable objects
+            if (!(list is IIndexable)) throw new Exception(string.Format("Can't set indexed elements on an object of type {0}", list.GetType()));
```
Dunbaratu: I think this interferes with the new associtive array, which does allow string indeces.

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -577,9 +577,17 @@ public override void Execute(ICpu cpu)
             object popValue = cpu.PopValue();
 
             var specialValue = popValue as ISuffixed;
+            
             if (specialValue == null)
             {
-                throw new Exception(string.Format("Values of type {0} cannot have suffixes", popValue.GetType()));
+                if (popValue is String)
+                {
+                    specialValue = new StringValue((String)popValue);
```
Dunbaratu: This works as a temporary for now, and is pretty clever now that I understand what you're doing (this is invoked every time you attempt to use the colon operator).  I think as a future refactor it should be changed to actually being on the stack as a StringValue, for other reasons, but doing so takes a bit of work (i.e. it has to be stored in KSM files that way, it has to be wrapped everywhere in the code we're using it. etc).  The reason for my wanting it is to help make kOS able to do types properly (i.e. have an isType(foo) operator), where all types in the system are made by us.  But that's a larger project, and in order to just get string operations working at all, I think your solution is a good 'shim' fix for now.

doc/source/structures/misc/string.rst
```diff
@@ -0,0 +1,311 @@
...
+    SET s TO "Hello, Strings!".
+    
+
+Strings are immutable. This means, once a string has been created, it 
+can not be directly modified. However, new strings can be created out
+of existing strings. For example::
+
+    // Create a new string with "Hello" replaced with "Goodbye"
+    SET s TO "Hello, Strings!".
+    SET t TO s:REPLACE("Hello", "Goodbye").
```
Dunbaratu: I think I'll also add an example of what *doesn't* work in addition to this example of what *does*, to help those unfamiliar with languages that use immutable strings understand how it works better.  (i.e. maybe show that `set s[0] to "A"` is the sort of thing we're talking about not working when we say "immutable".)

src/kOS.Safe/Compilation/Opcode.cs
erendrake: i will say that it did require a little work to make it work :)

PR #1160, opened 2015-09-16
Fix SASMODE radial orientation

### LINKED ISSUE #1130: SASMODE mixed up
Using the field SASMODE to use a probe core for attitude control has a bug. Setting SASMODE to "radialout" will activate the mode "Anti-Radial", which is the one pointing inwards. Analogous setting SASMODE to "radialin" will activate the mode "Radial" which is the one point outwards.

How to reproduce:
1. put vessel with probe core capable of radial attitude control and kOS computer in orbit
2. open console and enter:
  i. set sas to true.
  ii. set sasmode to "radialin".
3. vessel will orientate along the marker called "Radial", which is away from the body

observed on Windows 7 64-bit with KSP 1.0.4 32-bit and kOS 0.17.3
additional mod: Remote Tech (had connection to KSC)

hvacengi:
Fixes #1130
For some reason KSP seems to be reversing the naming convention for radial in and radial out internal to their own code.  When we set the enum value to RadialOut it points radialy in, and vice versa.  So I swaped the enums for RadialOut and RadialIn and added a TODO note to check it in the future.

### DIFF
--- a/src/kOS/Binding/FlightControlManager.cs
+++ b/src/kOS/Binding/FlightControlManager.cs
@@ -237,10 +237,11 @@ namespace kOS.Binding
                         SelectAutopilotMode(VesselAutopilot.AutopilotMode.Antinormal);
                         break;
                     case "radialin":
-                        SelectAutopilotMode(VesselAutopilot.AutopilotMode.RadialIn);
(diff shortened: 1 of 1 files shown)


erendrake:
That is so strange that they did this.

PR #1161, opened 2015-09-16
Issue 1099 dockingport spelling

### LINKED ISSUE #1099: dockingport:aquire[force,range,torque] is a misspelling of "Acquire"
Confirmed misspelled in both game and docs

hvacengi:
Fixes #1099
* Corrects the spelling of acquire in the suffixes and throws a deprecation exception if called with the old spelling.
* Added suffixes of `NODEPOSITION` and `NODETYPE` to make docking scripts a little easier and more accurate.
* Also adds a `HASMODULE` suffix to PartValue, so to easily determine if the part has a specified module.


src/kOS/Suffixed/Part/PartValue.cs
```diff
@@ -13,7 +15,7 @@ namespace kOS.Suffixed.Part
 {
     public class PartValue : Structure, IKOSTargetable
     {
-        private readonly SharedObjects shared;
+        protected readonly SharedObjects shared;
```
erendrake: no non-private fields.

hvacengi: Access from an inherited class is a little bit of a hybrid issue.  I agree in concept with not publicly exposing the member variable itself, and I don't think we want this variable publicly exposed anyways.  Protected fields are only accessible to inherited classes, which makes it helpful to not hold a 2nd handle to the object.  I think I've always treated protected fields similarly to private.  Would you prefer that I implement a protected property, or just add the 2nd handle to the DockingPortValue class?

erendrake: protected property with private setter would be fine.

hvacengi:
Assuming no further review questions pop up, this should be ready.

src/kOS/Suffixed/Part/DockingPortValue.cs
```diff
@@ -18,9 +18,12 @@ public DockingPortValue(ModuleDockingNode module, SharedObjects sharedObj) : bas
 
         private void DockingInitializeSuffixes()
         {
-            AddSuffix("AQUIRERANGE", new Suffix<float>(() => module.acquireRange));
-            AddSuffix("AQUIREFORCE", new Suffix<float>(() => module.acquireForce));
-            AddSuffix("AQUIRETORQUE", new Suffix<float>(() => module.acquireTorque));
+            AddSuffix("AQUIRERANGE", new Suffix<float>(() => { throw new kOS.Safe.Exceptions.KOSDeprecationException("18.0", "AQUIRERANGE", "ACQUIRERANGE", string.Empty); }));
+            AddSuffix("AQUIREFORCE", new Suffix<float>(() => { throw new kOS.Safe.Exceptions.KOSDeprecationException("18.0", "AQUIREFORCE", "ACQUIREFORCE", string.Empty); }));
+            AddSuffix("AQUIRETORQUE", new Suffix<float>(() => { throw new kOS.Safe.Exceptions.KOSDeprecationException("18.0", "AQUIRETORQUE", "ACQUIRETORQUE", string.Empty); }));
```
Dunbaratu: These should be "0.18.0" not "18.0".

src/kOS/Suffixed/Part/PartValue.cs
```diff
@@ -61,7 +64,7 @@ private PartModuleFields GetModule(string modName)
             foreach (PartModule mod in Part.Modules)
             {
                 SafeHouse.Logger.Log(string.Format("Does \"{0}\" == \"{1}\"?", mod.moduleName.ToUpper(), modName.ToUpper()));
```
Dunbaratu: We should remove these logs.  They were there from when I was debugging this.  It's probably the main reason GetModule is slow, since using Unity's logger is a very expensive operation (it basically re-opens the file on each operation, then closes it again, rather than leaving the file open through the life of the program.  This is done so that logged messages exist immediately in the file with no buffering, so that in a crash you have an accurate picture of how far the program got).

src/kOS/Suffixed/Part/PartValue.cs
```diff
@@ -61,7 +64,7 @@ private PartModuleFields GetModule(string modName)
...
-                if (String.Equals(mod.moduleName, modName, StringComparison.CurrentCultureIgnoreCase))
+                if (String.Equals(mod.moduleName, modName, StringComparison.OrdinalIgnoreCase))
                 {
                     SafeHouse.Logger.Log("yes it does");
```
Dunbaratu: see above comment.   We should remove this.

src/kOS/Suffixed/Part/PartValue.cs
```diff
@@ -61,7 +64,7 @@ private PartModuleFields GetModule(string modName)
             foreach (PartModule mod in Part.Modules)
             {
                 SafeHouse.Logger.Log(string.Format("Does \"{0}\" == \"{1}\"?", mod.moduleName.ToUpper(), modName.ToUpper()));
-                if (String.Equals(mod.moduleName, modName, StringComparison.CurrentCultureIgnoreCase))
+                if (String.Equals(mod.moduleName, modName, StringComparison.OrdinalIgnoreCase))
```
Dunbaratu: This should be a refactor everywhere, or nowhere.  This isn't the only place to use CurrentCultureIgnoreCase, and if we switch we should switch consistently rather than having code that half does it one way and half does it the other.

src/kOS/Suffixed/Part/PartValue.cs
hvacengi: Agreed.  I have a separate issue started for me to do a full audit of all calls to the logger.  But we can remove theses references as we go instead.

src/kOS/Suffixed/Part/PartValue.cs
hvacengi: Since we don't use have anything that's culture dependent,OrdinalIgnoreCase will be faster for the comparisons.  I've made an effort to make the change whenever I come across an instance, but we can certainly do a full refactor to catch all of them.

src/kOS/Suffixed/Part/DockingPortValue.cs
hvacengi: That's odd for me to make that kind of mistake.  Do you want to make the change in you're review commit, or do you need me to make the change?

Dunbaratu: I won't be in a position to edit code until tomorrow morning.  I can do it then or you can do it before if you like.

hvacengi:
@Dunbaratu should be good to go here.

PR #1170, opened 2015-09-26
Use the new RemoteTech SetAntennaTarget API

tomekpiotrowski:
Takes advantage of the new SetAntennaTarget API added in RemoteTechnologiesGroup/RemoteTech#493.

Uses the new API to set antenna target (instead of the old, hacky way). It now supports custom ground stations (previously only the default one was supported).

There is a one breaking change: `mission-control` is now not a valid value for the `target` field. The full name of a ground station must be used, in this case `Mission Control`.

The RT API is scheduled to be released with RT 1.7 (just as the other RT PR I've submitted), so we need to postpone merging this till the release.

### DIFF
--- a/doc/source/addons/RemoteTech.rst
+++ b/doc/source/addons/RemoteTech.rst
@@ -19,15 +19,24 @@ If you launch a manned craft while using RemoteTech, you are still able to input
 
 It is possible to activate/deactivate RT antennas, as well as set their targets using kOS::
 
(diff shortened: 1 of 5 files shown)


erendrake:
I am going to mark it 'not ready' because RT 1.7 is not out yet

erendrake:
@tomekpiotrowski are any of these RemoteTech PRs unblocked now that 1.6.9 is out?

tomekpiotrowski:
Sadly both are in the 1.7 branch.

PR #1176, opened 2015-09-28
updated exception names

erendrake:
Its a silly change but i thought it would be nice to be consistent


erendrake:
the error came from accidentally pushing the branch to origin  then deleting it before the build could clone it.

PR #1177, opened 2015-09-29
Fixes #1116 with a reference direction.

### LINKED ISSUE #1116: Make orbit's LAN useful by exposing Reference Direction to scripts, and making an AN vector.
Read this user's problem:

http://forum.kerbalspaceprogram.com/threads/130163-Finding-normal-vector-to-arbitrary-orbit?highlight=kos

RIght now the longitude of the ascending node is a useless suffix of ORBIT because it's an angle measured from a super secret Reference Direction that scripts CANNOT find out what it is.

We need to do two things to fix this:

(1) Expose users to a vector that represents the current Reference Direction of the solar system, expressed in whatever the current Ship-Raw coords are rotated to at the moment. There seem to be hints from user-community docs at the "antaid" site that Planetarium.Right might be the KSP solar system's Reference Direction.

This value, it should be stressed, is NOT a property of any particular orbit.  It's a global game property of the solar system as a whole, even though its value expressed in ShipRaw coords will keep changing as the X and Z axes keep rotating in that lovely KSP wacky way every time you land/liftoff from a body.  But the point is it should not be a suffix of ORBIT or of anything else.

(2) While we're at it, we should just make an ascending node vector suffix for orbit.  ORBIT:ANVECTOR or something like that.  It would be the vector pointing from the orbit's SOI body's center toward its ascending node (either a unit vector, or the actual distance to the orbit at that point if we think we can calculate that fast enough).  We could calculate this ourselves by use of the LAN combined with the reference direction mentioned in (1) above.

Why we need both (1) and (2) - (2) is way nicer and simpler as an orbit suffix, but it doesn't help in the scenarios where the orbit doesn't really exist yet because it's not attached to a current object, as in the case of the user who's post I linked to at the top.  For that user they'll need (1) in order to calculate what the LAN means because there is no way to obtain an ORBIT without a vessel or body to get it from.  (Besides, what would the concept of the "current" position of such an orbit be?  Without that defined, much of the data of Orbit is not specifiable.)

Dunbaratu:
iI decided to drop the "give location of Ascending Node" part of the issue, as it's not quite the same remit, and it's possible for a user to calculate it now, as foillows:

```
set planetary_lan to ship:obt:lan - ship:body:rotationangle.
set an_position to latlng(0, planetary_lan):position.
```

### DIFF
--- a/doc/source/bindings.rst
+++ b/doc/source/bindings.rst
@@ -359,6 +359,37 @@ But kOS does not support this at the moment so in kOS if you set
 the LOADDISTANCE, you are setting it to the same value
 universally for all situations.
 
+.. _referencevector:
+
+REFERENCEVECTOR
+---------------
+
+Gives the Reference :struct:`Vector` for the Solar System itself, in
+current Ship-Raw XYZ coordinates.
+
+Both the :attr:`Orbit:LONGITUDEOFASCENDINGNODE` orbit suffix and the
+:attr:`Body:ROTATIONANGLE` body suffix are expressed in terms of
+degree offsets from this *Universal Reference Vector*.
+
+What is the Reference Vector?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The reference vector is an arbitrary vector in space used to measure
+some orbital parameters that are supposed to remain fixed to space
+regardless of how the planets underneath the orbit rotate, or where the
+Sun is.  In a sense it can be thought of as the celestial "prime
+meridian" of the entire solar system, rather than the "prime meridian" of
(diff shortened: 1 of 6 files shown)


Dunbaratu:
I'm reluctant to use FirstPointOfKries because calling something "Kries" is sort of creating "canonical" naming for the stock game that isn't really there.  Its a bit like when people started calling the sun "Kerbol" even though nothing in the game does that or says that.

"Universal" something is good, but "origin" is the wrong term because it's not a single spot in space but a directional vector.

How about ``UniversalZeroLongitude``, aliased as ``UniversalZeroLng`` ?

Issue #1079, opened 2015-06-28
Fix LOADDISTANCE to be in compliance with how KSP 1.0 does it.


Dunbaratu:
We made a dirty fix a while back to get things to compile with LOADDISTANCE when KSP 1.0 came out, but never revisited the issue of the fact that LOADDISTANCE is really not a single setting anymore and what kOS does with it is wrong.  We should go back and address this.

(This is something I realized as I was going over the docs.)

Dunbaratu:
Whomever works on issue #1128 should probably work on this issue at the same time, in the same PR.  They are inevitably intertwined issues.

PR #1178, opened 2015-09-29
Implement KUniverse and revert

### LINKED ISSUE #1148: Expose the ability to revert to launch or vab/sph
Provide methods for the user to revert to the launchpad/runway or back to the vab/sph from script.  While I don't know that reverting to the vab/sph will be all that useful for autonomous flight, reverting to launch would be very useful for iteratively running benchmarks on vessels.  And if we're going to implement one type of revert, it makes sense to implement the other.

We probably should make a way to query whether or not you are able to revert at a given time as well, rather than just throwing an exception.

hvacengi:
This fixes #1148
Adds a new structure (KUniverseValue) from which you can access information about reverting to launch.  You can determine if reverting is possible, and then invoke the revert.
- [x] New KUniverseValue class
- [x] Can revert detection
- [x] Invoke revert
- [x] Documentation
- [x] Fix bug when reverting from action groups

### DIFF
--- a/doc/source/structures/misc/kuniverse.rst
+++ b/doc/source/structures/misc/kuniverse.rst
@@ -1,0 +1,47 @@
+.. kuniverse:
+
+KUniverse 4th wall methods
+==========================
+
+
+.. structure:: KUniverse
+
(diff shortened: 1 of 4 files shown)


hvacengi:
Closes #1180

hvacengi:
Merged as part of #1079.

Issue #1128, opened 2015-08-23
LOADDISTANCE doesn't do any good without changing PACK distance as well.


Dunbaratu:
Each of the load distances also has a pack distance, for example:
```
    PhysicsGlobals.Instance.VesselRangesDefault.orbit.load
    PhysicsGlobals.Instance.VesselRangesDefault.orbit.unload
    PhysicsGlobals.Instance.VesselRangesDefault.orbit.pack
    PhysicsGlobals.Instance.VesselRangesDefault.orbit.unpack
```
In our bound variable LOADDISTANCE, we set both the ```load``` and the ```unload``` distances, but never touch the ```pack``` and ```unpack``` distances.  The effect this has is that you can have vessels loaded (their parts exist) but they are frozen in place as one fixed glued object that refuses to pay attention to any inputs of any sort.

Example where this came up:

Two airplanes parked at the ground at KSC, about 400m apart, one on the runway, the other on the grass to take off next to it.   Both running the same script to take off.  Both of them LOCK THROTTLE TO 1.  But only the currently active vessel actually throttles up.  The other one is not fully "working" even though you can see it.  This is the "loaded but not unpacked" state.

As the currently active vessel gets near the second vessel and starts to pass it by in the midst of its takeoff run, it gets within about 300m of the second vessel, and the second vessel suddenly throttles up and starts moving.   Because it started later, it is slower to get going, and the first vessel soon leaves it behind.  When they get more than 300m apart again, the second vessel's engines magically stop, and the second vessel freezes in place, wheels stuck to the ground, nose partly off the ground as it was in the midst of pulling up for takeoff.  You can see it (parts loaded), and you can even interact with it's kOS terminal, but KSP prevents it from moving.

When you crash the first vessel deliberately, and then switch back to the second one that was frozen in mid-takeoff. suddenly the second one "unsticks" the moment you switch to it, and it finishes taking off.

This evidence seems to point to the idea that:
```
    PhysicsGlobals.Instance.VesselRangesDefault.landed.pack
```
is only 300m, even though I've change the LOADDISTANCE to 15000.

Thus the whole intent behind the LOADDISTANCE setting is rendered moot.  It won't let you actually control a ship that's far away - just LOOK at it and run its terminal window, but any commands that require the vessel to physically move or change don't actually take effect.

Fix: maybe just adjust all 4 (pack,unpack,load,unload) to the same number?  For now?

hvacengi:
Closed with #1179

PR #1179, opened 2015-09-29
Fix support for load distance

### LINKED ISSUE #1079: Fix LOADDISTANCE to be in compliance with how KSP 1.0 does it.

hvacengi:
Fixes #1079 and #1128 
Creates new `LOADDISTANCE` and `SITUATIONLOADDISTANCE` structures for controlling the load, unload, pack, and unpack behaviors of vessels.
- [x] Implement load distances for all 7 situations
- [x] Provide access to the default values of load distance
- [x] Provide access to the load distance values directly on vessels
- [x] Allow the the active vessel to be set from scripts
- [x] Documentation


hvacengi:
Closes #1178

doc/source/structures/misc/kuniverse.rst
```diff
@@ -0,0 +1,117 @@
...
+    :return: none
+
+    Revert to the provided editor.  Valid inputs are `"VAB"` and `"SPH"`.
+
+.. attribute:: KUniverse:ORIGINEDITOR
+
+    :access: Get
+    :type: string.
+
+    Returns identifier of the orginating editor based on the vessel type.
```
Dunbaratu: is this "VAB" or "SPH"?  I'm guessing that, but the doc doesn't make it clear.

doc/source/structures/misc/loaddistance.rst
```diff
@@ -0,0 +1,109 @@
...
+          - scalar
+          - Get/Set
+          - The pack distance
+
+.. attribute:: SituationLoadDistance:LOAD
+
+    :access: Get/sET
+    :type: scalar
+
+    Get or set the load distance.  This is the distance at which a :struct:`Vessel` will load into physics (as opposed to "on rails").  This value must be greater than :attr:`UNLOAD`, and will automatically be adjusted accordingly.
```
Dunbaratu: I'll have a go at explaining this a bit more explicitly.  The term "on rails" is confusing here because a loaded but packed vessel still behaves in a very 'on rails' sort of way - in the sense that you cannot move it or execute its kOS programs.

doc/source/structures/misc/kuniverse.rst
hvacengi: Yes, those should be the only two values returned, though technically the enum includes a value for `NONE`.  Again, do you want to add it to your review?

doc/source/structures/misc/loaddistance.rst
hvacengi: I welcome your attempt to clarify the definitions.  The concept is kind of difficult.

I did not play around much with the loaded vs packed values.  I nearly wrote the same protection into them that I already included with the load/unload and pack/unpack pairs.  I may take another swing at that.

hvacengi:
You can seem my comments in response to @Dunbaratu's review PR here: https://github.com/hvacengi/KOS/pull/7

These revisions have been uploaded and are ready for final review.

Dunbaratu:
> These revisions have been uploaded and are ready for final review.

The way you wrote them seems to imply you still were planning on one more edit to the docs first before it's final.

hvacengi:
Oh, I just meant that I made a couple of changes, so you should review those changes to make sure every thing still looks OK.  Unless you disagree with any changes I made, I think I'm good with calling this final.

Dunbaratu:
I followed the link you provided which only showed the comments on what I sent back to you, not the new version where you had made the edits you were talking about.  I hadn't seen that you did in fact make them and not just talk about them.
 never mind.

Dunbaratu:
looks good.  I'll declare it done and pull it.

PR #1180, opened 2015-09-30
Opcodes can now abort execution

### LINKED ISSUE #1120: SHUTDOWN doesn't take effect immediately.
Based on Reddit discussion at https://www.reddit.com/r/Kos/comments/3g3nfq/this_very_short_script_i_wrote_to_abort_the_whole/ , `SHUTDOWN` takes effect only at the end of the physics tick where it executes.

The desired and intuitive behavior is for SHUTDOWN to immediately abort any code that's running, so the final state is deterministic.

hvacengi:
Fixes #1120
Modify the Opcode class so that any opcode can abort program or context execution and change the cpu to use this new property.  This allows the shutdown and reboot functions to prevent execution of future opcodes as described in the referenced bug.

### DIFF
--- a/src/kOS/Execution/CPU.cs
+++ b/src/kOS/Execution/CPU.cs
@@ -1009,6 +1009,8 @@ namespace kOS.Execution
             {
                 try
                 {
+                    // If the program is ended from within a trigger, the trigger list will be empty and the pointer
+                    // will be invalid.  Only execute the trigger if it still exists.
                     if (currentContext.Triggers.Contains(triggerPointer))
                     {
                         currentContext.InstructionPointer = triggerPointer;
@@ -1035,6 +1037,8 @@ namespace kOS.Execution
                 }
             }
 
+            // since `run` opcodes don't work in triggers, we can use the opcode count to determine if the
+            // program has been aborted.  If the count isn't right, leave the pointer where it is.
             if (oldCount == currentContext.Program.Count)
             {
                 currentContext.InstructionPointer = currentInstructionPointer;
--- a/src/kOS/Execution/CPU.cs
+++ b/src/kOS/Execution/CPU.cs
@@ -1063,9 +1063,21 @@ namespace kOS.Execution
             }
             try
             {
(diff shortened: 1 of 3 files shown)


src/kOS/Execution/CPU.cs
```diff
@@ -1031,7 +1037,12 @@ private void ProcessTriggers()
                 }
             }
 
-            currentContext.InstructionPointer = currentInstructionPointer;
+            // since `run` opcodes don't work in triggers, we can use the opcode count to determine if the
+            // program has been aborted.  If the count isn't right, leave the pointer where it is.
```
Dunbaratu: I'm not comfortable with layering a good feature on top of a bad misfeature.  It's true we can't RUN from a trigger, but that's unfortunate, not desirable.  Making use of that as a core assumption in implementing something else useful makes it harder to fix the misfeature later.

hvacengi: It's been a while since I wrote this part.  I will have to review to see if there's a good way to get the Abort info back out of `ExecuteInstruction` As implemented, we just get the binary "continue or not" return value.  The problem is that right now the program has already been set back up after the execution of the reboot call opcode.  So If we reset the instruction pointer, it causes a misalignment.  Maybe we could switch to returning an enum instead of a bool.

For what it's worth, I'm not a fan of `run` working inside of triggers.  Interrupt and event driven programming should avoid long and/or complicated execution if possible, and I would say running a new program is by definition complicated.  But I won't argue with you on the feature.  I know that in my early kerbscript days I tried to do it, but that was before user functions.  I'm sure it would be useful to some one else, so leaving the door open for the feature is a good idea.

Dunbaratu: I guess for now we can accept it and flag it as a thing to test in the future after triggers stop being limited by IPU.

Dunbaratu:
I tried it, tested it, and see nothing to change in the code, so I'll pull it as-is right on the website.

Issue #1068, opened 2015-06-13
A kOS part shouldn't be able to control other vessels directly, but it is.


Dunbaratu:
(I thought this bug was already reported ages ago but I read through the issues list and couldn't see it, so I'm reporting it again.)

### To recreate:

Make two airplanes, one with a kOS computer and one without.  Put them within loading range so both vessels' parts are fully loaded at the same time (i.e. have them both parked on the runway, for example).  Give them movable elevator control surfaces so you can easily see when they're being pitched up or down.

Call the vessel with a kOS computer "ship A" and the other one without a kOS computer "ship B".

At the interactive prompt In the kOS terminal for ship A, just type this command:
```
set vessel("ship B"):control:pitch to -1.
```
Then type this:
```
set vessel("ship B"):control:pitch to 1.
```
You can prove that the kOS computer on ship A has the full ability to control ship B by watching ship B's elevator surfaces when you do this.  They respond and move to your commands typed in ship A's terminal, even though ship A has no wiring connecting it to ship B's controls.

### Why is that a bug?

While sending wireless messages between kOS computers so they can coordinate with each other is a neat thing we should support later, it shouldn't just work magically like this.  It should require an antenna connection from source to dest, and it should require some kOS script on the destination vessel to hear the messages the other kOS vessel is sending it and IT moves the controls, not the kOS script on the source vessel.  (i.e later we support some message queue between vessels and let people write their own scripts to decide what to do with the messages).

But in the meantime, this behavior of magically allowing a kOS computer to get direct control of any vessel it's not part of, using the exact same API it uses to control itself, should be considered a bug and disabled.

This immediately raises the question of, in addition to preventing controlling foreign vessels, perhaps preventing querying too much data about them.

Getting the position and velocity of a foreign vessel (and therefore predicting its future position) should be possible just by taking remote readings so that can stay, but things like querying a foreign vessel's internal resources list ("how much charge is left in your batteries, ship B?") shouldn't be possible automatically without even having a connection.  Whether we decide to make that entirely a script-land thing or give it to users automatically as long as its connected can be decided later.  For *now* the fact that you can just get the data magically for a vessel you have no contact with is a bug.

erendrake:
closed with #1182

PR #1182, opened 2015-10-02
Throw exceptions when controlling another vessel

### LINKED ISSUE #391: Only allow partmodules to be altered for CURRENT ship
It is currently possible to use the new partmodule system to fiddle with the controls of parts your CPU isn't actually connected to because they're on a different ship.

For example:
```
// open the shroud on a docking port of the target vessel:
target:partsdubbed("that dockingport")[0]:doevent("open").
```

That should only be done by the SCS that is ON that target ship.  A check should be put in place that for a DOEVENT, GETFIELD, SETFIED, or DOACTION to be legal, the part must be a member of the same vessel as the vessel running the code (the shared.vessel).

hvacengi:
Fixes #391 and #1068.
Adds a new exception class for cases where a script tries to access a suffix on another vessel that should only be accessible to the local cpu vessel.

The VesselTarget, PartValue, and PartModuleFields classes now implement a method for checking the vessel, and have had the check added to some suffixes:
- `vessel:control`
- `part:controlfrom`
- `part:tag` (set only)
- `partmodule:doaction`
- `partmodule:doevent`
- `partmodule:setfield`

I think the correct way to document this is to add a warning to each of these suffixes documentation identifying that access is restricted to the local cpu vessel.

I've labeled the PR as breaking, because there may be scripts that previously took advantage of this bug that will be broken.

I welcome any and all suggestions for additional suffixes (including on other structures) to black list.


Dunbaratu:
I was considering self-assigning this and pulling it, but the only reason I haven't is that it has the "Pending Documentation" tag, which hints at it maybe not being ready yet?  If it's ready, I'm happy to pull it and write the documentation.

hvacengi:
I want to get this merged after we do most of the other changes.  I was kind of hoping that others would chime in with more suffixes that should be protected by this code, so I didn't want to get this branch itself done and then need to go through and add all of the exceptions again.  The other option is that we update what we can in current develop, and then we do the check on new submissions as part of the review.  Let me know what you want, I have a blurb prepared to put it the effected suffixes' text.

erendrake:
other than listing docking ports, im not sure i see why any of a craft's parts would be available to be queried.

Are we worried about reducing the chance to do cool stuff if we hobble the data available to cores on other craft?

Dunbaratu:
> other than listing docking ports, im not sure i see why any of a craft's parts would be available to be queried.

Docking.

When docking with another vessel, you need to know it's docking port parts' locations and orientations and not just the entire target vessel as a single unit.

erendrake:
Alright, then i think this is good enough for now and we should get this documented and merged. We can discuss other suffixes later if we find a need. @hvacengi do you need anything from us on this?

hvacengi:
I don't think I'll need anything.  I'll merge the current develop tonight, then get the documentation noted.  I kind of like the idea of waiting to add more suffixes, mostly because we can see if there's any backlash for the core change itself.

hvacengi:
Documentation has been committed and the most recent develop branch merged.  This should be ready.

PR #1185, opened 2015-10-10
Fixes #1184 with new ``once`` keyword for ``run``.

### LINKED ISSUE #1184: Implement 'run unless already run' kerboscript feature
For libraries, it can be useful to have something that will run them but only if they haven't already been run once during the current instance of the *Program Context*.

To understand this issue's description, it's vital to know the following terminology:

**terminology:** *Program Context* - The array of kRisc Opcodes that represents the current compiled program content in memory.  There are two different program contexts: the *interpreter context* and the *program's context*  The interpreter context is the one that holds the current interpreter code.  It is designed to grow slowly, one command-line at a time.  The program's context is the context the interpreter clears out and starts anew every time the *interpreter* issues the ``run`` command.

The behavior of ``run`` is different when issued from the interpreter versus when it is when issued from a program.  In the interpreter it makes a NEW program context, compiles the program into it, and runs it there in that new context, which will be thrown away and started over the next time it issues a new run  command.  (Thus edits to the script file will take effect immediately on the next run when it recompiles the script.)  However, when you ``run`` from a program (one program running another program), it's a bit different.  It *appends* the sub-program's opcodes to the current program context, rather than making a new one (if it compiled the program before in a previous instance of the ``run`` command (i.e. if you run something in a loop), it skips this step and doesn't compile it again, instead just using the code that's already there in memory from last time.).  Once the code is present, appended to the current program context, it jumps to the sub-program's mainline code entry point instruction.

This avoids the compile step when the sub-program was already run previously in a different sub-program.  Thus if you have a sub-program that contains only functions and no mainline code, and you run it a second time from some other part of the current program context, nothing really happens on the second run, since there's no compile step done.  It just jumps to the already existing mainline code for it, which since there was no mainline code, is just a dummy stub program containing an end of program statement, which returns right away.

This model works great for the loading of libraries *that consist of no mainline code at all and are just functions*.  You can just put the ``run mylib.ks.`` at the top of every sub-program file that needs it, and not care whether or not some other sub-program already ran it in this context or not, since running them won't execute anything meaningful, and the expense of compiling the functions doesn't happen the second time.

But it is a problem for loading library files that, in addition to having functions, *also contain some mainline setup code or initialization code* that you only meant for them to run once.

For them , it would be nice to have a command that means "run, but only if it hasn't already been run".

It's a bit like the `require` or `import` keywords in some languages, except that they imply a compile-time importation, and this would be a runtime importing, and would actually RUN the code if there is any there to be run, but just the once and not again on the second or third occurrence of the same run command.
.
I propose the syntax: ``run once progname`` 

(Basically, making a new reserved word `once`, and having it be an optional modifier to the `run` command.  If the `once` keyword is present, it means the progname will not be run at all if it already has been before during this program context.)

Dunbaratu:
Means of change:

### Previously, a run command worked as follows:

   First, regardless of how many instances of the same run command
   exist, make exactly one function in the header declarations area that
   looks like so:

      // if programname is already compiled, tested by seeing
      // if the programname's entry point IP is still the dummy value of -1:
              push progname-prog*
              push -1
              equals
      // then skip the section that compiles and loads:
              br.false runIt
      // section that does the load:
              push programname-prog*
              push ARGBOTTOM
              push null
              call load()
              pop
              store // stores entry point IP from load() function
                    // in programname-prog*

      // section where the running happens:
      runIt:  call progrname-prog*
              pop // ignore return value
              push 0 // replace it with hardcoded 0
              return

Then, every time the ``run`` command is encountered, it calls that
function built above:

    call (the IP of the code mentioned above)

### This is the new change this PR makes:

The built run code looks like so now:

      // if programname is NOT already compiled
              push progname-prog*
              push -1
              equals
      // then skip the section that does the 'once' check:
              br.true loadIt
      // If it gets here, that proves the compile already happened.
      // Now pull the 'once' arg off the stack use it to decide whether
      // to run the code or just return now:
              br.false runIt
              push 0  // return value from the not-run 'run' command.
              return
      // section that does the load:
      loadIt: pop   // ignore the 'once' argument on the stack and throw it away
              push programname-prog*
              push ARGBOTTOM
              push null
              call load()
              pop
              store // stores entry point IP from load() function
                    // in programname-prog*

      // section where the running happens:
      runIt:
              call progrname-prog*
              pop // ignore return value
              push 0 // replace it with hardcoded 0
              return

Then every time the 'run' command is encountered, if does this:

If it wasa 'run once', then it does this:

              push True
              call (the IP of the code shown above)

or if it does a 'run' without a 'once', it does this:

              push False
              call (the IP of the code shown above)


Dunbaratu:
Note that because I made my PR from the current Develop branch, and the current Develop branch does not yet have my #1133 PR merged into it, there may be a small merge conflict between this and #1133 due to the fact that ARG_MARKER_STRING changes with #1133, and this code makes some references to it working the old way (because I have to make it using current develop, which is missing my other PR).

Dunbaratu:
Oops - forgot to write user docs for ``once``.  Let me do that and add a supplemental commit.

doc/source/language/syntax.rst
```diff
@@ -31,11 +31,11 @@ operator symbols:
...
-    set  to  if  else from until step do lock  unlock  print  at  on  toggle
-    wait  when  then  off  stage  clearscreen  add  remove  log
-    break  preserve  declare  parameter  switch  copy  rename
-    volume  file  delete  edit  run  compile  list  reboot  shutdown
-    for  unset  batch  deploy  in  all
+    add all at batch break clearscreen compile copy declare delete
+    deploy do do edit else file for from from function global if
+    in list local lock log off on once parameter preserve print reboot
+    remove rename run set shutdown stage step switch then to toggle
+    unlock unset until volume wait when
```
Dunbaratu: Not directly related to this PR, but when I added ``once`` to this list, I decided it was about time we re-sorted it to be alphabetical.

doc/source/language/user_functions.rst
```diff
@@ -104,14 +104,96 @@ and once its been "parsed" by the compiler, the function can be called
...
+
+    prog1:      next counter ID = 1
+    prog1:      next counter ID = 2
+    prog1:      next counter ID = 3
+    subprogram: next counter ID = 4
+    subprogram: next counter ID = 5
+    subprogram: next counter ID = 6
+    
+whereas, had you used just ``run counterlib.`` instead of
+``run once counterlib.``, then it would have printed this:
```
Dunbaratu: dang it - this needs two colons. not one.

erendrake:
This is my bad for being slow on merging this but there is a conflict on this merge now that i would rather have you do if you dont have objections. Sorry :(

Dunbaratu:
I'll have a look at the merge conflicts tonight.  I have a suspicion it's just that the arg bottom marker has changed from string to a new class.

Dunbaratu:
Merge conflict looked into and fixed.  Also ended up exposing an error in the original fix that hadn't been noticed because it was implemented prior to the enforcement of matching arg count in run statements.

src/kOS.Safe/Exceptions/KOSOnceInvalidHereException.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// A version of KOSCommandInvalidHere describing an attempt to use
+    /// the WAIT keyword when in a trigger.
+    /// </summary>
+    public class KOSOnceInvalidHereException : KOSCommandInvalidHereException
+    {
+        public override string HelpURL
+        {
+            get { return "http://ksp-kos.github.io/KOS_DOC/summary_topics/CPU_hardware/index.html#WAIT"; }
```
Dunbaratu: crap.  this is a cut and paste error I didn't catch.  It's going to direct users to the wrong part of the docs because this is a copy from a different exception.

PR #1189, opened 2015-10-14
Delegates

### LINKED ISSUE #740: [proposal] functions as callbacks
see [this wiki page](http://en.wikipedia.org/wiki/Callback_(computer_programming))
Not in next update but at some point you might want to introduce callbacks.
How I would like to see it:
```
declare function foo  // declare a library function
{
  declare parameter callback.
  print callback(42).
}
declare function bar  // declare a callback
{
  declare parameter x.
  return x*x - x.
}
foo(bar). // pass the callback into library function as parameter
```
The current implementation of functions will think that `foo(bar).` is `foo(bar())`. But considering `bar` and `bar()` the same prevents you from introducing callbacks later (or at least you will have to add some complex syntax like `foo(bar as callback)`). May be it is not to late to make parentheses required?

tomekpiotrowski:
Fixes #740 

Not ready yet, but enough for somebody to have a look. Syntax:

```
set d to delegate somemethod.
d("argument").

set l to list(delegate somemethod).
//This does not work for now:
l[0]("argument").
//You have to:
set d2 to l[0].
d2("argument").
```


hvacengi:
Interesting.  I hadn't noticed the previous conversation about user delegates, but @Dunbaratu and I were talking about them briefly the other day on slack.  He would be a better person than I to look at the implementation in the compiler and cpu execution.  I'm working on getting involved with that stuff, but it's still a ways off.

I like the decision to use a kOS value rather than trying to trick the compiler into not executing the user function.  I also really like the idea of being able to use the keyword within a call (like in your example `set l to list(delegate somemethod).`.  Are you thinking this would be able to work with function suffixes as well?  `set d to delegate list():add.`  In that instance, we might need to watch out for orphaned objects like the on in my example.  That might be something to watch out for in your example of the delegate keyword inside of a method call.  How is the object created/stored at that point, or is it essentially re-created every time you call it?

What will really take a little time and/or though is the keyword itself.  `delegate` is very recognizable to some one familiar with C#, but I don't know how "plain language" recognizable it would be.  Off the top of my head, other options might be `callback`, `functionpointer`, or maybe even `userfunction` or `function` itself.  `function` would require the use of the keyword anytime you want to treat the function as a delegate/callback/whatever, and I think that's less than ideal.  I like that your syntax allows for reusing the delegate object.

I think @erendrake is going to reach out to you about joining our slack channel.  It might make working out the syntax itself a little easier if we can communicate in real time.

tomekpiotrowski:
The change itself was very easy actually. The underlying UserDelegate was already implemented for other purposes, I've just added the syntax and the wrapper class.

`delegate` was one of two keywords proposed in the original issue, so I went with that, but we can change it of course.

About the suffixes, they surely don't work now - I expect a user identifier. I would have to take a look again at how they're implemented to evaluate how difficult it would be to make them work, I'm still new to kOS internals :)

Dunbaratu:
**On trying to also support Suffixes:**

User functions and suffix methods are still different enough from each other to make it hard to make the same mechanism work for both.  Maybe it could be done with some magic inside UserDelegate to make it so that it can alternatively store a method call.  (i.e. have it have a flag for "is suffix" versus "is user function", and if it is a suffix, have it store a ref to the parent object and its suffix string name (has to be string name because not all suffixes use the new AddSuffix() system, so you can't reliably use the actual C# Delegate)).

**On the naming:**

I'm a little unsure about the name 'delegate'.  First of all, it's a C# specific term.  Most other languages call it something else, making it less universally understood by people who don't do C#.  Secondly, because of this, it holds certain assumptions wrapped up inside it that won't be true that might trip up people who *do* know C#.  It won't do operator overloading.  It won't know the arguments list.  It won't know the return type.  "Delegate" implies all these things are doable.  It's much more like a C function pointer, plus maybe some Closure information.

But the keyword is easy enough to change later if we don't like it.  It's not set in stone until there's a public release of this code.

erendrake:
@tomekpiotrowski i am sorry i didnt get this in before we had a conflict, would you mind catching this branch up to develop please?

tomekpiotrowski:
This PR still needs some work. Syntax may have to be changed and there are a couple of minor issues. I'll get back to it after I'm done with the stuff I'm working on now.

tomekpiotrowski:
This is superseded by @Dunbaratu's work.

PR #1192, opened 2015-10-17
Implement CORE:VOLUME

tomekpiotrowski:


hvacengi:
Glad you caught and closed this, I forgot to mention that the PR became redundant when I set `Core` to inherit from the `kOSProcessorFields`

Issue #1168, opened 2015-09-22
Key-value store


tomekpiotrowski:
Some of my scripts need data to be stored between executions. Up till now I've been using `log "set ..." to file. ... run file.` to store simple data in a file, but that's less than ideal. I thought about providing a simple API that would allow the user to store key-value entries in a file on the volume. That file could use JSON internally. The API would look more or less like this:

```
store:put("phase", "parking_orbit").
store:put("list_data", list(25, 15, 25)).

store:has("phase").
store:get("list_data").
```

Alternatively we could provide a lower level File API, that would let users create/read/write to files.

Thoughts?

Dunbaratu:
I figure reading a text file and reading user input keys is really the same thing, and should be addressed together.

Then when it can be done low-level, that's when it's time to consider layering things like this on top of it.

erendrake:
@tomekpiotrowski we have been working on an associative array type for users. that coupled with the list type would let us have a full bi-directional serialization between a JSON file stored on the disk and these two types in memory. 

I think that could be a pretty cool tool for ingesting data into kOS from the archive

tomekpiotrowski:
@erendrake Yes, I know, and I've been thinking about that as well :)

@Dunbaratu I'm not sure if the lower level stuff is that necessary in our case. Most people manage without storing anything and storing simple key-value entries covers majority of any use cases I can think of. I think very few people actually need proper, low-level file access. To be honest I'm not sure if it's worth spending time on. Limiting ourselves to serializing stuff to JSON could save us a lot of effort. Or is there something I'm missing?

Anyway, I'm willing to put some time into this (if I have some to spare), so if we decide on what to do here I can give it a shot.

Dunbaratu:
Here's why it's not as simple as you're both implying:

The associative arrays can store any data type as their value.
They can use any data type as their key.

To automagically transform any arbitrary associative array into a file or visa versa first requires reliable serialization of all types kOS variables can hold.

i.e. How do you print a VesselTarget to a file?  How do you read a PartValue from a file?  The fact that we didn't handle serialization of these things well at all is why we removed the load/save of program state when you switch scenes.

It would require limiting the associative array to a small subset of the possible data values - simple ones like strings and numbers - to make it easy to implement.

[...]

Issue #353, opened 2014-11-10
Inter-CPU communication can be made possible now


Dunbaratu:
The big hurdle to inter-CPU communication had always been the inability to NAME which CPU you meant.

Now you can do it with kOSNameTags.  We could send messages between CPU's now because you can name the SCS parts in the editor and then say "send this message to 'MyCPU1'.".

Tabris83:
this would also enable detecting decoupling. (ie. when a probe becomes a separate vessel then run program.)

PR #1193, opened 2015-10-17
Set volume name to kos tag

### LINKED ISSUE #354: Naming volumes from KOSNametags?
Now that parts can be named with KOSNameTags, should we implement the ability to refer to a volume by its nametag?  (The Part on which the KOSProcessor PartModule exists should also have a kOSNameTag partmodule on it too now since we attached KOSNameTags to all parts now.).

tomekpiotrowski:
Fixes #354 

Sets the initial volume name to processor's kOS name tag.

### DIFF
--- a/doc/source/general/volumes.rst
+++ b/doc/source/general/volumes.rst
@@ -120,6 +120,9 @@ Therefore using the RENAME command on the volumes is useful when dealing
 with multiple CX-4181's on the same vessel, so they all will refer to
 the volumes using the same names.
 
+If a kOS processor has a name tag set, then that processor's volume 
+will have its name initially set to the value of the name tag.
+
 Archive
 -------
 
--- a/src/kOS/Module/kOSProcessor.cs
+++ b/src/kOS/Module/kOSProcessor.cs
@@ -293,6 +293,12 @@ namespace kOS.Module
             if (HardDisk == null)
             {
                 HardDisk = new Harddisk(Mathf.Min(diskSpace, PROCESSOR_HARD_CAP));
+
+                if (!String.IsNullOrEmpty(Tag))
+                {
+                    HardDisk.Name = Tag;
+                }
+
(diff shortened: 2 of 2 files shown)


tomekpiotrowski:
Ah, this actually depends on #1195

hvacengi:
I'm not sure I like this being a set in stone feature.  I personally use the processor's tag to store other data (such as the name of the station it is to dock with).  The user doesn't necessarily want to have the volume named "Babylon 5" just because they put that in the name tag of the processor.  On the other side of the coin, unless the user takes the time to set the volume name directly they probably wouldn't even notice that the name is different.

I think I'd prefer a better way to opt in to this feature, like a config flag.  The alternative would be to recommend that users place the following line at the opening of their boot file: `if core:volume:name = "" set core:volume:name to core:part:tag.`

tomekpiotrowski:
I was thinking that in the end we could totally eliminate volume names. See #1196.

Tags would only be used for identification. I'm working on storing data here #1168 and I have #353 implemented. Those 2 would eliminate the need to use tags for anything else than ids.

Dunbaratu:
I'd rather tags not be used to store data.  It's really a roudabout way to deal with things.  They're supposed to just be used to identify WHICH thing you have, not to store data IN the thing.  I think using it for the volume name makes perfect sense, but it does have one consequence to think about:

Right now we *happen to have* a one-to-one relation between volumes and cpu's.  There's no rule that says it has to stay that way forever.  Maybe we'll have a part that is just a hard drive and stores several volumes?  But using the tag name makes it so that the one-to-one relation is set in stone.

tomekpiotrowski:
@Dunbaratu I see your point.

Dunbaratu:
In retrospect I think it's fine to have the one-to-one relationship because if we ever have multiple hard drives in one unit we can just add more than one PartModule to it, one per drive.  So never mind, the design is fine.

PR #1195, opened 2015-10-17
Get processor by volume or tag, kOSProcessorFields, list processors

### LINKED ISSUE #788: Feature request: volume to kOS processor part and processor to volume mapping
Sometimes it can be useful to have something like this:
```
list volumes in volumes.
for volume in volumes
{
  if volume:renameable // if not archive
  {
    if volume:part:tag = "some tag"
    {
      do stuff
    } 
  }
}
```
or like this: `kos_processor:volume:files`

tomekpiotrowski:
Fixes #788

Makes it possible to query, list kOS processors etc. Basically stuff that I needed to make inter-cpu communication possible.


tomekpiotrowski:
Would appreciate a speedy review and merge. I have an inter-cpu communication PR ready that depends on this.

erendrake:
As luck would have it I have a bunch of time this weekend set aside to merge stuff. This will be behind a few others but it should be possible to get many of them in

src/kOS/Module/kOSProcessor.cs
```diff
@@ -30,6 +30,15 @@ public class kOSProcessor : PartModule, IProcessor, IPartCostModifier, IPartMass
 
         public Harddisk HardDisk { get; private set; }
 
+        public string Tag
+        {
+            get
+            {
+                KOSNameTag tag = part.Modules.OfType<KOSNameTag>().FirstOrDefault();
+                return tag != null ? tag.nameTag : null;
```
hvacengi: Since kOS doesn't give the user any way to interact with null values, this should return an empty string.  There is no guarantee that the tag will be present on a given part, because the MM script was intentionally written so that the user could remove the name tag from any part using the :final keyword.

From the PartValue class: `return tagModule == null ? string.Empty : tagModule.nameTag;`

space-is-hard: @hvacengi 

>Since kOS doesn't give the user any way to interact with null values

I believe that was actually fixed with #1135

hvacengi: No, I'm pretty sure that just gives the ability to tell if a variable has been defined.  There was a conscious decision at the time to not pass null values to the user, nor give the user an ability to access them.  Some way to null check will probably come when we implement a type structure for the users.

PR #1199, opened 2015-10-18
Account for time warp when updating observers

### LINKED ISSUE #723: RemoteTech delay should be in game time not real time
From Reddit.

I didn't see it listed in there, and I'm not sure if it's been addressed yet, but something that really bugs me with regards to RT integration is that when kOS is waiting for the delay to finish it uses real time as opposed to in-game time. So if you non-physical timewarp at all then suddenly everything gets out of sync and actions you queued in the RT flight computer for after a script starting will suddenly happen first (since RT works on in-game time) while kOS will still be using real-time.
Waiting a minute and a half and not being able to timewarp through it isn't fun. And that's only at Duna orbit. If I ever have to do anything manual with my Eeloo probe, it'll be awful. During testing I'd turn off RT integration to do manual commands (copying from Archive, renaming files, starting scripts, etc) then turn it back on so that kOS could actually control the ship, otherwise I'd have to wait several real minutes between every command.

http://www.reddit.com/r/Kos/comments/31amif/does_anybody_know_when_017_will_come_out/cq02l4s

tomekpiotrowski:
Fixes #723

### DIFF
--- a/src/kOS/Execution/CPU.cs
+++ b/src/kOS/Execution/CPU.cs
@@ -891,7 +891,7 @@ namespace kOS.Execution
 
             if (showStatistics) updateWatch = Stopwatch.StartNew();
 
-            currentTime = shared.UpdateHandler.CurrentTime;
+            currentTime = shared.UpdateHandler.CurrentFixedTime;
 
             try
             {


hvacengi:
Interesting, so the `TimeWarp` deltas change based on what warp you're in, while those values on `Time do not.  It kind of makes you wonder why they even have the values available on `Time`.

I like fixing it for all observers.  This should also correct a bug I never had time to verify and report: `wait 10.` would wait in real time, not in warp time.

tomekpiotrowski:
From what I can see `Time` is a Unity class, while `TimeWarp` was added in KSP, so it makes sense that `Time` does not handle time warp.

I will test the `wait` bug when I get a chance.

tomekpiotrowski:
While doing some testing before I've noticed that resource transfers are not affected by time warp as well. Looking at the source code deltaTime is not used in any calculations in `ResourceTransferValue.cs`. I guess it's not that big of a deal since transfers are usually quick anyways.

hvacengi:
The transfer is actually managed in Execution/TransferManager.cs, which implements `IFixedUpdateObserver`.  It calls the `ResoureceTransferValue`'s update method using the `deltaTime` passed to `UpdateHandler.UpdateFixedObservers`.  So you're changes should have already fixed resource transfer :smile:

tomekpiotrowski:
You're right, but if you look inside `ResourceTransferValues#Update()` the `deltaTime` is not used. I've tested transfers with that fix before and they were still not affected by time warp.

hvacengi:
Interesting.  Looking deeper into the code (I should have done that before) it looks like @erendrake hard coded the transfer to be limited by `RESOURCE_SHARE_PER_UPDATE` which is 0.5% per update.  If we just use `deltaTime` to adjust that rate, that should make resource transfers respect timewarp:
```
        private double PullResources(IList<global::Part> parts, double transferGoal, double deltaTime)
        {
            double toReturn = 0.0;
            var availableResources = CalculateAvailableResource(parts);
            foreach (var part in parts)
            {
                var resource = part.Resources.Get(resourceInfo.id);
                if (resource == null) continue;

                var thisPartsPercentage = resource.amount/availableResources;

                // Throttle the transfer
                var thisPartsShare = transferGoal*thisPartsPercentage;
                var thisPartsRate = resource.maxAmount*RESOURCE_SHARE_PER_UPDATE*deltaTime/0.02;
                
                // The amount you pull must be negative 
                thisPartsShare = -Math.Min(thisPartsShare, thisPartsRate);
                // the amount is subject to floating point lameness, if we round it here it is not material to the request but should make the numbers work out nicer.
                thisPartsShare = Math.Round(thisPartsShare, 5);

                toReturn += part.TransferResource(resourceInfo.id, thisPartsShare);
            }
            return toReturn;
        }
```
I multiplied the rate by `deltaTime/0.02` because at least in my game the default timescale is 0.02.  Alternatively it could be written `deltaTime/Time.fixedDeltaTime` which would scale the rate based on the local minimum frame time.

tomekpiotrowski:
Looks like it should work :)

hvacengi:
Do you want to add that as part of your update, or should I submit a separate PR?

tomekpiotrowski:
I think it's a separate issue and deserves its own PR :)

PR #1206, opened 2015-10-22
Add Missing Flight Parameters

### LINKED ISSUE #1205: kOS unable to lock throttle or steering after undocking
Was asked by hvacengi to post this issue as a reminder.

After docking and then undocking without ending a kOS program, the core is left unable to lock throttle or steering. Even ending the program and trying to lock throttle or steering manually does not work, only rebooting the core "fixes" the issue. Something related to the FlightControlManager according to hvacengi.

hvacengi:
Fixes #1205.

Some times UnbindUnloaded will remove the flight parameters after docking changes, but the processor will switch to the new vessel.  In these cases, the FlightControlManager is not reinitialized, so we now re-add any missing parameters after attempting to update the parameters' vessel properties.

### DIFF
--- a/src/kOS/Binding/FlightControlManager.cs
+++ b/src/kOS/Binding/FlightControlManager.cs
@@ -153,7 +153,7 @@ namespace kOS.Binding
 
         private void AddNewFlightParam(string name, SharedObjects shared)
         {
-            flightParameters.Add(name, new FlightCtrlParam(name, shared));
+            flightParameters[name] = new FlightCtrlParam(name, shared);
         }
 
         private void AddMissingFlightParam(string name, SharedObjects shared)
--- a/src/kOS/Binding/FlightControlManager.cs
+++ b/src/kOS/Binding/FlightControlManager.cs
@@ -110,6 +110,12 @@ namespace kOS.Binding
 
             foreach (var param in flightParameters.Values)
                 param.UpdateFlightControl(currentVessel);
+
+            // If any paramers were removed in UnbindUnloaded, add them back here
+            AddMissingFlightParam("throttle", Shared);
+            AddMissingFlightParam("steering", Shared);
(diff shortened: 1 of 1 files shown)


src/kOS/Binding/FlightControlManager.cs
```diff
@@ -150,6 +156,14 @@ private void AddNewFlightParam(string name, SharedObjects shared)
             flightParameters.Add(name, new FlightCtrlParam(name, shared));
```
Dunbaratu: Would it be better to do this as:
```
flightParemeters[name] = new FlightCtrlParam(name,shared);
```
That way, you don't have to bother having protections checking if the key is already present, because unlike Add(), the index operation will happily clobber a value in-place if it already exists (where Add() throws an exception).

The effect this change would have is that every time you call AddNewFlightParam, you'd be guaranteed that it *WOULD* ensure that the value is updated to the new thing, even if it already existed.  Thus preventing potential stale data.  What this class has to do at the moment (As seen below in AddMissingFlightParam() is this:
```
            if (!flightParameters.ContainsKey(name))
            {
                AddNewFlightParam(name, shared);
            }
```
That has the effect that it will refuse to *update* a stale old value if, say, the ``shared`` object is now something different and needs updating (i.e. like after a dock/undock maybe?)

erendrake:
There are still open comments on this one so for now i am going to remove it from this current milestone.

hvacengi:
@Dunbaratu your comment on the dictionary key access has been implemented.

I don't think it will be possible to entirely eliminate the key check.  The main issue is that parameters need to be unbound before they can be fed to the garbage collector.  At the very least, the new steering manager should be properly disposed/unbound, since all instances are held within a dictionary until removed.  That leaves the door open for a potential memory leak, and other unexpected behaviors.

PR #1214, opened 2015-10-29
Fixes #813 with protection against running past array size.

### LINKED ISSUE #813: error printing resources
PRINT SHIP:RESOURCES AT (0, 0).

causes 

[Log]: System.ArgumentException: Destination array was not long enough. Check destIndex and length, and the array's lower bounds
  at System.Array.Copy (System.Array sourceArray, Int32 sourceIndex, System.Array destinationArray, Int32 destinationIndex, Int32 length) [0x00000] in <filename unknown>:0 
  at kOS.Safe.Screen.ScreenBufferLine.ArrayCopyFrom (System.Char[] source, Int32 sourceStart, Int32 destinationStart, Int32 length) [0x00000] in <filename unknown>:0 
  at kOS.Safe.Screen.ScreenBuffer.PrintLine (System.String textToPrint) [0x00000] in <filename unknown>:0 
  at kOS.Safe.Screen.ScreenBuffer.Print (System.String textToPrint, Boolean addNewLine) [0x00000] in <filename unknown>:0 
  at kOS.Safe.Screen.ScreenBuffer.PrintAt (System.String textToPrint, Int32 row, Int32 column) [0x00000] in <filename unknown>:0 
  at kOS.Screen.Interpreter.PrintAt (System.String textToPrint, Int32 row, Int32 column) [0x00000] in <filename unknown>:0 
  at kOS.Function.FunctionPrintAt.Execute (kOS.SharedObjects shared) [0x00000] in <filename unknown>:0 
  at kOS.Function.FunctionManager.CallFunction (System.String functionName) [0x00000] in <filename unknown>:0 
  at kOS.Execution.CPU.CallBuiltinFunction (System.String functionName) [0x00000] in <filename unknown>:0 
  at kOS.Safe.Compilation.OpcodeCall.Execute (ICpu cpu) [0x00000] in <filename unknown>:0 
  at kOS.Execution.CPU.ExecuteInstruction (kOS.Execution.ProgramContext context) [0x00000] in <filename unknown>:0

Dunbaratu:
This should get rid of the bug.

### DIFF
--- a/src/kOS.Safe/Screen/ScreenBufferLine.cs
+++ b/src/kOS.Safe/Screen/ScreenBufferLine.cs
@@ -59,7 +59,8 @@ namespace kOS.Safe.Screen
 
         /// <summary>
         /// Perform the Array.Copy() function.
-        /// Copies a subrange of values from a source array into this array.
+        /// Copies a subrange of values from a source array into this array,
+        /// with safety checks to truncate the copy length if it's out of range.
         /// <param name="source">copy from here</param>
         /// <param name="sourceStart">starting at this index</param>
         /// <param name="destinationStart">putting it into this index of me</param>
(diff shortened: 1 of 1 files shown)


src/kOS.Safe/Screen/ScreenBufferLine.cs
```diff
@@ -67,13 +68,22 @@ public ScreenBufferLine(int size)
...
+
+            Array.Copy(source, sourceStart, charArray, destinationStart, effectiveLength);
             TouchTime();
         }
 
         /// <summary>
         /// Perform the Array.Copy() function.
-        /// Copies a subrange of values from a source array into this array.
+        /// Copies a subrange of values from a source array into this array,
+        /// with safety checks to truncate the copy length if it's out of range.
```
erendrake: you added this to the summary without actually changing any of the code in the method like you did above? is that correct?

Dunbaratu: It is correct because if you look inside, it makes a call to the other method where I *diid* change it.  I was just documenting it properly for the xml docs so the person calling this method knows it will happen, without caring about the fact that it's actually happening due to what changed in a method call nested inside.  From the "public" viewpoint, the fact that that's why the behavior changed isn't relevant.

erendrake: Sorry i quickly looked at the body and thought it read `Array.Copy` not `ArrayCopyFrom` :P

PR #1218, opened 2015-10-30
Fixes #1208.

### LINKED ISSUE #1208: SHIP:GROUNDSPEED returns NaN instead of zero when parked with brakes - sometimes.
Park a plane on the runway, brakes on.

In a terminal, type:
``print ship:groundspeed.``

Use up-arrow and hit return to repeat the command.
agian.
again.
again.
again.

It seems to give a NaN error about a third or so of the time.  The rest of the time it gives a very small magnitude number near zero, as one would expect for a parked plane.

Dunbaratu:


src/kOS/Suffixed/VesselTarget.cs
```diff
@@ -504,8 +504,17 @@ private double GetHorizontalSrfSpeed()
...
-            return System.Math.Sqrt(Vessel.srfSpeed*Vessel.srfSpeed - Vessel.verticalSpeed*Vessel.verticalSpeed);            
+            //    Since C in our case is the vertical speed we want to remove, we get the following formula:
+
+            double squared2DSpeed = Vessel.srfSpeed*Vessel.srfSpeed - Vessel.verticalSpeed*Vessel.verticalSpeed;
+
+            // Due to floating point roundoff errrors in the KSP API, the above expression can sometimes come
+            // out slightly negative when it should be nearly zero.  (i.e. -0.0000012).  The Sqrt() would
+            // return NaN for such a case, so it needs to be protected from ever going negative like so:
+            if (squared2DSpeed < 0)
+                squared2DSpeed = 0;
```
Dunbaratu: I decided to go with this instead of System.Math.Abs() because a "less than literal zero" test is super fast.  The compiler can optimize it better than it can a method call.

PR #1221, opened 2015-11-02
Kuniverse debug

Dunbaratu:


src/kOS/Suffixed/KUniverseValue.cs
```diff
@@ -117,5 +119,10 @@ public void ForceSetActiveVessel(VesselTarget vesselTarget)
                 FlightGlobals.ForceSetActiveVessel(vessel);
             }
         }
+        
+        public void DebugLog(string message)
+        {
+            SafeHouse.Logger.Log("(KUNIVERSE:DEBUGLOG) " + message);
```
Dunbaratu: The reason I force the prepending of this string is so that it is very clear these are NOT messages coming from our own code, but from the script.

PR #1234, opened 2015-11-06
Queue and Stack data structures

tomekpiotrowski:
@erendrake thought it might be a good idea to add those data structures. I had them lying around because I was experimenting with them in some older code.

### DIFF
--- a/doc/source/structures/misc/stack.rst
+++ b/doc/source/structures/misc/stack.rst
@@ -4,7 +4,7 @@ Stack
 ====
 
 A :struct:`Stack` is a collection of any type in kOS. Stacks work according to `Last-In first-out <https://en.wikipedia.org/wiki/FIFO_and_LIFO_accounting>`_ principle. It may be useful to contrast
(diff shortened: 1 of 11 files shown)


doc/source/structures/misc/stack.rst
```diff
@@ -0,0 +1,126 @@
+.. _stack:
+
+Stack
+====
+
+A :struct:`Stack` is a collection of any type in kOS. Stacks work according to `Last-In first-out <https://en.wikipedia.org/wiki/FIFO_and_LIFO_accounting>`_ principle. It may be useful to contrast
+:struct:`Stack` with :struct:`stack` to better understand how both structures work. You can read more about stacks on `Wikipedia <https://en.wikipedia.org/wiki/Stack_(abstract_data_type)>`_.
```
space-is-hard: >It may be useful to contrast  `Stack` with `stack` to better understand how both structures work 

I assume you meant `Queue` for the second link there?

tomekpiotrowski: Thanks :)

PR #1249, opened 2015-11-15
Engine update

AlchemistCH:
Support for MultiMode engines (RAPIER).
Installed Gimbal suffix into Engine (if the part has the module)

New ENGINE suffixes for end-user:
for all engines:
MULTIMODE - true for multimode engines, false otherwise (read only)
MODES - list of modes for multimode engine (2 strings), otherwise returns a list of 1 string "Single mode"
HASGIMBAL - is there a gimbal module on this part

for multimode only
MODE - name of the current mode (same as in the right-click menu)
TOGGLEMODE - mode toggle, same as the event in right-click menu
PRIMARYMODE - true for primary, false for secondary (get/set) can use set to toggle to specific mode.
AUTOSWITCH - true for automatic, false for manual switching (get/set)

only if has Gimbal:
GIMBAL - this engine's Gimbal


hvacengi:
Sorry for the long delay in getting this reviewed.  I'm going to pull your modifications and get started with a review.  After #1400 gets's merged, I may ask you to update to the most current develop, unless it looks like a pretty clean merge (I don't want to accidentally mess up some of your code while merging a fairly major suffix revision).

Thanks for submitting the PR.

AlchemistCH:
OK, I'll check into this (it's basically changing data type definitions in suffixes, right?).

I was also thinking of one detail: 
here it's using IModuleEngine interface that handles ModuleEngines and ModuleEnginesFX, in other places (like the steering controller) it's all just handled as ModuleEngines
Is this just legacy of the time when inheriting one partModule from another one didn't work properly? (When they first added these modules, first experiment I tried to do was inheriting from ModuleEngines - it failed with a ton of null references. I also kept running in a ton of bugs with config nodes at the time, resulting in me never properly updating the plugin I had back then to the new system... and most of its functionality got eaten by stock anyway)

hvacengi:
Squad updated the ModuleEnginesFX class so that it inherits from ModuleEngines in v1.0 iirc.  Or at least, that's when I found the change because I was working on the engine thrust changes.  So, yes you should be able to only check for `ModuleEngines` classes.  When I found the issue before, I chose not to make the change because it would increase the scope of my edit at the time, and frankly I've never come back to it to evaluate.

When I looked at it last night, I thought I might be able to clean up some of the methods too.  I'm not sure if `Part.Modules.GetModules<ModuleEngines>()` will return the instances of `ModuleEnginesFX`, but if it does that would have potential for cleaning things up.

I kind of thought that it was only going to look like a modification of some type definitions, but I'm always wary of large merges ;-)

I'll have some time later tonight to look at the PR itself.  I'll either comment or submit you a PR with any changes I see.  If you want to mess around with it before I get to it, that would be cool too.

Issue #1251, opened 2015-11-15
LIST ENGINES (PrintList.cs) - what exactly it's supposed to be?


AlchemistCH:
I tried to analyze GetEngineList function in PrintList.cs and it seems to be quite a mess, to the point it probably doesn't match its original intention.

Let's compare:
**LIST ENGINES IN variable.** (mostly done by PartsToList in EngineValue.cs) 
- current release: creates a list of ENGINE (EngineValue) consisting of every engine module found on the vessel (supposedly to list every engine part as EngineValue, but has issues with Rapiers due to multiple modes)
- my proposed update ( #1249 ): creates a list of ENGINE (EngineValue) consisting of every engine part found on the vessel (first checking the parts for MultiModeEngine, if none, then for usual engine module - now Rapier also handled properly)

**LIST ENGINES.** (mostly done by GetEngineList in PrintList.cs)
- checks only already active engines (why?)
- creates an instance per engine module (double listed Rapiers)
- prints part UID, stage, and name of the module (which is practically the same for everything! Shouldn't name of the part be more useful?)

Should it be changed to print all the engines (not only the active ones) and give part names, instead of module name?
My proposition in  #1252

PR #1252, opened 2015-11-15
Update for LIST ENGINES

AlchemistCH:
My proposition for #1251

### DIFF
--- a/src/kOS/Function/PrintList.cs
+++ b/src/kOS/Function/PrintList.cs
@@ -231,21 +231,30 @@ namespace kOS.Function
             list.AddColumn("Stage", 8, ColumnAlignment.Left);
             list.AddColumn("Name", 28, ColumnAlignment.Left);
 
-            foreach (Part part in VesselUtils.GetListOfActivatedEngines(shared.Vessel))
+            foreach (Part part in shared.Vessel.Parts)
             {
+                bool enginefound = false;
                 foreach (PartModule module in part.Modules)
                 {
-                    if (module == null) continue;
-
-                    var engines = module as ModuleEngines;
-                    if (engines != null)
+                    if (!enginefound)
                     {
-                        list.AddItem(part.uid(), part.inverseStage, engines.moduleName);
+                        var enginesMM = module as MultiModeEngine;
+                        if (enginesMM != null) { enginefound = true; }
+                        else
+                        {
+                            var engines = module as ModuleEngines;
+                            if (engines != null) { enginefound = true; }
+                            else
+                            {
+                                var enginesFX = module as ModuleEnginesFX;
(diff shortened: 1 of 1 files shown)


Dunbaratu:
@hvacengi, Is this a candidate for 0.19.0, since it's in testing phase?

hvacengi:
Yeah, it's a candidate.  I've been thinking about the implications of it.  Obviously the list needs to be updated to properly handle the `MultiModeEngine`.  But I don't know why it previously only listed active engines, so I don't really have the other half of the argument to weigh.  

I've honestly never really understood why the two methods (`list [parts] in [list var]` and `list [parts]` didn't actually use the same method, so I'd be tempted to modify it so that it uses the same part factory method.  That would mean only one location in the code to maintain the list generation, which would prevent it from getting out of sync in the future.

AlchemistCH:
Yeah, maybe calling the same list generation and then just displaying the list by these functions would be better.
Ah, the amount of frustration trying to to fix one function while unknowingly calling another one for tests. And then finally finding this mess.

hvacengi:
Does that mean you can update it to call that method instead, or should I make the edit and submit a PR to you @AlchemistCH?

AlchemistCH:
OK, I'll deal with this file.

AlchemistCH:
Here, made LIST ENGINES function to use the one from EngineValue. 
(Is it OK to take contents of ListValue like this? Seems perfectly functional)

Was also considering doing the same to LIST SENSORS, but that would require changing some access levels in SensorValue. Besides that's way simpler function.

src/kOS/Function/PrintList.cs
```diff
@@ -276,7 +273,7 @@ private kList GetConfigList()
             list.AddColumn("Name", 34, ColumnAlignment.Left);
             list.AddColumn("Value", 6, ColumnAlignment.Left);
 
-            foreach (ConfigKey key in Config.Instance.GetConfigKeys())
+            foreach (ConfigKey key in SafeHouse.Config.GetConfigKeys())
```
hvacengi: The interface exposed by `SafeHouse.Config` does not include `GetConfigKeys()`.  I'm going to revert this back to `Config.Instance` in the merge.

AlchemistCH: Ok, I've put it back. Shouldn't have copied over the entire file when the difference on my side was a forgotten 'uses' reference.

hvacengi:
I just wanted to share an example of my testing of the PR:
screenshot55

hvacengi:
Apparently I fail at merging.  I merged this into my develop branch and pushed, and the commits are now shown in the history of develop... but the merge did not show up here.

I'll try to figure that out.

PR #1258, opened 2015-11-18
Update changelog for 0.18.2

hvacengi:


CHANGELOG.md
```diff
@@ -1,6 +1,32 @@
...
+* New suffixes for interacting with kOS Processor modules (including `core`).
+* `debuglog(...)` function to print directly to the KSP log file.
+* New `queue` and `stack` data structures.
+
+### BUG FIXES
+* The processor's mode (on/off/starved) is now saved and restored ( https://github.com/KSP-KOS/KOS/issues/1172 )
+* Fixed stage resources again to address a change in KSP 1.0.5 ( https://github.com/KSP-KOS/KOS/issues/1242 )
+* Fix occasional instances of flight controls getting disabled during a docking/undocking/staging event ( https://github.com/KSP-KOS/KOS/issues/1205 )
+* kOS can now trigger module events with RemoteTech installed and no KSC connection ( https://github.com/RemoteTechnologiesGroup/RemoteTech/issues/437 )
+* Volumes and processors are now mapped together ( https://github.com/KSP-KOS/KOS/issues/788 )
```
tomekpiotrowski: I think this is already covered by 'Volumes and processors have better integration.' Even if we want to have a separate bullet point for this then I would suggest moving it to features.

CHANGELOG.md
```diff
@@ -1,6 +1,32 @@
...
+* You can now switch to a volume using the volume object itself ( https://github.com/KSP-KOS/KOS/issues/789 )
```
tomekpiotrowski: Same here, I think this is a feature rather than a bug.

CHANGELOG.md
```diff
@@ -1,6 +1,32 @@
...
+------------------------------
+
+### BREAKING CHANGES
+* As usual, you MUST recompile all KSM files before running them on the new version.  Some of the changes have altered how the VM works.
+* Nothing else... we hope.
+
+### NEW FEATURES
+* Compatibility with KSP version 1.0.5
+* `run once ...` syntax to run a script only once per session.
+* Volumes and processors have better integration.
```
tomekpiotrowski: Maybe we could mention `LIST PROCESSORS` and `PROCESSOR()` explicitly here?

hvacengi: Once @Dunbaratu has the documentation pages generated, I'm going to link to your documentation pages.  Do you think we need to be more explicit if we have that link?

CHANGELOG.md
hvacengi: Yeah, I guess this could be removed.  It was tagged as an enhancement instead of a bug, I just always thought of it as a bug in my mind.

PR #1269, opened 2015-11-24
Serialization

### LINKED ISSUE #1168: Key-value store

tomekpiotrowski:
Fixes #1168

I will write docs and do some extra testing in a day or two, but generally this is ready for review.

This PR adds a serialization/deserialization mechanism as well as `WRITEFILE` and `READFILE` functions available to the user.

Couple of comments:
* I use SimpleJSON (https://github.com/facebook-csharp-sdk/simple-json). I had problems with other libs - JSON.net and FastJsonParser turned out to be incompatible with Unity. Someone on KSP forums recommended SimpleJSON so I went with that. Works well. Only drawback I've noticed is that it does not indent generated files. I don't think it's a big deal. My Monodevelop has added the full source code to the project even though I added it as a Nuget dependency. I guess that is supposed to work this way?
* The main entrypoints for serialization are `SerializationMgr#Serialize()` and `SerializationMgr#Deserialize()` 
* Classes that implement `IDumper` can be serialized (they are expected to have a default constructor). Objects are first dumped to a `Dictionary`, which is later serialized by a `Formatter`. This allows us to easily add formatters. Currently we have `JSONFormatter` and `TerminalFormatter`.
* Types currently handled by the serializer:
  * primitives, strings
  * lexicons, lists, stacks, queues
  * `BodyTarget`
  * `VesselTarget`
  * `Vector`
  * `GeoCoordinates`
  * `OrbitInfo`
  * `TimeSpan`
* Type information is stored in `$type` field. Obviously we need the type name to create an instance of the class while deserializing.
* `Lexicon` in kOS can have any kind of object as a key. In JSON only strings can be keys, so lexicons with funky keys would not survive serialization/deserialization. Actually our `Lexicon` used to be a parameterized type, so it was not deserializable in a general case. I have dropped the type parameters as we were only using `<object, object>` in the project anyway. Keeping `Lexicon` parameterized would require introducing some other class for serialization purposes anyways.

Example:
```
set l to lexicon().
set nested to lexicon().
set nestedList to list().
set nestedStack to stack().

l:add("key1", "value1").
l:add("key2", nested).
l:add("key3", nestedList).
l:add("key4", nestedStack).

nested:add("nestedkey1", "nestedvalue1").

nestedList:add("one").
nestedList:add(2).

nestedStack:push(1).
nestedStack:push(2).

writefile(l, "lexicon.json").
```

results in (indented by hand):

```
{
  "key1":"value1",
  "key2":
  {
    "nestedkey1":"nestedvalue1",
    "$type":"kOS.Safe.Encapsulation.Lexicon"
  },
  "key3":
  {
    "0":"one",
    "1":2,
    "$type":"kOS.Safe.Encapsulation.ListValue"
  },
  "key4":
  {
    "0":1,
    "1":2,
    "$type":"kOS.Safe.Encapsulation.StackValue"
  },
  "$type":"kOS.Safe.Encapsulation.Lexicon"
}
```

This can be read and printed:

```
set read to readfile("lexicon.json").
print read.
```


Dunbaratu:
How possible is it to take a JSON serialization and encode it in a ConfigNode?  The reason I ask is that down the road in the future we will need to store queue information for inter-vessel comms in the saved game file (so you can send a message to a distant on-rails vessels which doesn't arrive there instantaneously), and the persistent.sfs file that stores the saved game is essentially one big ConfigNode file.  If you can't fit a piece of game-saved data in a ConfigNode, you need a very ugly messy way to store it alongside the saved game file in a separate file.  I've seen mods that try this and it's always fraught with problems because sometimes the two files are out of sync.

tomekpiotrowski:
I've been thinking about that specific use case when implementing this. Basically I think we would only need to implement `ConfigNodeFormatter` and any implementations of `IDumper` (and primitive types of course) could be stored in such a queue. I haven't actually tested this so I'm not 100% sure if it will be that easy.

hvacengi:
Worst case scenario, you could hex encode the byte array of the string in the config node (using 0-9, a-f).  It consumes more space, but it should be a simple matter of writing a method to create the hex string from a byte/char array and then to read the hex string back into the byte/char array.  I think I may have one of those somewhere in my old projects.  Don't we already treat strings as Unicode internally?  

I guess I'd have to look at that again.  My implementation was to encapsulate data in a QR code using a serial bus (which didn't like it if you used anything that wasn't a number, letter, or comma.

Dunbaratu:
True we could do the Base64 encoding technique - we in fact already do that for storing the arbitrary text file content of the local volumes (which are stored on individual vessel instances and therefore have to be encoded in the persistent.sfs savegame file).    The technique used is to run them through a compression followed by a Base64 encoding.

tomekpiotrowski:
Maybe we could wait a bit before merging this. While working on inter-vessel stuff I actually made some improvements to how serialization works. I will try to push a commit this weekend.

erendrake:
if you think this is useful and works we can always take an improvement in another PR if you want. The reason i ask is i am hoping to have time to merge stuff this weekend

tomekpiotrowski:
Ok, I've pushed what I wanted to push. This should be ready for review.

src/kOS/Serialization/SerializationMgr.cs
```diff
@@ -0,0 +1,48 @@
...
+            this.sharedObjects = sharedObjects;
+        }
+
+        public override IDumper CreateInstance(string typeFullName, IDictionary<object, object> data)
+        {
+            var deserializedType = Type.GetType(typeFullName);
+
+            if (deserializedType == null)
+            {
+                SafeHouse.Logger.Log("#### " + typeFullName + ", " + typeof(SafeSerializationMgr).Assembly.FullName);
```
tomekpiotrowski: That line shouldn't have been pushed :/

src/kOS.Safe/Serialization/Formatter.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace kOS.Safe.Serialization
+{
+    public interface Formatter
```
erendrake: Interface should begin with an `I` eg `IFormatter`

src/kOS.Safe/Serialization/Formatter.cs
```diff
@@ -0,0 +1,12 @@
...
+    {
+        string Write(IDictionary<object, object> value);
+        IDictionary<object, object> Read(string input);
```
erendrake: we only have 2 implementations of this interface and we already have segmentation in that `TerminalFormatter` cant read. we should split the interface with a reader and a writer

src/kOS.Safe/Exceptions/KOSSerializationException.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
```
erendrake: What is the extra benefit of having this derived class if we arent doing anything with it?

src/kOS.Safe/Serialization/JSONFormatter.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        private object MakeStringDictionaries(object value)
+        {
+            if (value is IDictionary<object, object>)
+            {
+                Dictionary<string, object> stringKeys = new Dictionary<string, object>();
+
+                foreach (KeyValuePair<object, object> entry in value as IDictionary<object, object>)
+                {
+                    stringKeys[entry.Key.ToString()] = MakeStringDictionaries(entry.Value);
```
erendrake: recursive call with no protection gives me the willies

src/kOS.Safe/Serialization/JSONFormatter.cs
```diff
@@ -0,0 +1,72 @@
...
+            
+        private object Unwrap(object read)
+        {
+            if (read is IDictionary<string, object>)
+            {
+                Dictionary<object, object> result = new Dictionary<object, object>();
+
+                foreach (KeyValuePair<string, object> entry in read as IDictionary<string, object>)
+                {
+                    result[entry.Key] = Unwrap(entry.Value);
```
erendrake: recursive call with no protection gives me the willies

src/kOS.Safe/Serialization/TerminalFormatter.cs
```diff
@@ -0,0 +1,82 @@
...
+                if (value is IDictionary<object, object>)
+                {
+                    string header = Environment.NewLine;
+
+                    if (value is DictionaryWithHeader)
+                    {
+                        header = (value as DictionaryWithHeader).Header + Environment.NewLine;
+                    }
+
+                    valueString = header + WriteIndented(value as IDictionary<object, object>, level + 1);
```
erendrake: recursive call with no protection gives me the willies

src/kOS/Serialization/IDumperWithSharedObjects.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using kOS.Safe.Encapsulation;
+
+namespace kOS.Serialization
+{
+    public interface IDumperWithSharedObjects : IDumper
+    {
+        void SetSharedObjects(SharedObjects sharedObjects);
```
erendrake: if we are exposing currently private properties/fields of implementing classes. I would prefer turning this into a property setter requirement.

src/kOS/Suffixed/OrbitInfo.cs
```diff
@@ -113,5 +127,44 @@ public override string ToString()
...
+            return dump;
+        }
+
+        public void LoadDump(System.Collections.Generic.IDictionary<object, object> dump)
+        {
+            name = "<unnamed>";
+
+            double inclination = Convert.ToDouble(dump[DUMP_INCLINATION]);
+            double eccentricity = Convert.ToDouble(dump[DUMP_ECCENTRICITY]);
+            double semi_major_axis = Convert.ToDouble(dump[DUMP_SEMI_MAJOR_AXIS]);
```
erendrake: snake case! eww

src/kOS.Safe.Test/Serialization/TerminalFormatterTest.cs
```diff
@@ -0,0 +1,40 @@
...
+
+            lex["key1"] = "value1";
+            lex["key2"] = 1;
+            lex["key3"] = nested;
+
+            nested["nested1"] = "nested1value";
+            nested["nested2"] = "nested2value";
+
+            Assert.AreEqual("LEXICON of 3 items:\n[\"key1\"]= value1\n[\"key2\"]= 1\n[\"key3\"]=" +
+                " LEXICON of 2 items:\n  [\"nested1\"]= nested1value\n  [\"nested2\"]= nested2value", Serialize(lex));
```
erendrake: this test depends on \n as the newline character. this is contrasted with the `Environment.NewLine` that we use in code to determine newline based on OS. So this test fails on windows.

Dunbaratu: Hang on a second here.  Where is the standard that says we'll use Environment.Newline everywhere?  Because we certainly don't.  I know many places that hardcode '\n' in our code,  One of them is the PRINT statement printing to the terminal - it assumes that if it sees a '\n' you will get a \r\n effect even if the \r is missing.

erendrake: I am sorry if i was unclear. The test fails on windows because of a newline character mismatch. I dont honestly care which we pick i would just like the test to not depend on OS in order to pass.

Dunbaratu: I'm really not in favor of the serialization file coming out different depending on what OS you produced it on, and being only able to read in files made by the same OS.  It's going to eventually be part of the save state of a game when it is used for comms (at least that's our plan), and as such should be copyable between different users on different machines.  ("Here's my bug report, with a copy of my .craft and my saved game folder....") About the only text reading or JSON reading tool out there that can't handle the files either way around is Windows' crappy Notepad.  Everything else can deal with text files written by "foreign" OS'es.

tomekpiotrowski: Environment.NewLine is only used when printing to terminal. JSON files actually do not contain any new line characters - SimpleJSON doesn't indent the output.

Dunbaratu: If you mean the in-game GUI terminal, it' written to expect just '\n' without the '\r' and be okay with it.

PR #1290, opened 2015-12-01
Cpu safe

erendrake:
After all the work building up kOS.Safe i finally was able to move the goal of the whole project. CPU.cs

there is more to do because i added a new config that needs to get documented and i havent even tried it in game yet.

### DIFF
--- a/doc/source/structures/misc/config.rst
+++ b/doc/source/structures/misc/config.rst
@@ -65,6 +65,10 @@ Configuration of kOS
           - boolean
           - True
           - Force the telnet server to use loopback (127.0.0.1) address
+        * - :attr:`DEBUGEACHOPCODE`
+          - boolean
+          - false
+          - Unholy debug spam used by the kOS developers
 
 .. attribute:: Config:IPU
 
@@ -208,4 +212,18 @@ Configuration of kOS
 
     To make the change take effect you may have to
     stop, then restart the telnet server, as described above.
+
+.. attribute:: Config:DEBUGEACHOPCODE
+
+    :access: Get/Set
+    :type: boolean
+
+    Configures the ``DebugEachOpcode`` setting.
+
+    If true, each opcode that is executed by the CPU will be accompanied by 
+    an entry in the KSP log. This is a debugging tool for those who are very 
+    familiar with the inner workings of kOS and should rarely be used outside
+    the kOS dev team
+
+    This change takes effect immediately 
 
--- a/doc/source/structures/misc/config.rst
+++ b/doc/source/structures/misc/config.rst
@@ -220,10 +220,12 @@ Configuration of kOS
 
     Configures the ``DebugEachOpcode`` setting.
 
+    NOTE: This makes the game VERY slow, use with caution.
+
     If true, each opcode that is executed by the CPU will be accompanied by 
     an entry in the KSP log. This is a debugging tool for those who are very 
     familiar with the inner workings of kOS and should rarely be used outside
-    the kOS dev team
+    the kOS dev team.
 
-    This change takes effect immediately 
+    This change takes effect immediately.
 
--- a/src/kOS.Safe/Execution/CPU.cs
+++ b/src/kOS.Safe/Execution/CPU.cs
@@ -1,16 +1,15 @@
-﻿using System;
+﻿using kOS.Safe.Binding;
+using kOS.Safe.Compilation;
+using kOS.Safe.Exceptions;
+using kOS.Safe.Utilities;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Text;
-using kOS.Safe.Binding;
-using kOS.Safe.Compilation;
-using kOS.Safe.Exceptions;
-using kOS.Safe.Execution;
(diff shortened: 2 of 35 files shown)


doc/source/structures/misc/config.rst
```diff
@@ -209,3 +213,17 @@ Configuration of kOS
...
+
+    :access: Get/Set
+    :type: boolean
+
+    Configures the ``DebugEachOpcode`` setting.
+
+    If true, each opcode that is executed by the CPU will be accompanied by 
+    an entry in the KSP log. This is a debugging tool for those who are very 
+    familiar with the inner workings of kOS and should rarely be used outside
+    the kOS dev team
```
Dunbaratu: I'd also be sure to mention to people that it greatly lags the game by a lot, just to help discourage it's use.

erendrake: darn, i meant to do that. Thanks!

erendrake: done

src/kOS.Safe/Execution/CPU.cs
```diff
@@ -704,16 +695,17 @@ public object GetValue(object testValue, bool barewordOkay = false)
...
             // flags used internally:
-            if (testValue is string &&
-                ((string)testValue).Length > 1 &&
-                ((string)testValue)[0] == '$' &&
-                ((string)testValue)[1] != '<')
+            var identifier = testValue as string;
+            if (identifier == null ||
+                identifier.Length <= 1 ||
+                identifier[0] != '$' ||
+                identifier[1] == '<')
```
Dunbaratu: Presumably this will be a merge conflict with what @hvacengi is doing.

src/kOS.Safe/Execution/CPU.cs
```diff
@@ -704,16 +695,17 @@ public object GetValue(object testValue, bool barewordOkay = false)
             // If it's a variable, meaning it starts with "$" but
             // does NOT have a value like $<.....>, which are special
             // flags used internally:
```
Dunbaratu: Given that you flipped the meaning of the 'if' condition below to mean the opposite, that makes this comment misleading now because it's describing the old if condition and you didn't edit it to match.

But this all may be moot as I expect @hvacengi gutted this whole check in favor of something that doesn't need to examine the string prefixes anyway.

src/kOS.Safe/Execution/CPU.cs
```diff
@@ -884,18 +876,18 @@ public void EndWait()
...
         {
-            bool showStatistics = Config.Instance.ShowStatistics;
+            bool showStatistics = SafeHouse.Config.ShowStatistics;
             Stopwatch updateWatch = null;
             Stopwatch triggerWatch = null;
             Stopwatch executionWatch = null;
-            double triggerElapsed = 0.0;
-            double executionElapsed = 0.0;
+            var triggerElapsed = 0.0;
+            var executionElapsed = 0.0;
```
Dunbaratu: Man I really hate how the automated tool you run does this.

src/kOS.Safe/Execution/Stack.cs
```diff
@@ -61,13 +60,14 @@ private object ProcessItem(object item)
...
 
-                if (!Double.IsNaN((double)item) && outOfBounds)
+                if (!double.IsNaN(doubleItem) && outOfBounds)
                 {
-                    int intPart = Convert.ToInt32(item);
-                    if ((double)item == intPart)
-                        item = intPart;
+                    int intItem = Convert.ToInt32(item);
+                    if (doubleItem == intItem)
+                        item = intItem;
```
Dunbaratu: This has nothing to do with your change, but I suspect this naming of variables was wrong all along.

I had a hard time following the logic here just now until I realized that the way it's written, `outOfBouds` is *actually* a boolean that means `isInBounds` - exactly the opposite of what it says.

src/kOS.Safe/Execution/CPU.cs
hvacengi: so far I'm keeping my hands off the cpu class.  I'm planning on merging this PR with mine after I've tested it out, so I'll handle any merges that conflict like this.  The way I have it currently implemented however, this change shouldn't be a problem.  Because I'm using the new methods directly from compile time, only system values will cast as a string (even though `StringValue` has an implicit conversion to string, this is not taken into account by the `as` operator, or so the documentation says).

src/kOS.Safe/Execution/CPU.cs
hvacengi: I expect that yes, the prefix check will be obsolete.  I noted it as such with a TODO in my first run through, but as noted above the CPU changes have been put on hold so I didn't commit those changes.

src/kOS.Safe/Execution/Stack.cs
hvacengi: I caught that too, if you look at the implementation in `ScalarValue.Create` I changed the name to `inBounds`, I did not refactor the method the same way that @erendrake did however.  I'll plan on copying this syntax to my implementation for consistency sake, but it will probably end up deleted from the stack when my PR is ready to merge.

Issue #1285, opened 2015-11-30
Open brace and open paren halt execution in strings


gisikw:
Using an unclosed paren or brace within a string expression causes execution to freeze.

```
print "(". // No output
1/0.       // Still no output
).         // Now, it crashes
```

Same with the open brace, and when used as part of an expression (not simply a print issue)

```
set x to "{". // No output
print x.      // No output
print "}".    // Outputs {, then }
```

## Workaround
Using the terminating paren/brace as a no-op expression gets the parser to behave correctly, i.e.
```
set x to "BODY(Mun)":split("(")[0]. ")".
print x. // "BODY"

print "{". "}". // Prints "{"
```

Dunbaratu:
I'm surprised it doesn't barf and instead hangs.  If it was a problem, I'd expect it to fail utterly, not get stuck, since this:
```
       (".
1/0.
)
```
isn't a valid thing to be printing.  It must be coming from somewhere other than the actual parser.

Dunbaratu:
Confirmed.  I was able to reproduce the problem.  Weird.  Looks like I need to fire up TinyPG and carefuly examine the parse tree it produces on this input syntax.

Dunbaratu:
Interestingly, if you do this:
```
print "(".
print "stuff".
print ")".
```
Then it *does* print correctly like so:
```
(
stuff
)
```
However, it doesn't print any of that out until *after* it encounters the lastmost line.  If you type those lines in one at a time interactively, you end up with this:
```
print "(".
print "stuff".
print ")".
(
stuff
)
```
When since it was interactive, it shouldn't have waited until the last line .  IT should have been printing the strings out immediately like this:
```
print "(".
(
print "stuff".
stuff
print ")".
)
```

Dunbaratu:
Verified that the TinyPG parse tree is actually correct here.  It is separating the 3 print statements as 3 different statements, and isn't parsing the stuff between parens as a unit.  So.... whatever it is is broken *after* the parse... I suspect because it *does* perform the code correctly eventually, but too late, that it's got something to do with the system that allows you to continue an incompleted command on the next line in the interpreter rather than calling it an error.

Dunbaratu:
Problem analysis:

[...]

PR #1291, opened 2015-12-02
Fixes 1285 better detection of incomplete command

Dunbaratu:
Now allows any of the following to be typed on the interpreter command line, with the interpreter being able to detect that a command is in need of continuing rather than just spitting an error:

```
if x = 2 {
   print "two".
}

print ((6 / 2)
  + 5 ).

print "this string
spans several
lines before closing".
```

This problem, that it didn't do this sort of thing right before, was at the heart of what was wrong with #1285 - its 'dumb' string scanner was seeing the parenthesis in the string as a parenthesis in need of continuation and wasn't smart enough to realize it doesn't count when it's inside a string literal like that.

### DIFF
--- a/src/kOS.Safe/Compilation/KS/KSScript.cs
+++ b/src/kOS.Safe/Compilation/KS/KSScript.cs
@@ -95,40 +95,25 @@ namespace kOS.Safe.Compilation.KS
 
         public override bool IsCommandComplete(string command)
         {
-            // Good FUture Refactor Opportunity:
-            //
-            // It's not good that this is
-            // essentially doing kerboscript-aware syntax thinking
-            // somewhere outside the parser generator TinyPG.  That
-            // puts the same logic in two places of the code, which
-            // is usually a bad thing.  If someone can think of a
-            // better way to query the parser to ask it "is this
-            // statement incomplete?" rather than putting this logic
-            // here, that would be a good opportunity for a refactor.
-            //
-            // The difficulty is that the parser returns "syntax erorr"
-            // the same regardless of whether it's because it's incomplete
-            // or because there's an error.  So we can't say "if parser
-            // doesn't think it's done" wihtout the side effect of it hanging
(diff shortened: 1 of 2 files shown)


src/kOS/Screen/Interpreter.cs
```diff
@@ -64,6 +64,10 @@ public override bool SpecialKey(char key)
             if (key == (char)UnicodeCommand.BREAK)
             {
                 Shared.Cpu.BreakExecution(true);
+                LineBuilder.Remove(0, LineBuilder.Length); // why isn't there a StringBuilder.Clear()?
```
erendrake: I was confused because there is one. then i realized it isnt available in our ancient version of c#

erendrake:
Is this ready? i dont see your customary battery of tests?

Dunbaratu:
I couldn't give a bunch of test scripts for this one because it's a problem that occurs entirely in the interactive interpreter.

src/kOS.Safe/Compilation/KS/KSScript.cs
```diff
@@ -97,35 +97,79 @@ public override bool IsCommandComplete(string command)
...
+            // TinyPG C# source code and changing the way it encodes a ParseError
+            // so it stores that sort of thing as separate pieces of data in its members.
+            
+            if (!inQuotes)
+            {
+                ParseTree parseTree = parser.Parse(command);
+                
+                foreach (ParseError err in parseTree.Errors)
+                {
+                    System.Console.WriteLine("eraseme:  err msg: " + err.Message);
```
Dunbaratu: aahhh damn.
I forgot to remove this!!

Dunbaratu: I should make a small bash script to wrap around my 'git push' es  that greps for my 'eraseme' string throughout all the diffs and warns me if there are any present, asking for confirmation before actually pushing the change.

PR #1292, opened 2015-12-03
Add ScalarValue structure

hvacengi:
ScalarValue.cs
* New class to encapsulate number types (int, double, float)
* Class will implicitly convert to both int and double types
* Implements IConvertible to enable Convert.ToDouble and the like to work
* Provide access to the int, double, and validity information
* ISVALID provides a way for users to disable number safe mode and monitor safety of numbers themselves.
* Implement applicable operators for manipulation and equality testing
kOS.Safe.csproj
* Add reference to ScalarValue.cs

### DIFF
--- a/src/kOS.Safe/Encapsulation/ScalarValue.cs
+++ b/src/kOS.Safe/Encapsulation/ScalarValue.cs
@@ -4,43 +4,34 @@ using kOS.Safe.Exceptions;
 
 namespace kOS.Safe.Encapsulation
 {
-    public class ScalarValue : Structure, IConvertible
+    abstract public class ScalarValue : Structure, IConvertible
     {
-        private int? internalInt;
-        private double? internalDouble;
+        protected object internalValue;
 
-        public bool IsInt { get { return internalInt.HasValue; } }
+        abstract public bool IsInt { get; }
 
-        public bool IsDouble { get { return internalDouble.HasValue; } }
+        abstract public bool IsDouble { get; }
 
         public bool IsValid
         {
             get
             {
-                if (internalInt.HasValue) return true;
-                else if (internalDouble.HasValue)
+                if (IsInt) return true;
+                if (IsDouble)
                 {
-                    if (double.IsInfinity(internalDouble.Value) || double.IsNaN(internalDouble.Value)) return false;
+                    double d = GetDoubleValue();
+                    if (double.IsInfinity(d) || double.IsNaN(d)) return false;
                     return true;
                 }
                 return false;
             }
         }
 
-        public object Value
-        {
-            get
-            {
-                if (IsInt) return internalInt.Value;
-                if (IsDouble) return internalDouble.Value;
-                throw new kOS.Safe.Exceptions.KOSException("Scalar value contains no double or int value");
-            }
-        }
+        public object Value { get { return internalValue; } }
 
-        public ScalarValue(object value)
-            : base()
+        protected ScalarValue()
         {
-            SetValue(value);
+            InitializeSuffixes();
         }
 
         public void InitializeSuffixes()
@@ -50,7 +41,7 @@ namespace kOS.Safe.Encapsulation
             AddSuffix("ISVALID", new Suffix<bool>(() => IsValid));
         }
 
-        protected void SetValue(object value)
+        public static ScalarValue Create(object value)
         {
             if (value is float)
                 value = Convert.ToDouble(value);
@@ -59,26 +50,23 @@ namespace kOS.Safe.Encapsulation
                 bool inBounds = Int32.MinValue < (double)value && (double)value < Int32.MaxValue;
                 if (inBounds && !double.IsNaN((double)value))
                 {
-                    // I still don't quite get what this part is doing... I'm thinking of rewriting it using modulous instead
+                    // Convert the double to an int, and check and see if they are still equal.
+                    // if so, treat the double as if it was an int.
                     int intPart = Convert.ToInt32(value);
                     if ((double)value == intPart)
                     {
-                        internalInt = intPart;
-                        internalDouble = null;
-                        return;
+                        return new ScalarIntValue(intPart);
                     }
                 }
-                internalDouble = (double)value;
-                internalInt = null;
+                return new ScalarDoubleValue((double)value);
             }
             else if (value is int)
             {
-                internalInt = (int)value;
-                internalDouble = null;
+                return new ScalarIntValue((int)value);
             }
             else if (value is ScalarValue)
             {
-                SetValue(((ScalarValue)value).Value);
+                return ScalarValue.Create(((ScalarValue)value).Value);
             }
             else
             {
@@ -88,21 +76,12 @@ namespace kOS.Safe.Encapsulation
 
         public int GetIntValue()
         {
-            if (internalInt.HasValue)
-            {
-                return internalInt.Value;
-            }
-            else
-            {
-                return Convert.ToInt32(internalDouble);
-            }
+            return Convert.ToInt32(Value);
         }
 
         public double GetDoubleValue()
         {
-            if (IsDouble) return internalDouble.Value;
-            if (IsInt) return Convert.ToDouble(internalInt.Value);
-            throw new kOS.Safe.Exceptions.KOSException("Scalar value contains no double or int value");
+            return Convert.ToDouble(Value);
         }
 
         public override string ToString()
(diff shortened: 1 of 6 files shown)


src/kOS.Safe/Encapsulation/ScalarValue.cs
```diff
@@ -0,0 +1,456 @@
...
+        public object Value { get { return internalValue; } }
+
+        protected ScalarValue()
+        {
+            InitializeSuffixes();
+        }
+
+        public void InitializeSuffixes()
+        {
+            AddSuffix("ISINT", new Suffix<bool>(() => IsInt));
```
erendrake: Did we want to remove these suffixes for the time being?

src/kOS.Safe/Encapsulation/ScalarValue.cs
```diff
@@ -0,0 +1,456 @@
...
+                }
+                return new ScalarDoubleValue((double)value);
+            }
+            else if (value is int)
+            {
+                return new ScalarIntValue((int)value);
+            }
+            else if (value is ScalarValue)
+            {
+                return ScalarValue.Create(((ScalarValue)value).Value);
```
erendrake: I do not understand the point of this recursive call. How would you get out of it?

src/kOS.Safe/Encapsulation/ScalarValue.cs
```diff
@@ -0,0 +1,456 @@
...
+        public override string ToString()
+        {
+            if (IsInt) return GetIntValue().ToString();
+            else if (IsDouble) return GetDoubleValue().ToString();
+            return "NaN";
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (obj == null) return false;
```
erendrake: a `null` check then an `is` then an `as` cast? this can be simplified.

src/kOS.Safe/Encapsulation/ScalarValue.cs
```diff
@@ -0,0 +1,456 @@
...
+                value = Convert.ToDouble(value);
+            if (value is double)
+            {
+                bool inBounds = Int32.MinValue < (double)value && (double)value < Int32.MaxValue;
+                if (inBounds && !double.IsNaN((double)value))
+                {
+                    // Convert the double to an int, and check and see if they are still equal.
+                    // if so, treat the double as if it was an int.
+                    int intPart = Convert.ToInt32(value);
+                    if ((double)value == intPart)
```
erendrake: the chance of this being true will be vanishingly small once users start doing any math operations on it.

src/kOS.Safe/Encapsulation/ScalarValue.cs
```diff
@@ -0,0 +1,456 @@
...
+            if (obj is ScalarValue)
+            {
+                ScalarValue val = obj as ScalarValue;
+                if (val != null)
+                {
+                    if (this.IsInt && val.IsInt)
+                    {
+                        return this.GetIntValue() == val.GetIntValue();
+                    }
+                    return this.GetDoubleValue() == val.GetDoubleValue();
```
erendrake: the chance of this being true will be vanishingly small once users start doing any math operations on it.

src/kOS.Safe/Encapsulation/ScalarValue.cs
```diff
@@ -0,0 +1,456 @@
...
+        }
+
+        public static ScalarValue operator ^(ScalarValue val1, ScalarValue val2)
+        {
+            return Power(val1, val2);
+        }
+
+        public static bool operator ==(ScalarValue val1, ScalarValue val2)
+        {
+            return val1.Equals(val2);
```
erendrake: Potential NullReferenceException

src/kOS.Safe/Encapsulation/ScalarValue.cs
```diff
@@ -0,0 +1,456 @@
...
+        }
+
+        public static bool operator ==(ScalarValue val1, object val2)
+        {
+            return val1.Equals(val2);
```
erendrake: Potential NullReferenceException

src/kOS.Safe/Encapsulation/ScalarValue.cs
Dunbaratu: True but the only way to fix that is to decide FOR the user what the epsilon will be where things get called identical.  That's not something we should be doing.  Either it's exactly equal or it's not.  Just like native doubles.  There are *some* native doubles where it does work exactly - basically integers small enough to fit in the floating point mantissa work (i.e. 0, 1, 2, 3,...) , and that's quite a few.  You have to start getting pretty large before a floating point representation won't store an integer properly.

src/kOS.Safe/Encapsulation/ScalarValue.cs
```diff
@@ -0,0 +1,456 @@
...
+            return Convert.ToUInt32(GetIntValue());
+        }
+
+        ulong IConvertible.ToUInt64(IFormatProvider provider)
+        {
+            return Convert.ToUInt64(GetIntValue());
+        }
+    }
+
+    public class ScalarIntValue : ScalarValue
```
erendrake: No multi-class files

src/kOS.Safe/Encapsulation/ScalarValue.cs
```diff
@@ -0,0 +1,456 @@
...
+            get { return true; }
+        }
+
+        public ScalarIntValue(int value)
+        {
+            internalValue = value;
+        }
+    }
+
+    public class ScalarDoubleValue : ScalarValue
```
erendrake: No multi-class files

src/kOS.Safe/Encapsulation/ScalarValue.cs
```diff
@@ -0,0 +1,456 @@
+﻿using System;
+using kOS.Safe.Encapsulation.Suffixes;
+using kOS.Safe.Exceptions;
+
+namespace kOS.Safe.Encapsulation
+{
+    abstract public class ScalarValue : Structure, IConvertible
+    {
+        protected object internalValue;
```
erendrake: No non-private non-const fields

src/kOS.Safe/Encapsulation/ScalarValue.cs
erendrake: yeah, this is not really something we need to change. I was just pointing it out. picking an epsilon would be pretty ballsy. Floating point math is not a newbie friendly topic anyway.

src/kOS.Safe/Encapsulation/ScalarValue.cs
hvacengi: Suffixes were removed in the combined PR instead of this one, mostly because I didn't want to do another round of branch switching.  I can make that change in this PR itself if it will make it easier for you.

src/kOS.Safe/Encapsulation/ScalarValue.cs
hvacengi: I suppose I can modify the `Value` property to make it have a `protected` setter instead of exposing `internalValue` to subclasses.  In the case of `abstact` classes however it makes sense to me that `protected` fields be allowed, since by definition the `abstact` class itself cannot be instantiated.  I will make the change to `Value`.

src/kOS.Safe/Encapsulation/ScalarValue.cs
hvacengi: Agreed.  But until we get it set up that we can run basic scripts outside of kOS (a stand alone interpreter) we cannot easily use the debugger to confirm this is the case.  I'm of the mindset to leave these checks in until we know for sure that we have caught every method of instantiating.

src/kOS.Safe/Encapsulation/ScalarValue.cs
hvacengi: The null check prior to casting is needed for null safety.  I should remove the `is` however, since the `as` and `!= null` are equivalent, which means we're casting twice instead of once.

src/kOS.Safe/Encapsulation/ScalarValue.cs
hvacengi: I see below that you are referring to the floating point rounding, rather than the conversion itself.  I'll make comments in reply to the post that you and dunbaratu have already be conversing in.

src/kOS.Safe/Encapsulation/ScalarValue.cs
hvacengi: You two understand floating point math itself better than I.  If there is no loss of precision, I would suggest we just treat everything like an double.  There is nowhere in the code that we limit the operations based on being an integer, and in fact operations like `Power` and `Divide` require us to treat all values like doubles already.  `Divide` is probably the most notable, since the value is very different in integer math than it is in floating point math (`3/2 = 1`, rather than `1.5`).  But it would be important that the sum of the integer values would equal an integer parse (i.e. `100000 + 228000 = 328000` evaluates to `true`).  Since I'm not familiar with the inner workings of floating points, I don't know at which point that becomes an issue, maybe it only breaks down beyond the storage capacity of the integer anyways.

I should also point out that right now there is nothing to address overflow, so I think it is possible to add two positive integer values and return a negative integer value.

src/kOS.Safe/Encapsulation/ScalarValue.cs
hvacengi: Fixed in the combined PR #1294, I can port back here if you prefer.

src/kOS.Safe/Encapsulation/ScalarValue.cs
hvacengi: I'll put them into their own files.

src/kOS.Safe/Encapsulation/ScalarValue.cs
Dunbaratu: It won't be a problem.

32-bit floats can store an accurate integer up to (2^24 - 1).
64-bit floats can store an accurate integer up to (2^53 - 1)

The standard floating point number formats are essentially just scientific notation, with the following two important differences:

1 - The mantissa is expressed only as a fractional part between 0 and 1, rather than with having one digit to the left of the point as it usually is with scientific notation. (i.e. normal scientific notation would express the number 123 as 1.23e2, while the way computer numbers do it would express it more like 123e3, with the nomenclature using a shorthand that there's an implied "0." preceding the integer 123, that just isn't shown in the format.  This nomenclature allows the format to never mention a decimal place, and just write it out with integers.)
2 - Except it's not in base 10 of course, it's base 2.  So the nomenclature 123e3 mentioned above would actually be more like this: 1111011e111 ( 0.123, (stored as 1111011), shifted 7 (stored as 111) binary places).
3- It's a bit more complex when the number is negative because it does negatives using 2's compliment representation of the mantissa integer, but that's the general idea.

Thus the largest exact integer it can hold is just however many binary digits are dedicated to the mantissa part.  Looking up the IEEE standards on it, the 64 bits of a double are allocated as 10 bits for the exponent part, and 54 bits for the mantissa integer part (one of which is the sign bit so it's really 53 bits for the max int, minus 1 for how 2s complement works, thus the value (2^53 - 1) I mentioned above.)

I went through working this out a while back when I was worried about how we were letting people index arrays and lists using doubles instead of ints, and how I thought it would break because it can't actually convert down into the integers the array indeces expect, but then I did that above work and realized the integers would have to get enormous before it would truly break.

src/kOS.Safe/Encapsulation/ScalarValue.cs
hvacengi: Turns out this was done in the other branch as well.  I'm merging in the fixes as "cherry-pick" commits

src/kOS.Safe/Encapsulation/ScalarValue.cs
hvacengi: That makes it sound like I can remove the `if (val1.IsInt && val2.IsInt)` checking and just treat everything as a double.  I don't think the performance difference between integers and doubles would be that notable given the bottle neck we already have in executing the opcode instructions.

PR #1294, opened 2015-12-03
New primitive type structures

hvacengi:
Includes #1293 and #1293

Modified execution to use structures for built in primitive types.  This should allow us to add suffixes to numbers and booleans.  It also means that any "user" value can now be treated differently from our internal instructions, so strings don't need to be qualified against special characters used by the VM.


src/kOS.Safe/Compilation/Calculator.cs
```diff
@@ -54,293 +51,90 @@ public static Calculator GetCalculator(object argument1, object argument2)
         }
     }
 
-    public class CalculatorIntInt : Calculator
```
erendrake: No Multiclass files.

src/kOS.Safe/Encapsulation/ISerializableValue.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace kOS.Safe.Encapsulation
+{
+    public interface ISerializableValue
```
erendrake: this could be done with the https://msdn.microsoft.com/en-us/library/system.serializableattribute(v=vs.110).aspx class attribute

src/kOS.Safe/Encapsulation/Structure.cs
```diff
@@ -157,5 +158,59 @@ public override string ToString()
         {
             return "Structure ";
         }
+
+        public static StringValue operator +(Structure val1, Structure val2)
+        {
+            return new StringValue(val1.ToString() + val2.ToString());
+        }
+
+        public static object FromPrimitive(object value)
```
erendrake: I am not convinced that this is the right class for this.

hvacengi: I don't know that I'm convinced it's the perfect class for it, but I've never been huge on "utility" classes that exist simply to provide miscellaneous helper methods.   I didn't want three separate methods on `StringValue`, `ScalarValue`, and `BooleanValue` because that would just mean chaining together a bunch of convert attempts.   My logic for putting it into the `Structure` class was that it was the first common inherited class, and it is used primarily to ensure an object is a `Structure` so that it has suffixes.

Dunbaratu: I think it's totally appropriate to have multiple classes per file provided only one of them is publicly exposed.  If you have a class who's essentially used by one other class and one other class only and is internal to that class's thinking, I don't think it has to be in a separate file.  A helper class needs to be separate when it's something you expect might be passed back to a caller outside the class, or be passed in by a caller outside the class.  

Essentially, the answer to the question, "what file is this ThatNameClass I'm seeing here defined in?", should always be "look for a file called ThatClassName.cs" and not "look for a file called SomeOtherClassName.cs".  It's only when nobody outside the one place it's used locally will ever see the class and need to find its definition that I'd consider it okay to put it in another class's file.

src/kOS.Safe/Encapsulation/ISerializableValue.cs
hvacengi: I may be wrong, but I don't think that our serialization hooks into .net's serialization methods.  In this particular case, the interface is being used by @tomekpiotrowski to tell his serializer to reduce the structure to the built in type.

I haven't spent enough time with the serialization stuff to know how structures opt in (or out) of having suffixes serialized, so I don't know if the shift to the attribute system is worth the cost.

tomekpiotrowski: I'm using that interface to basically group all new encapsulation types as they had no common parent class that I could use to recognize them. I'm not familiar with C# attributes, so I don't know if we could use them to replace that.

erendrake: Im not saying we should use the built in serializer framework. just the attribute. When i see an empty Interface it makes me want to make it an attribute. we could make a `kOSSerializable` if we want.

erendrake: @tomekpiotrowski if you look at our binding classes or our functions. They are using our own Attributes.

PR #1315, opened 2015-12-16
Delegation

Dunbaratu:
Phew.  Documenting this took more words than the actual code changes.  It's not an easy concept to ELIK to the new programmers out there.

Implements the creation of delegates in kerboscript programs for at least user functions and builtin functions, but not for suffixes.


doc/source/language/delegates.rst
```diff
@@ -0,0 +1,390 @@
...
+    local vecxyz is vecxy:bind(1).  // vecxyz is a KOSDelegate hardcoding x to 10, y to 5, and z to 1, taking no args.
+    local vec is vecxyz:call(). // makes a V(10, 5, 1).
+
+    // The above chain of bindings could have been chained together on one line like so:
+    local vec is V@:bind(10):bind(5):bind(1):call().
+
+The technique of transforming a function that takes many arguments into
+a nested succession of functions that each only take one argument has a
+name.  It's called :ref:`Currying <https://en.wikipedia.org/wiki/Currying>`_.
+(It's named after mathemetician
```
tomekpiotrowski: typo

doc/source/structures/misc/kosdelegate.rst
```diff
@@ -0,0 +1,77 @@
...
+        * - :meth:`CALL(varying arguments)`
+          - same as the function this is a delegate for.
+          - calls the function this delegate wraps.
+        * - :meth:`BIND(varying arguments)`
+          - another KOSDelegate
+          - creates a new KOSDelegate with some arguments predefined.
+
+.. note::
+
+    This type is not serializable.
```
tomekpiotrowski: When a type is not serializable I've been simply omitting this note. I think it could be potentially confusing, someone could have a quick glance, see a 'serializable-something' note and think it means that a type IS serializable.

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -1469,10 +1554,10 @@ public static object ExecuteDelegate(ICpu cpu, Delegate dlg)
...
                     dlg.DynamicInvoke(argArray);
                     return null; // So that the compiler building the opcodes for a function call statement doesn't
-                                 // have to know the function prototype to decide whether or
-                                 // not it needs to pop a value from the stack for the return value.  By adding this,
-                                 // it can unconditionally assume there will be exactly 1 value left behind on the stack
-                                 // regardless of what function it was that was being called.
+                    // have to know the function prototype to decide whether or
+                    // not it needs to pop a value from the stack for the return value.  By adding this,
+                    // it can unconditionally assume there will be exactly 1 value left behind on the stack
+                    // regardless of what function it was that was being called.
```
Dunbaratu: dangit I got lazy and told the editor to auto-indent this section, and it did this to my nicely lined up comment.

src/kOS.Safe/Encapsulation/KOSDelegate.cs
```diff
@@ -0,0 +1,154 @@
...
+            cpu.PushStack(new KOSArgMarkerType());
+            foreach (object item in preBoundArgs)
+            {
+                cpu.PushStack(item);
+            }
+            foreach (object item in aboveArgs) // Because this was pushed to a stack, this should show in reverse order.
+            {
+                cpu.PushStack(item);
+            }
+        }
```
Dunbaratu: It occurs to me that PushUnderArgs() is only being called from Call(), not from here.  That may explain why I'm having some troubles with cases where I don't explictitly state the ``:call`` suffix.  I'll have to double-check this tomorrow.

src/kOS.Safe/Encapsulation/KOSDelegate.cs
```diff
@@ -0,0 +1,154 @@
...
+        /// If the derivative class needs to put anything on the stack underneath the
+        /// KOSArgMarkerType and the args, it's given an opportunity to do so here by
+        /// overriding this method.
+        /// </summary>
+        public abstract void PushUnderArgs();
+        
+        /// <summary>
+        /// Should return a new instance of self, with all fields copied.
+        /// </summary>
+        /// <param name="KOSDelegate">Should be of type whatever the actual derived type is, not a raw KOSDelegate</param>
```
Dunbaratu: hmm. this parameter was deleted, but it looks like I never deleted the xml comment to go with it.

src/kOS.Safe/Encapsulation/Suffixes/VarArgsVoidSuffix.cs
```diff
@@ -0,0 +1,19 @@
...
+        {
+            this.del = del;
+        }
+
+        public override object Get()
+        {
+            return del;
+        }
+    }
+}
```
Dunbaratu: I should point out that I inherited this from @tomekpiotrowski 's original branch but because of a redesign of the CALL suffix, it ended up never getting used yet.  We might want to just ditch this file entirely from the PR.

src/kOS.Safe/Encapsulation/UserDelegate.cs
```diff
@@ -1,17 +1,18 @@
 ﻿using System.Collections.Generic;
+using kOS.Safe.Execution;
+using kOS.Safe.Compilation;
 
-namespace kOS.Safe.Execution
+namespace kOS.Safe.Encapsulation
```
Dunbaratu: Rationale behind why this moved to a new location:  Previously  UserDelegate had no user-exposed parts inside.  Now it's a kind of KOSDelegate and therefore has suffixes users can access.

src/kOS/UserIO/TelnetWelcomeMenu.cs
```diff
@@ -152,7 +152,6 @@ private bool CPUListChanged()
                     if (newList[i] != availableCPUs[i])
                         itChanged = true;
 
-            Console.WriteLine("in CPUListChanged(): itChanged = {0}, newList.Count={1}, availableCPUs.Count={2}", itChanged, newList.Count, availableCPUs.Count);
```
Dunbaratu: caught this when I did a scan through all the code for cases where there was a WriteLine, to see if any of them were leftover debug crud, and this one was.

src/kOS/UserIO/TelnetMainServer.cs
```diff
@@ -57,7 +57,6 @@ public TelnetMainServer()
             
             DontDestroyOnLoad(transform.gameObject); // Otherwise Unity will stop calling my Update() on the next scene change because my gameObject went away.
 
-            Console.WriteLine("kOS TelnetMainServer class exists."); // Console.Writeline used because this occurs before kSP's logger is set up.
```
Dunbaratu: caught this when I did a scan through all the code for cases where there was a WriteLine, to see if any of them were leftover debug crud, and this one was.

tomekpiotrowski:
Worth noting that this fixes #740

Dunbaratu:
I just found a bug where this works fine:
```
print "which evals to " + outer(10):call().
```
but this fails with a complaint about missing arguments:
```
print "which evals to " + outer(10)().
```
(Note, `outer` is a curried delegate here.  The bug is that I seem to have found a case where currying doesn't work right when calling directly without the `call` suffix.  I though I'd covered all those cases but I must have missed one.  I'm investigating now.

src/kOS.Safe/Encapsulation/KOSDelegate.cs
Dunbaratu: Nope, that wasn't the problem.  never mind

src/kOS.Safe/Encapsulation/Suffixes/VarArgsVoidSuffix.cs
Dunbaratu: I'm ditching it.

Dunbaratu:
Okay, @erendrake.  It's ready.

src/kOS.Safe/Execution/CPU.cs
```diff
@@ -614,8 +615,19 @@ private Variable GetVariable(string identifier, bool barewordOkay = false, bool
...
                 return null;
-            else
-                throw new KOSUndefinedIdentifierException(identifier.TrimStart('$'), "");
+            // In the case where we were looking for a function pointer but didn't find one, and would
+            // have failed with exception, then it's still acceptable to find a hit that isn't a function
+            // pointer (has no trailing asterisk '*') but only if it's a delegate of some sort:
+            if (identifier.EndsWith("*"))
+            {
+                string trimmedTail = identifier.TrimEnd('*');
+                Variable retryVal = GetVariable(trimmedTail, barewordOkay, failOkay);
```
erendrake: both barewordOkay and failOkay will always be false at this point because we test them both for true and return if they are.

Just raising a flag on the off chance you didnt mean for this to happen.

Dunbaratu: This is why I don't always agree with the attitude that you should always change
```
if (foo)
   return something;
else
{
   do something;
}
```
into just
```
if (foo)
   return something;
do something;
```
Even though they have the same effect, there is a conceptual difference in the programmer's mind between those two.  The first form conveys the intent that ``do something`` was intended as being directly married to the ``if (foo)`` check, and thus if you move that if-check at a later date, it should probably move the else clause with it so it stays attached.  The second form divorces the ``do something`` from the ``if (foo)`` check and makes it look as if it should be left behind if there's a reason to move the ``if (foo)`` check later.

In this case, I really see the call to GetVariable as being not directly "else-tied" to the ``if (failOkay)`` check above.  In other words, passing in the variables anyway even though they *happen to be* both false in this case is a form of future-proofing in case sections of his code get moved and shuffled around later.

PR #1323, opened 2015-12-20
Ranges

### LINKED ISSUE #808: (enhancement) Python-like "for i in range(begin, end, step)"
Similar to this issue: #40
It would be nice if instead of this:
```
set i to 0.
until i >= n
{
  do stuff.
  set i to i + 1.
}
```
I could just type:
```
for i in range(n)
{
  do stuff.
}
```

tomekpiotrowski:
Fixes #808

### DIFF
--- a/src/kOS.Safe/Encapsulation/EnumerableValue.cs
+++ b/src/kOS.Safe/Encapsulation/EnumerableValue.cs
@@ -1,25 +1,27 @@
 ﻿using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
 using kOS.Safe.Encapsulation.Suffixes;
 using kOS.Safe.Serialization;
+using System.Linq;
 
 namespace kOS.Safe.Encapsulation
 {
-    public abstract class EnumerableValue<T, TC> : SerializableStructure, IEnumerable<T> where TC : IEnumerable<T> where T : Structure
+    public abstract class EnumerableValue<T, TE> : SerializableStructure, IEnumerable<T> where TE : IEnumerable<T> where T : Structure
     {
-        protected TC Collection { get; private set; }
+        protected TE Enum { get; private set; }
+        private readonly string label;
 
-        protected EnumerableValue(TC collection)
+        protected EnumerableValue(string label, TE enumerable)
         {
-            Collection = collection;
+            this.label = label;
+            Enum = enumerable;
 
             InitializeEnumerableSuffixes();
         }
 
         public virtual IEnumerator<T> GetEnumerator()
         {
-            return Collection.GetEnumerator();
+            return Enum.GetEnumerator();
         }
 
         IEnumerator IEnumerable.GetEnumerator()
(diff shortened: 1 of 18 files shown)


tomekpiotrowski:
I would suggest merging #1360 first, after that I will push a commit that resolves all conflicts.

tomekpiotrowski:
This currently conflicts with #1393, I'll submit an update when #1393 gets merged.

src/kOS.Safe/Encapsulation/EnumerableValue.cs
```diff
@@ -1,27 +1,27 @@
...
 using kOS.Safe.Serialization;
+using System.Linq;
 
 namespace kOS.Safe.Encapsulation
 {
-    public abstract class EnumerableValue<T, TC> : Structure, IEnumerable<T>, IDumper where TC : IEnumerable<T>
+    public abstract class EnumerableValue<T, TE> : Structure, IEnumerable<T>, IDumper where TE : IEnumerable<T>
     {
-        protected TC Collection { get; private set; }
+        protected TE Enum { get; private set; }
```
hvacengi: I think I'd prefer that this identifier was refactored so it doesn't look like a potential collision with the `enum` type (I know, it doesn't actually conflict with it, I just think it's a bit confusing because of the similarity).  Maybe as simple as changing it to `Enumer`, or `InnerEnumerable`?  Let me know if I sound paranoid about the naming, maybe nothing needs to happen.

tomekpiotrowski: Sure, I can change the name. Please mind that this PR is in need of rebasing, which I'm postponing until file read/write PR gets merged.

hvacengi: Ah, I was just noticing that too.  I guess I'll have to jump ahead on the PR list then.

src/kOS.Safe/Encapsulation/EnumerableValue.cs
```diff
@@ -1,25 +1,27 @@
...
 using kOS.Safe.Serialization;
+using System.Linq;
 
 namespace kOS.Safe.Encapsulation
 {
-    public abstract class EnumerableValue<T, TC> : SerializableStructure, IEnumerable<T> where TC : IEnumerable<T> where T : Structure
+    public abstract class EnumerableValue<T, TE> : SerializableStructure, IEnumerable<T> where TE : IEnumerable<T> where T : Structure
     {
-        protected TC Collection { get; private set; }
+        protected TE Enum { get; private set; }
```
hvacengi: It looks like this was changed to InnerEnum in your subclasses, but not here in EnumerableValue.  I'll correct the issue with a refactor.

tomekpiotrowski:
I must have missed a file when creating a commit, sorry for that

hvacengi:
No worries.  I'm actually refactoring it anyways to be `InnerEnumberable` just for the sake of clarity on my end, so it isn't a big deal to fix.

PR #1330, opened 2015-12-27
Fixes #1289 (buffer overrun in command history)

### LINKED ISSUE #1289: terminal weirdness - uparrow at bottom line - doesn't redraw command.
To reproduce:

- Run some commands interactively until the cursor scrolls far enough down that it has caused the terminal to scroll some, and the cursor is on the bottom row.

- Use up-arrow to retrieve the previous command.

The command is invisible but it is in the buffer.  If you hit enter it will run again.  You can edit it, but you have to edit it blindly - i.e. backspace some and then retype stuff.

The moment you make an edit that will insert a new character into the line, THEN it redraws the line.  Before that it does not.  Deleting a char doesn't redraw it - only adding a new nonblank character does.

Dunbaratu:


src/kOS.Safe/Screen/ScreenBuffer.cs
```diff
@@ -99,10 +99,13 @@ public virtual int ScrollVertical(int deltaRows)
...
         /// <param name="startRow">Starting with this row number</param>
-        /// <param name="numRows">for this many rows</param>
+        /// <param name="numRows">for this many rows, or up to the max row the buffer has if this number is too large</param>
         public void MarkRowsDirty(int startRow, int numRows)
         {
-            for( int i = 0; i < numRows ; ++i)
+            // Mark fewer rows than asked to if the reqeusted number would have blown past the end of the buffer:
+            int numSafeRows = (numRows + startRow <= buffer.Count) ? numRows : buffer.Count - startRow;
+            
+            for( int i = 0; i < numSafeRows ; ++i)
```
Dunbaratu: I wasted some time trying to re-arrange the order in which things happen, to ensure the buffer is enlarged to the new size first before this gets called, but then ran into a chicken-and-egg problem (two steps both had to be done before each other).  I realized that was pointless because the only rows that need to be marked dirty are rows that previously existed anyway.  A newly added row starts off 'dirty' by default anyway.  Thus it's fine to call this before the buffer is expanded to the new size, provided I protect against blowing past the buffer, which was what caused the exception before.

PR #1331, opened 2015-12-27
Fixes #1317 : now all logged exceptions can beep.

### LINKED ISSUE #1317: error beep doesn't sound on some internal errors
The purpose of the error beep was to warn the player when their script crashed and isn't in control of the ship anymore.  A sort of "autopilot disengaged" warning.

But I'm finding a few circumstances where the beep does not occur on some exceptions.  It seems to be exceptions that aren't derived from KOSException, such as when we have an error in our own code that throws a nullref.  Regardless of whether the crash was the user's fault or our own, anything that stops the script needs to have the beep.

Dunbaratu:


src/kOS/Logger.cs
```diff
@@ -58,7 +58,7 @@ public virtual void Log(Exception e)
                 message += LINE_RULE;
             }
 
-            if (SafeHouse.Config.AudibleExceptions && e is KOSException)
+            if (SafeHouse.Config.AudibleExceptions)
```
Dunbaratu: Previously it deliberately avoided beeping non-KOSExceptions but that's silly.  If the code path gets here, that means it WILL print the exception unconditionally to the terminal, so that's the point at which a beeping is always appropriate regardless of the type of exception.  If we were suppressing an exception message it would never have gotten this far.

erendrake: you are right, that was silly!

Dunbaratu: well, i was the one that was being silly, originally.

PR #1332, opened 2015-12-27
Fixes #1283 : better msg for KOSCompileException

### LINKED ISSUE #1283: Compiler ad-hoc thrown exceptions don't give user the script line/col of the error.
When the grammar file finds a syntax error, it makes a message that will show line number and dump the offending line to the screen.

But those syntax errors we throw in the compiler as a special case, like KOSWaitIinvalidHere, or my new KOSDefaultParameterNotAtEndException for example, don't display the offending line.  We need to fix that.

Dunbaratu:
An example of the sort of error that is now reported better is this:
```
when x = 0 then {
  print "x is zero".
  wait 2.  // compile error: a wait inside a trigger.
}
```
and this:
```
if x = 0 {
  break. // compile error: you can't break when not in a loop
}
```
Basically, it's all the stuff the parse rules say is fine but
for further contextual reasons the parser isn't smart enough to
understand, it's considered a compile error.

### DIFF
--- a/src/kOS.Safe/Compilation/KS/KSScript.cs
+++ b/src/kOS.Safe/Compilation/KS/KSScript.cs
@@ -26,8 +26,16 @@ namespace kOS.Safe.Compilation.KS
                 var compiler = new Compiler();
                 LoadContext(contextId);
 
-                // TODO: handle compile errors (e.g. wrong run parameter count)
-                CodePart mainPart = compiler.Compile(startLineNum, parseTree, currentContext, options);
+                CodePart mainPart;
+                try
+                {
+                    mainPart = compiler.Compile(startLineNum, parseTree, currentContext, options);
+                }
+                catch (KOSCompileException e)
+                {
+                    e.AddSourceText((short)startLineNum, scriptText);
+                    throw e;
+                }
 
                 // add locks and triggers
                 parts.AddRange(currentContext.UserFunctions.GetNewParts());
@@ -148,7 +156,6 @@ namespace kOS.Safe.Compilation.KS
                 
                 foreach (ParseError err in parseTree.Errors)
                 {
-                    System.Console.WriteLine("eraseme:  err msg: " + err.Message);
                     if (err.Message.StartsWith("Unexpected token 'EOF'"))
                     {
                         if (err.Message.Contains("Expected CURLYCLOSE") ||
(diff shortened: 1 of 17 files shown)


src/kOS.Safe/Compilation/KS/LineCol.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace kOS.Safe.Compilation.KS
+{
+    /// <summary>
+    /// Just a dumb simple tuple of line and column, to be returned in
+    /// places where the compiler needs to treat the pair of them like
+    /// a single return value from a method.
+    /// </summary>
+    public class LineCol
+    {
+        public short Line { get; set; }
```
erendrake: could these have private setters to mke the object immutable?

Dunbaratu: doesn't really matter.  There's currently no use of their mutability so it could be changed.

src/kOS.Safe/Compilation/KS/KSScript.cs
```diff
@@ -26,8 +26,16 @@ public override List<CodePart> Compile(string filePath, int startLineNum, string
...
-                CodePart mainPart = compiler.Compile(startLineNum, parseTree, currentContext, options);
+                CodePart mainPart;
+                try
+                {
+                    mainPart = compiler.Compile(startLineNum, parseTree, currentContext, options);
+                }
+                catch (KOSCompileException e)
+                {
+                    e.AddSourceText((short)startLineNum, scriptText);
+                    throw e;
```
erendrake: i assume you are trying to throw the same exception with the original origin? if so you should remove the `e`

```catch (KOSCompileException e)
                {
                    e.AddSourceText((short)startLineNum, scriptText);
                    throw;
}```

src/kOS.Safe/Compilation/KS/LineCol.cs
erendrake: We have a few line column pairs floating around and many are ints. these are short. Do we have a real expectation that these types should be different?

Dunbaratu: It matches limits in KSM

src/kOS.Safe/Compilation/KS/KSScript.cs
Dunbaratu: I keep forgetting c# has this strange syntax of throw with no argument.

Dunbaratu: Go ahead and change it

Issue #1340, opened 2015-12-31
Rover driving cooked steering is badly documented.


Dunbaratu:
The wheelthrottle and wheelsteering cooked steering values are really not mentioned at all on the cooked steering page and they really should be.  They're only documented if you know exactly where to go to find out that they exist.

Dunbaratu:
This should have closed when PR #1341 was merged.

Issue #1326, opened 2015-12-25
Q reading doesn't match kPa


Gaiiden:
Here is what I get when I print out `ship:Q` under the latest kOS and FAR in 1.0.5

![untitled](https://cloud.githubusercontent.com/assets/7699577/12004375/3e72f8f6-ab22-11e5-9d80-a0b569ce37f6.png)

The kOS screen refreshes once a second so the value is behind that's not the issue - the issue is what's up with the decimal placing?

Dunbaratu:
I just had a read through the code and `Q` is being returned in units of ATM's not kPa's.  The documentation is basically lying.  I think it *should* be kPa's and the docs are right and the code is wrong.

Dunbaratu:
How many scripts are alive out there that are using `Q` and expecting it to be ATM's?  Because I don't want complaining by users who's scripts broke when we fix this.

Dunbaratu:
Okay after slack discussion, we're keeping it as ATM's and changing the documentation to match.  If you want to get it in kPA, you can multiply it by `constant:AtmToKpa`.

Gaiiden:
Whoops forgot to keep tabs on this while traveling. Sounds good, especially since it's a fix that doesn't need a new code release

Issue #1301, opened 2015-12-04
Bound Variables docs don't mention kuniverse


Dunbaratu:
the bound variables page has a section like so:

> System Variables
> 
> This section is about variables that describe the things that are slightly outside the simulated universe of the game and are more about the game’s user interface or the kOS mod itself. They represent things that slightly “break the fourth wall” and let your script access something entirely outside the in-character experience.

This section makes no mention of ``kuniverse`` at all, and in fact the entire bound variables page never mentions it.

hvacengi:
Close via #1341

Issue #1299, opened 2015-12-04
update or rewrite the links to kslib page of the docs


Dunbaratu:
This page:
http://ksp-kos.github.io/KOS_DOC/library.html

Has hardcoded links into the kslib docs.  That's not good futureproofing, as kslib is re-arrangable and things can change.  Also it only mentions a few of the things there, and one of them is the now obsolete ``pid_lib.ks``.

Maybe it should say less and just point directly to kslib?

hvacengi:
Close via #1341

PR #1341, opened 2016-01-01
some_documentation_fixes:  Fixes #439 #655 #1021 #1236 #1254 #1263 #1278 #1288 #1297 #1299 #1301 #1326 #1340

### LINKED ISSUE #439: Help the user realize that 2-D ground HEADING and BEARING requires horizontal orientations.
I just spent a long time debugging a small dumb rover script and couldn't figure out why it started driving in a circle as soon as tried to climb a slight hill slope.

After a lot of work I finally realized that kOS was calculating the BEARING suffixes based on the topward direction of the rover rather than its forward direction, and on the flat land it worked okay because the front wheels were slightly lower than the rear, tilting the craft slightly forward.  This slight 1 degree tilt forward made it possible to get the "right" answer because when it projects the topward direction down to the 2D surface, it gets a very small vector in the correct direction, just barely.  But as soon as the slope tilts the craft back just a little, that projection was now backward of the forward direction.  When on a slope, it will drive in a circle if you use the BEARING to your target to steer from, because while going upslope the bearing is 180 degrees flipped from what it is when going downslope.

But when I told it to "control from here" on a docking port facing forward, the script worked just fine.

Because it's so easy to get it to work on flat land by random accident when used on a vertically oriented core, and vertically oriented cores work fine in manual driving mode, it's really quite confusing when it suddenly goes all wrong on a slight upslope.  The user has no idea what's wrong.  Heck, I've read the C# code and I had no idea for several hours what was wrong.

We should have some sort of warning check that when someone tries to set the WHEELSTEERING, either through cooked or raw control mode, it detects if SHIP:FACING:VECTOR is aimed perpendicularly (or nearly so) from the wheel orientation and issues a warning in that case.

Maybe the warning should be in the form of an onscreen hover message rather than a script-killing exception.

Dunbaratu:
Lots and lots of small doc edits.

### DIFF
--- a/doc/source/tutorials/designpatterns.rst
+++ b/doc/source/tutorials/designpatterns.rst
@@ -57,7 +57,7 @@ Here, we introduce IF/ELSE logic into UNTIL loops:
     STAGE.
     UNTIL SHIP:ALTITUDE > 20000 {
         IF SHIP:ALTITUDE > 10000 {
-            LOCK STEERING TO R(0,0,-90) + HEADING(90,45).
+            LOCK STEERING TO TO R(0,0,-90) + HEADING(90,45).
         }
         IF STAGE:LIQUIDFUEL < 0.1 {
             STAGE.
--- a/doc/source/tutorials/pidloops.rst
+++ b/doc/source/tutorials/pidloops.rst
@@ -3,6 +3,22 @@
 PID Loops in kOS
 ================
 
+.. versionadded:: 0.18.1
+
+    Note, this is an older tutorial.  As of 
+    kOS version 0.18.1 and up, a new :struct:`pidloop`
+    feature was added to kOS to allow you to use a built-in PID
+    controller that executes very quickly in the kOS "hardware"
+    rather than in your script code.  You can use it to perform
+    the work described in detail on this page.  However, this
+    tutorial is still quite important because it walks you through
+    how a PID controller works and what it's really doing under the
+    hood.  It's probably a good idea to use the built-in
+    :struct:`pidloop` instead of the program shown here, once you
+    understand the topic this page describes.  However, it's also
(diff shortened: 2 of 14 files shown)


doc/source/tutorials/designpatterns.rst
```diff
@@ -57,7 +57,7 @@ Here, we introduce IF/ELSE logic into UNTIL loops:
     STAGE.
     UNTIL SHIP:ALTITUDE > 20000 {
         IF SHIP:ALTITUDE > 10000 {
-            LOCK STEERING TO R(0,0,-90) + HEADING(90,45).
+            LOCK STEERING TO TO R(0,0,-90) + HEADING(90,45).
```
erendrake: TO TO?

doc/source/tutorials/pidloops.rst
```diff
@@ -3,6 +3,22 @@
...
+    rather than in your script code.  You can use it to perform
+    the work described in detail on this page.  However, this
+    tutorial is still quite important because it walks you through
+    how a PID controller works and what it's really doing under the
+    hood.  It's probably a good idea to use the built-in
+    :struct:`pidloop` instead of the program shown here, once you
+    understand the topic this page describes.  However, it's also
+    a good idea to have a read through this page to get an 
+    understanding of what that built-in feature is really doing.
+
```
erendrake: Im glad we didnt simply get rid of the old tutorial. it is still useful for understanding PID  concepts

doc/source/tutorials/designpatterns.rst
Dunbaratu: You can fix the obvious error here if you like so you can merge it.  I wont be at my computer until tomorrow.

PR #1343, opened 2016-01-01
Char

gisikw:
Provides a `char(65). // "A"` function, which allows for snagging characters such as quote, newline, etc.

`print char(34) + "Foo" + char(34).`

Probably doesn't entirely belong in the Math functions, but we don't really have a section for casting, and this seemed the nearest fit.

Thanks to @hvacengi for the encouragement!

### DIFF
--- a/doc/source/math/basic.rst
+++ b/doc/source/math/basic.rst
@@ -74,7 +74,7 @@ constants about the universe that you may find handy in your math operations.
     This constant is provided mainly for the benefit of people who are
     playing with the mod "RemoteTech" installed, who may want to perform
     calculations about signal delays to hypothetical probes.  (Note that
-    if the probe already has a connection, you can 
+    if the probe already has a connection, you can
     :ref:`ask Remotetech directly <remotetech>` what the signal delay is.
 
 .. global:: Constant:AtmToKPa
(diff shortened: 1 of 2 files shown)


erendrake:
If we dont have a good section for something. dont hesitate to add one! Math 

So the result of the PR would be to allow for anything from UTF-32 to be stored in a variable based on its integral value? I like it!

doc/source/math/basic.rst
```diff
@@ -294,3 +295,12 @@ Trigonometric Functions
         PRINT ARCTAN2(0.67, 0.89). // prints 36.9727625
 
     The two parameters resolve ambiguities when taking the arctangent. See the `wikipedia page about atan2 <http://en.wikipedia.org/wiki/Atan2>`_ for more details.
+
+.. function:: CHAR(a)
```
erendrake: This not only puts it in math which is kinda close. but in trig functions which i dont think we want at all. ill move it up into the basic math stuff

Dunbaratu:
It would make a lot of sense to include the reciprocal operation wth this.  If you can get a "A" from a 65, then you should be able to get a 65 from an "A".

gisikw:
Sure, I should be able to rig up an `unchar("A").`

PR #1348, opened 2016-01-04
Issue #1321 undock suffix

### LINKED ISSUE #1321: DockingPort:undock causes crash
Can be reproduced on a simple CPU+Probe+DockingPort craft on the launchpad.

`SHIP:DOCKINGPORTS[0]:UNDOCK`

The part does indeed undock, but the program crashes with the following error:
```
(Cannot show kOS Error Location - error might really be internal. See kOS devs.)
```

Encountered this behavior both on develop, and in a clean install of the 0.18.2 release.

hvacengi:
Fix exception when using the docking port's `UNDOCK` suffix.

This PR fixes #1321.


hvacengi:
Fixes #1321

PR #1349, opened 2016-01-05
Issue 1279 rcs vecter renderers

### LINKED ISSUE #1279: RCS display vectors are strange when RCS is off.
No redraw

it looks like these are stale values that dont get cleared out. The issue could be solved by simply hiding vectors that are shown by `STEERINGMANAGER:SHOWRCSVECTORS` while RCS is turned off.

EDIT: I just realized that cropping the video destroyed the context that RCS is off the whole time. and the issue only appears when i come out of timewarp and start slewing to a new position. If you turn on RCS all of the yellow and pink vectors snap into place

hvacengi:
Fix RCS vector renderers showing up when the RCS action group is off.

Fix RCS and Engine vector renderers persisting after staging event.

This PR fixes #1279.

### DIFF
--- a/src/kOS/Binding/SteeringManager.cs
+++ b/src/kOS/Binding/SteeringManager.cs
@@ -570,6 +570,7 @@ namespace kOS.Binding
                 rcsModules.Clear();
                 engineModules.Clear();
                 gimbalModules.Clear();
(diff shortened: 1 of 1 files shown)


hvacengi:
Fixes #1279

PR #1371, opened 2016-01-13
Fix vectors multiplied by ScalarValue

### LINKED ISSUE #1370: Encapsulated vectors return int's when multiplied by encapsulated scalars
For some reason doing something like `print (v(0.75,0.75,0. 75) * 1.01)` is returning `v(1,1,1)` with the new encapsulated values.  Visual studio says that it should be using the scalar `*` operator, which should be using `double` values for the math, but for some reason it isn't working like that.  I've already implemented a fix, and the PR will follow shortly.

hvacengi:
Fixes #1370
Vector.cs
* modify the ScalarValue "*" operator to return the value of the double "*" operator.

### DIFF
--- a/src/kOS/Suffixed/Vector.cs
+++ b/src/kOS/Suffixed/Vector.cs
@@ -186,7 +186,7 @@ namespace kOS.Suffixed
 
         public static Vector operator *(Vector a, ScalarValue b)
         {
-            return new Vector(a.X * b, a.Y * b, a.Z * b);
+            return a * b.GetDoubleValue();
         }
 
         public static Vector operator *(float b, Vector a)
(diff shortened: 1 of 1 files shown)


hvacengi:
@erendrake in case merging of this is held up by the failed build by Travis, I'd like to point out that the error does not seem to be with my code, but rather an issue in the pre-build steps.  I don't know if you can force Travis to try and build again, but this compiles fine on my machine.

Dunbaratu:
I'm going to work on some encapsulation stuff tonight and I'll probably pull this into my local branch as I do so.  It looks correct visually and trivial without any design changes that impact anything.

Issue #1303, opened 2015-12-05
Make a new 4th generation low-mass computer part (was: Support Tweakscale)


dcrockwell:
I'm making drones in the game, but the smallest computer part I can use weighs 80kg

If I could use tweakscale, I could make the computer smaller to fit the drone.

Alternately, create a very small controller unit for miniature drones that weights only a few grams.

Dunbaratu:
> the smallest computer part I can use weighs 80kg

The radial mount computer is 30kg  (called "Compomax Radial Tubeless").

dcrockwell:
Ideally, I'd like one that is less than 1kg.

As it is, the CPU is the heaviest component on the drone I'm building, and my hopes to create a mini drone are not possible.

Dunbaratu:
You can change the mass of parts with ModuleManager, or by editing the .cfg files that kOS ships with under kOS/Parts/.

We might make a 4th model for higher up the tech tree that' is lightweight, but in the meantime you can just change the ``mass = `` line in the configs.  (It's in megagrams, so 0.03 means 30 kg, for example.)

TDW89:
I believe tweakscale support is handled at the tweakscale side of things, although they probably rely on other mod creators and 3rd parties to submit module manager files to them.

Stick this file in the tweakscale patches folder. (name it kOS_TweakScale.cfg)
```
@PART[kOSMachine0m] // Small Inline kOS core
{
    %MODULE[TweakScale]
    {
        type = stack
        defaultScale = 0.625
    }
}

@PART[kOSMachine1m] // Medium Inline kOS core
{
    %MODULE[TweakScale]
    {
        type = stack
        defaultScale = 1.25
    }
}

@PART[kOSMachineRad] // Radialy attached kOS core
{
    %MODULE[TweakScale]
    {
        type = free
    }
}
```

pellinor0:
TweakScale can scale any kspField inside a partModule, it just needs the configs to tell it what to do. To scale values of the KOS module you can copy from ScaleExponents.cfg in the TweakScale folder. These patches can go anywhere inside Gamedata. 

I have my doubts how well scalable processors fit into the concept of KOS, but you can always write your own custom patches.

dcrockwell:
> I have my doubts how well scalable processors fit into the concept of KOS, but you can always write your own custom patches.

[...]

PR #1380, opened 2016-01-15
#1303 Next Generation Part

pgodd:
I hope my code is bearable.
It works!


src/kOS/LightColor/kOSLiteModule.cs
```diff
@@ -0,0 +1,90 @@
+using System;
+using UnityEngine;
+using KSP.IO;
+
+namespace kOS.LightColor
+{
+    class kOSLiteModule : PartModule
```
Dunbaratu: Pure pedantry, I know, but spelling it "kOSLiteModule" makes it sound like a stripped-down smaller footprint version of kOS, rather than a light as in illumination.  I.e. "lite" as in "lite beer", not "light" as in "turn on the light".

At any rate, it should probably match the namespace spelling one way or the other.  Right now it's a class called "lite" stored in a namespace called "light".

pgodd: Easy to change.

PR #1384, opened 2016-01-18
Documentation updates

### LINKED ISSUE #1324: vessel:UNPACKED not documented
The UNPACKED suffix for vessel doesn't seem to be in the sphinx docs.

### LINKED ISSUE #1383: The new `:HASTARGET` and `:HASNODE` are not documented
Should have been part of #1368 but we missed it.

hvacengi:
Fixes #1324
Fixes #1383

Document ship:UNPACKED, HASTARGET, and HASNODE

### DIFF
--- a/doc/source/bindings.rst
+++ b/doc/source/bindings.rst
@@ -27,8 +27,18 @@ TARGET:
 | **Variable Name**: TARGET
 | **Gettable**: yes
 | **Settable**: yes
-| **Type**: `Vessel <structures/vessels/vessel.html>`__ or `Body <structures/celestial_bodies/body.html>`__ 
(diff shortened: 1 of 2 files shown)


doc/source/bindings.rst
```diff
@@ -27,8 +27,18 @@ TARGET:
...
+target vessel limit the implementation to working with only the active vessel.
+
+HASTARGET:
+
+| **Variable Name**: TARGET
+| **Gettable**: yes
+| **Settable**: no
+| **Type**: boolean
+| **Description**: Will return true if the ship has a target selected.  This will always return false
+when not on the active vessel, due to limitations in how KSP sets the target vessel.
```
erendrake: wouldnt we want to limit this to the active vessel regardless of KSP's API problems?

hvacengi: We might.  That would probably be a bit of debate in the community.  This closes the debate, it isn't at the discretion of the kOS developers, but rather a decision made for us by Squad (which we probably agree with).  

As things were written before, setting the target from the non-active vessel would change the target for the active vessel instead.  The value returned by TARGET was also the target of the active vessel, so you could argue that there was some utility there.

PR #1389, opened 2016-01-23
LIST:JOIN

tomekpiotrowski:


src/kOS.Safe/Encapsulation/ListValue.cs
```diff
@@ -95,6 +95,7 @@ private void ListInitializeSuffixes()
             AddSuffix("INSERT",   new TwoArgsSuffix<ScalarValue, T>     ((index, toAdd) => Collection.Insert(index, toAdd)));
             AddSuffix("REMOVE",   new OneArgsSuffix<ScalarValue>        (toRemove => Collection.RemoveAt(toRemove)));
             AddSuffix("SUBLIST",  new TwoArgsSuffix<ListValue, ScalarValue, ScalarValue>(SubListMethod));
+            AddSuffix("JOIN",     new OneArgsSuffix<StringValue, StringValue>     ((separator) => Join(separator)));
```
hvacengi: If you make the `Join` method accept and return return a `StringValue` instead of a `string`, you can remove the anonymous method here and simply pass it as 
```
AddSuffix("JOIN", new OneArgsSuffix<StringValue, StringValue>(Join));
```

tomekpiotrowski: Ok, submitted a new version.

hvacengi:
Marking as not ready until it's updated following the merge of #1393

tomekpiotrowski:
Updated.

PR #1391, opened 2016-01-24
add vessel drag to kOS

### LINKED ISSUE #1390: Add vessel drag to kOS
Since 1.0.0 this can't be calculated via formula and needs to be read off of the vessel.

lamont-granquist:
closes #1390

### DIFF
--- a/doc/source/structures/vessels/vessel.rst
+++ b/doc/source/structures/vessels/vessel.rst
@@ -47,6 +47,7 @@ All vessels share a structure. To get a variable referring to any vessel you can
      :attr:`VERTICALSPEED`                 scalar (m/s)              How fast the ship is moving "up"
      :attr:`GROUNDSPEED`                   scalar (m/s)              How fast the ship is moving "horizontally"
      :attr:`AIRSPEED`                      scalar (m/s)              How fast the ship is moving relative to the air
+     :attr:`DRAG`                          scalar (m/s)              Drag on the ship due to air
      :attr:`TERMVELOCITY`                  scalar (m/s)              terminal velocity of the vessel
      :attr:`SHIPNAME`                      string                    The name of the vessel
      :attr:`NAME`                          string                    Synonym for SHIPNAME
@@ -210,6 +211,13 @@ All vessels share a structure. To get a variable referring to any vessel you can
     :access: Get only
 
     How fast the ship is moving relative to the air. KSP models atmosphere as simply a solid block of air "glued" to the planet surface (the weather on Kerbin is boring and there's no wind). Therefore airspeed is generally the same thing as as the magnitude of the surface velocity.
+
+.. attribute:: Vessel:AIRSPEED
+
+    :type: scalar (m/s)
+    :access: Get only
+
+    The drag on the ship due to the atmosphere.
 
 .. attribute:: Vessel:TERMVELOCITY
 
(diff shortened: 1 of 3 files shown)


doc/source/structures/vessels/vessel.rst
```diff
@@ -211,6 +212,13 @@ All vessels share a structure. To get a variable referring to any vessel you can
 
     How fast the ship is moving relative to the air. KSP models atmosphere as simply a solid block of air "glued" to the planet surface (the weather on Kerbin is boring and there's no wind). Therefore airspeed is generally the same thing as as the magnitude of the surface velocity.
 
+.. attribute:: Vessel:AIRSPEED
```
tomekpiotrowski: should be `Vessel:DRAG`

Dunbaratu: Is this a cut and paste mistake?  This is the same name as the suffix above.

lamont-granquist:
well, i thought this would work...

anyone wanna throw me a bone and help me out with compiling this on a mac so i can test locally?

hvacengi:
Travis build system claims that `Vessel` does not have a member of `drag`.  I checked the KSP assembly with ILSpy and it doesn't appear to exist in the current libraries.

Dunbaratu:
Is there a chance you're using an older version of KSP?  Supposedly `Vessel.drag` used to exist in the past but doesn't anymore.

lamont-granquist:
kinda, i was basing this off reading the internet and probably hit old documentation...

researching a correct fix...

lamont-granquist:
meh, it looks like this isn't easily accessible other than doing a lot of calculations...

https://github.com/johnfink8/GravityTurn/blob/ec64a25eb6f4de19e249fd85a666af889fbd59b9/GravityTurn/GravityTurn/VesselState.cs#L574-L795

https://github.com/MuMech/MechJeb2/blob/0a46a67eb2de6c38891d601bf980450b5e8dbdd6/MechJeb2/VesselState.cs#L588-L793

too much work for this afternoon...

Dunbaratu:
Yeah `Vessel.drag` was from the previous primitive atmospheric  model they were using before KSP 1.0, back when it was that easy.

lamont-granquist:
Yeah, but KSP itself has to calculate the drag and apply it somehow, it could have stuffed the new results in there...

Dunbaratu:
In KSP the `Vessel` class is really a convenience euphemism for "here's a collection of parts glued together".  Most of the actual physics work is done part-by-part.  For example, when you thrust an engine, the engine part pushes on the fuel tank part which pushes on the decoupler part which pushes on.... etc.  It doesn't apply the force to the whole vessel as one single unit.  So for drag they might not have needed to store it at the vessel level since that's not where the drag force gets applied - it gets applied part by part by part, and the effect on the vessel as a whole happens because of the pulling of parts on other parts.

hvacengi:
In addition, KSP has some values which are stored in members that are inaccessible to mods due to obfuscated names.  In those cases we can't even read the "name" in a way that it can mean anything to us.  Some of the names use escaped characters, so the variable may technically be named `\r\n`.

lamont-granquist:
Ah, that makes perfect sense.  So this number doesn't actually exist in the underlying engine then...

Might revisit this later, I was hoping to be able to quickly add some drag readouts to a kOS launch script i've been hacking on all morning, and I need a break from KSP now...

Dunbaratu:
(Added "wontmerge" label because of the abovementioned problems with it not compiling on recent versions of KSP's DLL's.)

Issue #1309, opened 2015-12-11
clear a file


dcrockwell:
I'm using a file to dump CSV data from each flight to plot, but I have to manually go into the file and delete the previous flight's data on each flight.

Is there a way to clear a file?

hvacengi:
I just delete the file when I need to clear it.  That doesn't work well if you're using another program to effectively stream the file data (since removing the file would likely thrown an error in the other program) but if you're just opening the csv file after the mission, `delete datalog.csv` would work.  The bigger trick is that it's a long process right now to find out if a file exists (which will be fixed eventually) so really I typically do this `log "\\" to datalog.csv.  wait 1. delete datalog.csv.`.  You can probably safely remove the wait, but I keep my script folder synced using dropbox, and the file is usually locked until it has been uploaded.

dcrockwell:
Even clearing the data in the file would be a big help.

```
CLEARFILE data.txt.
```

I have the CSV streaming the file into LiveGraph, so deleting the file doesn't work right. Either it doesn't actually delete silently, or it's unable to create a new log of the same name.

TDW89:
not sure if this will work with streaming but it might.
```
switch to 1.
log "" to datalog.csv.
switch to 0.

//then to clear datalog.csv
copy datalog.csv from 1.
```
it may be a bug but copy does not produce the file already exists that rename does it just overwrites the file it may be fast enough that it wont interrupt streaming.

Dunbaratu:
I think that regardless of the existence of these workarounds, it's still a good idea to add a clear file functionality to the system.

dcrockwell:
@TDW89 Cool workaround. I'm going to try it out now :)

dcrockwell:
It's weird, but @TDW89's code didn't work.

```
SWITCH TO 1.
LOG "" TO datalog.csv.
SWITCH TO 0.
COPY datalog.csv FROM 1.
````

Results in:

```
Object reference not set to an instance of an object
At interpreter history, line 20
COPY datalog.csv TO 0.
                    ^
```

Issue #1277, opened 2015-11-27
LIST FILES IN var. tweak?


crafty-geek:
(I'd add the Enhancement tag if I knew how to.)

Any chance `LIST FILES IN allFiles.` could be either changed or (more likely) extended in light of the recent LEXICON addition?  My suggestion: If the `allFiles` variable is already initialized to a Lexicon, have the C# code speedily generate and return a `Lexicon<string, FileInfo>` mapping `NAME` to the fileinfo that contains that name.

Currently, to check if a file exists (so as not to error out on a COPY or RUN or similar), one must loop through a (potentially long) list of fileinfos in Kerboscript, checking each entry against the file of interest - an O(N) operation at Kerboscript speeds.  The Lexicon, however, provides (I assume) a C# backend which, from a Kerboscript perspective, would be an O(1) operation or damn near close to it.

I understand you don't want to break legacy code, hence why I'm not suggesting changing the output altogether; making it type-sense the variable it's asked to populate almost by definition means it won't break legacy code, as I see no scenario in which someone would choose to have their lexicon overwritten and List-ified unless they were then going to pass it off to a Lexicon-izing function, eg what I currently do to make my filesystem-related code run reasonably fast:
````
global directory is lexicon().//at least it is most of the time
//Populates directory with the local drive's contents.
FUNCTION FS_UPDATE_DIRECTORY{
  LIST FILES IN directory.//No need for a local
  SET directory TO _lexify_fileinfos(directory).
}
FUNCTION _LEXIFY_FILEINFOS{PARAMETER fList.
  LOCAL retLex IS Lexicon().
  FOR fInfo IN fList{ retLex:ADD(fInfo:NAME, fInfo). }
  RETURN retLex.
}
````

If type-sensing like I describe is nontrivial, perhaps a new 'phrase' that doesn't invoke the LIST behavior could do the same thing?

Also, might this behavior be useful for some other Listable keywords?  (Bodies and Vessels have no need of this with the ability to do, eg `BODY("MUN")`, but it would be nice to extend to Processors, Resources, Parts [which until the String update you had to search for with the full name anyway], Elements, and subtypes of Part)

Dunbaratu:
[...]

Issue #1238, opened 2015-11-09
Reading files to string


thexa4:
KOS currently allows you to write to files but there doesn't seem to be a way to read them back in which would be great for things like preprocessors during launch and as a substitute for user input by giving the user an edit window during program execution.

space-is-hard:
Currently you can read from files in a certain way. If you need to log a string for future use, you can

    LOG "GLOBAL inputString TO " + outputString + "." TO inputLog.ks.

and then later

    RUN inputLog.ks.
    PRINT outputString.

erendrake:
@thexa4 we are hoping to address this before the next major release.

Dunbaratu:
Check this issue again after we've fully tested the changes associated with PR #1269

PR #1393, opened 2016-01-27
Read and write file contents

### LINKED ISSUE #1095: LIST:CONTAINS() always false for file names.
when I call `myList:contains()` on a list that was obtained via `LIST FILES IN myList.` it always returns false even when iterating over the elements finds a match.

example:
```
log "hi" to "testfile.txt".

list files in haystack.
local needle is "testfile.txt".

print "looking for '"+needle+" in "+haystack.

print " results in "+haystack:contains(needle).

for f in haystack {
	print "match for '"+f+"' is "+(f=needle).
}
```
kos_contains_filename

On a related note: could we get a way to check if a file exists? listing files and, as a workaround for this ticket, iterating over them seems like a lot of work. (i want to force a rename and MAY have to delete an existing file and to delete it, it has to exist)

Edit: kOS v. 0.17.3.0, the newest version on CKAN as of yesterday.

tomekpiotrowski:
Fixes #1095, #1277, #1238, and #1309.

**Breaking changes**:
* `FileInfo` was removed, all it's functionality has now moved to `VolumeFile`
* `VOLUME:FILES` is now a Lexicon instead of a List
* default extensions are now not added when new files are created, they are still however added when accessing files (`CREATE TEST.` will create a file named `test` instead of `test.ks`, but `RUN TEST.` will try running `test` first and `test.ks` if `test` does not exist)

It would be really cool if someone found a moment to test this on Windows.

My test script as an example:
```
set filename to "testfile".
set original_volume to core:currentvolume.

function perform_test {
  set f to create(filename).
  f:writeln("xyz").
  f:writeln("xyz").

  print f:size = 8.

  set c to f:readall.
  print c:length = 8.

  set count to 0.

  for line in c {
    set count to count + 1.
    print line = "xyz".
  }

  print count = 2.

  f:clear.
  print f:size = 0.
  print f:readall:empty.

  set f to open(filename).
  print f:size = 0.
  f:writeln("abc").
  print f:size = 4.
  log "log" to filename.
  print f:size = 8.

  writejson(f:readall, "content.json").
  set f to open("content.json").
  set copied to create("content_copy.json").
  copied:write(f:readall).

  set readcontent to readjson("content_copy.json").

  print readcontent:length = 8.

  print core:currentvolume:files[filename]:size = 8.

  print core:currentvolume:delete(filename).
  print core:currentvolume:delete("content.json").
}

print "All lines should read 'True'".

print "--- Testing on Archive ---".
switch to 0.
perform_test().

print "--- Testing on Harddisk ---".
switch to 1.
perform_test().

switch to original_volume.
```


tomekpiotrowski:
Btw, this introduces 2 minor breaking changes. It also changes the internals of how files are handled in kOS (`ProgramFile` is gone for example). I will do a more detail write-up when I will push the final version.

Dunbaratu:
I'm writing up the changelog - can you remind me what the breaking feature is that the 'Breaking' tag was about?  I usually like to mention them in the changelog as a warning.

tomekpiotrowski:
There's a 'Breaking changes' section in the first post :)

Dunbaratu:
Sorry you're right.  I was going through these too fast and didn't catch it.

src/kOS/Function/BuildList.cs
```diff
@@ -40,7 +40,7 @@ public override void Execute(SharedObjects shared)
                     list = shared.Vessel.PartList(listType, shared);
                     break;
                 case "files":
-                    list = ListValue.CreateList(shared.VolumeMgr.CurrentVolume.GetFileList());
+                    list = ListValue.CreateList(shared.VolumeMgr.CurrentVolume.FileList.ToList());
```
hvacengi: I'm getting an encapsulation exception when running `list files in filelist.`.  It's complaining about a key value pair, so I suspect that this line is supposed to be `FileList.Values.ToList());` which should only put the `VolumeFile`s themselves into the list.  I will test locally to confirm.

tomekpiotrowski: I think you're right, somehow I haven't tested this.

hvacengi: Confirmed, `Values` works.

I found it because I specifically wanted to test to make sure the old work around still works.  So I'm running it on old scripts without implementing the new functionality first.

PR #1395, opened 2016-02-01
Issue #1009 update ec usage

### LINKED ISSUE #1009: Update EC usage to be based on actual processing
Now that EC usage is updated using fixed update, it would be nice to also adjust consumption based on actual instructions per update.  Currently consumption is based only the required power of the current volume.  It would be nice to have a hybrid model:
`[Required power] = [volume power per seconds] * [time delta] + [part power per instruction] * [number of instructions]`

Care will be needed with all aspects of balancing the consumption.  We should make sure specifically to balance the volume's requirement against the large time delta's that come with high rates of timewarp (maybe even a timewarp reduction factor) to prevent consumption from ballooning to unreasonable levels.

In the case of Archive access, it currently has a large power draw to account for the need to communicate with Kerbin.  If RemoteTech is installed (which enforces power requirements for antennae) should we reduce or eliminate this penalty?

hvacengi:
Fixes #1009 
Moves the power consumption from being a hard coded number to something configurable in the part.cfg file.  There are two values, `ECPerInstruction` and `ECPerBytePerSecond`.  For the first pass, these values were set to 0 and 0.000004 respectively, to match the current balance.  Ideally, we would like to reduce the dependence on hard drive based power and focus on instruction based power, but the exact balance, and more importantly the individual part balance, will need to be worked on.

@Dunbaratu expressed interest in reducing the hard drive based number to zero, making mass and cost the only penalties for increasing drive capacity via tweakables.

TODO:
- [x] Base EC usage on instructions/bytes
- [x] Move EC usage info to KSP fields to house in part.cfg
- [x] Update all part.cfg files
- [ ] Balance EC usage for parts
- [ ] Documentation


src/kOS/Module/kOSProcessor.cs
```diff
@@ -75,6 +75,19 @@ public string Tag
         [KSPField(isPersistant = true, guiActive = false)]
         public int MaxPartId = 100;
 
+        // This represents how much EC to consume per executed instruction.
+        // This would be the "variable" component of the processor's power.
+        // TODO: This value needs to be changed to a non-zero number and balanced.
```
hoylemd: Did you create an issue for this?  It should be linked to in this PR.

hvacengi: The todo is part of this PR.  It needs to be done before merging.  But I didn't want to presume to change it myself without showing it to every one else first.

src/kOS/Module/kOSProcessor.cs
```diff
@@ -75,6 +75,19 @@ public string Tag
         [KSPField(isPersistant = true, guiActive = false)]
         public int MaxPartId = 100;
```
hoylemd: This is *kind* of a magic number.  Perhaps defining constants should be done all together, so when someone balances this they don't have to hunt all over the file.

src/kOS/Module/kOSProcessor.cs
```diff
@@ -628,12 +637,29 @@ private void ProcessElectricity(Part partObj, float time)
...
-
-            var newMode = (result < 0.5f) ? ProcessorModes.STARVED : ProcessorModes.READY;
-            SetMode(newMode);
+            double volumePower = 0;
+            var volume = shared.VolumeMgr.CurrentVolume;
+            if (volume.Name == "Archive")
+            {
+                // TODO: Like with the default value of ECPerBytePerSecond, this corresponds to the existing value.
+                // It should be adjusted for balance.
+                volumePower = 50000 * ECPerBytePerSecond;
```
hoylemd: Magic number here.  that 50k should have a name, so it's immediately clear what it represents.

src/kOS/Module/kOSProcessor.cs
hoylemd: :+1:

Dunbaratu: While it's true that we need to fiddle with the number to find the right balance, the placeholder we use in the meantime should probably be something other than just `0`, so we can see it at least having an effect of some sort and be able to test that the code is using it properly.

Dunbaratu: In fact, now that I think of it, we really should just do that during testing.  Let's leave this as hardcoded to `0F` in the code now and forever, and instead play with values by fiddling with the part config files - that way we're doing the same thing the end user can, by "pulling on" the same handle they have available to tweak, and therefore testing even more of the whole chain of events.

Just put a comment in here in the code mentioning that the only reason this field is zero in the code is because the Part.cfg files override it, and it's not *really* zero when the game is running.  (otherwise reading the code may give people the wrong idea down the road).

src/kOS/Module/kOSProcessor.cs
```diff
@@ -82,7 +95,7 @@ public void Activate()
             OpenWindow();
         }
 
-        [KSPField(isPersistant = true, guiName = "Required Power", guiActive = true)]
+        [KSPField(isPersistant = true, guiName = "kOS Required Power", guiActive = true, guiActiveEditor = true)]
```
Dunbaratu: Are there any means of having tooltips on these fields?  It would be nice if we could explain that this is the power if and only if the cpu is running at 100% and not doing any waits.

src/kOS/Module/kOSProcessor.cs
hvacengi: That's exactly what I was thinking.  This is just a default value.  I'm not going to make a whole commit for just this, but I will add a documentation note.

src/kOS/Module/kOSProcessor.cs
```diff
@@ -628,12 +637,29 @@ private void ProcessElectricity(Part partObj, float time)
...
+                volumePower = 50000 * ECPerBytePerSecond;
+            }
+            else
+            {
+                volumePower = volume.Capacity * ECPerBytePerSecond;
+            }
+            RequiredPower = (float)(volumePower + SafeHouse.Config.InstructionsPerUpdate * ECPerInstruction / TimeWarp.fixedDeltaTime);
+            var request = volumePower * time + shared.Cpu.InstructionsThisUpdate * ECPerInstruction;
+            var available = partObj.RequestResource("ElectricCharge", request);
+            if (available / request > 0.5f)
```
Dunbaratu: Why base this on a ratio instead of just stopping when `available` goes to zero or goes negative?

src/kOS/Module/kOSProcessor.cs
hvacengi: Actually, as written now:
```
RequiredPower = (float)(volumePower + SafeHouse.Config.InstructionsPerUpdate * ECPerInstruction / TimeWarp.fixedDeltaTime);
```
So, when the game is actually running, it is the real current power requirement.  When in the editor, it is "max" power (based on current IPU), but in flight it is the real value.  You may also notice that I made it visible in the editor, so when you change the "max ipu" setting (which is not yet obsolete) or the disk size you can see the effect.  I expect that we will set the IPU per cpu using a tweakable as well, so the ability to observe how your changes effect power draw may be very useful, especially when designing RemoteTech satellites.

Dunbaratu: > So, when the game is actually running, it is the real current power requirement.

In both places where the code is calculating this `RequiredPower` value for display, both here AND down at line 653 below where it is calculating it during flight, it's using this same formula.  It never uses the real runtime drain calculation based on `shared.Cpu.InstructionsThisUpdate`.  It uses `Safehouse.Config.InstructionsPerUpdate` in both places.

That was the difference I was talking about.  A script that is doing a `wait until foo.` is yielding most of its IPU.

I made my comment about the tooltip because I thought this was deliberate and you *wanted* to display the same calculation in the ediitor as during runtime.  But your comment that you wanted to display the actual value at runtime makes me think there may be a cut and paste error here because I see InstructionsPerUpdate being used the same way in both places.

hvacengi: Yeah, I updated the comment after I posted and noticed the config.  It will be getting fixed shortly.

src/kOS/Module/kOSProcessor.cs
hvacengi: The "0.5" ratio is just what we had in there previously, and I think I've seen it used when poking around the source of other mods.  Part of me thinks that there is some practicality when dealing with time-warp conditions.  But I didn't give it much thought.

Dunbaratu: It looks like the logic is something along the lines of finding a way to take the time when power depletion will happen and "round it" up or down to the nearest FixedUpdate boundary.  If there's enough EC left for somewhere between 50% and 100% of a full physics tick, then "round up" to doing one more full physics tick.  If there's between 0% and 50% as much as you need for a full physics tick, then "round down" to zero more ticks.

But in our case I think this logic may not be totally appropriate as we have a varying drain per update.  The current update tick may have taken a lot more power drain than the next one will.  

But in the end it probably doesn't matter much.  Nobody's going to notice that the CPU got power starved half a second sooner than it should have because it treated a battery charge of 0.1% as if it was 0.0%.  Never mind my complaint.  Just leave it.

src/kOS/Module/kOSProcessor.cs
hvacengi: OK, I was just about to make this edit, when I realized you're right.  I did indeed make this the "Max" power, and that was intentional (at least at the time).  I did that so that the number wasn't bouncing every single update, because if you can't read the value it doesn't do any good.  Imagine the display rapidly changing between "0.2", "0.01", and "0.000001".

In that case, while we can't do a tool tip we can modify the string to convey the meaning.  Maybe "kOS Max Power".  I'm also going to add `guiUnits = "EC/s"` to the definition, and round the display value to make it more readable.

I apologize for the confusion.  I wrote the code a bit before I actually submitted the PR and it wasn't until I dug into the code that I recognized what I did and why.

Dunbaratu: Maybe 'kOS Peak Consumption"?

Be careful about rounding the value that it doesn't round down to zero in cases where the value is very small and people have reduced the power setting a lot, like RO seems to be planning to do.  Maybe round it by number of significant digits rather than by a fixed decimal place, so when the number is small it gets more places to the right of the decimal point.

It occurs to me it may be useful for us to maintain a memory of how much power got used during the last few fixedupdates, and say every 10 or 20 fixedupdates, get the average and display it in the context menu alongside the max value.  That's good enough for visual display and yet it smooths out the bouncy number problem you're talking about.

(And maybe we do actually want to expose the most recent power drain value to scripts too, as a suffix of `core`.)

hvacengi:
One of the things we have to address in the balance is the mismatch between what the editor claimed was the max power, and what was reported by the module itself in flight.

Editor: All parts report "Max Power consumption, EC/s: 0.2"

Flight: Parts report "Required Power: [x]" where x = disk size * 0.04 / 10,000.

Max power consumption appears to align with the power consumed when the archive is the currently selected volume.  At one point this was indeed the highest drain any processor may experience, but not with the larger capacities of newer parts.

So, when we are adjusting the balance, I am presuming that the actual base line power consumption is 0.04EC/s (which is the actual power requested by the 1.25m part, with a 10k drive).  If we presume zero for drive power, then that works out to 0.000004EC / instruction, at 200 instructions / frame and a frame time of 0.02s., or 0.000008EC / instruction at a frame of 0.04s.  I would still like to add a small component based on drive capacity, but this would be a good place for us to start.  I'll update the PR with changed part.cfg files.

src/kOS/Module/kOSProcessor.cs
hvacengi: Actually, I wrote a "Moving Average" class for use with the steering manager anyways, we could certainly use that class to keep track of the historical values.  In fact, that's what I do with it, to de-bounce the "observed" torque and the "update time".  I'll update to use a moving average.

I didn't notice your comment about title before committing "Max" but I like "Peak" better.  I'll probably plan on changing that.

Regarding your last point, while going through this, I decided I'd submit an issue for a "debug info" type class which could keep track of "instructions so far", "total instructions so far", variable count, trigger count, and any other information that we might be interested in knowing about operation.

src/kOS/Module/kOSProcessor.cs
hvacengi: I don't really know what this field does, and since it isn't really related to the PR I'm going to leave it as is for now.  If you want to track it down and get it changed, we'd certainly appreciate the assist.

hvacengi:
I just pushed a revision to the logic.  I implemented a few of @Dunbaratu's suggestions regarding averaging, and I fixed a major bug introduced in the last commit: requesting zero power resulted in permanently being in a STARVED state.

I like the way it's working right now.  The average power is a nice touch, and seems to stay pretty steady (when limited to 3 decimal digits).  There's still a lot of work to do with deciding the power requirements of each processor, but that's probably going to be a bit of a team effort.

I also still need to document this info for the end user.

src/kOS/Module/kOSProcessor.cs
Dunbaratu: I'd say get rid of it.  A search of the code shows that it's literally not used anywhere.  This is the only place it gets mentioned.  I have no clue why it's there.

It is, however, mentioned in the tutorial craft file, here: doc/source/_images/tutorials/quickstart/MyFirstRocket.craft , but that's only because being a KSPField, KSP automatically spits it out into the craft file, and when that tutorial was written it just took the craft file from KSP.  (So it would have to be removed from there too).

Which brings to mind a question.  If a save file or craft file contains a KSPField that got removed between versions, does it break the ability to reload the save?  If so then we may have a problem with removing this.

hvacengi: I believe my experience has been that extra field info is ignored if there is no matching field on the C# class.  The same is true for unrecognized modules.  It is possible that the next save will not include the data, but that shouldn't be a problem if we remove the reference entirely.

Dunbaratu:
In responding to the issue above about MaxPartId being removed from the craft file for the tutorial, a potential problem occurred to me:

Earlier we'd agreed that `ECPerInstruction` should just be zero in the code, and be overridden by what ever is in the config files, and we put the real setting in the config files instead.

But I just realized that doesn't work for loading up existing crafts and existing saves that were made prior to kOS 0.19.0 coming out.  The existing saves and craft files won't have this field in them, so the hardcoded default from the code will get used instead.

Thus when we do decide on the proper setting to default it to, we do have to make the ECPerInstruction field of the class also use that value as its default.  We can't leave the hardcoded default at zero because the default gets used by all old files that are missing the value.  Old vessels and old crafts would have kOS CPU's on them that get to run for free, without using power at all.

Dunbaratu:
Is this still "notready"?

hvacengi:
I'd say it isn't ready until we do some work to balance the EC usage.

Regarding the parts already in flight, because the EC fields are marked with persistence set to false, the values should always read from the part.cfg file, not from the save file.  I will do further testing to confirm.  But if this is true, then we have instant backwards compatibility.

Dunbaratu:
I'd like to get this in the 0.19.0 release.  But it should be flagged as Breaking too, just because in the past IPU setting had zero effect on power usage, and now it will.  So it's possible for some previous designs to now eat more power than they used to and thus not work as well.

Dunbaratu:
I always thought that our current usage was a bit high:  at 200 IPU the new settings would consume 1 EC per 50 seconds (presuming a program with no waits in it.)  Lately a lot of people are writing code assuming higher IPU, but even so, the mere fact that while waiting and doing nothing a long time you won't eat nearly as much power I think will have the effect of lessening power usage quite a lot even *with* the same power consumption we had before.  If you have no triggers active and are just in deep space doing a ``wait until ship:body = Mun.``, for example, you'll only be using about 5 instructions per update: push ship, push body, getmember, push Mun, equals.  (when that evals to false, it doesn't do any more instructions.)

At 5 IPU, we end up needing 2000 seconds (~ 33 min) to consume 1 EC.  If you haven't built enough recharge panels for that then that's too bad.

So I think that, balance wise, it's fine.  It's as much drain as it always was when at full load doing something like landing or taking off, and waay less when in wait state.  It's these long deep space waits that present the biggest problem with power running out anyway, and we've greatly reduced their usage unless the script is actively busy-looping during that time, which is sort of the user's own fault.  (When we document, we will explain that using wait to yield time reduces electric usage.)

The only problem I see is with the people who have been running code at higher IPU, like 2000, and expect everything to work like it always did.  But they will have the config file and MM if they want to change their settings.

I think the only real balance issue to discuss is whether we want the different upgrades of the part to have differing power drains.  I think we might, since one day we want to have the higher models allow faster IPU than the lower models, and if they do so only at higher power drain, it's sort of punishing and feels a bit backward.  Modern tech requires less power than older tech to get the same result when it comes to computing.  Right now, even without the code for making different CPUs have different IPU, we still get pretty much the same effect from having different players choose different IPU settings.  We just tell the ones who want to use higher IPU settings that those settings are really meant for the higher tech tree parts (which we will have a reduced ECPI (Electric charge per instuction), and they can reduce the drain by using the higher parts.

In summary, I think the balance is actually fine like it is in your code, in comparison to current behavior.  And for the RO people who want it changed, remember that they're never happy with what we pick anyway, and they can override it with what they like.

Dunbaratu:
I just realize I merged this while it still said "Pending Documentation".
I'll make documentation a second separate issue.

PR #1400, opened 2016-02-04
Collections force encapsulation

Dunbaratu:
This is the big refactor for all user variables and stack values to be encapsulated.
The goal is to make it so that the user never sees a raw `object` again.  All values that can end up in a user variable or in a stack'ed expression they will see must derive from Structure.

Also includes some refactors done by erendrake to the suffix system.


Dunbaratu:
Hang on, @hvacengi  I didn't realize this would have merge conflicts.  I thought it was sourced entirely from develop and develop hadn't changed since then, but apparently it has.  I'll merge develop into my branch first so you don't have to deal with that.  adding the 'dont merge' tag for now.

Dunbaratu:
Okay, merge conflict resolved.  It was an easy one - one of those where git saw a conflict but there really wasn't one - just to unrelated edits that were near each other.

hvacengi:
I will start testing and reviewing tonight.  I should have plenty of time to do so over the next 2 days, with any luck we can get it merged before Saturday evening.

thexa4:
I seem to get an error on this branch when using delegates:

    function foo {}
    local bar to foo@.
    bar:call().

Throws: impossible_FromPrimitiveWithAssert(kOS.Safe.Execution.KOSPassThruReturn) was attempted.

It doesn't seem to happen on develop and 4c170b761e8d59adc986225bd21dcab6a09cadd1.

Dunbaratu:
Thanks for the report.  We're still working through problems in that branch.

Dunbaratu:
@thexa4 I am able to recreate the problem on the branch I have so I'll try to solve it there and hope @hvacengi can merge my branch into what he's doing when I'm done.  I'm a revision behind what he was working on, but his work isn't pushed to his fork yet so I can't work from it.

Dunbaratu:
@hvacengi and @thexa4   : I discovered the cause of the problem @thexa4 reported above and pushed a fix to my fork's branch `collections_force_encapsulation`.  However that branch doesn't have the recent work done by @hvacengi so it has to be merged with what he did.  Either he can merge what I did, or I can merge what he did - either way is fine, but I won't be around to do it Sunday until evening so it would probably get done sooner if @hvacengi merges it into his work.


The gist of the probem is that when @erendrake updated all the suffixes to force args and return types to be derived from `kOS.Safe.Encapsulated.Structure`, he missed a spot - the VarArgsSuffix used by the `:CALL()` and `:BIND()` suffixes of KOSDelegate.  Thus when those suffixes still returned a non-`Structure` return type, this fact wasn't caught by the compiler.  It's not until runtime that it noticed it can't convert their return types into the `Structure` return type mandated by his new `ISuffixResult.value` property.

They return a magic 'dummy' return type called `KOSPassThruReturn`, which is what was appearing in the error message.  This return type means "this C# Delegate can't give you the return value because that's the job of the user code subroutine (not C# code) that is about to be jumped to starting with the next Opcode after this one.  Until that code runs, we don't have the return value.  Therefore unlike all the other suffixes, this one shouldn't push any return value on the stack, leaving that for the user code to do instead."  this dummy return, `KOSPassThruReturn` was not derived from `Structure`, and is not a primitive that can be changed into `Structure` by `FromPrimitiveWithAssert()`, thus the complaint.  My fast fix was to just make it be derived from `Structure` so `FromPrimitiveWithAssert()` lets it pass through without complaint, so it can reach the rest of the (already existing) code that detects the return value this magic dummy and acts accordingly.


If that was all I fixed it would have been a 1 line change.

But I decided to also fix the underlying problem that caused this case to slip through in the first place - which is that we had a few suffixes that were still allowing a generic type of `<object>` to exist.  The rest of my edits were about the cascading edits necessary to repair that problem and change the one or two other Suffix types that were missed.

hvacengi:
You can find my current review branch at this address: https://github.com/hvacengi/KOS/tree/review-1400-force_encapsulation

I'm going to merge in the changes that @Dunbaratu updated onto his PR, and then re-run all of my tests.

@thexa4 I'll try and update my test scripts to call delegates so that I can test them as well.  I'll post here when I'm finished with the update so that you may run your tests as well before I PR back to Dunbaratu.

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -685,8 +685,8 @@ private void PreProcessUserFunctionStatement(ParseNode node)
 
                 // build default dummy function to be used when this is a LOCK:
                 currentCodeSection = userFuncObject.GetUserFunctionOpcodes(0);
-                AddOpcode(new OpcodeArgBottom());
-                AddOpcode(new OpcodePush("$" + userFuncObject.ScopelessIdentifier)).Label = userFuncObject.DefaultLabel;
+                AddOpcode(new OpcodeArgBottom()).Label = userFuncObject.DefaultLabel;;
```
hvacengi: Won't this cause the ReplaceLabels method in ProgramBuilder to change the default label to point at "ArgBottom" opcode?

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -1141,14 +1165,17 @@ public class OpcodeLogicToBool : Opcode
 
         public override void Execute(ICpu cpu)
         {
+            // This may look like it's just pointlessly converting from a
+            // ScalarBoolean to a primitive boolean and then back into a
+            // ScalarBoolean, and in the case where the operand was already
+            // a ScalarBoolean that would be true.  But the purpose of this opcode
+            // is to also change integers and floats into booleans. Thus the call to
+            // Convert.ToBoolean():
```
hvacengi: Based on our conversation in slack the other day, I didn't think that using numbers as boolean values was supported.  But working with this branch, the following test works:
```
set x to 1.
if x and true { print "hello world". } // print's "hello world"
```
We could probably streamline this code, and make it less confusing, by using `Convert.ToType` and enabling a conversion to `BooleanValue` from `ScalarValue`.

But you and I had talked about omitting the feature, so I didn't want to update it as part of the review without discussing it first.

src/kOS.Safe/Compilation/KS/Compiler.cs
Dunbaratu: Correct.  That was the point.
The bug I was fixing is that it was jumping to the wrong line.  It's *supposed* to jump to the ArgBottom.  That's the first line of the routine.  By skipping past it before, it was causing the stack to not consume the argbottom mark.  Over time that clogged the stack with argbottoms that weren't being popped, each time you did a manual `run` commnd from the interpreter it left one behind for each lock steering and lock throttle in the programs (this bit of code is only run at the top of a new program context, so this only happened once per *manually* typed run command.  So it took time to build up enough cruft in the stack to notice the problem.

src/kOS.Safe/Compilation/Opcode.cs
Dunbaratu: Hmm. if it already works lets not omit the feature after all.

By the way, the OpcodeLogicToBool seems to be only used in one place, the `toggle` command.

src/kOS/Function/FunctionBase.cs
```diff
@@ -35,7 +35,7 @@ public object ReturnValue
                 internalReturn = value;
             }
         }
-        private object internalReturn = null;
+        private object internalReturn = 0; // really should be 'null', but kerboscript can't deal with that.
```
hvacengi: I think we need to make a decision about how to handle empty returns.  In some instances, we return "null" which the users can't deal with.  Setting the return to 0 doesn't really help them any better (is the return value 0 because it's really 0, or because there is no return?) but at least it provides an object.  If the return value is to be left as 0 in FunctionBase, I think that we should have line 139 of DelegateSuffixResult match.  I'll make that change, unless some one disagrees.

src/kOS/Function/Math.cs
```diff
@@ -14,7 +15,7 @@ public override void Execute(SharedObjects shared)
             double argument = GetDouble(PopValueAssert(shared));
             AssertArgBottomAndConsume(shared);
             double result = Math.Abs(argument);
-            ReturnValue = result;
+            ReturnValue = Structure.FromPrimitive(result);
```
hvacengi: The `ReturnValue` property definition automatically runs the `value` through `Structure.FromPrimitive`.  Should we be passing the value through it twice, or is it a cheap enough function that it's better to be safe than sorry?  This could effect all of the function's that change this value in the PR.

src/kOS/Function/FunctionBase.cs
Dunbaratu: I agree that they should match.  As to the conflation between a real value of 0 and a dummy value of 0, remember that these are only used in cases where the function is essentially void anyway, so there shouldn't be such a clash.  i.e. you'd have to be doing something like `set x to mylist:clear().` instead of just what you're supposed to do with is just `mylist:clear().` to be getting that return value.  So it's only happening in cases where the docs don't mention any intended return value you should be looking for.  There can't be a clash between "it returned meaningful zero" and "it returned dummy zero" because this will only happen in cases where there isn't a documented intended meaningful return value.

src/kOS/Function/Math.cs
Dunbaratu: Chances are that I started doing it this way first before noticing that ReturnValue does it for me.  I don't think it does that much harm, but it can be removed to make it cleaner.

The very first line of FromPrimitive is to abort and return as-is if the argument is already derived from `Structure`, so it's not going to be that expensive.  Still, for the sake of code clarity, go ahead and remove the redundant call.  (Note, *Slap* ReSharper, THIS is what the word "redundant" means - it means doing a thing two times instead of one.  If both the implicit and explicit way of doing something both still result in it happening only exactly once, that's NOT redundant, darn it.)

src/kOS.Safe/Compilation/KS/Compiler.cs
hvacengi: Roger.  Will leave it where it is.

src/kOS.Safe/Compilation/Opcode.cs
hvacengi: Odd, if it's only used for toggle, I don't know why the bool stuff works otherwise.  I'll update it to use `Convert.ChangeType` instead, just to clean it up.

src/kOS/Function/FunctionBase.cs
hvacengi: That logic works for me.  I'll update the `DelegateSuffixResult` to match.

src/kOS/Function/Math.cs
hvacengi: Oh, I actually picked it up visually (don't have ReSharper, just CodeMaid, and that only got run on a couple of the files I reviewed).  I don't have a problem leaving it there.  To some degree I like leaving things explicit.  But at the same time, we didn't modify most of the other functions to match.  So I think I'll pull it out of these classes for the sake of consistency.

Dunbaratu: Oh that wasn't aimed at you.  It was aimed at an edit that Resharper kept doing to my PRs when erendrake ran it that irked me.  Choosing whether an initializers is explicitly mentioned or not might not matter to the compiler, but it does matter to the human reading the code.  It communicates the difference between "initial value = The fact that it will start at zero is irrelevant" versus "initial value = my algorithm explicitly wanted it to be zero".  When my careful decision to mention some initializers explicitly when they were relevant to the algorithm kept getting squashed and removed under the false label of "redundant" by Resharper, it annoyed me.  Unless the compiler is really really dumb, writing something like `int x = 0;` isn't going to end up redundantly initializing it to zero twice.  It's going to notice there is an explicit initializer and use that *instead of* the implicit one, not *in addition to*, which is what the term 'redundant' would imply.

hvacengi:
@thexa4 My review branch has been updated with all of the review edits.  Would you mind verifying that delegates are operating as you expect with my changes? https://github.com/hvacengi/KOS/commits/review-1400-force_encapsulation

Issue #1257, opened 2015-11-18
A way to allow a RUN statement to work on runtime-changing instructions.


Dunbaratu:
Currently (0.18.1 and before) You cannot RUN on a varying filename string.

i.e. you can do this:
```
run myprog.
```
but you cannot do this:
```
set filename to "myprog".
run filename.
```

And the main reason for this restriction is that the *compiler*, at *compile time* builds the thing we've been calling the "maybe compile function" in the program header *before* the values of variables are known and it needs to know the hardcoded filename at that time, because it's using that to choose the name of the program instruction pointer variable.

But this seriously needs to be fixed before we can allow multiple directories to work.

Here's one approach:  When the compiler builds the "maybe compile function", instead of building one per instance of `run`, just build one global one only, no matter how many `run` statements there are.  However, make it able to receive the name of the program instruction pointer variable as an argument.  This is essentially an indirect pass by reference (i.e. pass in the instruction pointer variable so it can be set by the maybe-compile-function, rather than its value).  Such a pass by reference would be impossible in kerboscript directly, but we can do it in the underlying kRISC, by just pushing the variable name as a bareword string with an asterisk at the end of it.

gisikw:
Mind pointing me at the "maybe compile function"? I've been trying to trace it up, but got myself a bit confused, but would love to take a look at this, especially as it would offer a workaround for the lib_exec breakage (since users could time/version-stamp files prior to execution).

Dunbaratu:
@gisikw In general one hint about getting a dump of what the actual opcodes of a program look like is this:  Just put a runtime-error in the bottom of the script - something that will compile but will barf when it runs it.  I usually use ``print 1 / 0.`` to force a divide-by-zero runtime exception for this purpose.  Anyway, the point is that when you get such a runtime exception, the program dumps a bit of the program out to the output log file as part of the error.  If you're running the debug DLL (the one compiled from source here, rather than the public one) then it tends to print the entire program's opcodes out (the public one just prints a few lines of context, I believe) on any runtime error.

If you want to look at the section of the compiler that builds this "maybe compile function", it's in this file:
   src/kOS.Safe/Compilation/KS/Compiler.cs  -- look for a method titled PreProcessRunInstruction()

That method has a big chunk of opcodes it inserts into the program that effectively does this:

Create a low level "function" that will be put into the the top of the program as follows:

[...]

Issue #148, opened 2014-06-17
feature suggestion: IS NULL or isset()


ghost:
Maybe I'm mistaken but I think kOS lacks a way to check whether a variable is defined or not, like isset() or comparison against a NULL constant. All tests I've done yield a "variable myVar is not defined" and end the script execution, but I might have missed something. 

It'd become very handy in combination with variables that are passed to a script thru declared parameters, as it would essentially allow optional parameters.

abenkovskii:
Isn't it a duplicate: #11?

Dunbaratu:
Sort of.  #11 is a 'discuss', this is a 'do it'. issue.

PR #1403, opened 2016-02-06
Added assembly format

thexa4:
This allows people to work around certain features available in the machine code but not in the script language.

Examples of this are:
- embedding quotes in strings
- executing programs without caching #1257
- is NULL and isset checks #148 

This would work best alongside a feature to read a file to string.


Dunbaratu:
While this is a lot of great work, I really wish you'd started this off with an issue first for discussion, instead of silently working on such a big thing without telling anyone, and without any planning stage where we discuss how to make it look.  There's an enormous refactor going on under the hood that will possibly invalidate some of this work, or at least make it have to be edited a bit, before it can be merged.  Also, the lack of NULL checks is because of deliberate design by the mod organizer (that I don't agree with), not *inability* to implement it.

Also, "a feature to read a file to a string" is already in the issues list and there's already a PR for it.

But here's the biggest problem:  I've toyed with the idea of making an inborn ASM syntax inside of the actual script langauge (rather than with a separate parser grammar file so it has to be a separate compilation file), but was putting it off precisely because once you expose the opcodes directly to users to write programs with, they become a sort of public promise: an API.  That means changing how the VM works becomes a massive problem because it breaks backward compatibility in a big way.

Right now whenever we change how the VM works, it meas we include a note in the release notes that says "you have to recompile your KSM files for this release".  But with your proposal, people would also have to re-write their KASM hand-written code whenever the VM changes.  And that ties our hands a bit - it prevents us from making some kinds of VM changes down the road.


That's why I'm a bit conflicted about this.  It's wonderful, great work, and I love the idea.  But by having worked it all out ahead of time before announcing anything publicly at all about it, and only mentioning it now at the tail end of your work, it prevented any debate over the details of its design.  Now if I say things should be done differently (or worse yet that the idea shouldn't be done at all), then I feel like an ass because of all the work you've put into it.

thexa4:
I made this primarily for myself to create files like the char_spec library and because I encountered the quoting and run issue which you can work around if you edit the bytecode. I wouldn't mind if it isn't merged, I'll just keep it in my repo in that case. If you do decide it would be a worthwhile feature I'm willing to put in work to change it. 

As you currently need the spec_char file for some operations while it can't be created in game I feel that the opcodes are already used as a public API. If needed we can add a warning to the assemble and disassemble functions that the opcodes will most likely change between versions and that it should be used minimally.

Is there a branch where the refactor is happening?

Dunbaratu:
It's in the PR #1400 , which is about to get updated one more time by hvacengi today or tomorrow before it gets merged.

thexa4:
PR #1400 doesn't seem to change how the opcodes work. I have a merged branch here:
https://github.com/thexa4/KOS/tree/feature/assembly-fixed

Dunbaratu:
if you push to the same branch as the PR, the PR will automatically update.

hvacengi:
Re: character stuff: #1343 added a char function to help with this ability.  It enables `print char(34) + "Foo" + char(34).` to display "Foo" (with the quotation marks).

The `run` issue is a very complex issue, in terms of both maintaining instruction pointers and preventing program bloat.  I also second all of @Dunbaratu's concerns with the implications for backwards/forwards compatibility.  It's one of those things that is really difficult to take back and change once it's been made available.


I have not had any time to review this yet myself, does your method somehow prevent the opcodes of each sequential run from being kept in the program definition?

While we may not want to bring in the raw assembly language at this time, that doesn't mean that your techniques cannot be adapted to work with the compiled version of kerboscript.  After I finish the review of PR1400 I intend to spend some time looking at a few execution issues surrounding locks and script runs, I'll see if I can walk myself through this PR and maybe cherry pick some crucial changes.

thexa4:
Sorry for the late reply, my code doesn't prevent the opcodes from being kept in the program definition. It would allow someone to write a compile function that would replace the function pointer.

Dunbaratu:
we may want to talk after the release of 0.19.0 is out.  One of the things I want to look into is the way the system deals with whether a file has been compiled into memory already or not when trying to run a script.  it's a bit broken at the moment and needs an overhaul.  It may affect this.

thexa4:
The need for this has largely passed.

PR #1405, opened 2016-02-09
Some v0.19.0 changelog edits

### LINKED ISSUE #1433: Prepare and publish 0.19 credits
@Dunbaratu suggested that we should prepare a list of people who contributed to 0.19. We need to prepare the list and decide where to put it.

I've already asked all 9 people who have contributed commits to git since Nov 27th whether they want to be mentioned in credits.

Dunbaratu:
This first draft includes ALL the user-facing PR's more recent than Nov 25 (the date of the previous release v0.18.0), which we may or may not actually choose to include in the release.  We may need to trim this list down if we decide to leave some things out of the next release and not do them yet.  At the moment I don't yet know which items will be in release and which won't, so I just mentioned them all for now.

There are a few PR's left off because they contain fixes that the user doesn't really see.

Fixes #1433

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ release if the PR's about them don't get merged before release.
 * Automatic serialization system to save/load some kinds of data values to JSON-format files (TODO link to docs)
 * User Programs and Functions now allow trailing optional parameters with defaulted values. (TODO link to docs).
 * There are now some suffixes that work on all value types, even primitive scalars.  To accomplish this, a new "encapsulation" system has wrapped all kOS structures and primitive types inside a generic base type.  (TODO link to docs page about basic structure suffixes)
+* ENGINE type now supports multi-mode cases and has its gimbal accessible through :GIMBAL suffix (http://ksp-kos.github.io/KOS_DOC/structures/vessels/engine.html) 
+* Added GIMBAL:LIMIT suffix. (http://ksp-kos.github.io/KOS_DOC/structures/vessels/gimbal.html)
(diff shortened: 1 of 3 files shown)


AlchemistCH:
Here's my proposition for entries for my commits:

 ### NEW FEATURES
* ENGINE now fully supports multi-mode cases and has its gimbal accessible through :GIMBAL suffix. (http://ksp-kos.github.io/KOS_DOC/structures/vessels/engine.html) 
* Added GIMBAL:LIMIT suffix. (http://ksp-kos.github.io/KOS_DOC/structures/vessels/gimbal.html)

### BUG FIXES
* Removed the ability to activate both modes on multi-mode engine simultaneously.
* LIST ENGINES now lists all engines and displays part names instead of module names. (https://github.com/KSP-KOS/KOS/issues/1251)

hvacengi:
I hadn't thought about it until just now, but the nice thing about having a changelog PR sitting out there before the actual release is we can submit PR's to @Dunbaratu's fork, instead of having 10 different PR's all colliding upstream.  (Or we can just ask him to include something, as was done above)

tomekpiotrowski:
Fixes #1433

Dunbaratu:
I was going to add changes.rst to this PR, but actually I'd rather wait until all the other docs are in.

tomekpiotrowski:
There's no mention of the new lexicon syntax in the changelog.

Dunbaratu:
Okay I'll add  it in, but just for the record, to make things easier to read through for hvacengi I made a single merged document PR that has this and all the others in it.  That means I won't be fixing it *here*, but over there instead.

PR #1409, opened 2016-02-11
Change serialization API to produce/consume Structures

tomekpiotrowski:
Updated some serialization internals after #1400 was merged


src/kOS/Suffixed/OrbitInfo.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace kOS.Suffixed
 {
-    public class OrbitInfo : Structure, IDumperWithSharedObjects
+    public class OrbitInfo : SerializableStructure, IHasSharedObjects
```
Dunbaratu: I see a few fields that look like they wouldn't survive a serialization and reloading, like the `nextPatch` or `transition` information.

tomekpiotrowski: You might be right, I'll look into it.

tomekpiotrowski: I've created a separate issue for this as the problem existed before this PR.

Dunbaratu: The reason I noticed it in this PR is that this PR is the first time the code explicitly claimed this was derived from SerializableStructure, which was a class that didn't exist before this PR.  But conceptually, this is not really a serializable class, because of this.

tomekpiotrowski:
Unless there are more issues I think we can merge this.

Dunbaratu:
I'd be willing to merge it only if we stopped claiming `orbit` is serializable in its current state.  Right now it's not.  It claims to be, but what you get back is not quite what you saved.  If we told it to inherit OrbitInfo from just `Structure` for this release, and didn't make it inherit from `SerializableStructure` until that other issue is resolved, then I'm willing to merge.  It just means `OrbitInfo` becomes one of the several types that isn't serializable.

tomekpiotrowski:
This faulty serialization of `OrbitInfo` is already present in current develop and now we have a separate issue for it. I see no reason to keep a random PR on hold (which this really is, it is only by accident that you've noticed the bug here). The PR itself does what it states to. I can submit a separate PR that temporarily fixes #1410 as you suggest.

Dunbaratu:
This claim is incorrect:

> it is only by accident that you've noticed the bug here

For two reasons: 1 - I wasn't the one doing the merge of the previous PR, so this is the first serialization PR with me being the person doing it, and 2 - Whether you agree or not, the promotion of serialization from mere interface to actual class makes it seem like a much bolder assertion to me.  It is the reason why I spent the time looking at all the suffixes to ensure they really DID look like all the user-facing data would survive serialization properly..  This promotion caused the scrutiny.  So it's not really an accident that I noticed it on the first PR that contained that promotion.

But I do get your point that the problem still existed beforehand when it was a mere interface, (It is just that it was less important then, in my way of looking at it, but it still did exist) and the error of merging that has already happened and it's too late to back out the earlier merges without too many other cascading messy effects.

So we can let it pass for now, but only because the mechanical process of undoing it is uglier than the mechanical process of letting it be temporarily wrong for a bit before it gets fixed with another PR on top.

tomekpiotrowski:
Cool, thanks. I will try to submit a fix to #1410 asap.

Issue #1412, opened 2016-02-11
Fix all occurrences of bogus data types in documentation


tomekpiotrowski:
Because proper encapsulation types did not exist the docs often use things like `scalar`, `boolean`, `Boolean`, `scalar (metric tons)` to specify the type of a suffix parameter or a suffix return type. Now that we have proper user space types we should point to them.

As @hvacengi has pointed out we will probably want to take care of it after #1408 is done ;)

hvacengi:
I suppose it depends on what we end up with as the type names.  I think this will be related to the discussion in #1408.  But if the decision is that `ScalarValue` reports it's type to the user as `"scalar"`, then those references may (heavy on the *may*) be OK.  Either way, we need to walk it and make sure.    The references should be consistent for sure.

Dunbaratu:
I'm going to examine the possibility of scripting this edit with some clever regex search/replace scripts.

hvacengi:
Closed via #1493

PR #1414, opened 2016-02-12
Will be the PR that Fixes #1408 once it's done.

### LINKED ISSUE #1408: Give kOS.Safe.Encapsulated.Structure some generic suffixes
Now that PR #1400 is merged, we can start working on the suffixes intended for all user objects (which now all inherit from `kOS.Safe.Encapsulated.Structure`).

This issue superceeds the idea in issue #1237, which I just closed and marked as duplicate of this.

I was about to just jump in and get started on this myself, but in light of what I said in reaction to the KASM PR that arrived on our doorstep out of the blue a few days ago with no previous discussion or issue for it, I thought I should follow my own advice and start with an issue first before implementing it.

It is my hope that we can have at least part of this for the 0.19.0 release, otherwise all that refactoring work for `Structure` has no benefit the end-user can see yet and it feels like we didn't do much this release.  I want at least a few suffixes the end-user can see so there is something to announce with that refactor.

To that aim, I'd like to get a PR for this in very fast - in a day or two at most, which is why I want to break up the suffixes in terms of what can be done quickly now versus what takes a bit more work.  Discuss if you disagree with my ordering of what to do first, or if you think there are other good suffix ideas I've missed.

Text below this line is subject to being edited based on what's said in the comments to follow if people disagree or object:

-----

Ideas of Suffixes to implement, in order from probably easiest to probably hardest:

[ ] `:TOSTRING` - Essentially already exists anyway.  Just needs a user-visible suffix wrapped around it.  The only way people have to obtain it now is to concatenate with a string`("" + thing)`.
[ ] `:HASSUFFIX` - `anyobject:hassuffix("suffixname")` returns boolean.  True if the object has such a suffix name on it, false if it doesn't.  Should also support Vessel Resource suffixes, since it's called on an instanced object, not the class itself, so by then we now know which Resources it has.
[ ] `:TYPENAME` - returns a string (because we don't have enums in kerboscript) for the type of this object. (note (1))
[ ] `:ISTYPE("name here")` returns boolean true if the object is the given type *or is a class derived from it*, like asking if a ScalarIntValue is a "Scalar" should return true. (note (1))
[ ] `:ISSERIALIZABLE` - true if @tomekpiotrowski 's JSON outputter can work with this type.
[ ] `:SUFFIXNAMES` - a LIST() of all the suffix names as strings, but without the full signature info.
[ ] `:SUFFIXHELP` - a string dumping suffix signature info as best as we can, maybe without the descriptions for now, but descriptions could be added later. (note (2))

Notes:

(1) I'm still waffling on what the correct string to use for `:TYPENAME` and `:ISTYPE` should be.  Right now scripts don't see things like `VesselTarget` except in exception messages from time to time.  Instead they see the word just `Vessel`, as a constructor like `set ves to vessel("that ship").`  Similarly for `list()` rather than `ListValue`, and so on.  This would seem to imply that we should be using those strings not the C# names, in which case we should add that as a mandatory member of `Structure` - all structures must name themselves with a property string.

(2) It may be a good idea to reserve the term `:SUFFIXES` for use later but not implement it yet.  My idea for this is that once we have a `SuffixDelegate` as is a kind of `KOSDelegate`, then we can add a `:HELP` to KOSDelegate so that you can always dump the signature data about a delegate (even user ones?).  Once we have that, then I imagine that `:SUFFIXES` would be a LIST() of all the suffixes of the object rendered as KOSDelegates so you can then ask for their HELP one at a time.  This is why I'm reluctant to add the description info to SUFFIXHELP, as it would really belong here instead, and we're not ready to implement this with 0.19.0 yet.

Dunbaratu:
This checkin still isn't the final version yet.  It only implements 3 of the suffixes so far, but it shows that the idea is working for at least those 3.


tomekpiotrowski:
To implement `ISSERIALIZABLE` this should be fine: `this is SerializableStructure || this is ISerializableValue`. (Scalars do not descend from `SerializableStructure`, but are still considered serializable).

Dunbaratu:
> (Scalars do not descend from SerializableStructure, but are still considered serializable

Wait, implementing ISerializableValue doesn't mean a thing.  At all.  The definition of ISerializableValue doesn't say anything.  It's just this:
```

using System;

{
    public interface ISerializableValue
    {
    }
}
```
That's not asserting anything at all.  There is NO proof there that the object is in fact serializable.  In fact I wish I'd caught this before in the other PR - that doesn't mean anything at all so why is it still there?

Also, if you have to check EITHER SerializableStructure OR ISerializableValue then why not just make SerializableStructure itself be a class that claims to implement ISerializableValue and then you only have to make one check - for the interface.

hvacengi:
@Dunbaratu check today's slack chat.  I had some similar questions.

Dunbaratu:
@hvacengi but none of those questions included the point that claiming something implements ``ISerialableValue`` literally promises nothing at all because the interface definition is empty, and the point that if it wasn't empty and actually did mention the methods that ``SerializableStructure`` uses to do its work, then we could do the much cleaner ``if (this is ISerializableValue)`` instead of ``if (this is SerializableStructure || this is ISerializableValue)`` - I'm not understanding why it had to be designed with two different conditions to check for this.  If the important methods of SerializableStructure were mentioned in the interface ISerializableValue, and if SerializableStructure claimed it implemented ISerializableValue, then we'd have one clean universal check - just check if it implements the interface.

tomekpiotrowski:
Now that I think about it maybe we should present only types that satisfy `this is SerializableStructure` as serializable to the user. Encapsulation types will be serialized when inside collection types, but they can't be top level objects themselves (Json doesn't handle that).

Dunbaratu:
@tomekpiotrowski but what is the interface `ISerializableValue` even for, given that it's empty and has zero effect on anything?  Any class can claim to implement it and in so doing it is promising absolutely nothing.

tomekpiotrowski:
I'm using it in `SafeSerializableMgr` to differentiate between types that are to be serialized using `Structure.ToPrimitive` and the rest that will be serialized using `ToString()`. (See https://github.com/KSP-KOS/KOS/blob/develop/src/kOS.Safe/Serialization/SafeSerializationMgr.cs#L23)

@erendrake has previously pointed out that an attribute would be better suited for the job, but I didn't think it was such a big deal and left it as it is for now. Feel free to change it if you are strongly against what I've done.

Dunbaratu:
Anything that looks really weird like that should be at least explained well in comments adjacent to the weirdness, to explain to other programmers why it doesn't look like it makes much sense.  An interface that does nothing and only exists to be found in reflection walks is such a weird case.  An attribute would probably be better because then at least people would know it's for the purpose of signposting the class for finding via reflection.  In any case it's outside the remit of this PR so I won't be fixing it here.  But I do think something should be done.

Dunbaratu:
Newest push has everything working.  Only thing left to do is the documentation, which isn't as bad as I thought because I just went with everything matching what's already in the docs, and I gave it an alias system in which, for example, you can do `if foo:istype("geoposition")` or `if foo:istype("latlng")` and it means the same thing.

src/kOS.Safe/Utilities/KOSNomenclature.cs
```diff
@@ -0,0 +1,184 @@
...
+        {
+        }
+        
+        public static void PopulateMapping(params Assembly[] assemblies)
+        {
+            foreach (Assembly a in assemblies)
+            {
+                // Ensure an Assembly is never run through this twice in the life of the program:
+                if (seenAssemblies.Contains(a))
+                    return;
```
tomekpiotrowski: Did you mean 'continue' here?

Dunbaratu: Good point.  Hopefully I'll remember to go back in and fix that later when I get back to this in several hours.

src/kOS.Safe/Utilities/KOSNomenclature.cs
```diff
@@ -0,0 +1,184 @@
...
+using kOS.Safe.Encapsulation;
+using kOS.Safe.Exceptions;
+using kOS.Safe.Serialization;
+
+namespace kOS.Safe.Utilities
+{
+    /// <summary>
+    /// A class for the purpose of holding the mapping information that goes
+    /// back and forth between C# type names and thier KOS type name equivalents.
+    /// For the sake of this we are using unqualified C# names (without the namespace
```
Dunbaratu: I need to edit this comment - this was true but isn't anymore.  Now it thinks in the fully qualified C# names.

Dunbaratu:
Added the docs for it, and  merged in @tomekpiotrowski's latest change that removed the GimbalValue.cs file (I had been calling that the primary structure type you get from kos naming, and had to move that name to the Gimbal partmodule instead.)

@hvacengi @erendrake, this is ready for review and merge.

The one thing the docs could still use is a global replacement of all the places where the type of a function or suffix is given as "scalar" or "number" or "boolean", with a link to the new pages in the docs that describe those now.  But that seems more in the remit of issue #1412 instead, so I left that for later.

Dunbaratu:
@hvacengi I pushed my doc changes, but they also demanded that I fix the CRLF problems in unrelated files before git on Ubuntu would work properly.  If you look at the few .cs files in my recent documentation commit in detail, you'll see github says every line is different but if you look by eye at those diffs you see they're identical.  It's just flagging as different because the files currently in the repo have the wrong line endings and git on Ubuntu can't work properly unless they get committed with the change.  (else it forever flags them as having been "edited" and it won't let me switch branches.

Dunbaratu:
Updated branch to  merge in what has happened in develop over the last few days

PR #1417, opened 2016-02-13
Add more ways of constructing lexicons

### LINKED ISSUE #1381: Allow LEXICON() to accept an initializer list like LIST(), STACK(), and QUEUE() can.
We can do ``set L to list(1,2,3,4).``

We should have a similar ability for lexicons.  Right now we only let people construct empty ones that they then have to fill one item at a time.

Two possible syntaces are:

1 - Flattened list of pairs.  i.e. value0 is the key of value1, value2 is the key of value3, etc.  Like so:

```
set L to LEXICON("a", "ant", "b", "bear", "c", "cat").
```
Resulting in `L["a"] = "ant"`, `L["b"] = "bear"`, and `L["c"] = "cat"`.

2 -  If we added the ability to pass in a list of pairs to populate it like this:
```
set rawlist to LIST().
rawlist.add to LIST("a","ant").
rawlist.add to LIST("b","bear").
rawlist.add to LIST("c","cat").
set L to LEXICON( rawlist ).
```
Then we'd sort of automatically get a way to do that all in one go, but it would be a bit more wordy:
``set L to LEXICON(  list( list("a", "ant"), list("b", "bear"), list("c", "cat"))).``

(I noticed this in passing when I was running through all the constructor functions in Suffixed.cs for an unrelated refactor, but as this is outside the remit of that refactor I'm just recording it in an issue and moving on.)

tomekpiotrowski:
Fixes #1381

### DIFF
--- a/doc/source/structures/misc/lexicon.rst
+++ b/doc/source/structures/misc/lexicon.rst
@@ -51,6 +51,26 @@ LEXICON() built-in function::
     // Make an empty lexicon with zero items in it:
     set mylexicon to lexicon().
 
+If ``LEXICON()`` is given arguments then they are interpreted as alternating
+keys and values::
+
+    set mylexicon to lexicon("key1", "value1", "key2", "value2").
+
+Will have the same effect as::
+
+    set mylexicon to lexicon().
+    mylexicon["key1"] = "value1".
+    mylexicon["key2"] = "value2".
+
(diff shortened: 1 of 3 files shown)


Dunbaratu:
a lot of @tomekpiotrowski's other PRs depend on #1393 being merged first.  Does this one, or can I start reviewing it as an independent self-contained PR?

Issue #1225, opened 2015-11-05
Inter-CPU and inter-vessel communication


tomekpiotrowski:
Looks like there is consensus among kOS developers that we want this feature in the next release so it might be useful to have a place to discuss it. I know @Dunbaratu and @erendrake had some ideas regarding how we could go around implementing it.

So the goal is to provide means for kOS processors to send messages to each other within one vessel and to other vessels (respecting any RemoteTech constraints where appropriate).

Most importantly I think that we will have to provide two similar, but separate mechanisms. I realize that it would be preferable to have one, more general solution, but for the moment I don't see how we could make it work. Primarily because inter-vessel messages can't be addressed to a specific CPU - we can't assume vessels know about each other's CPUs. On the other hand inter-CPU messages need to have a specific CPU set as recipient. Inter-vessel messages need to be persisted, inter-CPU not necessarily, and inter-cpu communication doesn't have to worry about the connection between the sender and recipient.

**Inter-CPU** in this model is actually very simple to implement. We need a simple FIFO queue for every CPU and an API for the sender and the receiver to access that queue.

**Inter-vessel** would require a separate queue (one per vessel) and (probably) a separate API for that queue. The contents of that queue would be persisted. RemoteTech would be used here to check the connection and delay.

I have actually taken the liberty of implementing inter-CPU as described above, the interface is as follows:

Receiver:

```
CORE:MESSAGES:EMPTY.
CORE:MESSAGES:PEEK.
CORE:MESSAGES:POP.
CORE:MESSAGES:PUSH(msg).
```

and sender:

```
PROCESSOR:SENDMESSAGE(msg).
```

Example code:

```
// This code runs on 'first'
SET SECOND_PROCESSOR TO PROCESSOR("second").
SECOND_PROCESSOR:SENDMESSAGE("launch").
SECOND_PROCESSOR:SENDMESSAGE(10).

// This code runs on 'second'
WAIT UNTIL NOT CORE:MESSAGES:EMPTY.
SET MESSAGE TO CORE:MESSAGES:PEEK. // read 'count sheep'

IF MESSAGE = "launch" {
  CORE:MESSAGES:POP. // remove 'count sheep' from the queue
  WAIT UNTIL NOT CORE:MESSAGES:EMPTY. // make sure the count is available
  SET COUNT TO CORE:MESSAGES:POP. // read '10'

  PRINT "Commencing countdown!!!".

  UNTIL COUNT = 0 {
[...]

PR #1418, opened 2016-02-14
added two functions to handle keyboard input in user scripts

bulshavik:
**Getch()** will return characters from the in game terminal for use in a script. It is non-blocking and will return "none" if there isn't a key press availble. The buffer is FIFO.

**FlushKB()** will clear the contents of the buffer and should be called in a script before the user is expected to make a key press.
example use in ks file:
set quit to 0 . 
flushkb().
until quit
{
local s is getch(). 
if (s <> "none") {print s.}
wait .25. 
}
[getkeyboard.ks.txt](https://github.com/KSP-KOS/KOS/files/129535/getkeyboard.ks.txt)

### DIFF
--- a/src/kOS.Safe/Encapsulation/TerminalStruct.cs
+++ b/src/kOS.Safe/Encapsulation/TerminalStruct.cs
@@ -51,6 +51,12 @@ namespace kOS.Safe.Encapsulation
             AddSuffix("VISUALBEEP", new SetSuffix<BooleanValue>(() => Shared.Screen.VisualBeep,
                                                        value => Shared.Screen.VisualBeep = value,
                                                        "Get or set the value of whether or not the terminal shows beeps silently with a visual flash."));
+            AddSuffix("GETCH", new Suffix<StringValue>(() => Shared.Screen.getch(), "Retrieve a key stroke from the terminal(s)"));
+            AddSuffix("FLUSHKB", new NoArgsVoidSuffix(() =>shared.Screen.flushKB(), "Clear all pending key strokes from the terminal(s)"));
+            AddSuffix("KEYSTROKESREADY",new NoArgsSuffix<ScalarValue>( () => shared.Screen.KeyStrokesReady(),"Return the number of keystokes in buffer"));
+
+
+
         }
         
         public override string ToString()
--- a/src/kOS.Safe/Screen/IScreenBuffer.cs
+++ b/src/kOS.Safe/Screen/IScreenBuffer.cs
@@ -27,5 +27,10 @@ namespace kOS.Safe.Screen
         void AddResizeNotifier(ScreenBuffer.ResizeNotifier notifier);
         void RemoveResizeNotifier(ScreenBuffer.ResizeNotifier notifier);
         string DebugDump();
+        void putch(string ch);
+        string getch();
+        void flushKB();
+        int KeyStrokesReady();
+      
     }
 }
--- a/src/kOS.Safe/Screen/ScreenBuffer.cs
+++ b/src/kOS.Safe/Screen/ScreenBuffer.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Text;
 using kOS.Safe.UserIO;
 
+
 namespace kOS.Safe.Screen
 {
     public class ScreenBuffer : IScreenBuffer
@@ -13,7 +14,7 @@ namespace kOS.Safe.Screen
         private int topRow;
         private readonly List<IScreenBufferLine> buffer;
         private readonly List<SubBuffer> subBuffers;
-        
+        private List<String> CharForScript;   // this will be a list of characters( really is a unity keycode string)  availble to the script that calls it.
         public int BeepsPending {get; set;}
         
         // These next two things really belong in TermWindow, but then they can't be reached from
@@ -64,6 +65,7 @@ namespace kOS.Safe.Screen
 
             RowCount = DEFAULT_ROWS;
             ColumnCount = DEFAULT_COLUMNS;
+            CharForScript = new List<String>() ;  // start a new script screen buffer 
             InitializeBuffer();
         }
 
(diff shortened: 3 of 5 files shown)


Dunbaratu:
The problem with this is that it sets the telnet client up to be a second-class citizen because it only traps the Unity chars.  I think there has to be a way to trap it at a point where both telnet and Unity end up funneling their input together, or at least adding the same queuing code to the telnet input.

Part of why I didn't do input yet is because I want to put it on top of the file I/O that one of the other devs is doing, but I can see there's a strong desire for it sooner than that, so something like this might be a good idea. 

Also, maybe it could be a member of the Terminal class like `set ch to terminal:getch().` and then make getch just an alias name for doing that.

bulshavik:
I moved the functions into the terminal class and added support for telnet clients as was suggested. 
I also added an additional function **terminal:keystrokesready()**  that will return the number of keystrokes held in the buffer
here is an example usage. 

set quit to 0 . 
terminal:flushkb().
until quit
{
if terminal:keystrokesready()  {print terminal:getch().}
wait .25. 
}
 I am new to using a repo like this . I assume etiquette is not to issue another PR when its a modification to an existing request ? 

[getkeyboard.ks.txt](https://github.com/KSP-KOS/KOS/files/129856/getkeyboard.ks.txt)

Dunbaratu:
Generally if the original PR hasn't been merged yet, and is still being discussed, then the best thing is to do what you did and just keep modifying its branch in your fork that it came from.

(Also you can ignore the "all checks have failed" comment for the time being.  The automatic build with "Travis CI' is broken at the moment and is flagging everything as failing.)

Sorry I don't quite have time to look at the code in detail yet as I'm in the middle of some big changes getting ready for our next release.  If the queued keypresses are a, then b, then c, does getch() need to be called 3 times, once to get 'a', once to get 'b', and once to get 'c'?  Or does it return the string "abc" all at once in one call?  (one at a time would make for a cleaner system because it would work the same way from code no matter what).

Part of the reason I'm reluctant to merge this right away is that it would be *sooo* nice if we could treat reading chars from a file as being the same thing as reading chars from the terminal (i.e. 'stdin' style), but we don't yet have reading chars from a file as a thing (it's coming) and I'd like to release the two together, or at the very least know what the future file API will look like so the 'stdin' API looks identical to it and uses mostly the same calls.

But maybe that's not what people actually want.  Maybe they want a key-at-a-time ability that is divorced from "everything is a file" thinking.    I don't really know.

bulshavik:
the getch function will only return one key press at a time so in your example you would need to make 3 calls to get a , b and c.  I figured concatenating the string could be left to the player (at least for now).  I didn't think there was a datatype of char in the KOS environment so the returned character is actually a string.  Another point I think worth noting ; and I'm not sure if this is a good or bad thing , is that if a special key is pressed you get the keycode string for it.   That I think is cool for doing stuff in script like 
 If (Terminal:getch() = "TAB") then { do. some. stuff .}  // no need for escape sequences !! 
But there may be some drawbacks I'm not seeing. 

I'd see this as more of a 'conio' style. I think for the terminal it fits better than the file stream approach since I see it being used more for single keys and wait periods in between ( like navigating menus and such).  I could see the getch() function for just the console and a getline()/putline()  function implemented on both file operations and the consoles.

Dunbaratu:
Okay, the keycode thing makes it distinct from stdin, so I can see the point of it now. (being char at a time does not distinguish it from stdin, because there's traditionally all sorts of settings and modes you can change about an open file handle - like setting it to blocking or nonblocking mode, and so on, so you can read keypresses from stdin just fine.

But the one thing you can't do with stdin is read keypresses that aren't characters.  Like hitting the shift key, or the control key, which are meta keys and don't make characters when you press them.

Speaking of which, can you getch() a shift key without another key pressed with it?  if so, how would you distinguish between shift-A versus just hitting shift alone then letting up on the key?

There's typically two approaches to reading keyboard input:

1 - Stream: Treat it like a queue, read a character at a time: advantage: you can read something like "A" as a single character rather than having to track the state of the shift key separately.  disadvantage; you lose the ability to read meta keys as separate keys (shift, control, alt) this way since they won't send a character until combined with some other key.

2 - event handling, where you're told about each individual key-up and key-down action separately, so that for example SHIFT-A looks like keydown(shift), keydown(a), keyup(a), keyup(shift), and you can query for 'is key down right now?'.  Disadvantage: a pain to track these keys separately: Advantage: you can read meta keys as separate keys, and track keys video-game style this way, where holding down a key doesn't produce repeated keys in a queue.

It seems like what you're going for is style (1) (which it sort of has to be to support telnet), which is why I was calling it a streamed stdin-like interface.

Dunbaratu:
Sorry, your PR came at a sort of bad time.  We're in the middle of trying to merge in the final features for our 0.19.0 release and are pretty much frozen on new enhancements not already in the pipeline for that, but your work may appear in 0.19.1, or whatever else we call the next version.

I'd like to see us trying to release more often, with smaller changes.  But the timing makes it likely that Unity5 will be on us (KSP 1.1) between 0.19.0 and the next release, and that may very well completely change the entire terminal UI code.

bulshavik:
Thats alright, its not a lot of code. I'll probably maintain a current
source copy and keep adding in until something comparable is implemented.
another idea I wanted to look into is a set of local variables common to
the entire craft  ( so you can pass data between CPUs). I haven't looked
into what that would take and it may well be beyond what I am capable of
doing. Would that be something worth pursuing in your opinion ?

Dunbaratu:
We already have someone working on intra- and inter- vessel communication.  It just got put on hold because we need to get something out before KSP 1.1 drops, and then when it does it will likely occupy our time for a while just making everything work again with expected Unity5 breakages.

tomekpiotrowski:
See #1225

hvacengi:
With #1830 we implemented these methods another way.  Not the way we originally expected to, so some of the comments and rationale posted here don't necessarily apply, but the basic functionality now exists.  As such, I'm going to close this PR.

PR #1421, opened 2016-02-15
attempting to get encrypted store from another repo

erendrake:
lets try this out

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ notifications:
 
 # Before the install the library dlls need to be downloaded
 before_install:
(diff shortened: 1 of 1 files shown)


erendrake:
This is working now. the encrypted file is located over in a new repo over here

https://github.com/KSP-KOS/KSP_LIB

hvacengi:
Is there any reason for me to not merge this?  I'd like to know that it still works in "develop" and that it works for pending pull requests as well, since our current situation appears to be mutually exclusive.

erendrake:
there is no reason not to. It works and I have high confidence it will work for any branch that pulls this change in.

PR #1422, opened 2016-02-15
attempting to run unit tests

erendrake:
If this works ill be shocked


hvacengi:
Since I got unit testing to work in another PR I'm closing this one.  If you can figure out how to make nunit-console run with the 2.x framework in place of the 4.0 framework I'm using right now that would be awesome though.

PR #1426, opened 2016-02-16
Fix ckan version string

hvacengi:
Found one issue with the netkan file after talking to the ckan guys on IRC.

kOS.netkan
* Per the ckan devs, added a version edit parameter which strips the "v"
from "v0.18.2" to maintain compatibility with former numbering system


hvacengi:
I'll merge this myself in an hour if no one else does.

tomekpiotrowski:
Sounds like extortion to me.

PR #1429, opened 2016-02-19
Fix travis

hvacengi:
Another attempt at getting travis to work from pull requests.


hvacengi:
OK, at this point, I have it set up so that this will properly test on a pull request.  I have commented the unit tests and the document building.  Each has errors in the base repository, and I don't want users to see a build error when it isn't a result of their PR itself.  I know that @Dunbaratu has been working on the documentation stuff, and I'll review his information to see if it eliminates the warnings in the documentation.

I'm of the opinion that we can pull this without the unit tests and docs tests, and then simply enable them once we fix both issues in the develop branch.

@erendrake and @BertCotton as the two people who setup the travis builds to begin with, do you see any issues with what I have submitted for changes?

hvacengi:
I just confirmed that this also works when building from a branch owned by KSP-KOS.  So the only thing preventing it from being merged at this point is review.

PR #1431, opened 2016-02-19
Move KAL9000 to Automation, resolves #1430

### LINKED ISSUE #1430: KAL 9000 too early in tech tree
At the moment, the KAL 9000 part is under "Precision Engineering" (160-tier), with the KR-2042 b. Since this is the "god" unit, it should probably go in a later tier (the CompoMax is under "Unmanned Tech", 300-tier).

If we truly want this to be an endgame part, "Experimental Electronics" (final 1000-tier) makes the most sense. Otherwise, "Automation" (500-tier) might be appropriate as well.

Personally, my vote would be to throw it in the 1000-tier, since it really is the core to end all cores.

gisikw:
Moving KAL to further in the tech tree, as it's an endgame part.

### DIFF
--- a/Resources/GameData/kOS/Parts/kOSkal9000/part.cfg
+++ b/Resources/GameData/kOS/Parts/kOSkal9000/part.cfg
@@ -14,7 +14,7 @@ rescaleFactor = 1
 node_attach = 0.01, 0.0, 0.0, 1, 0, 0, 0
 
 // --- Tech tree ---
-TechRequired = precisionEngineering
+TechRequired = automation
 
 // --- editor parameters ---
 cost = 1200


hvacengi:
Any one have any issue with this part being in the 1000 level tier?  I'll merge later tonight unless I hear otherwise.

Tagging @Dunbaratu and @erendrake

tomekpiotrowski:
I personally prefer having kOS processors sooner rather than later. Waiting a long time for a part means using it less during a game.

I think the question should be what would users find this part useful for? Then we should try to put it in such a place in the tech tree that it would be available when needed (with a certain degree of probability of course).

gisikw:
I disagree. I think by the point someone has reached the 300-tier (for the CompoMax), the 1000-tier isn't really going to be a bottleneck if you're committed to it. This unit is orders of magnitude more powerful than the CompoMax, by virtue of being designed as an endgame part, and I think the tier should reflect that.

To be clear though, the unit isn't *qualitatively* different from the CompoMax, so it's not penalizing users who are still early-on in the tech tree, it's just offering a god-part for those who want to invest the science.

tomekpiotrowski:
AFAIK the original motivation for a new part was to support smaller vessels for which even CompoMax was too big, see #1303. Granted that we may have made a supercomputer out of it (maybe we shouldn't have?), but putting it at the end of the tech tree would mean that we have not really solved the original problem - users might want to build drones and rovers a lot sooner in the game.

hvacengi:
I haven't played a science game in the last 9 months, and I haven't played career since December 2014.  So I'm going to trust those of you who actually come across this part of game play to say where it should belong.  But I will offer my observations:

To some degree, I agree with @gisikw's logic that this is the "top tier" processor.  So putting it in an end game spot makes sense.  By the same token though, I don't know if there is anything truly useful in this research node, and I don't think I want people feeling like they need to spend the full 1000 science on just our single part.  The goal is to make it "tough" but not "frustratingly challenging".

#1303 doesn't really discuss the question of where in the tech tree the part should belong.  My take from the discussion was that they just want a small part, and that it would be higher in the tree than current parts are.

Performance wise, the parts will actually be more differentiated in the next release, when we introduce per-cpu ipu limits, which will allow some parts to execute more instructions in an update than other parts.

gisikw:
The Experimental Electrics tier is the RTG, so I don't think it's really a science sink.

I play pretty much exclusively in career, though I'd suspect that the drone example was something created either for endgame career or sandbox. What I can say is that in career, the instant HAL is unlocked, there's absolutely no reason to use any of the other CPUs, which I think kinda ruins progression if it's not made somewhat difficult to achieve.

hvacengi:
Oh, I definitely agree with that last point.  There is literally no reason to use one of the other processors at that point.

Dunbaratu:
@giskw: Price.  The way to make the KAL9000 not necessarily always the right choice all the time once you have it, and therefore leave the other ones still being worthwhile, is to give it a higher money price.  Make it more expensive (I know that's not realistic - computers got cheaper as they got higher tech, but in game progression that's how everything else in the tech tree works.  If a later part is literally better in every way than a previous part, then it's usually more funds.)

But I wouldn't add that cost in just yet until later when the parts really *do* differ quite a bit in speed as well, as is a future planned feature.  Then the "I only need the cheaper, slower part for this" consideration means something.  Once we make them run at differing speeds that will cause a need for another rebalancing decision about their costs and weights.  RIght now if we spend too long on that we end up balancing for a situation that is temporary anyway (i hope).

As for it being at the 1000 tech level, maybe we can knock it back to 500?  (as long as that doesn't "tie" with any other CPU part - I don't have the game in front of me to check.)  I think if we wait until the final tier to award it, it just doesn't get used much.

gisikw:
Automation is a 500-tier, that comes with the Drone Core. My vote would still be to keep it an endgame part, but if there's a consensus, I can change it.

Dunbaratu:
I prefer 500 to 1000, but not strongly enough to raise a big stink over it.  If more people want 1000, that's fine by me.  I do agree with the argument that says chances are the part will be used more by people in sandbox than in career anyway.

gisikw:
Updated to put KAL in Automation, a 500-tier.

PR #1435, opened 2016-02-20
Fixes #1388 by documenting MISSIONTIME

### LINKED ISSUE #1388: Request: accessable mission elapsed time
MET doesn't start until the first stage is activated, and seems like a handy thing to have access to.

Dunbaratu:


doc/source/bindings.rst
```diff
@@ -291,6 +291,23 @@ Controls that must be used with LOCK
 Time
 ----
 
+MISSIONTIME, or MET
```
tomekpiotrowski: This would suggest that `MET` is a valid alternative to `MISSIONTIME`, but looking at the code and in-game `MET` does not exist.

Dunbaratu: Feel free to change it.  I mention "MET" down below in the paragraph, so that should be enough so it doesn't have to be in the title as well.

PR #1436, opened 2016-02-20
Fixes #1365 explicitly document trigger can't wait.

### LINKED ISSUE #1365: Update documentation to remind users not to not feed PIDloop() a long function
If you update a PID with a function that takes longer than a tick to run, PIDloop will return 0. I suggest adding a note in the documentation to this effect.

Dunbaratu:


AlchemistCH:
I was thinking maybe it's worth adding the following code example to WHEN trigger:

The following code won't work:
```
WHEN abort THEN
 {
  do_now().
  WAIT 3.
  do_later().
 }
```
However, you can achieve delaying execution of some of the instructions by creating another trigger:
```
WHEN abort THEN
 {
  do_now().
  SET abortTimeStamp to TIME:SECONDS.
  WHEN TIME:SECONDS>=abortTimeStamp+3 THEN
    {do_later().}
 }
```

doc/source/language/flow.rst
```diff
@@ -391,7 +391,10 @@ Note that any ``WAIT`` statement, no matter what the actual expression is, will
     The :ref:`WAIT <wait>` command only causes mainline code
     to be suspended.  Trigger code such as WHEN, ON, LOCK STEERING,
     and LOCK THROTTLE, will continue executing while your program
-    is sitting still on the WAIT command.
+    is sitting still on the WAIT command.  Furthermore, you cannot
```
tomekpiotrowski: This sounds a bit redundant to me. The part after `Furthermore` pretty much just repeats what the first part states. (Apart from adding 'LOCK WHEELSTEERING' and 'LOCK WHEELTHROTTLE' to the list of locks).

Dunbaratu: Not redundant.
The first form describes this:
```
when x < 0 then {
  print "x is negative."
}
set x to -1.
wait 5.  // does not stop the trigger from firing.  Still happens right away.
```
The second phrasing is describing this:
```
when x < 0 then {
  print "x is negative."
  wait 5. // this doesn't do anything.
}
set x to -1.
```

tomekpiotrowski: Ah, alright, I get it now :)

PR #1444, opened 2016-02-22
Added basic gitattributes

erendrake:
c# template from https://github.com/Danimoth/gitattributes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+uto detect text files and perform LF normalization
+# http://davidlaing.com/2012/09/19/customise-your-gitattributes-to-become-a-git-ninja/
+* text=auto
+
+*.cs     diff=csharp


.gitattributes
```diff
@@ -0,0 +1,5 @@
+uto detect text files and perform LF normalization
```
Dunbaratu: I'm pretty sure this is a typo and this was meant to begin with a '#' to be a comment line.

hvacengi:
I submitted a PR to @erendrake to fix the typo.  It also added 2 more lines, which is why I didn't simply fix and merge.  If he agrees with those changes he can merge, otherwise he can just edit the file as noted above.

Someone on a linux machine should download the file and see if it causes any conflicts on their machine with existing files.  There is a chance that users may need to manually update their line endings on files with uncommitted changes in the working directory.

Dunbaratu:
It won't be possible to test @hvacengi's edits until the repo itself is repaired to get the files to all have the same line endings.  There's a few that are messy right now,.  Until they're fixed in the first place, we can't tell if it's the .gitattributes that caused a file to have the wrong endings or the problem is already there in the repo data itself.  Any of the recent PRs from me that have documentation in them will also have the re-written line endings of the .cs files in them too because it was the only way to get git to stop claiming I'd edited files I hadn't.

hvacengi:
@Dunbaratu once this is merged though, wouldn't that mean you could merge the upstream develop with the docs PRs and change the line endings at that point?  The blame would still point to your last commit, but the PR's diff would only show your changes (because the diff would see it as you changing the line endings and then changing them back).

PR #1445, opened 2016-02-22
Fixes 1412 docs rework

Dunbaratu:
This is *meant* to be just diffs of the .rst files, but because we got some files in our repo that have CRLF endings, a few 'bogus diffs' are in here too that do nothing but change the line endings back to LF, and have nothing to do with the intent of the PR.  (If working on Unix, if I don't change the files' endings, then git blows up and refuses to allow me to ever switch branches so I can't work on anything.  See my long saga in slack chat from yesterday to see the frustration with this.  in the end I just decided to accept the the PR will have these phantom "changes" in it for a while until develop gets fixed.)


AlchemistCH:
I added a little pull request for commands/flight/systems.rst 
It contains the change from #1447 and couple cases of string references on that page - I think it's better to put it in here altogether.

PR #1449, opened 2016-02-22
Inter-CPU and inter-vessel communication

### LINKED ISSUE #353: Inter-CPU communication can be made possible now

### LINKED ISSUE #1225: Inter-CPU and inter-vessel communication

tomekpiotrowski:
Fixes #1225
Fixes #353

I've decided to finally upload a working version of it as it's been sitting in my repo for about 3 months now.

All the functionality works, but it needs a couple of improvements here and there. We also need to discuss the API.

Currently the user api is as follows:

*Inter-CPU*:

```
// sender

SET MESSAGE TO "HELLO". // can be any serializable value or a primitive
SET P TO PROCESSOR("other").
P:SENDMESSAGE(MESSAGE).

// receiver

WHEN NOT CORE:MESSAGES:EMPTY { // one could use 'WAIT UNTIL' as well
  SET RECEIVED TO CORE:MESSAGES:POP. // or do something else...
  PRINT "Sent by " + RECEIVED:SENDER:NAME + " at " + RECEIVED:SENTAT. // same ship obviously
  PRINT RECEIVED:CONTENT.
}
```

*Inter-vessel*:

```
// sender

SET MESSAGE TO "HELLO". // can be any serializable value or a primitive
SET C TO VESSEL("SOMEVESSEL"):CONNECTION.
PRINT C:DELAY.
C:SENDMESSAGE(MESSAGE).

// receiver

WHEN NOT SHIP:MESSAGES:EMPTY {
  SET RECEIVED TO SHIP:MESSAGES:POP. // or do something else...
  PRINT "Sent by " + RECEIVED:SENDER:NAME + " at " + RECEIVED:SENTAT.
  PRINT RECEIVED:CONTENT.
}
```


Dunbaratu:
I like the API on the receiver's side and (mostly) the sender's side, but is there a way to make the API the same once a connection has been gotten, and the only difference is the getting of the connection in the first place?

i.e. both inter-  and intra- vessel comms have you grab a "CONNECTION" object, which has a :DELAY suffix, and a :DESTINATION suffix, but for intra- vessel comms the :DELAY is obviously always zero, and for intra- vessel comms the :DESTINATION is a CPU core, and for inter- vessel comms the :DESTINATION is a Vessel (now that we can do :TYPENAME and :ISTYPE, these sorts of dual-duty suffixes are more workable in scripts.), but both of them have you do a :CONNECTION to get the connection object in the first place, it's just that one is a :CONNECTION of the vessel and the other is a :CONNECTION of the CPU core part.

tomekpiotrowski:
No, `Connection` is only used when sending inter-vessel messages. To send intra-vessel you only need a `Processor` (see the first code example).

Dunbaratu:
@tomekpiotrowski I know that's how it currently works.  I'm describing a different way I'd prefer.

tomekpiotrowski:
Ah, sorry.

I think we could make it work your way I guess. I'll think about it more after we're done with 0.19 :)

Dunbaratu:
Yeah we can talk about it later.  What I'm going for is the "everything is a file" sort of model.  Get handle on a connection, then write things to that handle.  The "write things to that handle" could work the same way regardless of how the handle got constructed.  Thus the only difference between sending to a local CPU and sending to a remote vessel is entirely in the line that constructs the handle.  So you'd have a Connection object either way, and the difference between whether that's a Connection to a specific CPU part or a Connection to an entire remove vessel is abstracted away after that unless the script explicitly wants to query it and ask the Connection what sort of thing it's connected to.

The reason I like that model is that it lets you obtain a `Connection` then pass it in to a function that uses it, and the function from then on can choose to not care whether it's a local or remote thing it's talking to, and thus you can make re-usable script code.

tomekpiotrowski:
Alright, this should be ready for review.

Currently messages arrive even when the receiver (CPU or vessel) has no power. I'm not sure if that's something that we want to change. Surely detecting if vessel has power at the moment of receiving a message is going to be difficult.

src/kOS.Safe/Encapsulation/PrimitiveStructure.cs
```diff
@@ -0,0 +1,11 @@
...
+
+namespace kOS.Safe.Encapsulation
+{
+    [kOS.Safe.Utilities.KOSNomenclature("Structure", KOSToCSharp = false)]
+    public abstract class PrimitiveStructure : Structure
+    {
+        public abstract object ToPrimitive();
+    }
+}
+
```
hvacengi: It looks to me like you're replacing the `IserializableValue` interface with the `PrimitiveStructure` inherited class.  Since this class only serves to force the object to implement `ToPrimitive` is there a reason you didn't add the method to the previous interface?

src/kOS.Safe/Encapsulation/Structure.cs
```diff
@@ -337,20 +337,10 @@ public static Structure FromPrimitiveWithAssert(object value)
...
-            var booleanValue = value as BooleanValue;
-            if (booleanValue != null)
-            {
-                return booleanValue.Value;
-            }
-            var stringValue = value as StringValue;
-            if (stringValue != null)
-            {
-                return stringValue.ToString();
+                return primitive.ToPrimitive();
```
hvacengi: The reason I originally left the `ToPrimitive` method as a static method on the `Structure` class was because we cannot implement `FromPrimitive` on the individual structures.  While I agree that this logic is better suited to be in the individual classes (since it's part of the class definition) I rather liked the fact that both `From` and `To` were in the same place, making it so that edits to the system only needed to be done in a single file.  Was this done to facilitate easier serialization?  Does it maybe make sense to treat the on structure `ToPrimitive` logic strictly in a serialization context?

src/kOS/Communication/ConnectivityManager.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace kOS.Communication
+{
+    public interface ConnectivityManager
+    {
+        double GetDelay(Vessel vessel1, Vessel vessel2);
+    }
+}
```
hvacengi: Can we move this into kOS.Safe.Communication?

src/kOS/Communication/InterProcCommand.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace kOS.Communication
+{
+    public class InterProcCommand
+    {
+        public virtual void Execute(SharedObjects shared) { }
+    }
+}
+
```
hvacengi: Can we move this into kOS.Safe.Communication?

src/kOS/Communication/MessageQueueStructure.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }
+
+        public override string ToString()
+        {
+            return messageQueue.ToString();
+        }
+    }
+}
+
```
hvacengi: Can we move this into kOS.Safe.Communication?

src/kOS/Communication/MessageStructure.cs
```diff
@@ -0,0 +1,78 @@
...
+            return dump;
+        }
+
+        public override void LoadDump(Dump dump)
+        {
+            Message = dump[DumpMessage] as Message;
+        }
+    }
+}
+
```
hvacengi: I guess the reliance of `MessageStructure` prevents moving the `MessageQueueStructure` into kOS.Safe.

src/kOS/Communication/RunCommand.cs
```diff
@@ -0,0 +1,18 @@
...
+
+        public override void Execute(SharedObjects shared)
+        {
+            if (shared.Cpu != null)
+            {
+                shared.Cpu.RunProgram(Program);
+            }
+        }
+    }
+}
```
hvacengi: I've noticed that there is an inter-cpu command for running a program but I don't see any documentation of such a feature.  I thought that we had discussed this option and weren't going to include that feature, that all of the communication would be done via messages and that enacting commands on other processors would have to be managed by a listening script.  Am I mis-remembering?

tomekpiotrowski: You remember correctly. I've decided not to remove it in this PR. I've created a separate package for communication-related stuff and I've simply moved it there for now.

src/kOS/Communication/ConnectivityManager.cs
tomekpiotrowski: I guess so.

src/kOS/Suffixed/VesselTarget.cs
```diff
@@ -470,6 +473,16 @@ private void InitializeSuffixes()
             return crew;
         }
 
+        public MessageQueueStructure GetMessages()
+        {
+            if (Shared.Vessel.id != Vessel.id)
+            {
+                throw new KOSException("You can only access the message queue of the current vessel");
```
hvacengi: Should we maybe make some kind of `KOSCommunicationException` (or maybe vessel/processor specific versions) instead of using the raw `KOSException`?  Is there any real benefit other than formatting? (might really be a question for @erendrake )

src/kOS/Communication/ConnectivityManager.cs
hvacengi: I'm only wondering because of my work to get a command line instance running.  I'm of the mindset to put everything we can into Safe, that way we can run unit tests, check performance, and do faster debugging.  It isn't super critical right now, but I probably will want to get there eventually.

src/kOS/Suffixed/VesselTarget.cs
tomekpiotrowski: Yeah, there could be a benefit - someone might want to catch only `KOSCommunicationException`s. We don't have exception catching in kerboscript for now so I wasn't concentrating on this. I'll have a look at it again later, I'll add a new exception if it makes sense.

src/kOS/Communication/RunCommand.cs
hvacengi: Ah, so it's there so we don't loose the logic, but it isn't in use.  It might be good to add a comment to that effect.

I haven't spent much time looking at this yet, but I don't think it's even exposed to the scripts right now, is that correct?  We've had a few times where people have submitted bug reports on features that we just didn't document instead of actually removing access, so I'd prefer there isn't an easy way for a user to randomly try to make it work.  I think that is already the case, but thought I should mention it in case you know otherwise.

tomekpiotrowski: I'm not sure if we want to keep the logic. I think @Dunbaratu explicitly asked me to remove this. I simply wanted to ask for the final confirmation later and push a separate PR if needed.

src/kOS/Communication/ProcessorManager.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using kOS.Module;
+using kOS.Safe.Compilation;
+using kOS.Safe.Persistence;
+using kOS.Safe.Utilities;
+
+namespace kOS.Communication
+{
+    public class ProcessorManager
```
hvacengi: This looks like a duplicate of `kOS.InterProcessor.ProcessorManager`, but I don't see the old manager deleted.  If we remove the run command, is this even needed?  It doesn't appear to connect to your messaging stuff directly.

src/kOS.Safe/Encapsulation/Structure.cs
tomekpiotrowski: IMO my change simplifies the fact that a given type is a primitive. Previously that information was stored in 2 places - `ISerializableValue` - used by serialization and `Structure.ToPrimitive`. This could cause bugs where a type was handled by `ToPrimitive`, but not marked with `ISerializableValue` or vice-versa (as actually is the case now with `BooleanValue`). Now if a type is a `PrimitiveStructure` it must implement `ToPrimitive`.

src/kOS.Safe/Encapsulation/PrimitiveStructure.cs
tomekpiotrowski: The pattern of a `Structure` implementing a `ISerializableValue` would be repeated a few times so I thought it might make sense to have a common parent class.

src/kOS.Safe/Encapsulation/Structure.cs
hvacengi: OK, I may mess with this later to try and make it so inheriting from `PrimitiveStructure` can some how opt in to the `FromPrimitive` option.  This fixes half of the issue, because it unifies type serialization but doesn't unify de-serialization (unless I'm missing something, which I easily could, I really need to spend more time looking at serialization).

src/kOS/Communication/ProcessorManager.cs
tomekpiotrowski: You're right, looks like I forgot to delete this whole package when creating a commit.

src/kOS/Communication/ConnectivityManager.cs
tomekpiotrowski: Ah, nope - `Vessel`... :stuck_out_tongue:

hvacengi: Oh yeah... huh.  I guess it's just used for inter-vessel communications, that doesn't have to go to Safe.  So long as we can try to move inter-processor communications into Safe eventually (doesn't have to happen before we merge this, unless it's easy enough to do).

Dunbaratu: What are you seeing as the meaning of "inter-processor" in command-line mode, given that the command-line mode just has the one kOS CPU?

hvacengi: For now.  We can pretty easily adapt the command line to use multiple processors in an effort to test the communication stuff.  Like I said, doesn't need to happen right away, just something to think about.  I think it's a good idea to consider whether or not something can reside in safe for all new additions.

doc/source/commands/communication.rst
```diff
@@ -0,0 +1,129 @@
...
+-----------
+
+:struct:`Connection` structure represents your ability to communicate with a processor or a vessel. Whenever you want
+to send a message you need to obtain a connection first.
+
+Inter-processor communication
+-----------------------------
+
+First we'll have a look at the scenario where we want to send messages between two processors on the same vessel. As
+a first step we must obtain the :struct:`kOSProcessor` structure associated with the target CPU.
```
tomekpiotrowski: Should be 'the'

doc/source/addons/RemoteTech.rst
```diff
@@ -29,6 +32,16 @@ It is possible to activate/deactivate RT antennas, as well as set their targets
 
 Acceptable values for `"target"` are: `"no-target"`, `"active-vessel"`, `"mission-control"`, a :struct:`Body`, a :struct:`Vessel`, or a string containing the name of a body or vessel.
 
+Communication
+~~~~~~~~~~~~~
+
+RemoteTech will influence :ref:`communication <communication>` between vessels. In order to send a message to another vessel a valid RemoteTech connection will have to exist between them
```
Dunbaratu: I'd add the qualifier that you mean *if it's installed* then a RemoteTech connection is required.  It's my understanding that you can still do inter-vessel comms without RemoteTech, but you just don't get delays and connection checks then.  (Is this right?).

doc/source/commands/communication.rst
```diff
@@ -0,0 +1,129 @@
...
+
+kOS allows you to write scripts that communicate with scripts running on other processors within the same vessel
+(inter-processor communication) or on other vessels (inter-vessel communication).
+
+Messages
+--------
+
+The basic unit of data that is sent between two entities (CPUs or vessels) is called a :struct:`Message`.
+Messages can contain any primitive (scalars, strings, booleans) data types as well as any
+:ref:`serializable <serialization>` types. This allows for a lot of flexibility. It is up to you as a developer
```
Dunbaratu: Maybe we need a term to distinguish "person who works on the mod itself" from "person who uses the mod to make scripts".  Here you mean "script author" when you say "developer" but in other places the word "developer" means working on the C# under the hood.

doc/source/commands/communication.rst
```diff
@@ -0,0 +1,129 @@
...
+    PRINT "Message sent!".
+  }
+
+The CPU on the receiving vessel will use :attr:`SHIP:MESSAGES <Vessel:MESSAGES>` to access the vessel's message queue::
+
+  WHEN NOT SHIP:MESSAGES:EMPTY {
+    SET RECEIVED TO SHIP:MESSAGES:POP. // or do something else...
+    PRINT "Sent by " + RECEIVED:SENDER:NAME + " at " + RECEIVED:SENTAT.
+    PRINT RECEIVED:CONTENT.
+  }
```
Dunbaratu: I suspect more people will be interested in inter-vessel than intra-vessel comms.  (I know I am).  It might make more sense to reverse the order of explaining things.  (Give inter-vessel comms first, then explain how intra-vessel comms differ.)  Also, this gets the user to see useful code faster, higher on the page which is usually a good thing.  Intra-vessel comms have the extra complexity of explaining how to get which processor you're talking about, while inter-vessel comms only need people to know how to get a vessel by name, and know that ``SHIP`` is the name of the current vessel.  (So not only do I think it's the more useful one of the two, I also think it takes less text to explain it which would be a second reason to put it first).

doc/source/structures/communication/connection.rst
```diff
@@ -0,0 +1,76 @@
...
+use them to find out whether such communication is currently possible and send messages.
+
+Obtaining a connection
+----------------------
+
+There are 2 types of connections. The first type is used to communicate with other :struct:`processors <kOSProcessor>` within the same vessel.
+You can obtain a connection by using `CONNECTION` suffix. Assuming your vessel has a second processor tagged `'second'` it could look like this::
+
+  SET PROCESSOR TO PROCESSOR("second").
+  SET CONNECTION TO PROCESSOR:CONNECTION.
```
Dunbaratu: Does this actually work?  (Using the same identifier name "processor" as both the name of a built-in function AND the name of a user variable.)  Even if it does, I think it would help avoid confusion if the example didn't use the same name for both.

doc/source/structures/communication/connection.rst
```diff
@@ -0,0 +1,76 @@
...
+
+The second type are connections to other vessels. Assuming you have a rover on duna named `'dunarover'` you could obtain a connection to it like this::
+
+  SET VESSEL TO VESSEL("dunarover").
+  SET CONNECTION TO VESSEL:CONNECTION.
```
Dunbaratu: I have a similar comment here as the one above.  Using the same identifier, `vessel`, as both the name of the built-in function and the name of a user's variable doesn't really help distinguish the difference between them in the example.

doc/source/structures/communication/connection.rst
```diff
@@ -0,0 +1,76 @@
...
+
+    Destination of this connection. Will be either a vessel or a processor.
+
+.. method:: Connection:SENDMESSAGE(message)
+
+    :parameter message: :struct:`Structure`
+    :return: (:struct:`Boolean`) true if the message was successfully sent.
+
+    Send a message using this connection. Any serializable structure or a primitive (:struct:`String`, :struct:`Scalar` or :struct:`Boolean`) can be given as an argument.
+    It is always worth checking the return value of this function. A returned false value would indicate that the message was not sent for some reason.
```
Dunbaratu: When sending, if ``:ISCONNECTED`` is false, does that mean ``:SENDMESSAGE`` will fail?  If so, it would be good to mention it explicitly in a sentence here, to help connect the two suffixes to each other and make it clear what the relationship between them is.

About the underlying architecture:  I haven't gotten that far yet, but what is the intention when there *was* a connection, the message got sent, but there was a delay and the connection becomes false before the message delay is over?

In other words, does the connection have to exist throughout the entire duration of the delay, or just at the start of it, or just at the end of it?  Whichever answer it is, it might be good to mention which way around it works here in the documentation.

doc/source/structures/communication/message_queue.rst
```diff
@@ -0,0 +1,86 @@
...
+
+.. attribute:: MessageQueue:LENGTH
+
+    :type: :struct:`Scalar`
+
+    Number of messages in this queue.
+
+.. method:: MessageQueue:POP()
+
+    Returns the first message in the queue and removes it. Messages in the queue are always ordered by their arrival date.
```
Dunbaratu: It may be less ambiguous if the word "oldest" was used instead of "first".  Sometimes queue terminology refers to the "front" as being the place you push things onto it with the "back" being the place you pop things from.  Using that terminology, it would imply that "front" is the pushing end, not the popping end.   It depends on who wrote the explanation, and it varies confusingly.  Saying "oldest" would make it unambiguous.

doc/source/structures/communication/message_queue.rst
```diff
@@ -0,0 +1,86 @@
...
+
+.. method:: MessageQueue:PEEK()
+
+    :return: :struct:`Message`
+
+    Returns the first message in the queue without removing it from the queue.
```
Dunbaratu: (similar: maybe say "oldest"?)

Dunbaratu:
When looking at the diffs, this PR appears to contain a lot of changes to serializations and structures that are not directly connected to doing comms.  Is this just because this PR is old and serialization has changed in the meantime in develop (meaning if there's a merge conflict in those sections I should keep the develop version, not this PR's version), or is it because those changes are actually necessary for comms to work and they need to be changed with this PR?

tomekpiotrowski:
There are 2 reasons this PR touches serialization IIRC. One is that I've added a new formatter - `ConfigNodeFormatter` and some things needed updating. Even though the whole system was designed from the beginning to handle multiple formatters few things needed fixing. Then there were a couple of minor fixes here and there that I've noticed while writing the PR that I just didn't want to move to a separate PR.

doc/source/structures/communication/connection.rst
tomekpiotrowski: I've added a few more sentences to `ISCONNECTED` and `SENDMESSAGE`.

src/kOS.Safe/Communication/Connection.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using kOS.Safe.Encapsulation;
+using kOS.Safe.Encapsulation.Suffixes;
+
+namespace kOS.Safe.Communication
+{
+    [kOS.Safe.Utilities.KOSNomenclature("Connection")]
+    public abstract class Connection<S> : Structure where S : SharedObjects
```
Dunbaratu: This template usage ``<S>``seems like a very strange thing to do when I don't think ``SharedObjects`` has any other classes derived from it, or planned to be derived from it.  Or to put it another way, what are you expecting someone might use in place of the ``S`` here other than ``SharedObjects`` ?

src/kOS.Safe/Communication/BaseMessage.cs
```diff
@@ -0,0 +1,85 @@
...
+            set {
+                if (value is PrimitiveStructure || value is Dump)
+                {
+                    content = value;
+                } else
+                {
+                    throw new KOSException("Message can only contain primitives and serializable types");
+                }
+            }
+        }
```
Dunbaratu: Minor nitpick : I think everyone else on the project is using the format that puts the opening ``{`` on its own separate line.  I don't care for it myself because it introduces a lot of extra vertical whitespace and thus reduces what can fit on the screen, but that does seem to be the standard on the project.

src/kOS.Safe.Test/Serialization/JSONFormatterTest.cs
```diff
@@ -1,105 +1,24 @@
...
-            Lexicon deserialized = Deserialize(Serialize(lex)) as Lexicon;
-
-            Assert.AreEqual(new StringValue("value1"), deserialized[new StringValue("key1")]);
-            Assert.AreEqual(new ScalarIntValue(1), deserialized[new StringValue("key2")]);
-            Assert.IsTrue(deserialized[new StringValue("key3")] is Lexicon);
-            Assert.AreEqual(new StringValue("nested1value"), (deserialized[new StringValue("key3")] as Lexicon)[new StringValue("nested1")]);
+        protected override IFormatReader FormatReader {
+            get {
+                return JsonFormatter.ReaderInstance;
+            }
```
Dunbaratu: (See elsewhere in the PR comments for the comment with the word "curly brace" in big letters.)

src/kOS.Safe.Test/Serialization/JSONFormatterTest.cs
```diff
@@ -1,105 +1,24 @@
...
-
-            ListValue deserialized = Deserialize(Serialize(list)) as ListValue;
-
-            Assert.AreEqual(new StringValue("item1"), deserialized[0]);
-            Assert.AreEqual(new ScalarIntValue(2), deserialized[1]);
-            Assert.IsTrue(deserialized[2] is ListValue);
+        protected override IFormatWriter FormatWriter {
+            get {
+                return JsonFormatter.WriterInstance;
+            }
```
Dunbaratu: (See elsewhere in the PR comments for the comment with the word "curly brace" in big letters.)

src/kOS.Safe/Communication/GenericMessageQueue.cs
```diff
@@ -0,0 +1,151 @@
...
+        /// We need the queue to be constructed this way because SortedList doesn't allow for duplicate keys.
+        /// </summary>
+        private SortedList<double, List<M>> queue = new SortedList<double, List<M>>();
+        private CurrentTimeProvider timeProvider;
+
+        private List<M> Messages {
+            get {
+                return queue.Aggregate(new List<M>(), (acc, l) => { acc.AddRange(l.Value); return acc; } );
+            }
+        }
```
Dunbaratu: (See elsewhere in the PR comments for the comment with the word "curly brace" in big letters.)

src/kOS.Safe/Communication/GenericMessageQueue.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        private List<M> ReceivedMessages {
+            get {
+                return Messages.Where(IsReceived).ToList();
+            }
+        }
```
Dunbaratu: (See elsewhere in the PR comments for the comment with the word "curly brace" in big letters.)

src/kOS.Safe/Communication/GenericMessageQueue.cs
```diff
@@ -0,0 +1,151 @@
...
+            this.timeProvider = timeProvider;
+        }
+
+        private void RemoveMessage(KeyValuePair<double, List<M>> queueItem, M message) {
+            queueItem.Value.Remove(message);
+            if (queueItem.Value.Count() == 0)
+            {
+                queue.Remove(queueItem.Key);
+            }
+        }
```
Dunbaratu: (See elsewhere in the PR comments for the comment with the word "curly brace" in big letters.)

src/kOS.Safe/Encapsulation/Structure.cs
Dunbaratu: @hvacengi this is one of those hanging comments I was referring to in slack.  I don't know if you have work you wanted to do with this before I allow this to get merged.

src/kOS/Suffixed/VesselTarget.cs
Dunbaratu: I'd much rather have it implemented as a separate exception now while we're still looking at the code, rather than merging it and then forgetting it later on.  If and when we ever do more with exceptions in kerboscript, we may have a big task on our hands to go back through piles and piles of code to turn all the generic ``KOSException``s into specific exceptions.  It just makes more sense to have them already split out into separate types when they first get written, for all new code.  This is because at the time you write the code it's more fresh in your mind what the exception is about than months later when refactoring things.

src/kOS/Communication/InterVesselManager.cs
```diff
@@ -0,0 +1,87 @@
+﻿using System;
+using System.Collections.Generic;
+using kOS.Safe.Encapsulation;
+using kOS.Safe.Serialization;
+using kOS.Serialization;
+using kOS.Safe;
+
+namespace kOS.Communication
+{
+    [KSPScenario(ScenarioCreationOptions.AddToAllGames | ScenarioCreationOptions.AddToExistingGames, GameScenes.FLIGHT, GameScenes.SPACECENTER, GameScenes.TRACKSTATION)]
```
Dunbaratu: Was the goal here to make there always exist exactly one and only one instance of this class during the life of the KSP game?  Because I think there's a better simpler set of attributes for that, that we use for things like the telnet server, where you don't have to mention all the GameScenes one by one.  It looks like what you want for this class is for an instance of it to always exist and always respond to any load or save action anywhere.  I'm a little afraid that by mentioning explicit scenes it might prevent the saves from happening correctly if the save occurs on any other scene that wasn't mentioned in the list above.

src/kOS/Communication/InterVesselManager.cs
```diff
@@ -0,0 +1,87 @@
...
+        private static string MessageQueue = "messageQueue";
+
+        private Dictionary<string, MessageQueue> vesselQueues;
+
+        public static InterVesselManager Instance { get; private set; }
+
+        static InterVesselManager() {
+            // normally we do this in SerializationMgr, but KSPScenarios run before we create any instances
+            SafeSerializationMgr.AddAssembly(typeof(SerializationMgr).Assembly.FullName);
+        }
```
Dunbaratu: (See elsewhere in the PR comments for the comment with the word "curly brace" in big letters.)

src/kOS/Serialization/ConfigNodeFormatter.cs
```diff
@@ -0,0 +1,184 @@
...
+using kOS.Safe;
+using kOS.Safe.Utilities;
+using System.Linq;
+using System.Collections;
+using kOS.Safe.Encapsulation;
+
+namespace kOS.Serialization
+{
+    public class ConfigNodeFormatter : IFormatWriter, IFormatReader
+    {
```
Dunbaratu: Does this class guarantee that none of the content inside the data dumped to the config node will break config node syntax?  I know there's a few characters that you can't have in a string value in a ConfigNode because they confuse the ConfigNode format itself.  We ran into this problem when trying to save file content in local volumes, which is why the file contents get base64 encoded when storing them in the ConfigNode.  I did a quick scan by eye and didn't see any similar protection here, but then again I might have missed it if it was buried inside the Dump methods or serialization methods themselves.

Specifically, I'm thinking about when you dump strings to the ConfigNode.  Since the strings are open to the user putting whatever they like into them via the `CHAR()` function, we can't make any assumptions about their content being safe.

src/kOS/Communication/MessageStructure.cs
```diff
@@ -0,0 +1,78 @@
...
+        public override string ToString()
+        {
+            return "Message(" + Message.Vessel.ToString() + ")";
+        }
+
+        public override Dump Dump()
+        {
+            Dump dump = new DumpWithHeader {
+                Header = "Message"
+            };
```
Dunbaratu: (See elsewhere in the PR comments for the comment with the word "curly brace" in big letters.)

src/kOS/Communication/ProcessorConnection.cs
```diff
@@ -0,0 +1,66 @@
...
+    [kOS.Safe.Utilities.KOSNomenclature("Connection", KOSToCSharp = false)]
+    public class ProcessorConnection : Connection<kOS.SharedObjects>
+    {
+        private kOSProcessor processor;
+
+        public override bool Connected {
+            get {
+                return IsCpuVessel();
+            }
+        }
```
Dunbaratu: (See elsewhere in the PR comments for the comment with the word "curly brace" in big letters.)

src/kOS/Communication/ProcessorConnection.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        public override double Delay {
+            get {
+                return IsCpuVessel() ? 0 : -1;
+            }
+        }
```
Dunbaratu: (See elsewhere in the PR comments for the comment with the word "curly brace" in big letters.)

src/kOS/Communication/ProcessorManager.cs
```diff
@@ -53,7 +53,7 @@ public void RunProgramOn(List<Opcode> program, Volume volume)
         {
             kOSProcessor processor = GetProcessor(volume);
             var runCommand = new RunCommand {Program = program};
-            processor.ExecuteInterProcCommand(runCommand);            
+            processor.ExecuteInterProcCommand(runCommand);
```
Dunbaratu: I'm thinking we should remove this thing about running a command on another CPU given that it hasn't been updated and doesn't really quite work.  By now so much has changed about the architecture that whatever the old code did would need to be mostly ignored and re-done from scratch anyway.

With the new comms system one could make a "server" program that just receives command-lines as message strings and faithfully logs whatever was in the message out to a filename and then runs the filename.

src/kOS/Communication/VesselConnection.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    [kOS.Safe.Utilities.KOSNomenclature("Connection", KOSToCSharp = false)]
+    public class VesselConnection : Connection<kOS.SharedObjects>
+    {
+        private Vessel vessel;
+
+        public override bool Connected {
+            get {
+                return shared.ConnectivityMgr.GetDelay(shared.Vessel, vessel) != -1;
+            }
```
Dunbaratu: (See elsewhere in the PR comments for the comment with the word "curly brace" in big letters.)

Dunbaratu: For the greatest degree of "put all the code in just one place and don't repeat anything" design, perhaps instead of calling ``shared.ConnectivityMgr.GetDelay()``, this should actually call the ``Delay`` Property below in this same class, which in turn calls the ConnectivityMgr.

Dunbaratu: Given that the delay is a double and not an int, it's probably good form to either do a ``< 0`` comparison so you're not comparing to a direct number, or to at least use ``-1.0`` or ``-1d`` rather than the integer ``-1``.  My preference is for ``< 0`` rather than just explicitly checking for a single value of ``-1.0``, just because I dislike treating doubles like exact enums, for a variety of reasons.

src/kOS/Communication/VesselConnection.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        public override double Delay {
+            get {
+                return shared.ConnectivityMgr.GetDelay(shared.Vessel, vessel);
+            }
+        }
```
Dunbaratu: (See elsewhere in the PR comments for the comment with the word "curly brace" in big letters.)

src/kOS/Communication/VesselConnection.cs
```diff
@@ -0,0 +1,64 @@
...
+        public override string ToString()
+        {
+            return "VESSEL CONNECTION(" + vessel.vesselName + ")";
+        }
+
+        protected override BooleanValue SendMessage(Structure content)
+        {
+            double delay = shared.ConnectivityMgr.GetDelay(shared.Vessel, vessel);
+
+            if (delay == -1)
```
Dunbaratu: Similar to above.  I'd prefer to just call the entire range of negative values 'unconnected' by saying ``< 0`` here.

src/kOS/Serialization/ConfigNodeFormatter.cs
```diff
@@ -0,0 +1,184 @@
...
+            // ConfigNode doesn't handle arrays so we wrap everything in an extra node
+            // and give it a special name to mark this data set as an array
+            ConfigNode innerNode = new ConfigNode();
+            innerNode.name = ListKey;
+
+            var indexedList = list.Select((value, index) => new { Value = value, Index = index.ToString() });
+
+            foreach (var item in indexedList) {
+                HandleValue(innerNode, item.Index, item.Value);
+            }
```
Dunbaratu: (See elsewhere in the PR comments for the comment with the word "curly brace" in big letters.)

src/kOS/Serialization/ConfigNodeFormatter.cs
```diff
@@ -0,0 +1,184 @@
...
+
+        private ConfigNode ToConfigNode(string name, Dump dump)
+        {
+            ConfigNode configNode = new ConfigNode();
+
+            configNode.name = name;
+
+            foreach (var entry in dump) {
+                HandleValue(configNode, entry.Key.ToString(), entry.Value);
+            }
```
Dunbaratu: (See elsewhere in the PR comments for the comment with the word "curly brace" in big letters.)

src/kOS/Serialization/ConfigNodeFormatter.cs
```diff
@@ -0,0 +1,184 @@
...
+            {
+                configNode.AddNode(key, ToConfigNode(key, value as List<object>));
+            } else if (value is int) {
+                configNode.AddNode(key, ScalarToConfigNode(key, ScalarKey, Convert.ToString(value)));
+            } else if (value is double) {
+                configNode.AddNode(key, ScalarToConfigNode(key, ScalarKey, ((double)value).ToString("R")));
+
+            } else if (value is bool) {
+                configNode.AddNode(key, ScalarToConfigNode(key, BooleanKey, Convert.ToString(value)));
+            } else
```
Dunbaratu: (See elsewhere in the PR comments for the comment with the word "curly brace" in big letters.)

This, specifically: the case of if/else ladders, is why I'm not a fan of the formatting standard we're using on this project and actually prefer the kind you're trying to use here.  Putting all the braces on their own separate lines makes if/else ladders waste LOTS of screen height.  It makes it take 3 lines to express a single ``else if ....`` that way, as opposed to just one line the way you're doing it.

However, like it or not, that does seem to be the standard the rest of the project is using so it should probably be changed.

src/kOS/Serialization/ConfigNodeFormatter.cs
```diff
@@ -0,0 +1,184 @@
...
+                } else if (configNode.HasNode(i.ToString()))
+                {
+                    result.Add(ObjectFromConfigNode(configNode.GetNode(i.ToString())));
+                } else
+                {
+                    break;
+                }
+
+                i++;
+            }
```
Dunbaratu: (See elsewhere in the PR comments for the comment with the word "curly brace" in big letters.)

Dunbaratu: I'm not a fan of writing a finite loop by using what looks like an infinite loop in the header (``while (true)``) but with an explicit conditional break, unless there's a really complex set of circumstances causing that break that would greatly uglify the loop header, or it's something like an event dispatcher or a daemon that's expected to stay alive a long time.

This doesn't look like one of those cases.  It just looks like a finite loop where it consumes input until input is exhausted, and it's just that you can't see ahead of time how many iterations that will take.

I'd prefer something like:
```
bool moreValues = true;
while (moreValues)
{
    if .....
    else if ......
    else
        moreValues = false;
}
```
It's generally easier for people reading it to follow the logic that way.

Dunbaratu: (I also think that if you definitely do have an index variable that gets initialized and counts up per iteration, as you have here, then the 3-part for loop makes more sense than "while", but I know I'm not in the majority on that opinion.)

src/kOS/Serialization/SerializationMgr.cs
```diff
@@ -11,22 +12,19 @@ public class SerializationMgr : SafeSerializationMgr
     {
         private readonly SharedObjects sharedObjects;
 
+        static SerializationMgr() {
+            SafeSerializationMgr.AddAssembly(typeof(SerializationMgr).Assembly.FullName);
+        }
```
Dunbaratu: (See elsewhere in the PR comments for the comment with the word "curly brace" in big letters.)

src/kOS.Safe/Encapsulation/Structure.cs
hvacengi: No, I don't really have any work with it.  It's just a matter of wishing we had a better way to interface this.  I don't like having half of the feature residing on the Sub-class while the remainder resides on the base class.  It means having to remember 2 different locations if we need to change the logic, rather than keeping it in one place.  The only way to get the conversion from a `double` to a `ScalarValue` to be implemented on the `double` object is to use an extension method or reflection.  Neither of which work well on an un-cast `object` type.  

I'm also used to a serialization structure that gets annotated onto classes using attributes instead of by inheritance and interfaces.  So any time we move a "core" function to make serialization easier it seems a little weird to me.  But it will function fine.

So I guess I'd punt.  No changes necessary on my end.  Maybe a comment in the `FromPrimitive` method to point to the `PrimitiveStructure` implementation.  But that isn't much of a benefit either.

src/kOS.Safe/Communication/Connection.cs
tomekpiotrowski: We have 2 classes named `SharedObjects` one in kOS and one in kOS.Safe (the former inherits from the latter). `Connection` uses kOS.Safe.SharedObjects as a base class for the parameter. This allows us to keep it in kOS.Safe. Granted, currently this doesn't give us much, but I was thinking about writing unit tests that might involve this class some time in the future. Anyway, the more stuff is in kOS.Safe the better, right?

src/kOS.Safe/Communication/BaseMessage.cs
tomekpiotrowski: Those formatting issues are a combination of improper MonoDevelop settings and my lack of discipline. I'll try to correct them all.

src/kOS.Safe/Communication/Connection.cs
Dunbaratu: If you just tell it to use the kOS.Safe.SharedObjects class wouldn't it still allow you to assign a kOS.SharedObjects into it just fine?
Using the 'generics' syntax basically means you want there to be able to exist more than one variation of the class.  It means you want there to be two *different kinds* of `Connection` - one that can work on kOS.Safe.SharedObjects and a *different one* that can work on kOS.SharedObjects.  Is there a reason you have to make two versions of the Connection class, one for use with kOS.Safe and one for use with kOS (unsafe), or will it work just fine to have only one version, using the kOS.Safe version, and then have the kOS (unsafe) version just use that directly (it should be perfectly happy with letting you store an (unsafe) SharedObjects in a member designated as a kOS.Safe.SharedObjects)?

tomekpiotrowski: Yeah, `ProcessorConnection` and `VesselConnection` both need to access fields `kOS.SharedObjects`.

I just had another look at the `Connection` class. The `shared` field is not used in there, only in the subclasses, so I've moved it there. I think the `<S>` parameter was a remnant of some older code that I must have had refactored at some point. Anyway, I have removed it now.

Dunbaratu: Okay that makes a lot more sense then.  So then you only have to store the `shared` member in the derived class because none of the parent stuff that's defined in the abstract base class explicitly tries using it - shared is only used in the bodies of some of the overriding methods and properties in the derived class, not in the base class.

src/kOS/Communication/InterVesselManager.cs
tomekpiotrowski: I've looked at some different mod and have stolen that piece of code.

You're right: I need one instance and load/save capability. I don't know how else I can achieve that. Feel free to change this part if you find something better.

src/kOS/Communication/ProcessorManager.cs
tomekpiotrowski: Yeah, I know. For now I've just moved it. I'd prefer to remove it in a separate PR.

src/kOS/Serialization/ConfigNodeFormatter.cs
tomekpiotrowski: You're right, I've forgotten about that.

tomekpiotrowski: This is now fixed.

src/kOS.Safe/Communication/GenericMessageQueue.cs
tomekpiotrowski: This is now fixed, GH didn't auto-close the comment.

PR #1450, opened 2016-02-23
Fixes #1438 docs for ec

### LINKED ISSUE #1438: Documentation for EC drain #1395
#1395 changed the EC calculation, and I merged it without noticing it lacked documentation (my bad).  The following two things need to be documented about it:

1 - Explain power drain rules to people as part of the CPU hardware.

2 - For modders, and users of ModuleManager, a page that explains the part.cfg file options (while we're at it, we don't have explanations of the other values in part.cfg either and probably should.)

Dunbaratu:


doc/source/general/kospartmodule.rst
```diff
@@ -0,0 +1,108 @@
+.. _kospartmodule:
+
+KOS Processor PartModule Configuration Fields
+=============================================
+
```
Dunbaratu: This new file currently generates the following Sphinx warning, which is normally important:

/home/madings/GitHub/KOS-1/doc/source/general/kospartmodule.rst:: WARNING: document isn't included in any toctree

It means there is no mention of this file in any table of contents and therefore it is orphaned other than for deep links that point into sections of it.  That would normally be a big deal.  But in this case I believe one of the other doc PR's that hasn't been merged yet provides a change that moves to a new system where every directory automatically gets scanned for its contents to form a toc tree.  I think this warning goes away after everything is put together.

But just in case I'm wrong, this should be looked at later after everything is merged together.

PR #1451, opened 2016-02-23
Removed delay on event invoking.

AlchemistCH:
It turns out that `PartModule.Invoke(eventName, 0)` introduces 1-frame delay. 
Changed it to the method used in `partmodule:DOEVENT` - getting the event itself and calling  `BaseEvent.Invoke()`, which works immediately.


AlchemistCH:
Ok, I think `First` is good enough then, thanks for clarifying the difference.
BTW, it also looks like it's possible to use `PartModule.Events[EventName]`.

PR #1454, opened 2016-02-24
Fixes 1423 pretty stuff

Dunbaratu:
1 - Logo now included in upperleft box.
  2 - First page is a splash page.  contents is next.
  3 - Now supports kerboscript syntax highting in code sections.
  4 - Repaired a couple of typos discoverd BY said highlighting.
  5 - Added more videos to the show and tell.
  6 - Replaced KerbalStuff link with Spacedock link.

If anyone wants to make a better logo, feel free to push a new PR that replaces the file doc/source/_images/kos_logo.png.


doc/source/structures/misc/time.rst
```diff
@@ -32,6 +32,8 @@ Special variable TIME
 
     Any time you perform arithmetic on **TIME** you get a result back that is also a :struct:`TimeSpan`. In other words, TIME is a :struct:`TimeSpan`, but TIME + 100 is also a :struct:`TimeSpan`.
 
+.. highlight:: none
+
     Note that Kerbals do not have the concept of "months"::
```
tomekpiotrowski: Sphinx throws an error here. Apparently it thinks that 'Note that kerbals....' belongs to the highlight section, because it is indented. As a result this section of the file is not added to the output html.

doc/source/structures/misc/config.rst
```diff
@@ -127,11 +127,15 @@ Configuration of kOS
 
     Configures the ``EnableSafeMode`` setting.
 
+.. highlight:: none
+
     If true, then it enables the following error messages::
```
tomekpiotrowski: Same as below. Sphinx thinks that 'If true, then it enables' belongs to highlight and throws an error.

doc/source/structures/misc/config.rst
```diff
@@ -127,11 +127,15 @@ Configuration of kOS
...
 
         Tried to push NaN into the stack.
         Tried to push Infinity into the stack.
 
+.. highlight:: kerboscript
+
```
tomekpiotrowski: Same indentation problem.

doc/source/contents.rst
```diff
@@ -0,0 +1,31 @@
...
+    Addons <addons>
+    Contribute <contribute>
+    Changes <changes>
+    About <about>
+
+Indices and tables
+==================
+
+* :ref:`genindex`
+* :ref:`modindex`
```
tomekpiotrowski: How should this work? My Sphinx did not generate a `modindex` ref and as a result the link doesn't work.

doc/source/downloads_links.rst
```diff
@@ -47,18 +48,20 @@ Show and Tell, and Tutorials
 **Tutorials**
 
     - `Seth Persigehl's Tutorial playlist <https://www.youtube.com/playlist?list=PLEpK8nolWr1rghS5cOEOL-JQ33h4CgPAP>`__
+    - `CheersKevinGames <https://www.youtube.com/watch?v=fNlAME5eU3o&list=PLb6UbFXBdbCrvdXVgY_3jp5swtvW24fYv>`__
+      A channel devoded to a career mode game played with kOS from day 1, in which no
```
tomekpiotrowski: typo

doc/source/structures/misc/time.rst
hvacengi: I wonder if these instances of "none" should be using "..codeblock:: none" instead.  Also, it looks like the order is backwards here.  Highlighting is set to "none" before the kerboscript code block, and then back to kerboscript after the block.

tomekpiotrowski: I would guess that the default highlighting is kerboscript, but we choose to disable it temporarily.

tomekpiotrowski:
I've found a minor issue with the highlighting: when a suffix contains a keyword within it, for example `:ORBIT` contains an `OR`, `OR` is highlighted differently than the rest of the suffix name.

doc/source/index.rst
```diff
@@ -1,60 +1,79 @@
...
-Introduction to **kOS** and **KerboScript**
-===========================================
 
-**KerboScript** is the language used to program the CPU device attached to your vessel and **kOS** is the operating system that interprets the code you write. The program can be as simple as printing the current altitude of the vessel and as complicated as a six-axis autopilot controller taking your vessel from the launchpad to Duna and back! With **kOS**, the sky is *not* the limit.
+The do-it-yourself autopilot
+----------------------------
 
-This mod *is* compatible with `RemoteTech`_, you just have to make sure you copy the program onto the local CPU before it goes out of range of KSC.
+**kOS**, or Kerbal Operating System, is a community-supported mod for the popular game
+ `Kerbal Space Program <https://kerbalspaceprogram.com/>`_
```
tomekpiotrowski: This renders in a weird way. With a bit of indentation from the left, but no spacing from the top. Doesn't look good.

doc/source/structures/misc/time.rst
Dunbaratu: Correct.  It's turning kerboscirpt highlighting off, doing the block, then turning it back on.  This block is not quite kerboscipt syntax so the colored highlighting was borken.

I wish it was better "nested" than that ("please return highlighting to whatever it was" rather than picking the highlight explicitly and leaving it globally that way until picking a different highlight explicitly.)  But that seems to be the way it works.

As to it throwing an error - it didn't when I ran it.  I'll be on the lookout for it later when running it again, but it accepted it just fine when I ran it, and did produce the output file.

doc/source/structures/misc/config.rst
Dunbaratu: Again, I got no such complaints in my run.  I wonder if there's a sphinx version difference.

tomekpiotrowski: Sphinx 1.2.3 here.

Dunbaratu: 1.3.5 here

Dunbaratu:
I can probably fix the **:OR**BIT highlight.  It's a matter of finding out how Python regexes specify the "must be end of word boundary" rule, on the 'or' term.

Dunbaratu:
edits in response to @tomekpiotrowski;s comments have been pushed to the PR now.

PR #1472, opened 2016-02-27
Fixes #1467 - this one was very simple

### LINKED ISSUE #1467: SHIP is missing several Orbital properties. (Possible documentation issue)
This might go deeper, (as in related to basic object inheritance in the underlying code) or it's just a documentation error, but the following are undefined:
SHIP:HASORBIT
SHIP:HASOBT
SHIP:HASBODY

The following similar values DO work:
SHIP:OBT
SHIP:ORBIT
SHIP:BODY
As SHIP is a Vessel, and therefore also an Orbitable, all of these should work according to the docs.

Tested on kOS versions 0.17.3 (where SHIP:ORBIT did not work either)
and 0.18.2. Note: I doubt it matters, but KSP version was 1.0.4. I don't use 1.0.5 due to a game breaking bug.

Dunbaratu:
Somehow this got lost along the way, but our docs have been promising that it exists, and have been lying to people.  That's why I added this to merge in for the 0.19.0 milestone - it's was dead simple to implement and should be dead simple to merge in.


Dunbaratu:
To test, the suffixes (which are all aliases to the same thing) should return true for everything other than the Sun.  (as long as the object orbits another object, or is landed on another object, this is true).

PR #1479, opened 2016-02-29
Artwork cleanup and additions

pgodd:
Includes new kOS Logo and complete Unity Dev package in editable formats


Dunbaratu:
This PR contains ZIP files in it.

Is that correct?  I fear they may be zips that contain the same thing as the exploded files in the repo (in which case they don't belong in the repo), but I'm not entirely sure as these files are unknown to me.

pgodd:
Cleaned up zip files left over from old release, and changed filename to remove tilde ~ symbol

src/kOS/kOS.csproj
```diff
@@ -30,7 +30,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Assembly-CSharp">
-      <HintPath>..\..\Resources\Assembly-CSharp.dll</HintPath>
+      <HintPath>..\..\..\..\Library\Application Support\Steam\SteamApps\common\Kerbal Space Program\KSP.app\Contents\Data\Managed\Assembly-CSharp.dll</HintPath>
```
erendrake: you might want to revert this change

hvacengi: It got reverted in @Dunbaratu's review.

PR #1488, opened 2016-03-01
Change 0m part name back to original

pgodd:


hvacengi:
I'm not sure what's causing the conflict here, but I'm going to handle these changes by adding a 0m legacy part folder.

PR #1497, opened 2016-03-03
Added a 'pidloop' section to the PID Loops tutorial

### LINKED ISSUE #1298: should we rework pid loop tutorial to use new built-in PIDLoop?
There is a PIDLoop tutorial that runs people through making a PID loop themselves.  It definitely should be edited to *mention* the new PIDLoop structure if nothing else.  I'm still waffling on whether or not it should actually use it up front.  It may be useful to let people see the inner thinking of how a PID loop behaves and have them derive one themselves before then using the built-in one.

Maybe it needs to be split into two paths of teaching.  One for if you don't want to know the innards and just want to only know how to tweak the gains, and the other for if you want to actually know the innards under the hood.  It might be cool if the end result of the detailed tutorial ends up being exactly the kerboscript code shown at the bottom of the PIDLoop page.

meyerweb:
Related to #1298, though I didn’t act directly on that issue (I found it after I’d made my pull request).

(Edit by hvacengi to associate the issue with the PR:)
Fixes #1298


Dunbaratu:
Thanks for the docs.   The more docs the better.

PR #1499, opened 2016-03-03
Fix hard disk file rename

### LINKED ISSUE #1094: Boot files can be larger than the drive they're set to load on
screenshot1

### LINKED ISSUE #1498: Rename File fails for default use case.
Not sure if I'm just doing something horribly wrong or the newest update (0.19) broke "RENAME".

Reproduce:
```
copy anyfile from 0.
rename file anyfile to othername.
```
Error is: File does not exist: othername________________________________

The result is that anyfile gets deleted. I tried various other naming conventions, e.g.
```
rename file "anyfile" to othername.
rename file anyfile.ks to othername.ks.
rename file "anyfile,ks" to "othername.ks".
```
etc. I actually can't get rename file to work in any situation.

The first example structure comes from the docs: http://ksp-kos.github.io/KOS_DOC/commands/files.html#rename-file-oldname-to-newname

hvacengi:
Fixes #1498
Fixes #1094

Harddisk.cs
* Update Rename method to work with directly with the files dictionary and bypass the normal Delete/Save methods
* Update FileSearch to try and improve performance by avoiding extra HarddiskFile and Dictionary instantiation

kOSProcessor.cs
* Add a check for boot file size because saving the boot file is now size checked, and will throw an exception breaking processor initialization.

### DIFF
--- a/src/kOS/Module/kOSProcessor.cs
+++ b/src/kOS/Module/kOSProcessor.cs
@@ -341,7 +341,7 @@ namespace kOS.Module
                         FileContent content = bootVolumeFile.ReadAll();
                         if (HardDisk.IsRoomFor(bootFile, content))
                         {
-                            HardDisk.Save(bootFile, bootVolumeFile.ReadAll());
+                            HardDisk.Save(bootFile, content);
                         }
                         else
                         {
--- a/src/kOS/Module/kOSProcessor.cs
+++ b/src/kOS/Module/kOSProcessor.cs
@@ -338,7 +338,18 @@ namespace kOS.Module
                     var bootVolumeFile = archive.Open(bootFile);
                     if (bootVolumeFile != null)
                     {
-                        HardDisk.Save(bootFile, bootVolumeFile.ReadAll());
+                        FileContent content = bootVolumeFile.ReadAll();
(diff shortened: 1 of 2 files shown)


src/kOS/Module/kOSProcessor.cs
```diff
@@ -338,7 +338,18 @@ public void InitObjects()
                     var bootVolumeFile = archive.Open(bootFile);
                     if (bootVolumeFile != null)
                     {
-                        HardDisk.Save(bootFile, bootVolumeFile.ReadAll());
+                        FileContent content = bootVolumeFile.ReadAll();
+                        if (HardDisk.IsRoomFor(bootFile, content))
+                        {
+                            HardDisk.Save(bootFile, bootVolumeFile.ReadAll());
```
tomekpiotrowski: We already have `bootVolumeFile.ReadAll()` as `content`

hvacengi: good point, I added that variable to improve performance and then forgot to remove the original call.  I'll update that.

tomekpiotrowski:
I've tested renames and too large bootfiles. This PR has my :+1:

PR #1500, opened 2016-03-03
Fix unit tests

### LINKED ISSUE #1432: Make sure all unit tests pass
#1400 has broken our unit tests. Some of them were fixed but there are still 35 of them that fail.

Whether or not this should be added to 0.19 is to be decided.

hvacengi:
Fixes #1432
Many changes to the unit tests to get them running with the new encapsulated types as well as the new delegate architecture.  Some tests needed to have a cpu reference added, others just needed the assert tests to check against properly typed "expected" values, while still others needed to invoke a delegate.

Tagging @erendrake @Dunbaratu and @tomekpiotrowski specifically because they have more experience with unit tests than I have, and I just edited to make them work.


Dunbaratu:
So far on a quick eyeball check it looks good to me.

PR #1506, opened 2016-03-05
Raster mask

### LINKED ISSUE #1505: terminal embetterment
(Technically I'm already working on this but wanted there to be an issue to go with it.)

Some issues were on hold a long time due to KSP 1.1, but now that we know Unity5 isn't dropping the "do it all in code" paradigm, nor relegating it to a second-class citizen of calling it "legacy", I decided to address some of the things I wanted to do for a while now.  Maybe some of these will be moot in KSP 1.1 but at least it lets us leave a KSP 1.0.5 version behind that's as good as it can be while we work on KSP 1.1 later.

Issues with the terminal:

- Fixed font size is too small to read in high res situations.  Users can only adjust how many rows/cols, not how big the font is.
- Sometimes too dim to see.
- Focus is communicated by making the terminal too hard to read - all transparent and fuzzy - communicate it in a way that can still leave the unfocused terminal nice and clearly readable.
- Terminal becomes invisible when game is paused - it should be still visible.
- Terminal does not obey the F2 'hide controls' button.

Dunbaratu:
Fixes #1505.

Despite the branch being called "raster_mask", all the stuff
related to actually doing a raster mask has been removed,
to concentrate on just the other terminal fixes.

Note there is a slight weirdness with the window resizing
when you change the font.

Let's say the window is currently size 400 pixels wide, and is
holding 50 chars at 8 pixels each.  Then you increase it to 12
pixels each. 400 doesn't divide by 12 evenly, so when it recalcs
the new number of chars that fit it does some rounding.
Technically 33.33333333 chars fit now in the 400 pixel width,
but since that's not possible it cuts it down to 33 characters.
Okay, that's fine so 33 characters is the new width.  It calls the
routines to set the terminal width to 33, which in turn go ask the
terminal how many pixels that is.  Since it's not 33.333333 chars,
but is 33 chars, the number of pixels returned is 398, not 400.
So the terminal resizes itself to 398 pixels wide.

The same thing happens in vertical height too.

Fixing that would mean a massive overhaul to the logic that is
**suposed to resize** the terminal when the number of chars
changes via, say, telnet or via set terminal:width to 40.


src/kOS/Screen/TermWindow.cs
```diff
@@ -108,6 +131,39 @@ public void Awake()
             DontDestroyOnLoad(gObj);
             popupEditor = (KOSTextEditPopup)gObj.GetComponent(typeof(KOSTextEditPopup));
             popupEditor.SetUniqueId(UniqueId + 5);
+            
+            customSkin = BuildPanelSkin();
+
+            GameEvents.onHideUI.Add (OnHideUI);
+			GameEvents.onShowUI.Add (OnShowUI);
```
Dunbaratu: Odd... these two lines seem to be showing as not indented the same in here Github. They look indented the same in my editor.

hvacengi: They're indented properly on my editor too.  I wouldn't worry about it.

hvacengi: Oh, they're tabs, not spaces.  That's the issue.

Dunbaratu: Odd.  I have my editor configured to indent with only spaces.  Maybe these were tabs I actually typed in.

ohhhh... I cut-n-pasted from a code example I saw in the forums.  I bet the tabs were in the pasted text.

PR #1515, opened 2016-03-06
Fixes #1513 terminal font fails on low texture setting.

### LINKED ISSUE #1513: characters in terminal are rectangular blocks
from the forums:
This in game terminal looks like this:
img

Dunbaratu:
The problem was leaving the "mipmap" option set to true when cutting up the font_sml.png Texture2D into 128 separate smaller character Texture2D's.  That option made Unity apply the graphics quality settings at *that* point when what we really want is just a straight-up copy of every pixel as-is.  It's not until later when drawing the texture that we want to let Unity do texture quality manipulation if it wants to.

### DIFF
--- a/src/kOS/Screen/TermWindow.cs
+++ b/src/kOS/Screen/TermWindow.cs
@@ -13,11 +13,6 @@ namespace kOS.Screen
     // Blockotronix 550 Computor Monitor
     public class TermWindow : KOSManagedWindow , ITermWindow
     {
-        /// <summary>
-        /// Pixel size of one square section of the font template image file holding one character.
-        /// </summary>
-        private int charSourceSize;
-        
         private const string CONTROL_LOCKOUT = "kOSTerminal";
         private const int FONTIMAGE_CHARS_PER_ROW = 16;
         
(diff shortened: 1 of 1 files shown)


src/kOS/Screen/TermWindow.cs
```diff
@@ -14,9 +14,9 @@ namespace kOS.Screen
     public class TermWindow : KOSManagedWindow , ITermWindow
     {
         /// <summary>
-        /// Pixel size of one square section of the font template image file holding one character:
+        /// Pixel size of one square section of the font template image file holding one character.
         /// </summary>
-        private const int CHAR_SOURCE_SIZE = 8;
+        private int charSourceSize;
```
Dunbaratu: While I was at it, I also made it calculate this on the fly from the image file (under the assumption the image file contains exactly 16 character images per row) instead of hardcoding it.  That means we can drop-in replacement font images and see what it looks like, or even let the end-user do that, and they can be whatever size scale we like without recompiling the code.  Thus someone who wants to could make, say, a 192x192 image for font_sml.png, with each character being 12x12 instead of 8x8, and the system would happily use it without recompiling.

src/kOS/Screen/TermWindow.cs
```diff
@@ -155,7 +161,7 @@ private void LoadFontArray()
                 // SetPixel on the texture (which we do).  So we start it off as a ARGB32
                 // first, long enough to perform the SetPixel call, then compress it
                 // afterward into a DXT5:
-                Texture2D charImage = new Texture2D(CHAR_SOURCE_SIZE, CHAR_SOURCE_SIZE, TextureFormat.ARGB32, true);
+                Texture2D charImage = new Texture2D(charSourceSize, charSourceSize, TextureFormat.ARGB32, false);
```
Dunbaratu: The very last part of this one single line is actually the relevant bug fix - turning that `true` into a `false`.  Everything else in the PR is just the effect of making CHAR_SOURCE_SIZE no longer be hardcoded.

PR #1523, opened 2016-03-07
Issue 1520 calculator structure

mthiffau:
I've added .Equals and == operators to Direction and TimeStamp which didn't have them before. As I understand this is to help move away from TryOperation.

### DIFF
--- a/src/kOS/Suffixed/Direction.cs
+++ b/src/kOS/Suffixed/Direction.cs
@@ -174,6 +174,38 @@ namespace kOS.Suffixed
         {
             return new Direction(a.rotation.Inverse());
         }
+
+        public override bool Equals(object obj)
+        {
+            Direction d = obj as Direction;
+            if (d != null)
+            {
+                // Check the equality of the quaternion components
+                if (d.rotation.x == rotation.x &&
+                    d.rotation.y == rotation.y &&
+                    d.rotation.z == rotation.z)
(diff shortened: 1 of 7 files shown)


src/kOS/Suffixed/Direction.cs
```diff
@@ -174,6 +174,38 @@ private void DirectionInitializeSuffixes()
...
+
+        public override bool Equals(object obj)
+        {
+            Direction d = obj as Direction;
+            if (d != null)
+            {
+                // Check the equality of the quaternion components
+                if (d.rotation.x == rotation.x &&
+                    d.rotation.y == rotation.y &&
+                    d.rotation.z == rotation.z)
```
Dunbaratu: `UnityEngine.Quaternion` already has its own `Equals()` operator method defined.  It might be a good idea to just defer to using that instead of re-implementing it (and I had a peek in ILSpy and it turns out what you're doing here i almost the same thing Unity is doing for Equals, with the exception that you only compare the `x`, `y`, and `z` and Unity is comparing all 4 components (including the `w`).)

Dunbaratu:
Removed milestone because we want more time to work this through and test it - just took the infnite recursion fix and nothing else for now.

hvacengi:
I'm adding the 1.0 milestone so that we don't miss it in that version

PR #1526, opened 2016-03-07
Fixes #1525 by paying attention to current color again.

### LINKED ISSUE #1525: new terminal doesn't show color change on power-off
Pre 0.19.1 behavior:  terminal changed text to pale grey when CPU powered off.
0.19.1 behavior: terminal has the same exact green color regardless of if the CPU is on.

A very minor bug but it should be easy to find and fix.

Dunbaratu:


src/kOS/Screen/TermWindow.cs
```diff
@@ -862,10 +863,10 @@ protected void CheckResizeDrag()
             }
         }
         
-        void ShowCharacterByAscii(char ch, int x, int y, Color charTextColor, bool reversingScreen, int charWidth, int charHeight, float brightness)
+        void ShowCharacterByAscii(char ch, int x, int y, bool reversingScreen, int charWidth, int charHeight, float brightness)
```
Dunbaratu: Removed this arg when I realized it wasn't even being used.

src/kOS/Screen/TermWindow.cs
```diff
@@ -862,10 +863,10 @@ protected void CheckResizeDrag()
...
         {
             GUI.BeginGroup(new Rect((x * charWidth), (y * charHeight), charWidth, charHeight));
-            GUI.color = AdjustColor(reversingScreen ? bgColor : textColor, brightness);            
+            GUI.color = AdjustColor(reversingScreen ? bgColor : currentTextColor, brightness);
```
Dunbaratu: textColor is the TermWindow.cs field tht holds the normal text color (green).

currentTextColor is the TermWindow.cs field that swaps between either textColor or textColorOff depending on pwer state.  In v0.19.1, the code to set currentTextColor that way was still present and it was still doing it, but the code that actually uses it when drawing the letter was edited and ended up ignoring it and just using textColor directly.  This was the problem.

PR #1530, opened 2016-03-08
Update readme files

### LINKED ISSUE #1533: Update README files
We should review the various README style documents for the repository and update instructions to be more complete, and more current.

I've actually already submitted a PR for this issue, but since one of the elements in CONTRIBUTING.md is that all PR's need an issue, here's the associated issue.

hvacengi:
Fixes #1533 

Updates to readme files and .gitignore based on my experience with setting up a release.

Also included more instructions for new users to follow in setting up the solution and git repository.


CONTRIBUTING.md
```diff
@@ -1,32 +1,90 @@
...
+* [Setting Up Your Environment](#setting-up-your-environment)
+  * [Assumptions](#assumptions)
+  * [Your Repository](#setting-up-your-repository)
+  * [Solution Dependencies](#setting-up-the-solution-dependencies)
+
 Pull Requests
 =============
 
+* Please reference [Style Guide](STYLEGUIDE.md)
 * All PRs must be related to an open issue
```
Dunbaratu: I'd say append a second sentence here: "If you want to make a PR for something that doesn't have an issue yet, do it by first making a new issue for it, then submitting the PR."

Make it clear that we're not trying to say "only work on the already identified stuff".

CONTRIBUTING.md
```diff
@@ -1,32 +1,90 @@
...
 ===========================
+####Assumptions
+* `$KSP` is the full path to your Kerbal Space Program installation directory.
+  (i.e. `"C:\Program Files (x86)\Steam\SteamApps\common\Kerbal Space Program"`)
+* `$KOS` is the full path to your KOS git repository.
+* Other than examples of full paths for Windows, all paths are shown in this
+  guide using the unix directory seperator of `/`.  Please change all instances
+  of `/` and `\` to match the style of your development environment.
+* You already have a C# IDE and are familiar with how to use it to build a
+  Visual Studio solution.
```
Dunbaratu: I'd call it a ".net solution", just so it doesn't undermine the claim you can use use any IDE.

CONTRIBUTING.md
```diff
@@ -1,32 +1,90 @@
...
+
+####Setting Up Your repository
+1. Use the github web interface to create a fork of KSP-KOS/KOS
+2. Your fork should have a web address like `https://github.com/[username]/KOS`
+  and a git address like `https://github.com/[username]/KOS.git`
```
Dunbaratu: Mine's been 'KOS-1' instead of 'KOS' for ages and nobody complained.  I think it was like that because I'd already had a KOS one that was made from nivekk's sources earlier, and the name clash made github append the "-1".

CONTRIBUTING.md
```diff
@@ -1,32 +1,90 @@
...
+  ```
+  See the notes above regarding `rebase`
+8. You may push your branch with edits to your `origin` on github, and submit a
+  pull request to KSP-KOS/KOS `develop` for review to be included.
+
+####Setting Up The Solution Dependencies
 
-1. Copy `Resources/GameData/kOS` to `$KSP/GameData/`, `where $KSP` is your
-   Kerbal Space Program installation directory.
+1. Copy the folder `$KOS/Resources/GameData/kOS` to `$KSP/GameData/`
```
Dunbaratu: Uhm.  No?  I only put ModuleManager there because I supposedly had to.  KOSExtensions on the other hand, remains in the kOS subfolder on my install.
I thought the rule was that $KOS/Resources *must* match the installed folder layout.  That command you give above changes what folder a file goes in by moving it up one directory, from kOS to the parent GameData folder.

CONTRIBUTING.md
```diff
@@ -1,32 +1,90 @@
...
-          <PropertyGroup>
-            <ReferencePath>
-              /path/to/KSP/KSP_Data/Managed;
-              /path/to/KSP/GameData/kOS/Plugins
-            </ReferencePath>
-          </PropertyGroup>
-        </Project>
-   
+   https://github.com/Swamp-Ig/KSPAPIExtensions/releases, and copy
+   it to `$KSP/GameData/kOS/Plugins` and `$KOS/Resources`.
```
Dunbaratu: Again, this instruction makes the folder layout of the Resources folder differ from the installed folder.  If you want it to land in $KSP/GameData/ when it gets installed, then it needs to be in $KOS/Resources/GameData.  If you want it to land in $KSP/GameData/kOS/Plugins, then it needs to be in $KOS/Resources/GameData/kOS/Plugins.

Remember that we make the ZIP by essentially grabbing a copy of the Resources folder.  So it needs the same layout as what's to be installed.

doc/README.md
```diff
@@ -1,4 +1,60 @@
...
+  C:\>pip install sphinx
+  ```
+4. Ensure that the read the docs template is installed.
+  ```
+  C:\>pip install sphinx_rtd_theme
+  ```
+5. Switch to the docs directory and run the make batch file:
+  ```
+  C:\>cd KOS\docs
+  C:\KOS\docs>make html
```
Dunbaratu: I'd add the `make clean` beforehand just to be sure.  Sure the first time there's no difference, but after that sometimes you make a change that affects more than just the source file that was itself touched.  For example, if a file has a tag :ref: in it, then it needs to be recompiled when the target of that tag moves.  Even if the file pointing to it itself didn't change, the html built from it needs a new URL placed there.

doc/README.md
```diff
@@ -1,4 +1,60 @@
...
+  Python's included SimpleHTTPServer:
+  ```
+  C:\KOS\docs>cd gh-pages
+  C:\KOS\docs\gh-pages>python -m SimpleHTTPServer 8000
+  Serving HTTP on 0.0.0.0 port 8000...
+  ```
+  At which point you can point your browser to `http://localhost:8000`
+
+#Getting started on Linux
+1. TODO: WRITE LINUX INSTRUCTIONS WITH APT-GET AND PIP
```
Dunbaratu: Other than how you install the application in the first place, the rest of the instructions are actually the same.

doc/README.md
```diff
@@ -1,4 +1,60 @@
...
+
+#Publishing
+1. Clone KSP-KOS/KOS gh-pages branch if not done already, pull the most recent
+  version from the upstream repository if you already have a clone.
+2. The clone should either be in a different folder from the primary
+  repository or you need to copy the rendered `doc/gh-pages` contents to a
+  temporary folder
+3. Delete the contents of the repository root, except the `.git` folder,
+  `.gitatributes` file, and `.nojekyll` file.
```
Dunbaratu: I'd just say `checkout gh-pages`.
That will first re-arrage the folders as desired, and make sure that your changes are on the right branch.
Then wipe the contents and copy files back. 
For wiping contents it would be cleaner to just say "Delete everything, recursing through all directories,*except for* those files and directories that begin with a dot, *for example* the .git folder, .gitattributes, and so on."  Don't explicitly give a whitelist of only specifically named files to preserve and say wipe everything else, but rather mention the pattern of what files to preserve - all the ones starting with dot.  The unix wildcard '*' does this rule exactly (which is why that naming convention got used).  Of course I don't know about what windows does.

doc/README.md
```diff
@@ -1,4 +1,60 @@
...
+4. Copy the contents of `doc/gh-pages` into the `gh-pages` repository root folder
+5. Commit all of the changed files
+6. Push the gh-pages branch to your own kOS github repository
+7. You can test the documents at `https://[username].github.io/kOS`
```
Dunbaratu: uppercase 'K' on 'KOS'.  I don't know if github is using a case-sensitive web server or not.

doc/README.md
```diff
@@ -1,4 +1,60 @@
...
+
+#Getting started on Linux
+1. TODO: WRITE LINUX INSTRUCTIONS WITH APT-GET AND PIP
+
+#Publishing
+1. Clone KSP-KOS/KOS gh-pages branch if not done already, pull the most recent
+  version from the upstream repository if you already have a clone.
+2. The clone should either be in a different folder from the primary
+  repository or you need to copy the rendered `doc/gh-pages` contents to a
+  temporary folder
```
Dunbaratu: It may be better to just give one of these choices in order to make the instructions cleaner from here on down, and more specific.
I like making it an entirely separate second repository.  Less panic when the user sees, "oh no where'd all my work go?!", and less likely to wipe what they were working on if they forgot to commit, or didn't do the copy of gh-pages correctly.
Also, it would let us be very specific which repo copy we meant.  If the instruction for making the new clone gave it a specific name, we could then say filename paths using that exact name and be very clear exactly which repo we are asking people to delete files from.

CONTRIBUTING.md
hvacengi: I think I'll call it a ".sln solution" because technically the solution file is not tied to a project being .net either (there is a way to use it for arduino projects, for example).

CONTRIBUTING.md
hvacengi: This is the issue when dealing explicitly with paths.  The first path is the folder itself, not the contents.  So the instruction is to copy the **folder** into the GameData folder, not the contents.  MM should not be located inside of the kOS folder itself, but in the GameData folder.  

But when dealing with paths, people seem to interpret things differently.  This is why ckan is nice, there were way too many people ending up with folder structures like `$KSP/GameData/kOS-v19.3.0/GameData/kOS`.

In your case, you interpreted that the instructions were to copy the **contents** of the kOS folder into `GameData`.

Can you tell me a better way for me to explain this so it is absolutely clear?

CONTRIBUTING.md
hvacengi: I will update the readme to put them into the `$KOS/Resources/GameData/kOS/Plugins` folder, and update the solution file's hint path to match (since the hint currently points to just `Resources`)

doc/README.md
hvacengi: I'll look at just writing out a quick thing about apt-get.  You don't know off hand if `apt-get install python` does 2.7.x by default, or 3.x?

hvacengi:
Hm, I just found the problem with setting the hint path for KSPAPIExtensions.dll to be the plugin folder, that isn't where it is extracted during the travis build, so travis fails.

doc/README.md
hvacengi: Windows will not ignore files starting with `.`, it will however ignore "read only" files.  I would need to look into whether or not we can set the readonly attribute via git to allow that to work.  I used that phrasing because when I do the doc's release, I do it from windows explorer and `[Ctrl]+[A]`, then deselect the 3 files, then `[Del]`.

The bottom line is that this is a really confusing section of the process, because we need to delete the source before we commit the files themselves.  I could actually provide instructions using `git stash` which may turn out to be the easiest option to explain.

Dunbaratu: I wasn't saying "you should exclude all files starting with a dot because windows does".  I was saying you should because it catches everything, and it prevents accidentally adding files you don't want to add if you're on a unix system, where often tools add dot files all the time willy-nilly under the assumption it's always safe because nobody will see them unless they explicitly ask to.  When interacting with `git`, that changes because `git` won't treat them specially like most unix apps will.  So some other app may leave have a ".myprivate-settings" file or a ".backup_of_text_file_during_editing" file in the repo folder and you don't notice it.

hvacengi: Using stash won't work because of how file tracking is managed, so that idea is out.

>"Delete everything, recursing through all directories,except for those files and directories that begin with a dot, for example the .git folder, .gitattributes, and so on."

>I was saying you should because it catches everything, and it prevents accidentally adding files you don't want to add if you're on a unix system, where often tools add dot files all the time willy-nilly...

The way I read those two statements is contradictory.  First you tell me to give instructions to delete everything without a leading `.`, because you want to preserve those files.  Then you say that unix adds these `.` files every where and you want to avoid committing these extra files.

My understanding is that we want to only keep the files `.gitattributes` and `.nojekyll` and the directory `.git`.  Nothing else. Then we copy the contents of the gh-pages folder into the root.

If they didn't clean the working directory properly, we don't want to keep the `.ci` folder, nor do we want to include a random `.` file that their system added to the directory, which isn't tracked and therefore not deleted by the checkout command.  (this is probably the biggest issue, untracked files aren't touched by git commands).  So I think we do want to use a white list, so that there isn't a random file or folder kept in the folder which shouldn't be in the repo.  If we need to preserve additional files in the future, we would just need to update the list.

If what you really want me to do is tell them explicitly to delete all files with a leading `.`, excluding those on the list, then I will need you to give me an appropriate command to do so.

Dunbaratu: It's a little of both.  All dot files are presumed to not belong in the repo except where explicitly noted (``.gitignore`` is one exception we explicitly made).  That means both not adding dot files to the repo, and not assuming the repo will repopulate your dot files for you when you want to switch back to the normal branches and not work in gh-pages anymore.

It's the last one that I have a problem with in your instructions.  You're telling people to destroy their only copy of the dot file, under the assumption it will come back again when they switch branches back, and it won't because it's not a tracked file (actually, that same complaint extends to the entire way its being done here, where it's being done in the same location as the normal development, so all local untracked files have to be destroyed in the user's local working folder).

I strongly dislike the way github organized gh-pages for this reason.  It really should have put them in a separate subdirectory so you don't have to destroy all your local untracked files when you switch branches to it.

Here's how I deal with the gh-pages mess:  I make a totally separate secondary clone of the fork that I use ONLY for gh-pages, and then after the ``make html`` step, I do the copy from the main repo's gh-pages directory that ``make html`` created to this secondary repo's root folder, and perform the commit and push of gh-pages from the secondary one.  That way I don't need to clobber all my local untracked stuff.  I think that may make for simpler instructions, and make my comment about dot files moot.

hvacengi: So, if we recommend using the dual repositories idea, does that eliminate the concern for maintaining the dot files?  I'm thinking it's gonna get confusing if we have a side conversation discussing the files that are safe to delete compared to the files that are safe to add to the commit.

Dunbaratu: Yes, that's essentially why I do it that way (that plus it means never having to worry about deleting a file it turns out I forgot to put under repo control but should have).

hvacengi:
@Dunbaratu OK, check out the doc publishing guide now.  I updated it with instructions for a 2nd clone.

I also updated the link in the README.md file in the root of the repository, and merged develop into the branch.

CONTRIBUTING.md
```diff
@@ -1,32 +1,101 @@
...
+
 Pull Requests
 =============
 
-* All PRs must be related to an open issue
-* If a PR changes the scripting API, it must include all related changes to the documentation. 
+* Please reference [Style Guide](STYLEGUIDE.md)
+* All PRs must be related to an open issue.  If there isn't already an issue
+  that applies to your PR, please submit a new issue with the PR.
+* If a PR changes the scripting API, it must include all related changes to the documentation.
```
Dunbaratu: Maybe append the sentence: "The documentation is edited by changing or adding to the *.rst files under /doc/source."

hvacengi: OK, I can do that, or you can do it at the end of your review, whatever works better for you.

CONTRIBUTING.md
```diff
@@ -1,32 +1,101 @@
...
+
+  If you have not made any local conflicting changes to the `develop` branch
+  itself, the "fast forward" pull will update your local branch without adding
+  a merge commit.
+
+7. If you need to update a local branch you are working on, you may "pull" the
+  upstream branch or rebase:
+  ```
+  git pull upstream develop
+  ```
```
Dunbaratu: To be absolutely certain people are updating their local develop branches properly, to make their future branches use it, I usually prefer to do this in two steps:
step 1 : get my local develop up to date with upstream develop: ``git checkout develop ; git pull --ff-only upstream develop``
step 2 : merge local develop into local branch i'm working on: ``git checkout whatever-my-branch-is-called-again ; git merge develop``

Dunbaratu: The above also avoids doing the ``pull`` operation, which I normally really don't like as it tries to fetch and merge in the same step and if there's some ugly conflict and I want to undo it and think it over, it can be a PITA to undo.

CONTRIBUTING.md
```diff
@@ -1,32 +1,101 @@
...
+
+####Setting Up The Solution Dependencies
+1. Download the latest KSPAPIExtensions.dll from
+   https://github.com/Swamp-Ig/KSPAPIExtensions/releases, and copy
+   it to `$KSP/GameData/kOS/Plugins` and `$KOS/Resources/GameData/kOS/Plugins`.
+
+2. Copy the folder `$KOS/Resources/GameData/kOS` to `$KSP/GameData/`
+
+3. Copy `Unity-Engine.dll` and `Assembly-Csharp.dll` from `$KSP/KSP_Data/Managed`
+  into `$KOS/Resources`.  If you do not have a copy of KSP locally, you may
```
Dunbaratu: Append the phrase "from $KSP/KSP_Data/Managed/" maybe ?  A new modder won't realize these are bundled with the KSP main game.  I remember when I started, getting very frustrated at modding instructions that just said to copy these files, without telling me they're part of the main game.  I kept looking around for them online.  Especially since the name ``Assembly-Csharp.dll`` that SQUAD chose to use does not make it clear *at all* that this is a file they made.  It really sounds like the sort of thing you're supposed to go get from Microsoft to start working in CSharp.

RELEASECHECKLIST.md
```diff
@@ -3,16 +3,24 @@
...
 * Update AssemblyInfo for kOS project
 * Update AssemblyInfo for kOS.Safe project
-* Update Resources\GameData\kOS\kOS.version
-* Update doc\source\conf.py
-* Update CHANGELOG.MD
+* Update AssemblyInfo for kOS.Safe.Test project
+* Update `Resources\GameData\kOS\kOS.version`
+* Update `doc\source\conf.py`
+* Update `CHANGELOG.MD`
+* Update `doc\source\changes.rst`
```
Dunbaratu: I think *.md files want two backticks.  I'll make the edit before I merge.

CONTRIBUTING.md
hvacengi: What I like about using the pull on the branch itself is that it will always check the upstream branch, rather than requiring you to make sure that you already synced the local develop branch.  

It also means that the resulting commit message identifies the remote repository address ("Merge branch 'develop' of https://github.com/KSP-KOS/KOS").  And because this pull doesn't use either `--ff-only` or `--no-ff` you'll get a fast forward if it's possible which is one less commit.

If you feel strongly about it, feel free to make the change.  I think this is a matter of personal preference more than it is anything else, and I don't mind my preference being the example.

Dunbaratu: Well, I guess it's fine so long as the phrase "for example" is appended so it doesn't look like we demand it has to be done with that exact command.

hvacengi: Undoing things in git is not much of a PITA in my experience, unless you force something.  If you don't like the pull result, you can always `git reset [last good commit]`.

There is a large sect of git users who don't like using pull for anything, for the reason you just mentioned.  But all it really does is chain the two commands, it doesn't really do them at the same time.  It's functionally equivalent to
```
git fetch
git merge FETCH_HEAD
```
Again, I don't mind being overruled here if you feel strongly.

RELEASECHECKLIST.md
```diff
@@ -3,16 +3,24 @@
...
-* Name the zip file with the following pattern kOS.v<major>.<minor>.<patch>.zip (eg kOS.v0.14.2.zip )
-* Build the documentation in \docs\
+* Run all unit tests
+* Ensure that all required resources are in place
+  * `Resources/GameData/kOS/Plugins/KSPAPIExtensions.dll`
+  * `Resources/GameData/ModuleManager.dll`
+* Create zip file with a root starting in the `\Resources\` directory
+* The zip file should have the GameData folder in the root
+* Name the zip file with the following pattern `kOS-v<major>.<minor>.<patch>.zip` (eg kOS-v0.14.2.zip )
+* Build the documentation in `\docs\`
```
Dunbaratu: Since you're also writing up documentation on how to do this, it may be a good idea to link to that documentation from this line.

Dunbaratu: ah no, nevermind.  This is the 'how to release' checklist, not the instructions for newbs - so never mind, don't do that.

CONTRIBUTING.md
hvacengi: > Well, I guess it's fine so long as the phrase "for example" is appended so it doesn't look like we demand it has to be done with that exact command.

At that point we probably should just add a line at the beginning saying that these are just recommended git commands, the user is free to use their own style, experience, and knowledge.

Dunbaratu: I think it's still useful to say 'for example' on each command where we don't really care.  It matters because in this case, as you say, the way the commit history looks is actually *different* between doing it your way versus mine.  Someone who knows that can come away with the impression that we want the effect of making the commit history look exactly like this command makes it look, and therefore my way is "wrong".

doc/README.md
```diff
@@ -1,4 +1,129 @@
...
+  path character with `/` and adapting paths to reference your Linux file system.
+
+#Publishing
+1. We recommend creating a second clone repository for managing Github pages
+  publishing.  You should also add KSP-KOS/KOS as the upstream remote:
+  ```
+  C:\KOS>cd ..
+  C:\>git clone https://github.com/[username]/KOS.git KOS_DOC
+  C:\>cd KOS_DOC
+  C:\KOS_DOC>git remote add upstream https://github.com/KSP-KOS/KOS.git
```
Dunbaratu: I'm going to slightly edit this section and the lines below to pick a different name from "KOS_DOC".  (The problem being that KSP-KOS actually *has* a fork called KOS_DOC, and the fact that you're calling the local copy "KOS_DOC" despite the fact that it's actually for KOS and not for KOS_DOC could steer people the wrong way.

doc/README.md
```diff
@@ -1,4 +1,129 @@
...
+  ```
+
+3. Checkout the gh-pages branch and update it to match the upstream version:
+  ```
+  C:\KOS_DOC>git checkout gh-pages
+  C:\KOS_DOC>git pull --ff-only upstream gh-pages
+  ```
+
+  (You may delete other branches from this clone if you want, but make sure you
+  so not delete the remote branch)
```
Dunbaratu: "make sure you so not delete the..." doesn't parse.

hvacengi: do*

doc/README.md
```diff
@@ -1,4 +1,129 @@
...
+  At which point you can point your browser to `http://localhost:8000`
+
+#Getting started on Linux
+1. As with Windows above, install Python 2.7.  You may use your distribution's
+  package manager system, or download from: https://www.python.org/downloads/
+
+2. All other instructions are the same as above for windows, replacing the `\`
+  path character with `/` and adapting paths to reference your Linux file system.
+
+#Publishing
```
Dunbaratu: I think I'll add a clarifying preamble here that points out this is NOT something everyone has to do with every PR, that this is part of the release process and is usually not something every contributor has to do, just one person.

RELEASECHECKLIST.md
Dunbaratu: I just did a test and github DOES allow single-backtick to work in its MD files, but NOT in its comments (like this one) even though it claims both are using its Markdown.  I have no idea why it's inconsistent about this, but the point is, I don't have to change the single backticks after all.

PR #1534, opened 2016-03-09
Fixes #1532 implementing some profiling output.

### LINKED ISSUE #1532: Better profiling of kRISC program
Right now we have the ability to get an execution report at the tail end of a program run - how much time was spent on it.

I'd like to see that information broken down more specifically and I think it would be easy to implement.

A program is a `List<Opcode>` - if we could attach a number to each opcode in that list for "milliseconds_elapsed", and have the CPU add to it as it runs through executing each instruction, then it could give a report at the end showing the whole program with its timings in a column on the right.

Dunbaratu:
To use:

```
    set config:stat to true.
    run something.
    print ProfileResult().
    // or:
    log ProfileResult() to somefilename.txt.
```

### DIFF
--- a/doc/source/bindings.rst
+++ b/doc/source/bindings.rst
@@ -419,6 +419,51 @@ But kOS does not support this at the moment so in kOS if you set
 the LOADDISTANCE, you are setting it to the same value
 universally for all situations.
 
+.. _profileresult:
+
+PROFILERESULT()
+---------------
+
+If you have the runtime statistics configuration option
+:attr:`Config:STAT` set to ``True``, then in addition to
+the summary statistics after the program run, you can also
+see a detailed report of the "profiling" result of your 
+most recent program run, by calling the built-in function
+``ProfileResult()``.  *"Profiling"* is a programmer's term
+that means gathering data about how long the program is
+spending doing each piece of the program.  If you are trying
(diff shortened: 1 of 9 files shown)


Dunbaratu:
Reason for the "not ready" is just that I want to work on correcting the output formatting.  My understanding of how to use ``String.Format()`` is limited and a bunch of searching through the .net docs still doesn't help me figure out how to get what I want, which is this:

```
  0        3
  1.0     12
 12.01     3
123      789
123.423    4
```

But the best I can manage is this:
```
  0.000    3
  1.000   12
 12.010    3
123.000  789
123.423    4
```
Or this:
```
  0    3
  1.0   12
 12.01    3
123  789
123.423    4
```

I can't figure out how to tell it "yes, pad so the decimal point lines up with N digits after the point, but don't pad those N places with zeros - but definitely DO pad them so the next column lines up - just pad them with spaces not zeroes."

I also have no clue why the lines under the headings, that go "======" refuse to line up the same way the heading text does.

Dunbaratu:
I gave up on trying to get the numbers to line up  in text form nicely without the zero padding.  Now that it's going to be going to a CSV format anyway, that formatting doesn't matter too much.

PR #1542, opened 2016-03-15
Fixes #1537 interruptable triggers

### LINKED ISSUE #1376: `on stage:ready` doesn't work.
The on command is supposed to activate a trigger whenever a boolean changes state. `stage:ready` is a boolean, but for some reason `on stage:ready` always evaluates to true.

### LINKED ISSUE #1522: Alternative to WAIT inside WHEN/THEN and ON triggers?
The way it's currently possible to bypass the inability to `WAIT delaytime.` inside a trigger is to save a timestamp somewhere and create another `WHEN` trigger with condition based on the time: 
`WHEN TIME:SECONDS>=timestamp+delaytime THEN`
This way the new trigger will be called after the set amount of time and will continue the triggered sequence.
(and with cases of `WAIT UNTIL` you can just create the new trigger with the very expression)

As alternative, it could be useful to have a command that allows this to be done without multilevel triggers. A command (`DELAY`? `SUSPEND`? the same `WAIT` keyword?) would suspend the trigger body execution at this position (so that mainline and other triggers can execute in the meantime) and create another trigger (with timestamp or boolean expression) which will continue this trigger body from right after this command.

The primary need for this functionality is in multistage trigger sequences. For example, LES sequence could look like this:
```
ON ABORT
{ if LES_allowed()  //check it hasn't been jettisoned
{
UNLOCK steering. 
SAS OFF. 
decouple_capsule().
ignite_LES().
DELAY 1.5.
jettison_LES().
DELAY 3.
DELAY UNTIL chute_deployment_conditions().
CHUTES ON.
}}
```

### LINKED ISSUE #1537: An easy(?) way to make triggers interruptable, today, and get rid of the IPU limit problem.
(After some talking about it on Slack, I decided I really want to do this, and as per the rules, I'm making an issue describing it first before trying to actually make a PR and submit code for it.)

This may also make some old issues moot and allow us to remove them.  If implemented it will require going back through the issue history and revising or closing some of them.

The problem:  Right now we have the Instructions Per Update limit on triggers and on the lock expressions that are used to supply values to steering manager (``throttle``,``steering``,``wheelthrottle``,``wheelsteering``).

Fixing this was always put off as being something to wait for full threading to deal with it differently, but that may not be necessary.  A possibly easy fix may be doable.

First, I'll review why the limit exists in the first place (why all triggers have to finish within one update and thus cannot have more instructions than an update can do).  It exists because the FixedUpdate algorithm in CPU.cs is this:
```
Start of a FixedUpdate pass
    instructions so far this update = 0
    foreach magic control variable (throttle, steering etc)
        execute the user-land code to eval its lock expression
        (increment instructions so far this update by how many instructions that took)
    End-For
    foreach active trigger
        Run the user-land code to perform the Boolean condition check
        (increment instructions so far this update by how many instructions that took)
        If the Boolean check was true
            execute the whole trigger body to the bottom
            (increment instructions so far this update by how many instructions that took)
        End-if
    End-for
    Starting from where the instruction pointer left off last time in main code:
        While instructions so far this update < IPU setting
        |    Perform 1 mainline instruction
        |    If instruction said to please break (EOP) or yield (WAIT)
        |<------quit while-loop prematurely.
        |    move instruction pointer to next instruction
        |    increment instructions so far this update +1
        End-while
    Instruction pointer is now aligned with the next instruction to be executed next update.
End of a FixedUpdate pass
```

The problem is that you can't ``wait`` during a trigger body, or even just let it run out of IPU during a trigger body, purely because that instruction pointer work mentioned in the main loop body isn't done for triggers, and the system can't remember where it was in the midst of the trigger's body.

So my fix is this:

- We already *have* an infrastructure in place for doing subroutine calls that can call other subroutines that call other subroutines and so on, and all the stack tracking work needed to make that happen, and to make it return right where it was when it's done popping its way back from all those calls, with the stack left just like it was, and ready to resume.
- You can even call subroutines from the middle of other expressions: `set x to y + get_z(3) / 4.`, and the containing expression doesn't break by the funky stack values that were inserted by the function - because everything pops back the exactly the state it was in where it started the call from.
- This system is already very well tested and gets exercised a lot.
- So, why not make the bodies of triggers *be* subroutine calls in every single way, so they can interrupt mainline code wherever, by *being* treated like mainline code itself, as if the mainline code had inserted a subroutine call to the trigger right in the middle of whatever else it was doing?  Then they can get interrupted like mainline code, and resumed like mainline code.

That would change the algorithm to basically this:
```
Start of a FixedUpdate pass
    instructions so far this update = 0
    foreach magic control variable (i.e. throttle, steering. etc)
        execute the user-land code to eval its lock expression
        (increment instructions so far this update by how many instructions that took)
    End-For
    foreach active trigger
        Run the user-land code to perform the Boolean condition check
        (increment instructions so far this update by how many instructions that took)
        If the Boolean check was true
            + Insert the stack frame information that would happen for a function call.
            +  (this includes storing the current instruction pointer in the stack frame as the "come back to" IP.)
            + Set the instruction pointer to the start of the trigger's body, but do NOT execute it.
        End-if
    End-for
    + Starting from EITHER:
        + where the instruction pointer left off last time, OR
        + wherever the trigger check above just left the instruction pointer set to.
        While instructions so far this update < IPU setting
             Perform 1 mainline instruction
             move instruction pointer to next instruction
             increment instructions so far this update +1
        End-while
    Instruction pointer is now aligned with the next instruction to be executed next update.
End of a FixedUpdate pass
```

The idea is that you execute the body of the trigger in the mainline code because you have just simulated a function call to the trigger, where the trigger will return back to the interrupted spot in the main code the same way a function call would.

"Gotchas" to make sure to deal with when doing this:

- Make sure that the trigger's body code built by the compiler ends with a `return` instruction, which currently it doesn't.
- The fact that triggers return nothing and take no args simplifies things as it means we need not worry about that part of dealing with the stack, but we still need to ensure that interruptions where a trigger jumps in right in the midst of mainline code trying to push some args onto the stack for a function it's trying to call don't break things (it shouldn't, but test it).
- Two triggers can both get triggered in the same physics update.
    - It should deal with this just fine because the algorithm above will effectively chain the two triggers together as if one had called the other in a nested subroutine call.  When the inner one returns, the outer one will continue, then return, and now you're back to the main body of code having just run both triggers like would have normally happened.
- A trigger's condition is likely to still be true (i.e. `altitude > 70000`) while the trigger body is executing.  Ensure that if that trigger body runs into an IPU boundary, or yields deliberately with a `wait`, that the very same trigger won't re-start on the next physics update.
    - I plan to deal with this by disabling a trigger while it is running and then re-enabling it only at the bottom when it's done.  Right now the code that removes a trigger from consideration when you choose not to ``preserve`` it happens at the bottom.  The new edit would remove it right at the top, first thing, and then re-insert it at the bottom if you had said to ``preserve``.  Thus once a trigger starts, it won't re-trigger until it's done.
        - This effectively prevents recursion being possible in triggers, but since triggers are not meant to be directly called upon request, that's a good thing.
- A trigger could interrupt *another* trigger.
    - Actually, I don't see this as a problem, provided recursion is prevented by the technique mentioned above.  It's okay for trigger 3 to interrupt trigger 2 that interrupted trigger 1.  What's not okay is for trigger 1 to interrupt trigger 2 that interrupted trigger 1 in the first place, which the above technique should prevent.
- It's possible for frequently firing triggers to starve mainline code of any cpu time!!
    - Not likely - perhaps *a lot* but not *all* of the CPU time,  yes.  But unless the trigger code lasts *exactly* a number of instructions equal to IPU *every* time, there will always be some leftover for the main code to execute within that last update that the trigger was doing.
        - If this is considered a problem, it's actually fixable by adding a slight random fluctuation to the IPU (so with a setting of 200 it actually varies between, say, 199, 200, or 201 instructions depending), just to enforce that things can't land on the *exact* same boundary every time.
    - Remember that even if this isn't solved, it's talking about a scenario that already exists.  A trigger *today* could hypothetically last exactly long enough for the CPU to allow it under the IPU limit, but still have no instructions left for the main code.
- It is still possible for `lock steering.`, and `lock throttle.` to exceed IPU limit, unless we move their expressions to this new system as well.  (run them as functions with return statemements, from the mainline code's loop.)  I'm unsure if we want to do that or just leave them as being required to be under the limit still.
- It is still possible for the Boolean expressions within the triggers to exceed IPU limit if they're super complex.  I'm tempted to just punt on that and say if anyone is wacky enough to try making it perform a super complex query in that test, then their program still breaks (so the error message still does need to exist for that reason).

Dunbaratu:
Note this is built on top of the profile branch, so it contains changes from there too in the diff.
Don't merge until after the profile branch is merged first, as that should make the diffs clearer.

_EDIT by hvacengi: to associate with other issues_
Fixes #1376 
Fixes #1522


Dunbaratu:
TO-DO:  I actually *can* re-add the ability to do trigger timings separately now that I think of it:
Like so:
If ProcessTriggers() inserted at least one trigger 'function call', then start the timer in "trigger" mode, and record the IP of the original mainline code that we're interrupting.
Then, during ContinueExecution()'s loop, keep using the "trigger" timer until such a time as the IP becomes equal to the original mainline one that ProcessTriggers() remembered.  At that point, switch it back to using the mainline timer.

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -413,76 +413,85 @@ private void PreProcessChildNodes(ParseNode node)
         private void PreProcessOnStatement(ParseNode node)
         {
             NodeStartHousekeeping(node);
+            nextBraceIsFunction = true; // triggers aren't really functions but act like it a lot.
+
```
Dunbaratu: This change was to support having the ``return`` keyword work correctly from within the trigger body.

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -413,76 +413,85 @@ private void PreProcessChildNodes(ParseNode node)
...
+            AddOpcode(new OpcodeReturn((short)0));
+            
+            nextBraceIsFunction = false;
         }
 
         private void PreProcessWhenStatement(ParseNode node)
         {
             NodeStartHousekeeping(node);
+            nextBraceIsFunction = true; // triggers aren't really functions but act like it a lot.
+
```
Dunbaratu: This change was to support having the ``return`` keyword work properly within the trigger's body.

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -413,76 +413,85 @@ private void PreProcessChildNodes(ParseNode node)
...
+            // Use the other dup'ed copy of the new value to actually do the equals
+            // comparison with the old value that's still under it on the stack:
             AddOpcode(new OpcodeCompareEqual());
-            AddOpcode(new OpcodeLogicNot());
-            Opcode branchOpcode = AddOpcode(new OpcodeBranchIfFalse());
+            OpcodeBranchIfFalse branchToBody = new OpcodeBranchIfFalse();
+            branchToBody.Distance = 3;
+            AddOpcode(branchToBody);
+            AddOpcode(new OpcodePush(true));       // wasn't triggered yet, so preserve.
+            AddOpcode(new OpcodeReturn((short)0)); // premature return because it wasn't triggered
```
Dunbaratu: The edits from lines 423 to 441 (I'm using the new (green) numbering when I say that) are all because of the change to make ``on`` triggers work on any arbitrary expression, rather than requiring just an identifier variable.  Now it has to run the expression, eval the result (thus de-referencing it if it *is* just a variable name), and use the result of THAT as the value to compare against the previous stored value.

Possible weird side effect?  If the result of the expression is a reference to a thing (i.e. a Vessel or a Body or something like that), then as long as the ``on`` trigger is still active, that thing won't orphan and get garbage collected.

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -413,76 +413,85 @@ private void PreProcessChildNodes(ParseNode node)
...
             // make flag that remembers whether to remove trigger:
             // defaults to true = removal should happen.
-            string triggerRemoveVarName = "$remove-" + triggerIdentifier;
-            PushTriggerRemoveName(triggerRemoveVarName);
-            AddOpcode(new OpcodePush(triggerRemoveVarName));
-            AddOpcode(new OpcodePush(true));
+            string triggerKeepName = "$keep-" + triggerIdentifier;
+            PushTriggerKeepName(triggerKeepName);
+            AddOpcode(new OpcodePush(triggerKeepName));
+            AddOpcode(new OpcodePush(false));
```
Dunbaratu: The logic was inverted from before.  Before it was:

   keep trigger active until the end of the body, then remove trigger unless preserved.

The new logic is:

    remove trigger right away at the top, then re-add it at the bottom by choosing to return true or false from the trigger depending on if it wants to be preserved.

Now it's the Return opcode itself that chooses to re-insert the trigger, rather than it happening in the trigger's code.  This is to force it to be atomic, else the IPU boundary could have been hit in-between the re-insertion and the return.

(This same comment applies to the WHEN statement logic further down.)

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -2487,8 +2497,9 @@ private void VisitOnStatement(ParseNode node)
 
             if (triggerObject.IsInitialized())
             {
-                AddOpcode(new OpcodePush(triggerObject.VariableNameOldValue));
-                AddOpcode(new OpcodePush(triggerObject.VariableName));
+                // Store the current value into the old value to prep for the first use of the ON trigger:
+                AddOpcode(new OpcodePush(triggerObject.OldValueIdentifier));
+                VisitNode(node.Nodes[1]); // the expression in the on statement.
```
Dunbaratu: Arrgggggggg!!! dangit.  There should be an ``OpcodeEval`` inserted here (where this comment is, just after the ``VisitNode()`` call).  I didn't catch it until re-reading my own code.  The problem being that if ``node.Nodes[1]`` is nothing more than a single identifier, what goes on the stack will be something like ``$foo`` rather than the contents of ``$foo``.  For this check, we need the contents of it, just to be safe and make sure things still work if ``foo`` stops existing.

It's a super rare case where it could matter, but it could.  because triggers need the values to be global it doesn't seem likely that ``foo`` would stop existing, but it could.

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -2094,15 +2109,38 @@ public class OpcodeWait : Opcode
...
+                endTime = -1; // reset in case this is called again in a loop.
+                DeltaInstructionPointer = 1; // okay now move on.
+            }
         }
     }
 
-    
+    // TODO: Does this Opcode even need to exist?  What's the point?  You can't execute it while you're stuck inside
+    // a wait so.... when was anyone ever expecting it to happen?  Discuss if it can just be removed.  Nothing in
+    // the compiler currently builds any code that uses it:
```
Dunbaratu: I suppose if we did have an ``endwait.`` command, you could have a trigger interrupt the main code during the wait, and then tell the main code to stop waiting, but ... *shudder* that would be super complex to explain in documentation, and of limited use.

src/kOS.Safe/Execution/SurboutineContext.cs
```diff
@@ -1,39 +1,8 @@
...
-    /// possible to identify how much of the top of the stack to erease if we want to
-    /// abort prematurely and return (everything on top of the topmost SubroutineContext
-    /// is the stuff that was added by the current routine).</p>
-    /// 
-    /// <p>It also makes it possible to implement varying arguments passed to a function
-    /// this way, because we can find out how many of the objects on top of the stack are
-    /// the parameters to the current function - When the function starts everything in
-    /// the stack above where the SubroutineContext record is will be a parameter.  (Also,
-    /// it should be possible to detect the wrong number of arguments better this way, for
-    /// the same reason).</p>
```
Dunbaratu: I deleted most of this text purely because it was old and talking about "future plans" that have in fact already been implemented.

hvacengi:
Good news, this passed my "Duna transfer" test on the first go.  I heavily leveraged triggers in that set of scripts, so I'd say it's safe to say that existing functionality hasn't broken.  I'm gonna run a few more tests on existing function, and then I'll try to write some tests to check the new functionality.

doc/source/commands/flight/cooked.rst
```diff
@@ -157,35 +157,41 @@ Like all ``LOCK`` expressions, the steering and throttle continually update on t
...
 
-Then the ``WAIT`` command won't work.  You can't make the
-system pause execution while it's trying to run the
-expression that tells it what to do 25 times a second.
-
-The entire expression that you LOCK any of the four cooked
-controls to must execute and finish quickly enough that it can
-be called 25 times a second (on typical default game settings
-that's how often the steering manager will run your expression).
+Then the ``WAIT`` command will be getting hit *every single*
```
hvacengi: Should we mention that this will not technically happen every single update, that it will actually delay the evaluation of both the trigger and main line code, and may cause the evaluation to be out of sync with the current conditions?

doc/source/general/cpu_hardware.rst
```diff
@@ -142,40 +146,134 @@ Triggers are all of the following:
...
+complex than that - see :ref:`CPU Update Loop <cpu_update_loop>`
+elsewhere on this page).
+
+Each of the steering locks behaves like a function that returns
+a value, and is re-called to get the new value for this **physics
+tick**.  Each of the ``ON`` and ``WHEN`` triggers also behave
+much like a function, with a body like this::
+
+   if (not conditional_expression)
+       return 1.  // premature quit.  preserve and try again next time.
```
hvacengi: should this be `return true.` instead?  So as not to imply that `return 2.` might not mean "premature quit"?

doc/source/general/cpu_hardware.rst
```diff
@@ -142,40 +146,134 @@ Triggers are all of the following:
...
+
+.. _wait_in_trigger:
+
+Wait in a Trigger
+~~~~~~~~~~~~~~~~~
+
+It is possible for kOS to allow a trigger that takes longer than one
+*physics tick* to execute.  It just means the rest of the program is
+stuck until the trigger is done.  Triggers can interrupt mainline code, but
+mainline code can't interrupt triggers.  Thus using a ``wait`` in a trigger,
```
hvacengi: This is the only instance where `wait` is lower case.

As a secondary note, it looks like all of the explanations use `wait dt.` as the example, but don't specifically show that `wait until condition.` will do the same thing.  While I think it *should* be clear from the generic use of the term `WAIT`, I can see that being a user question.  I expected this behavior based on the docs, but tested to be sure.

doc/source/general/cpu_hardware.rst
```diff
@@ -142,40 +146,134 @@ Triggers are all of the following:
...
 
-If you suspect that your trigger body would have ended if it was allowed to run a little longer, try setting your :attr:`Config:IPU` setting a bit higher and see if that makes the error go away.
+This is because while you are in a trigger, ALL the other triggers aren't
+being fired, and the main-line code isn't being executed.  A trigger that
+performs a long-running loop will starve the rest of the code in your
+kerboscript program from being allowed to ever run again.
 
-If it does not make the error go away, then you will need to redesign your script to not depend on running a long-lasting amount of code inside triggers.
+This is a consequence of: *Main-line code gets interrupted by triggers,
+but triggers don't get interrupted by main-line code (or other triggers).*
```
hvacengi: It might be good to highlight the fact that a long running loop in triggers will specifically disable proper steering control.  While the explanation makes it clear that steering controls are triggers, this may warrant a special note.

doc/source/language/flow.rst
```diff
@@ -87,7 +87,9 @@ Every time you read the value of the "variable", it executes the expression
 again.
 
 .. note::
-    If a ``LOCK`` expression is used with a flight control such as ``THROTTLE`` or ``STEERING``, then it will get continually evaluated in the background :ref:`each physics tick <cpu hardware>`.
+    If a ``LOCK`` expression is used with a flight control such as ``THROTTLE`` or ``STEERING``, then it will get continually evaluated in the background :ref:`each physics tick <physics tick>`.
```
hvacengi: I would change the phrasing from "continually" to "repeatedly", and given the documentation for triggers specifically says that they do not run in the "background" it might be good to find another way to phrase the last part of this line.

doc/source/language/flow.rst
```diff
@@ -384,105 +386,324 @@ Halts execution for a specified amount of time, or until a specific set of crite
...
-``WHEN`` / ``THEN``
--------------------
+All conditional statements, like ``IF``, can make use of boolean operators. The order of operations is as follows:
 
-Executes a command when a certain criteria are met. Unlike ``WAIT``, ``WHEN``
-does not halt execution. It starts a check in the background that will keep actively looking for the trigger condition while the rest of the code continues. When it triggers, the body after the ``THEN`` will execute exactly once, after which the trigger is removed unless the ``PRESERVE`` is used, in which case the trigger is not removed.
+- ``=`` ``<`` ``>`` ``<=`` ``>=`` ``<>``
+- ``AND``
+- ``OR``
+- ``NOT``
```
hvacengi: So, according to this order of operations, this:
```
not x = y and a > b or c < d
```
is equivalent to:
```
not ( ( (x = y) and (a > b) ) or (c < d) )
```
Is that right?

doc/source/language/flow.rst
```diff
@@ -384,105 +386,324 @@ Halts execution for a specified amount of time, or until a specific set of crite
...
+.. warning::
+    Do not make the body of a ``WHEN``/``THEN`` take a long time to
+    execute. If you attempt to run code that lasts too long in the body
+    of your ``WHEN``/``THEN`` statement, it will cause the main
+    line code, and all other triggers (WHEN, ON, and cooked steering
+    locks) to be stuck unable to continue until it finishes.
+
+For the reason mentioned in the warning above, you also probably
+should not make the system execute a ``WAIT`` command when inside the
+body of a WHEN/THEN statement.
```
hvacengi: Since this references the warning message, maybe it should be moved into the warning message itself?

doc/source/language/flow.rst
```diff
@@ -494,56 +715,87 @@ Avoid looping during ``ON`` code blocks if you can. If you are going to make ext
...
 
-To alter this, execute the ``PRESERVE`` command anywhere within the body of the code being executed and it tells the **kOS** computer to keep the trigger condition active. When it finishes executing the code block of the trigger, if ``PRESERVE`` has happened anywhere within that run of the block of code, it will not remove the trigger. Instead it will allow it to re-trigger, possibly as soon as the very next tick. If the ``PRESERVE`` keyword is executed again and again each time the trigger occurs, the trigger could remain active indefinitely.
+To alter this, a new ability was added in kOS 0.19.3 and above to
+have triggers simply :ref:`return a true or false value <trigger_return>`
+to determine if they wish to be preserved.
 
-The following example sets up a continuous background check to keep looking for if there's no fuel in the current stage, and if there is, then it activates the next stage, but no more often than once every half second. Once more than ``NUMSTAGES`` have happened, it allows the check to stop executing but it keeps the check alive until that happens::
+But prior to kOS 0.19.2 the only way to do it in kerboscript was
```
hvacengi: Prior to 0.19.3, since 0.19.2 includes the restriction as well.

src/kOS.Safe/Compilation/KS/Compiler.cs
Dunbaratu: Never mind, false alarm.  The act of doing an opcodestore into the remember-old-value variable will effectively do its own eval.

doc/source/commands/flight/cooked.rst
Dunbaratu: I want to emphasize how often it gets hit - so maybe just "every single time the kOS system tries to re-run the lock throttle expression, which it will *attempt to* do every update *if it can*."  I don't want to introduce the complexity of how the update really works here, as that's documented elsewhere and is its own full topic.  Saying "attempt to" and "if it can" makes it technically not a lie.

doc/source/general/cpu_hardware.rst
Dunbaratu: Yes it should use 'true'.

doc/source/general/cpu_hardware.rst
```diff
@@ -142,40 +146,134 @@ Triggers are all of the following:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Even a trigger who's condition isn't true yet still needs to execute
+a few instructions into the trigger subroutine to *discover* that its
+condition isn't true yet.  The trigger still causes a subroutine call
+once per **physics tick** just to get far enough into the routine to
+reach the conditional expression check and discover that it's not
+time to run the rest of the body yet, so it returns.  An expensive
+to calculate conditional expression can really starve the system of
+instructions becuase it's getting run every single **physics tick**.
```
Dunbaratu: spelling.  'because' not 'becuase'.

doc/source/general/cpu_hardware.rst
Dunbaratu: maybe change one example somewhere to use wait until?  I don't think it warrants a special clause, but maybe make sure both styles are represented among the examples?

doc/source/general/cpu_hardware.rst
Dunbaratu: It is explicitly mentioned up above that *waiting* in a trigger causes the cooked steering to be stuck on the same old settings until the wait is over, so such a sentence could reference that and say that long looping has the same problem.

doc/source/language/flow.rst
Dunbaratu: maybe change it to "repeatedly evaluated ~~in the background~~".

doc/source/language/flow.rst
Dunbaratu: I did not edit this section.  This is git's diffing getting massively confused by the fact that I re-ordered things and moved the when/then section to the bottom of the page.  It therefore claims, incorrectly, that I added these lines.  What I did was delete the section that came before them, and put it at the end and edit it a lot, which effectively moves these lines up as far as git can tell.  Which it shows as "added in new location, deleted in old location".

So while you're right, the error was in there already.  We can fix it here if you like, but it's outside the remit of this PR.

doc/source/language/flow.rst
Dunbaratu: meh.  okay sure.

doc/source/language/flow.rst
```diff
@@ -384,105 +386,324 @@ Halts execution for a specified amount of time, or until a specific set of crite
...
+      if count > 0
+        RETURN true. // will keep the trigger alive.
+      else
+        RETURN false. // will let the trigger die.
     }
 
-.. warning::
-    DO NOT make the body of an ``ON`` statement take a long time to execute. If you attempt to run code that lasts too long in the body of your ``ON`` statement, :ref:`it will cause an error <cpu hardware>`. For general help on how to deal with this, see the :ref:`tutorial on design patterns <designpatterns>`.
+There is an alternate, older syntax you can use to do the same thing,
+called the :ref:`preserve keyword`.  You may see it used in a lot of
```
Dunbaratu: Oops, this needs a ``<preserve>`` link inside the :ref: section's backticks.

doc/source/language/flow.rst
Dunbaratu: good point.  My brain was thinking with 'less than or equal to' logic, but that's not how I wrote it.

doc/source/commands/flight/cooked.rst
hvacengi: I like the addition of "attempt", that makes it so it doesn't contradict other portions of the documentation.

doc/source/general/cpu_hardware.rst
hvacengi: That would be sufficient.  Just so some one see that we mention both forms.

doc/source/language/flow.rst
hvacengi: I realized that after I submitted the comment.  Then decided not to withdraw it.

Error?  are you saying that that order shown and which I listed is not correct in operation?

src/kOS.Safe/Execution/CPU.cs
```diff
@@ -1121,77 +1111,96 @@ private void PostUpdateBindings()
...
+            
+            // To ensure triggers execute in the same order in which they
+            // got added (thus ensuring the system favors trying the least
+            // recently added trigger first in a more fair round-robin way),
+            // the nested function calls get built in stack order, so that
+            // they execute in normal order.  Thus the backward iteration
+            // order used in the loop below:
+            for (int index = currentContext.ActiveTriggerCount() - 1 ; index >= 0 ; --index)
             {
-                try
```
hvacengi: It looks like an exception in a trigger will now break mainline execution, is that correct?  I like that shift.

doc/source/language/flow.rst
Dunbaratu: Oh, to be honest I didn't carefully analyze what you said when I said "error".  I just assumed you wouldn't have bothered to make a comment if it was correct, so the mere fact that you commented on it at all made me assume it was an error.

hvacengi: Honestly, I just wanted to make sure it was expressed as intended.  I did not test.  The only part that I can't confirm is the `or` precedence.  Honesty, I usually use the parenthesis to make sure that **I** get the intended order right.

src/kOS.Safe/Execution/CPU.cs
Dunbaratu: I hadn't realized that it didn't beforehand.  I swear I had cases where the exception in a trigger broke main code, rather than re-firing again and again.

hvacengi: It would break execution if it screwed up the stack by not popping all the args, or pushing things that didn't get cleaned up.  So often times execution broke, but some things were simple enough that they stuck around.  I'll do a quick check to make sure we don't claim that behavior in the docs.

hvacengi:
Something must be wrong here... my only substantial comments are documentation related.  I didn't see anything with nomenclature or implementation stuff.  I guess I could run it through code maid, but I think it would find more stuff that we didn't touch here.

Should I just make the doc changes we discussed and push upstream?

src/kOS.Safe/Execution/CPU.cs
hvacengi: I can't find any existing documentation that claims trigger exceptions will or will not terminate main line code, so this shouldn't require additional edits.

Dunbaratu:
Yes, just make the doc changes.  I don't have a high opinion of automated "fixing" tools anyway.  They keep requiring that I remove information from my code.

hvacengi:
I figured you'd be happy with skipping that step.

I should have this merged tonight then.

doc/source/general/cpu_hardware.rst
hvacengi: You're right, this is only a few paragraphs latter.  No need to re-reference it.  The line numbers made it feel like it was further away.

PR #1544, opened 2016-03-15
Issue 1492 steering manager disassociation

### LINKED ISSUE #123: Refactorize the FlightControlManager class to separate the RT2 code
We should refactorize the class to separate the RemoteTech2 code from the rest of the code and place its own class in the AddOns\RemoteTech2 namespace.
This could probably be achieved by extending the FlightControlManager class and using the factories to instantiate the correct object.

### LINKED ISSUE #1492: Steering manager becomes disassociated
Originally reported here: https://www.reddit.com/r/Kos/comments/3vk2tz/splitting_a_ship_in_two_kos_cores_controlling/

When staging with multiple probe cores, the steering manager some times looses it's proper control reference.  This has something to do with how the cpu vessel is assigned/updated and which vessel is the "dominant" vessel and retains the original underlying ID value.

I have already started working on a fix for this, and will focus on it significantly more after we release 0.19.0.

### LINKED ISSUE #1587: [KSP1.1] Update steering to account for new engine parameters
In 1.1 there is a new method for spreading out thrust across thrust transforms:
```
transformMultipliers
{
    trf0 = 0.4
    trf1 = 0.4
    trf2 = 0.1
    trf3 = 0.1
} 
```
We should update the steering logic to account for that.

It also supports variable gimbal for different axes, and we should account for that.

Reference: http://forum.kerbalspaceprogram.com/index.php?/topic/135769-11-new-moduleengines-features-examples-and-discussion/

### LINKED ISSUE #1608: PIDLoop integral accumulator
I was looking [here](https://github.com/KSP-KOS/KOS/blob/c3e842cd53a11ca4e7eace167ce554055dc62381/src/kOS.Safe/Encapsulation/PIDLoop.cs#L155) at the `Update` method of the PIDLoop and I noticed that is using `ITerm` as the integral accumulator instead of `ErrorSum`. Is like that by design and I'm missing something or is just a typo?

The current implementation has a couple of problems:
* The `ResetI` method is not actually resetting the accumulator.
* When the output is saturated the accumulator `ITerm` is set to the new value of `iTerm`, basically resetting most of the accumulated error.

hvacengi:
Fixes #123 
Fixes #1492 
Fixes #1587
Fixes #1608 

This is my "progress" update for the code to fix the issue of disassociation of the steering manager from the CPU vessel.  It serves as a fairly sweeping overhaul for how the steering manager is used, so that the manager itself exists only once and is managed by a `VesselModule`.  Eventually (possibly before this PR is complete) I intend to move everything that is "vessel" specific rather than "part" specific into this module (or at least to be managed by the module).  While working on some other items, I realized I was starting to diverge into some similar changes as what I did in this branch, so I wanted to make sure that I got it posted and published so that every one can kick me to make sure it gets finished.

Features:
- [x] Implement new `VesselModule`
- [x] Add new `IFlightControlParameter` interface
- [x] Move `SteeringManager` to `Control` namespace and implement `IFlightControlParameter
- [x] Use new KSP 1.1 `ITorqueProvider` interface for calculating potential torque
- [x] Remove now abandoned methods and fields with the elimination of our own torque calculation
- [x] Depreciate old suffixes that no longer function
- [x] Document depreciated suffixes
- [x] Fixes to `PIDLoop`


hvacengi:
This should now be ready for review and merging.  I will leave the migration of other steering parameters to another PR so that these rather important fix can get merged for the next release.

hvacengi:
This is technically breaking because I depreciated 2 suffixes (SHOWTHRUSTVECTORS and SHOWRCSVECTORS)

hvacengi:
I've updated this to throw an exception when setting the `steering` value to an unsupported type.  This will cause it to break execution, instead of silently disabling steering and continuing execution.

I also fixed it so the trace log will identify the error as part of the built in steering update.  I really wanted to make it properly identify the line causing the error, but that's really complicated given how the trigger is created.

src/kOS/Module/kOSVesselModule.cs
```diff
@@ -0,0 +1,346 @@
...
+                    {
+                        if (partLookup[id].HasFlightControlParameter(key))
+                        {
+                            // Only copy the data if the previous vessel module still has the parameter.
+                            IFlightControlParameter paramDestination = flightControlParameters[key];
+                            IFlightControlParameter paramOrigin = partLookup[id].GetFlightControlParameter(key);
+                            // We only want to copy the parameters themselves once (because they are vessel dependent
+                            // not part dependent) but we still need to iterate through each part since "control"
+                            // itself is part dependent.
+                            if (partLookup[id].ID != BaseId) paramDestination.CopyFrom(paramOrigin);
```
hvacengi: We may want to change this logic to only copy "enabled" parameters, or perhaps only "enabled" parameters who's controlling part is on the new vessel.  This could be done pretty easily by moving the copy line into the if block below.  I'm open to discussion on it though.

src/kOS/Binding/FlightControlManager.cs
```diff
@@ -319,16 +344,28 @@ public bool Enabled
...
+                            {
+                                param.EnableControl(shared);
+                            }
+                            else
+                            {
+                                param.DisableControl(shared);
+                            }
+                            return;
+                        }
+                        if (RemoteTechHook.IsAvailable(control.Vessel.id))
```
tomekpiotrowski: We have code like this in couple of other places - a core kOS class depends on RT, which is an addon. If one day we decided to remove RT we'd have to edit quite a number of classes. It would be better for those classes to expose some kind of an interface that other components could plug in to. We do something like this with `UpdateHandler` - classes can subscribe to certain events.

I'm not going to ask you to change it now, we need a more general solution in the whole project. Just something to keep in mind.

hvacengi: That's one of the things on the docket as part of my 3rd party api implementation.  One of the things I'd like to do is essentially create a way for dll's to hook our factory methods.  The other long goal here is that `FlightControlManager` stops handling the events itself, and that we move all of that function to `kOSVesselModule`.

PR #1545, opened 2016-03-15
IR ADDON: Restricting access to servos not on CPUVessel

### LINKED ISSUE #1540: IR part control of other vessels.
If you store the `IRServo` structure while it is part of your vessel (haven't found a way of accessing it when it's not yet), then you can continue to control it after undocking / decoupling. despite the core and the IR part now being on separate vessels.

ZiwKerman:
Fixes #1540


hvacengi:
I'm adding the "breaking" label since it is possible that existing scripts take advantage of this bug.

hvacengi:
Looks like this should be clean with the current develop branch.  Do you want me to test it using your pre-release of infernal robotics, or wait for the full release?

ZiwKerman:
Well, I moved around some things in IR API in new release (2.0), so it will break anyway, but this PR should work with any 0.21.x version.

I'd say merge it, and I'll submit a fresh PR to develop branch once I finalize some remaining fields in IR 2.0

hvacengi:
I'll try testing in my v1.0.5 archive before merging the current develop into it.  I'd prefer to test it with 1.1 since that's what all future releases will be, but it should be fine so long as we do a lot of testing after you update the API.

ZiwKerman:
It will not work fully with current 1.1 beta version of IR for sure. Some fields in API have been moved around (like isAxisInverted), so when you try calling them it will cause exceptions. But if you merge it, i can grab a snapshot from develop and have a PR for compatibility with latest IR beta ready in couple hours.

hvacengi:
No need to rush on the API on my accord.  I may not be able to focus on this for a few hours anyways, and it isn't critical.

PR #1549, opened 2016-03-16
Enhancement 1546; Add a way to get the ETA of SOI for any orbit

### LINKED ISSUE #1546: Add an eta:transition like variable which takes manouvre-nodes in account
When I set a manouvre node which will put the vessel on an escape orbit, I am not able to get the ETA of the transition via kOS (while able seeing it on the map overview). Perhaps we can add an ETA:transition which takes this in account? Another option would perhaps be to add this information to the Orbit structure, besides nextpatch something like nextpatchETA.

I'm currently familiarizing with kOS codebase and perhaps can help implementing it if someone tells what is the preferred API point.

riemervdzee:
Adds functionality suggested in #1546

I am unsure whether it should return an infinite value when a nextpatch does not exist however.

_Edit: fixes #1546 and fixes #550_

### DIFF
--- a/src/kOS/Suffixed/OrbitInfo.cs
+++ b/src/kOS/Suffixed/OrbitInfo.cs
@@ -53,6 +53,7 @@ namespace kOS.Suffixed
             AddSuffix("VELOCITY", new Suffix<OrbitableVelocity>(() => GetVelocityAtUT( new TimeSpan(Planetarium.GetUniversalTime() ) )));
             AddSuffix("NEXTPATCH", new Suffix<OrbitInfo>(GetNextPatch));
             AddSuffix("HASNEXTPATCH", new Suffix<BooleanValue>(GetHasNextPatch));
+            AddSuffix("NEXTPATCHETA", new Suffix<ScalarValue>(() =>(GetNextPatchETA())));
 
             //TODO: Determine if these vectors are different than POSITION and VELOCITY
             AddSuffix("VSTATEVECTOR", new Suffix<Vector>(() => new Vector(orbit.vel)));
@@ -105,6 +106,15 @@ namespace kOS.Suffixed
         private OrbitInfo GetNextPatch()
         {
             return ! GetHasNextPatch() ? null : new OrbitInfo(orbit.nextPatch,Shared);
+        }
+
+        /// <summary>
+        /// Returns the ETA of when the nextpatch will happen
+        /// </summary>
+        /// <returns>A double representing the ETA in seconds, or a zero if there isn't any.</returns>
+        private double GetNextPatchETA()
+        {
+            return GetHasNextPatch() ? (orbit.EndUT - Planetarium.GetUniversalTime()) : 0;
         }
 
         /// <summary>


src/kOS/Suffixed/OrbitInfo.cs
```diff
@@ -107,6 +108,12 @@ private OrbitInfo GetNextPatch()
             return ! GetHasNextPatch() ? null : new OrbitInfo(orbit.nextPatch,Shared);
         }
 
+
+        private double GetNextPatchETA()
+        {
+            return  ! GetHasNextPatch() ? Double.PositiveInfinity : orbit.EndUT - Planetarium.GetUniversalTime();
```
hvacengi: Since any value of infinity is an invalid in kerboscript and will throw an exception, it would be better for this to default to either zero or -1.  I think I would prefer it to be zero.  And while it appears the syntax was copied from above, I think it's cleaner if remove the `!` conditional and reverse the `? ... : ...` parameters.

hvacengi:
Edited your description to auto associate with the issue.

hvacengi:
Could you please push an update addressing the comment above.  Are you able to perform the appropriate documentation updates yourself, or will that need to be done at the time of merging?

Thanks for submitting this addition!

riemervdzee:
I pushed a new revision of the commit.
As for the documentation additions, do you mean the remotes/origin/gh-pages branch? or the KSP-KOS/KOS_DOC repo

hvacengi:
Just make the modifications in your existing branch, in the `doc/source` folder.  We generate the docs based on rst files using Sphinx.  When we publish the next version we'll generate the html from that source and push to the gh-pages branch and KOS_DOC repository (but you don't need to worry about that)

src/kOS/Suffixed/OrbitInfo.cs
Dunbaratu: I'd prefer a negative number.  a negative number *CANNOT* be a valid ETA time, while a zero could be valid, for just an instant, hypothetically.

riemervdzee:
I'll work on it this weekend

src/kOS/Suffixed/OrbitInfo.cs
riemervdzee: On one side I agree with Dunbaratu, you basically want to return something which is outside the "domain" of accepted ETA values. On the other hand, -1 is not a good indication something is wrong straight away. It could screw up calculations further up in the chain and would require a debug session.

So I am not sure of what is best. Is there perhaps a way to throw an exception?

hvacengi: I guess the question would be what happens if you keep the handle to the transition in memory, and then check `EndUT` after the transition.  That would be the only way that zero and negative values may be valid.  In fact, I'll want to test that anyways to see if KSP spouts out some silly value.  I'll report me results.

You can indeed throw an exception.  You can use `kOS.Safe.Exceptions.KOSSituationallyInvalidException`.  Just make sure to be clear in the text about the cause.  This will have the same effect as returning infinity, but with an error message that the user can actually use to correct their behavior.  I'm OK with throwing an exception because there is a way to detect the "error" case other than by the exception itself (via `HASNEXTPATCH`).

hvacengi:
If you prefer, I can handle the docs in my review commit.  I just figured I'd let you describe it in your own words.  Let us know if you need any help figuring out the doc system.

hvacengi:
I think I'm just going to write the docs for this when I review for merging.  I don't really want this one hanging out there any longer if I can help it.

riemervdzee:
I am sorry I could not report this earlier. I got stuck on testing the documentation edits on my local-machine and live got in the way.
Feel free to make any changes you see fit and sorry I cannot do these myself.

hvacengi:
Don't worry about it. It shouldn't be a problem for me to get it written.

Dunbaratu:
Did the pending documentation for this ever happen?

hvacengi:
Yes, documentation happened in my review commit 816b07869e5946d377ca8ce56f82b6aad47644f5

PR #1551, opened 2016-03-17
Fixes #1550 anonymous functions

### LINKED ISSUE #1550: Allow anonymous functions for delegates.
Let's say you have a user-defined sort function that takes a comparator delegate argument like so:
```
function my_compare { parameter p1,p2. return p1 > p2. }

mysort( mylist, my_compare@ ).
```
We should be able to support making that function in-line in the mysort call with some sort of syntax like so:
```
mysort( mylist, {parameter p1,p2. return p1 > p2.} ).
```

This would basically push a `UserDelegate` object onto the stack where that syntax is put.  This also means you can make a function that you store into a delegate right away without it having a "real" name, like so:

```
set my_delegate to {
  parameter p1, p2.
  return p1 > p2.
}.
```

I have an idea how I'd do it.  Just making the issue first so the PR will attach to an issue.

Dunbaratu:
Working, and it was really easy.

```
set foo to {parameter p. return "Hi, " + p.}.
print foo("Steve").
Hi, Steve
```
The basic technique is:
1 -  In the grammar file:  The topmost level of expression is now: instruction block or ordinary expression.
2 - If the expression is ordinary, descend into its children and act as before.
3 - If the expression is actually instruction block, then:
3.1 - Insert a branch opcode that will skip around the stuff in (3.2) and land on the stuff in (3.3)
3.2 - Compile the instruction block right here, in the context that it's a function body (so parameter and return commands work and are legal).
3.3 - Insert an opcode to push a UserDelegate of the code built in (3.2) above.

Thus the anonymous function's UserDelegate is now the thing atop the stack as the result of the expression.

To whomever reviews and merges this:

- As usual, with any change that involves altering the syntax rules, the easiest way to understand the change is to start from looking at what changed in the ``kRISC.tpg`` file first, even though the github diff tends to show it last.  Everything else that had to be changed in the other source files is a follow-up effect from the changes to that file.
- And of course, Scanner.cs, ParseTree.cs, and Parser.cs are autogenerated files from TinyPg, so reviewing them is pointless.


src/kOS.Safe/Compilation/KS/kRISC.tpg
```diff
@@ -95,8 +95,9 @@ COMMENTLINE  -> @"//[^\n]*\n?";
 // Rules
 // ===================================================
 Start -> (instruction)* EOF;
-instruction_block -> CURLYOPEN instruction* CURLYCLOSE EOI?;
-instruction -> set_stmt | 
+instruction_block -> CURLYOPEN instruction* CURLYCLOSE;
+instruction -> empty_stmt |
+               set_stmt |
```
Dunbaratu: These changes were because this syntax didn't work before:
```
set foo to {return 0.}.
```
and instead required two dots, which looked kinda dumb:
```
set foo to {return 0.}..
```
It required 2 dots because the old version of this rule insisted on taking the first dot as being part of the `instruction_block` instead of as the statement terminator, thus leaving the statement unterminated.
This change makes it cleaner by saying that the optional dot after the braces isn't in fact part of the instruction block at all, but is just an empty statement, which is now legal.

src/kOS.Safe/Compilation/KS/kRISC.tpg
```diff
@@ -137,8 +138,9 @@ directive            -> lazyglobal_directive; // Add to this list later if we ma
 
 // ------------ statements --------------------
 
+empty_stmt     -> EOI;
 set_stmt       -> SET varidentifier TO expr EOI;
-if_stmt        -> IF expr instruction (ELSE instruction)? EOI?;
+if_stmt        -> IF expr instruction EOI? (ELSE instruction EOI?)?;
```
Dunbaratu: To support the old backward compatibility where this was legal:
```
if x = 0 {
  print "true".
}. else {
  print "false".
}.
```
in which there was a dot *between* the if and the else.  This slight change has a minor effect on Compiler.cs in the VisitIfStatement() method.

src/kOS.Safe/Compilation/KS/kRISC.tpg
```diff
@@ -199,7 +201,8 @@ unset_stmt     -> UNSET (IDENTIFIER | ALL) EOI;
 // ---------- expressions ---------------------------
 
 arglist            -> expr (COMMA expr)*;
-expr               -> and_expr (OR and_expr)*;
+expr               -> (or_expr|instruction_block);
+or_expr            -> and_expr (OR and_expr)*;
```
Dunbaratu: Top level expression is now allowed to also be an instruction block with braces (thus the anonymous function syntax).

tomekpiotrowski:
This is great! And the PR is surprisingly short ;)

Do you think in the future it would be possible to shorten the syntax even more? The `parameter` and `return` keywords elongate the whole thing. Ultimately could we have something shorter like `{ |x| return x*x }` or even { |x| x*x }?

Dunbaratu:
The way I'd prefer to shorten it would be to shorten the way you do it for named functions too, rather than making the shortening be something that's unique to anonymous functions only.  (That's one of the beefs I have with C# syntax - the way to list parameters for an anonymous function is a totally different system altogether from how you do it for a named one, and there are cases where if you do use a semicolon in an anonymous function it's a syntax error where in a named one it would have been the other way around and been an error to NOT use it.  I don't see a good reason for that confusing inconsistency.)

doc/source/language/anonymous.rst
```diff
@@ -0,0 +1,142 @@
...
+Overview
+--------
+
+There is a certain kind of :ref:`Delegate <delegates>` in which you
+don't need to actually name the function in question, and can just
+assign it right into a variable *as* a delegate to begin with, or
+pass it *as* a delegate to another function call.
+
+This is referred to as an
+`Anonymous Funcion <https://en.wikipedia.org/wiki/Anonymous_function>`__.
```
Dunbaratu: @hvacengi Ooops.  Spelling.  "Funcion".  Since you already may be reviewing I don't want to push late changes now for something this minor.  Feel free to change it if you wish.

hvacengi: I'll grab it in my review.

hvacengi:
@Dunbaratu could you merge the current develop branch back in to the PR?  It looks like it should be pretty straight forward to merge but I think I'd feel more comfortable with you looking at the conflicts.

Dunbaratu:
I'll get to it tonight (bringing it up to date with current develop).

Dunbaratu:
Well the conflict was purely because we updated TinyPG and the conflicts are within the TinyPG-generated files.  Those werejust a matter of re-running TinyPG.

PR #1555, opened 2016-03-19
UniqueSet data structure

### LINKED ISSUE #1481: Add a `Set` structure
This is something that's missing from our collections. I think it should be easily doable.

tomekpiotrowski:
Fixes #1481


tomekpiotrowski:
Renamed and rebased on top of 1.1

PR #1556, opened 2016-03-20
Fixes serialization of boolean type

tomekpiotrowski:
I've extended my tests a bit and they revealed that boolean values currently don't serialize properly - they serialize as strings. Somehow `ISerializableValue` interface was not implemented.


tomekpiotrowski:
This would be deprecated by #1449

hvacengi:
Since we will be merging the inter-cpu PR for sure, should we just close this PR?

PR #1560, opened 2016-03-24
Add the ability to launch a new vessel from script

### LINKED ISSUE #1149: Provide the ability to launch a craft file via script
For the purpose of autonomous station building, it would be nice to be able to have a processor trigger launch a new vessel.  I think the KSP api looks pretty reasonable for doing this, but Kerbal Construction Time might be a reference if needed.

hvacengi:
Fixes #1149

This adds suffixes to the `KUNIVERSE` bound variable to allow for launching a new vessel from a craft file.  It also adds a new structure `CraftTemplate` which contains data about the craft file.  This information is accessed via KSP's `ShipTemplate` class rather than the `ShipConstruct` class because the template does not attempt to actually load each of the parts, it just reads the basic information off of the craft file itself.

Features:
- [x] New CraftTemplate structure
- [x] Get name, description, mass, cost, and part count of the template
- [x] Launch the craft from it's default launch site
- [x] Launch the craft from an arbitrary launch site
- [x] Launch with the craft's default crew (as would be selected automatically in the editor)
- [ ] Allow for assignable crew _(Optional for initial implementation, easy to add later)_
- [x] Documentation


hvacengi:
This should be ready, unless we really want to support assignable crew from the very beginning.

jonored:
Just wanted to note that I've been using this (I play a fair amount on a remote machine almost entirely via telnet, and this was a feature that I would have added locally if it hadn't been already implemented), and I seem to be observing some odd behavior: when I use this to launch a vessel using a kOS module attached to a manned station, the crew on the station disappear and reappear back at KSC. Not asking for a fix or anything, I'm just noting the test case and that it probably would be good to check that this behavior doesn't happen if/when this does get pulled into develop.

hvacengi:
@jonored could you give me a little more information about the crew behavior?  By "reappear back at KSC" I assume you mean that they are listed as available in the astronaut building?  If you don't assign them to a new ship and "fly" the station again, do they still show up as part of the crew?  Were these crew members on the station before the "flight" on which you called the new kOS function (did they fly to the station and then call the kOS function in the same flight, or were they on the station from a previous flight before calling the function)?

jonored:
@hvacengi - Apologies, I think this may have been an oops on my side - I didn't check properly with /just/ my local build of kOS and I think this probably was a different mod (USI-LS) changing behavior. and I didn't mean to hold this up as "not ready".

The behavior I see is that I switch to the station (with a secondary menu I've added to the telnet screen), call kuniverse::launchcraft, and then when I switch back to the station there is nobody there. I probably just didn't notice the changed behavior (after I caught that there had been a change I was going to try it with /just/ my build and make sure that it didn't happen there and remove my comment entirely, I just didn't get to it.)

hvacengi:
Oh, until we're sure it's not an issue on our end, I want the Not-ready tag to prevent some one from merging it (even though it shouldn't be merged until we have KSP1.1 compatibility merged, but it's still a possibility).

I'll try to test this tonight.

jonored:
Fair. I'll try to get a check with a sane configuration done on my end tonight and see about making sure that it was just a mistake on my part.

hvacengi:
It actually gives me a good excuse to test my entirely automated station building script, so I'm glad you found this and reported it.  I wanted to test it before merging anyways.

jonored:
Hah - that's pretty much exactly what I was up to when I hit it. Incidentally, until there's a proper KSC mainframe implemented, do you think a "focus vessel from telnet" sub-menu would be worth polishing up and sending in a pull request for? I ran into this because I've been basically using my main station (which I can just switch to because of the menu) as a substitute for the mainframe when I'm running remote & no-GUI.

hvacengi:
Actually, I would really like that option.  One of the things I wanted to do "long term" was make the telnet menu have the option of listing "loaded ships", "ships in flight" and "craft files to launch".  If you're comfortable enough messing with the telnet code, feel free to make a new issue and work on implementing it.

Dunbaratu:
If you want to mess with this, the class to look at is ``TelnetWelcomeMenu.cs``
Keep in mind that it is operating in the world of very primitive "un-cooked" bytes right from the socket, and therefore it has to do its input and output in very raw ASCII terms (i.e. it needs to use `\r\n` for end of lines, it can't assume the user has the ability to edit input lines, etc.)  That's why the only user interface it has at the moment is the rather crude "type some digits and hit Enter".

Dunbaratu:
Basically, when a ``TelnetSingletonServer`` notices that it isn't attached to a specific CPU terminal at the moment, then it attaches itself to a new instance of ``TelnetWelcomeMenu`` instead, until a CPU terminal is picked, and then it kills ``TelnetWelcomeMenu`` and attaches to a CPU terminal instead.  When you detach with control-D, the ``TelnetSingletonServer`` makes a new instance of ``TelnetWelcomeMenu`` again and attaches to that.  So what you need to know is that there exists one instance of ``TelnetWelcomeMenu`` per "pick".  Once a valid "pick" is made, ``TelnetWelcomeMenu`` is gone, and a new one will be constructed if the user comes back to the menu again.

jonored:
@Dunbaratu - I know - I'll just be cleaning up a change to those classes I already have working locally to support my play style, and fixing niceties like "to get back to the main menu you have to restart the telnet session", and "you can change focus to unowned objects". Ultimately, this seems redundant if a mainframe happens, but it gets you most of the way to the mainframe's basic usefulness if you've got a usable module /somewhere/ to swap to.

hvacengi:
I have now tested in KSP 1.1 and can confirm that the crew bug reported above does not appear to be an issue with kOS.  I was unable to replicate it.

I updated the logic to respect some things from the game settings (regarding stock craft) and cleaned up the documentation.

@Dunbaratu or @tomekpiotrowski this should be ready to merge if you don't see any issues in testing.

doc/source/structures/misc/kuniverse.rst
```diff
@@ -69,6 +69,22 @@ KUniverse 4th wall methods
...
+          - Method
+          - Get the file path for the craft with the given name, saved in the given editor.
+        * - :meth:`LAUNCHCRAFT(template)`
+          - none
+          - Method
+          - Launch a new instance of the given craft at it's default launch site.
+        * - :meth:`LAUNCHCRAFTFROM(template, site)`
+          - none
+          - Method
+          - Launch a new instance of the given craft at it's default launch site.
```
tomekpiotrowski: By the look of it this method does not launch from the default site.

doc/source/structures/misc/kuniverse.rst
```diff
@@ -192,6 +208,54 @@ KUniverse 4th wall methods
...
+    crew.
+
+    **NOTE:** Due to how KSP handles launching a new craft, this will end the
+    current program even if the currently active vessel is located within
+    physics range of the launch site.
+
+.. method:: KUniverse:LAUNCHCRAFTFROM(template, site)
+
+    :parameter template: :struct:`CraftTemplate` craft template object.
+    :parameter site: :struct:`String` editor name.
```
tomekpiotrowski: I assume you mean 'site name' here.

doc/source/structures/misc/kuniverse.rst
```diff
@@ -192,6 +208,54 @@ KUniverse 4th wall methods
...
+
+    Returns the :struct:`CraftTemplate` matching the given craft name saved from
+    the given editor.  Valid values for site include ``"RUNWAY"`` and
+    ``"LAUNCHPAD"``.
```
tomekpiotrowski: The first sentence needs updating here. Looks like it was copy-pasted from `GETCRAFT`.

doc/source/structures/vessels/crafttemplate.rst
```diff
@@ -0,0 +1,105 @@
...
+          - The default launch site for this craft
+        * - :attr:`MASS`
+          - :struct:`Scalar`
+          - The default mass of the craft
+        * - :attr:`COST`
+          - :struct:`Scalar`
+          - The default cost of the craft
+        * - :attr:`PARTCOUNT`
+          - :struct:`Scalar`
+          - The total number of parts n this craft.
```
tomekpiotrowski: 'n'

doc/source/structures/vessels/crafttemplate.rst
```diff
@@ -0,0 +1,105 @@
...
+
+
+.. attribute:: CraftTemplate:NAME
+
+    :access: Get only
+    :type: :struct:`String`
+
+    Returns the name of the craft.  It may differ from the file name.
```
tomekpiotrowski: extra space

doc/source/structures/vessels/crafttemplate.rst
```diff
@@ -0,0 +1,105 @@
...
+
+    Returns the description field of the craft, which may be edited from the
+    drop down window below the craft name in the editor.
+
+.. attribute:: CraftTemplate:EDITOR
+
+    :access: Get only
+    :type: :struct:`String`
+
+    Name of the editor from which the craft file was saved.  Valid values are
```
tomekpiotrowski: extra space

doc/source/structures/vessels/crafttemplate.rst
```diff
@@ -0,0 +1,105 @@
...
+    ``"VAB"`` and ``"SPH"``.
+
+.. attribute:: CraftTemplate:LAUNCHSITE
+
+    :access: Get only
+    :type: :struct:`String`
+
+    Returns the name of the default launch site of the craft.  Valid values are
```
tomekpiotrowski: extra space

doc/source/structures/vessels/crafttemplate.rst
```diff
@@ -0,0 +1,105 @@
...
+    ``"LAUNCHPAD"`` and ``"RUNWAY"``.
+
+.. attribute:: CraftTemplate:MASS
+
+    :access: Get only
+    :type: :struct:`Scalar`
+
+    Returns the total default mass of the craft.  This includes the dry mass and the
```
tomekpiotrowski: extra space

doc/source/structures/vessels/crafttemplate.rst
```diff
@@ -0,0 +1,105 @@
...
+    mass of any resources loaded onto the craft by default.
+
+.. attribute:: CraftTemplate:COST
+
+    :access: Get only
+    :type: :struct:`Scalar`
+
+    Returns the total default cost of the craft.  This includes the cost of the
```
tomekpiotrowski: extra space

src/kOS/Suffixed/CraftTemplate.cs
```diff
@@ -0,0 +1,141 @@
...
+            InitializeSuffixes();
+        }
+
+        public CraftTemplate(string filePath, ShipTemplate temp)
+        {
+            path = filePath;
+            template = temp;
+            if (template == null)
+                throw new KOSException("Failed to load template from given path:\n  " + filePath);
+            InitializeSuffixes();
```
tomekpiotrowski: It looks like we could simplify this constructor by making it call the first constructor and just overwrite the `template` field.

hvacengi: Do you mean:
```
        public CraftTemplate(string filePath, ShipTemplate temp) : this(filePath)
```
or
```
        public CraftTemplate(string filePath) : this(filePath, ShipConstruction.LoadTemplate(filePath))
```
I'd prefer to avoid the extra disk reading of the first option since we already have the template and would just discard the read information.

I suppose the 2nd option wouldn't be too bad.  The method would be clear enough that it can probably be called without hurting code readability.  The only concern I'd have is that we end up finding an exception thrown by the `LoadTemplate` method that we don't know about.  But we can revert the code back to this state if we need to add a `try/catch` block.

tomekpiotrowski: Right, the 2nd option looks better.

hvacengi: I've updated it (though these comments aren't hidden because the associated line itself wasn't modified).

doc/source/structures/vessels/crafttemplate.rst
hvacengi: This brought me to some interesting reading.  I assume that you are referring to the two spaces between sentences.  That was how I was originally taught to type over 20 years ago.  In various articles on the topic, it now appears that the two space rule has fallen out of favor in some cases.  It was apparently originally tied to the use of monospaced fonts on typewriters, as extra space was needed to make sentence separation more recognizable with the extra white space in the words.  Today many style guides point to single spaces after sentences, though no one can point to firm reason or study why other than the majority of type setters thinking it looks better.  Of note however is the APA style guide which returned to a two space recommendation in their 6th edition.

But none of that is really the issue here.  Honestly, I don't know that I can break my habit of double spaces as I'm typing at this point.  I know I did it anywhere that I had multiple sentences, and I don't think it's worth trying to go back and correct every instance of it in the doc source.  The rendering engine will fix it for us, and we can't do a simple "replace all" on the source since spaces are used for indentation as well.

If you'd like to remove the extra spaces before you merge, that's fine by me.

doc/source/structures/vessels/crafttemplate.rst
Dunbaratu: No, correct amount of space.
Please don't change these.  Even though the HTML will annoyingly strip out the proper spacing, it's still useful to have it here in the monospaced ASCII source ``*.rst`` file where we have to look at it when editing things.

tomekpiotrowski: It looks very strange to me, I've been working around computers for 20 years and I've never seen that to be honest :P It is only after few minutes of googling that I've found out it's mostly a US thing and a relict of the typewriter age.

Anyway, it's not that that important to me. Feel free to leave it as it is.

Dunbaratu: I find the notion of it being a relic of typewriters odd.  Because I find it even MORE important when using variable-width fonts than monospaced ones.  (Thus why I'm annoyed that HTML rendering prevents it.)  Variable-width fonts have narrower period characters than monospaced ones, making the end of the sentence stand out even less than it does in old-school typewriters.

Dunbaratu: A lot of US writing standards are kinda dumb (Writing dates in middle-endian order, putting the comma inside the quotation even when it's not part of what was quoted, etc) but this one makes a lot of sense.  It makes the break between words smaller than the break between sentences.

hvacengi: It appears to be a point of major debate for people who really care about typography.  By far the number of articles saying "If you're using 2 spaces between sentences you're doing it wrong" out numbered those supporting 2 spaces.

PR #1563, opened 2016-03-25
Issue 1553 scalar overflow

### LINKED ISSUE #1553: Integer overflow
in kOS 0.19.2 I'm getting the following multiplication overflow that did not happen in 0.18.x: 
```
print 60000 * 60000.
-694967296
print 60000.0 * 60000.0.
-694967296
print 60000 * 60000.1.
3600006000
```
Obviously kOS decides to treat both values as integers and does not check if the result fits in the type. By changing one value I can force it to do the whole calculation with float values, and get the correct result.

hvacengi:
Fixes #1553
Closes #1523

This includes the pull review for #1523, as well as optimizations for execution.  As part of the review, I updated the ScalarCalculator to fall back to object equality if an operator cannot be found (allowing all structures to compare the way they did previously).  The old `TryOperation` system is removed.

Optimizing scalar performance is really the big issue in this PR.  First of all, I eliminated doing integer math, meaning that everything performs like a floating point number.  Testing shows it isn't a significant performance hit.  More importantly it means that we can eliminate integer overflow situations, since the integer will always be converted to a double before doing the math, and return values that are outside of the limits of an integer will be kept as doubles.

The first big performance boosts come from eliminating an extra constructor when you attempt to create a clone of a `ScalarValue`.  While our `ScalarValue` is a reference type, none of the properties can be changed so we can effectively treat it like a value type so there isn't a need for another call to the constructor.

The second performance boost comes from allowing structures to lazy load suffixes.  Because `ScalarValue` is usually treated as a value type, they may frequently be created and manipulated without ever calling a suffix.  As a result, you can save a lot of execution time in the constructor by skipping adding the suffixes.  Then the first time that a suffix is called, the suffixes are added.  The execution cost for managing the suffixes is absorbed the first time you access a suffix instead of right away.

In the future, it would be nice to modify the system to allow suffixes to be declared statically, or maybe using attributes, in a way that allows you to pay for the suffix initialization exactly once, preferably during the KSP loading screen.

Features:
- [x] Include PR #1523
- [x] Optimize equality operations
- [x] Remove depreciated `TryOperation`
- [x] Eliminate integer overflow issues
- [x] Optimize scalar math operations
- [x] No documentation necessary


src/kOS.Safe/Encapsulation/ScalarValue.cs
```diff
@@ -67,12 +72,6 @@ public static ScalarValue Create(object value)
                 return new ScalarIntValue((int)value);
             }
 
-            var scalarValue = value as ScalarValue;
-            if (scalarValue != null)
-            {
-                return Create(scalarValue.Value);
-            }
-
```
Dunbaratu: I wonder how much savings there was just from moving this alone (previously it wasted time trying all the other types before finding out that no converting needed to be done after all).

hvacengi: I actually tested it both ways, and it was a benefit but not a huge benefit.  I figured even a small gain would be worth it.

src/kOS.Safe/Encapsulation/Structure.cs
```diff
@@ -26,7 +29,26 @@ static Structure()
...
+
+        private void callInitializeSuffixes()
+        {
+            if (needToInitializeSuffixes)
+            {
+                foreach (var callback in initializeSuffixCallbacks)
+                {
+                    callback.Invoke();
+                }
+            }
```
Dunbaratu: I like the system you're setting up here.  As long as we still also have an escape route to do it per-instance for those few suffixes that need it (like the resources in a fuel tank).

hvacengi: Yeah, that will still work.  The lazy-load will just happen after the overridden `GetSuffix` method calls `base.GetSuffix` The only way this lazy load would fail is if the base method isn't called.

src/kOS/Suffixed/Direction.cs
```diff
@@ -160,6 +160,16 @@ private void DirectionInitializeSuffixes()
...
+        public static Vector operator +(Direction a, Vector b)
+        {
+            return new Vector(a.Rotation * (Vector3d)b);
+        }
+
+        public static Vector operator +(Vector b, Direction a)
+        {
+            return new Vector(a.Rotation * (Vector3d)b);
+        }
+
```
Dunbaratu: I don't think I like this.  It's just turning the ``+`` operator into a synonym for the ``*`` operator, as far as I can tell, and that's .... sort of mathematically wrong.  You can multiply two matrices of different dimensions together, but you can't add them together.  I mean, we *can* do this.  I really don't think that we *should*.

src/kOS/Suffixed/Timespan.cs
```diff
@@ -94,6 +94,7 @@ public double ToUnixStyleTime()
         public static TimeSpan operator -(TimeSpan a, double b) { return new TimeSpan(a.ToUnixStyleTime() - b); }
         public static TimeSpan operator *(TimeSpan a, double b) { return new TimeSpan(a.ToUnixStyleTime() * b); }
         public static TimeSpan operator /(TimeSpan a, double b) { return new TimeSpan(a.ToUnixStyleTime() / b); }
+        public static TimeSpan operator +(double b, TimeSpan a) { return new TimeSpan(b + a.ToUnixStyleTime()); }
```
Dunbaratu: Really, we didn't have this before?  That seems like an odd omission.

src/kOS/Suffixed/Timespan.cs
```diff
@@ -128,48 +129,31 @@ public double ToUnixStyleTime()
...
-               if (other is double && op == "/") return (double)other / this; // Can't imagine why the heck you'd want to do this but here it is
-               if (other is TimeSpan && op == ">") return this < (TimeSpan)other;
-               if (other is TimeSpan && op == "<") return this > (TimeSpan)other;
-               if (other is TimeSpan && op == ">=") return (TimeSpan)other >= this;
-               if (other is TimeSpan && op == "<=") return (TimeSpan)other <= this;
-               if (other is double && op == ">") return (double)other > this;
-               if (other is double && op == "<") return (double)other < this;
-               if (other is double && op == ">=") return (double)other >= this;
-               if (other is double && op == "<=") return (double)other <= this;
+            return false;
```
Dunbaratu: Should we allow equality with a double, using ``ToUnixStyleTime()`` ?

src/kOS/Suffixed/Direction.cs
hvacengi: This is mostly for backwards compatibility.  Technically I should have supported _every_ operator between vectors and directions because the old code basically read "if other value is a vector, regardless of operator, apply the rotation to the vector".  I noticed there are a couple of random scripts out there that use the `+` operator.  I don't mind deleting this, but we'll need to add the breaking tag.

src/kOS/Suffixed/Timespan.cs
hvacengi: Well it supported the opposite order.  So `print 5 + time.` would display: "5TIME(...)" instead of adding 5s to the current time.  Just one of those fun order of operations issues.

src/kOS/Suffixed/Direction.cs
Dunbaratu: Really, this used to work?   Sorry I thought you were adding it because it's new (green) code, not noticing that you're just mimicing previous functionality that worked by a different mechanism.

Sigh.  Then we have to support it even though it's quite wrong.

hvacengi: It used to work from withing `TryOperation` which I deleted.

src/kOS/Suffixed/Timespan.cs
hvacengi: Actually, yeah, that would be a good idea.

src/kOS/Suffixed/Timespan.cs
Dunbaratu: Well good catch then,I suppose.
I sort of wish that operator overloading had a nice syntax that let you specify when order doesn't matter, so you didn't have to explicitly mention it both ways around.  Something like a `commutative` keyword.

src/kOS/Suffixed/Timespan.cs
Dunbaratu: Okay, after some testing, I concluded that it wouldn't work anyway.  The double timestamp is so precise that even if you print the ``:seconds`` suffix of the timestamp, then directly type in exactly what you see on screen as a double (so it should be ==), the comparison claims the timestamp is greater than the double.  So the mere tiny rounding that happens when printing the value as a decimal is enough to make equality fail, so it would need an epsilon in the comparison and that's just getting wonky.  It's better to force people to do their own epsilon logic with greater than and less than.

PR #1566, opened 2016-03-28
The changes to the KOS project to accompany new TinyPG.

Dunbaratu:
**Once this is merged, then from then all all kOS developers who edit the ``kRISC.tpg`` file must compile it using our forked version of TinyPG at https://github.com/KSP-KOS/TinyPG, and NOT the stock version of TinyPG.**

Specifically, this takes away the case-insensitive regex directive (``(?i)``) from the patterns, instead relying on the lowercasing technique used in the new TinyPG update we made.


Dunbaratu:
I assigned this to @hvacengi to merge because it's really his work - I just made a PR for it because I had to do this anyway in order to test his other work.

After you merge this, it might be a good idea to make a slack "everybody" announcement warning them of the information in the boldface text above.

Dunbaratu:
Okay, looking at the "files changed" list it's clear something wasn't right here.  The kRISC file is missing from the list of changed files.

That's why I marked it "not ready" until I investigate.

Dunbaratu:
Okay, that's better. fixed now.

Issue #1363, opened 2016-01-10
Start on archive yields "power starved" when powered up with no KSC connection.


AliceTheGorgon:
The expected behavior is for it to actually turn on, and just require you to switch to the local volume to run anything. Instead, it just gives the very unhelpful "power starved" message, even if there's plenty of power. (Edit: This is while there *is* a Kerbal in the command pod, so I do have local control of the ship.)

Note that an already powered on computer which loses connection works fine, and can still run scripts from the local volume. It's apparently only if it runs out of power, or has the power turned off, that it stops working.

Ship tested was:
1x Mk1 command pod
1x CX-4181 Scriptable Control System
1x SP-L 1x6 Photovoltaic Panels
1x Comms DTS-M1

Mods used:
KSP version 1.0.5
kOS version 0.18.2
RemoteTech version v1.6.9
ModuleManager version 2.6.17 (pulled in by ckan as a dependency)
HyperEdit version 1.4.2.0 (for quickly getting the craft into orbit to test)

AliceTheGorgon:
Just a quick update to confirm that the issue still exists in kOS 0.20.0.
As a minor note on reproducing the issue, I did have to restart KSP after enabling "Start on archive".

The ship was the same as last time:
1x Mk1 command pod
1x CX-4181 Scriptable Control System
1x SP-L 1x6 Photovoltaic Panels
1x Comms DTS-M1

The mods used where:
KSP version 1.1.2
kOS version 0.20.0
RemoteTech version v1.6.11
ModuleManager version 2.6.24 (pulled in by ckan as a dependency)
HyperEdit version 1.5.0 (for quickly getting the craft into orbit to test)

hvacengi:
I am unable to reproduce this with the current development build and the current pre-release of remote tech.  In order to work on this, we will more than likely need a copy of your log file and possibly a save file to do further testing.  I was unable to replicate the issue with or without a kerbal in the command pod.  It also didn't happen if I used a probe core without an omni-antenna built in (I edited it out of the MM patch).

Just by reading your description and the parts, my first guess is an issue with not having a probe core on board.  If you don't have a probe, there is no RT Signal Processing Unit, which prevents it from processing signals.  Essentially, any time that you want to do something that would require a signal back to KSC you **need** to have a probe core, or another signal processing part, on the vessel.

[...]

PR #1567, opened 2016-03-28
Subdirectories

### LINKED ISSUE #754: [proposal] a way to copy a file in the same volume
```
make copy of "my_file" as "some_other_name".
```

### LINKED ISSUE #778: Malicious kOS scripts
oooh this is fun i can `log "stuff that is not a script" to "/log/koslog.txt".` (log folder needs to exist first so can only be done on the archive). hadn't thought of doing that before.

### LINKED ISSUE #969: Null ref when trying to rename a file to empty string
screenshot7
It's one of situations that throw an exception that doesn't explain anything.

### LINKED ISSUE #1136: allow file copy command to be used to copy to new filename
Right now the ``COPY`` command only copies between volumes like so:

```
    copy aaaa to 1.
    copy aaaa from 1.
```
You can't do this:
```
    copy aaaa to bbbb. // now filename aaaa and filename bbbb contain the same contents.
```

Right now the only way to do this is with the following convoluted steps, with the aid of a second volume:
```
    // copy aaaa on vol 1 to name bbbb on vol 1:
    switch to 1.
    copy aaaa to 0. // temp place to hold aaaa
    rename aaaa to bbbb.
    copy aaaa from 0.
```

### LINKED ISSUE #1364: Subdirectories
This is something I've actually been working on for a [while](https://github.com/tomekpiotrowski/KOS/commit/8dd10ee0363feaa91495846c6cc465ab5b01ac76), but realised there's no issue for it. I have the most difficult internal stuff implemented, but there's still plenty of work left to do. I hope to make some progress in the coming weeks.

It will require #1257 to be fixed first.

### LINKED ISSUE #1536: volume:open(filename) throws encapsulation error when file doesn't exist
Instead of throwing an exception saying that the file does not exist, `archive:open("filename").` will throw an error that the value could not be encapsulated.  I suspect that open is returning `null` in this instance, which will throw that error message.

tomekpiotrowski:
Fixes #1364, fixes #778, fixes #754, fixes #1136, fixes #969, fixes #1536 

I've started to update my old subdirectories code. I should have time to work on it in the next few days and I hope to make solid progress on the whole subdirectories front.

### Breaking changes:
- Boot files are now expected to be placed in `boot` directory.
- `delete filename from volume`, `copy filename from/to volume`, `rename (both forms)` have been deprecated in favor of `deletepath(path)`, `copypath(fromPath, toPath)`, `movepath(fromPath, toPath)` and `set volume:name to newname`.

### Here's an outline of things to do:

- [x] Path handling types
- [x] `VolumeDirectory` type
- [x] Support directories on Archive
- [x] Support directories on Harddisk
- [x] Add directory support to Harddisk persistence
- [x] Substitute all occurences of string filenames with `GlobalPath`
- [x] Access path of the current script
- [x] Userspace `Path` type
- [x] `path()` command to create paths (`path` with no arguments creates a `Path` for the current path)
- [x] `VolumeDirectory` as a userspace type
- [x] Add a `volume()` command, to get a volume by name or id
- [x] `Volume` suffixes should accept strings representing `VolumePath`s
- [x] Expose all new stuff to the user, add new commands and change syntax wherever necessary
- [x] `copypath()` command
- [x] `movepath()` command
- [x] commands that accept string paths should also work with instances of `Path`
- [x] Make sure everything works (and doesn't work when it shouldn't) with RemoteTech
- [x] Test with compressed persistence disabled and enabled
- [x] Test everything with cooked filenames
- [x] Revert the change to KOSDeprecationException
- [x] List files from `boot` directory as bootfiles
- [x] test case sensitivity on archive
- [x] `Path` documentation
- [x] Document new `Volume` suffixes
- [x] Document `path()`
- [x] Document `volume()`
- [x] Document `scriptpath()`
- [x] Document `cd()`
- [x] Document `copypath()`, `movepath()`, `deletepath()`
- [x] Document `createdir()`
- [x] Document changes to `create()`, `exists()` and `open()`
- [x] `VolumeItem` documentation
- [x] `VolumeDirectory` documentation
- [x] Document new boot file handling
- [x] Review existing documentation and update if necessary

### DIFF
--- a/doc/source/structures/volumes_and_files/path.rst
+++ b/doc/source/structures/volumes_and_files/path.rst
@@ -44,6 +44,9 @@ Instances of this structure can be passed as arguments instead of ordinary, stri
         * - :attr:`PARENT`
           - :struct:`Path`
           - Parent path
+        * - :meth:`CHANGENAME(name)`
+          - :struct:`Path`
+          - Returns a new path with its name (last segment) changed
         * - :meth:`CHANGEEXTENSION(extension)`
           - :struct:`Path`
           - Returns a new path with extension changed
@@ -111,6 +114,13 @@ Instances of this structure can be passed as arguments instead of ordinary, stri
     :access: Get only
 
     Returns a new path that points to this path's parent. This method will throw an exception if this path does not have a parent (its length is 0).
+
+.. method:: Path:CHANGENAME(name)
+
+    :parameter name: :struct:`String` new path name
+    :return: :struct:`Path`
+
+    Will return a new path with the value of the last segment of this path replaced (or added if there's none).
 
 .. method:: Path:CHANGEEXTENSION(extension)
 
--- a/doc/source/structures/volumes_and_files/path.rst
+++ b/doc/source/structures/volumes_and_files/path.rst
@@ -1,0 +1,134 @@
+.. _path:
+
+Path
+====
+
+Represents a path. Contains suffixes that can be helpful when using and manipulating paths. You can use
+:ref:`path() <path_command>` to create new instances.
+
+Instances of this structure can be passed as arguments instead of ordinary, string paths, for example::
+
+  copypath("../file", path()).
+
+.. structure:: Path
+
+    .. list-table::
+        :header-rows: 1
+        :widths: 2 1 4
+
+        * - Suffix
+          - Type
+          - Description
+
+        * - :attr:`VOLUME`
+          - :struct:`Volume`
+          - Volume this path belongs to
+        * - :attr:`SEGMENTS`
+          - :struct:`List` of :struct:`String`
+          - List of this path's segments
+        * - :attr:`LENGTH`
+          - :struct:`Scalar`
+          - Number of segments in this path
+        * - :attr:`NAME`
+          - :struct:`String`
+          - Name of file or directory this path points to
+        * - :attr:`HASEXTENSION`
+          - :struct:`Boolean`
+          - True if path contains an extension
+        * - :attr:`EXTENSION`
+          - :struct:`String`
+          - This path's extension
+        * - :attr:`ROOT`
+          - :struct:`Path`
+          - Root path of this path's volume
+        * - :attr:`PARENT`
+          - :struct:`Path`
+          - Parent path
+        * - :meth:`CHANGEEXTENSION(extension)`
+          - :struct:`Path`
+          - Returns a new path with extension changed
+        * - :meth:`ISPARENT(path)`
+          - :struct:`Boolean`
+          - True if `path` is the parent of this path
+        * - :meth:`COMBINE(name1, [name2, ...])`
+          - :struct:`Path`
+          - Returns a new path created from this one
+
+.. attribute:: Path:VOLUME
+
+    :type: :struct:`Volume`
+    :access: Get only
+
+    Volume this path belongs to.
+
+.. attribute:: Path:SEGMENTS
+
+    :type: :struct:`List` of :struct:`String`
+    :access: Get only
+
+    List of segments this path contains. Segments are parts of the path separated by `/`. For example path `0:/directory/subdirectory/script.ks` contains the following segments:
+    `directory`, `subdirectory` and `script.ks`.
+
+.. attribute:: Path:LENGTH
+
+    :type: :struct:`Scalar`
+    :access: Get only
+
+    Number of this path's segments.
+
+.. attribute:: Path:NAME
+
+    :type: :struct:`String`
+    :access: Get only
+
+    Name of file or directory this path points to (same as the last segment).
+
+
+.. attribute:: Path:HASEXTENSION
+
+    :type: :struct:`Boolean`
+    :access: Get only
+
+    True if the last segment of this path has an extension.
+
+.. attribute:: Path:EXTENSION
+
+    :type: :struct:`String`
+    :access: Get only
+
+    Extension of the last segment of this path.
+
+.. attribute:: Path:ROOT
+
+    :type: :struct:`Path`
+    :access: Get only
+
+    Returns a new path that points to the root directory of this path's volume.
+
+.. attribute:: Path:PARENT
+
+    :type: :struct:`Path`
+    :access: Get only
+
+    Returns a new path that points to this path's parent. This method will throw an exception if this path does not have a parent (its length is 0).
+
+.. method:: Path:CHANGEEXTENSION(extension)
+
+    :parameter extension: :struct:`String` new path extension
+    :return: :struct:`Path`
+
+    Will return a new path with the extension of the last segment of this path replaced (or added if there's none).
+
+.. method:: Path:ISPARENT(path)
+
+    :parameter path: :struct:`Path` path to check
+    :return: :struct:`Boolean`
+
+    Returns true if `path` is the parent of this path.
+
+.. method:: Path:COMBINE(name1, [name2, ...])
+
+    :parameter name: :struct:`String` segments to add
+    :return: :struct:`Path`
+
+    Will return a new path created by adding segments to this path.
--- a/doc/source/structures/volumes_and_files/volume.rst
+++ b/doc/source/structures/volumes_and_files/volume.rst
@@ -116,7 +116,7 @@ Represents a :struct:`kOSProcessor` hard disk or the archive.
 
     :return: :struct:`VolumeItem`
 
-    Opens the file or directory pointed to by the given path and returns :struct:`VolumeItem`. It will fail if the file or directory doesn't exist.
+    Opens the file or directory pointed to by the given path and returns :struct:`VolumeItem`. It will return a boolean false if the given file or directory does not exist.
 
 .. method:: Volume:CREATE(path)
 
--- a/doc/source/structures/volumes_and_files/volume.rst
+++ b/doc/source/structures/volumes_and_files/volume.rst
@@ -34,27 +34,31 @@ Represents a :struct:`kOSProcessor` hard disk or the archive.
 
         * - :attr:`FILES`
           - :struct:`Lexicon`
-          - Lexicon of all files on the volume
+          - Lexicon of all files and directories on the volume
 
         * - :attr:`POWERREQUIREMENT`
           - :ref:`scalar <scalar>`
           - Amount of power consumed when this volume is set as the current volume
 
-        * - :meth:`EXISTS(filename)`
+        * - :meth:`EXISTS(path)`
           - :ref:`boolean <boolean>`
-          - Returns true if the given file exists
+          - Returns true if the given file or directory exists
 
-        * - :meth:`CREATE(filename)`
+        * - :meth:`CREATE(path)`
           - :struct:`VolumeFile`
           - Creates a file
 
-        * - :meth:`OPEN(filename)`
-          - :struct:`VolumeFile`
-          - Opens a file
+        * - :meth:`CREATEDIR(path)`
+          - :struct:`VolumeDirectory`
+          - Creates a directory
 
-        * - :meth:`DELETE(filename)`
+        * - :meth:`OPEN(path)`
+          - :struct:`VolumeItem`
+          - Opens a file or directory
+
+        * - :meth:`DELETE(path)`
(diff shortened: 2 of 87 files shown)


src/kOS.Safe/Exceptions/KOSDeprecationException.cs
```diff
@@ -6,7 +6,7 @@ public class KOSDeprecationException : KOSException
     {
         protected static string TerseMessageFmt = "As of kOS {0}, {1} is obsolete and has been replaced with {2}";
 
-        public KOSDeprecationException(string version, string oldUsage,string newUsage, string url) :
+        public KOSDeprecationException(string version, string oldUsage,string newUsage) :
```
Dunbaratu: I have two problems with this change:
(1) - It's got NOTHING to do with what you're trying to change in this PR.
(2) - If you had a problem with this being a parameter that wasn't used, the proper fix would have been to make it be used as it was intended (i.e. pass it down to the base KOSException class, and have KOSException populate helpURL from it), rather than remove it.  It's meant to be a feature we can hook into in a later refactor, yes the infrastructure to actually take advantage of it is only half there, but removing it entirely because it's only half done feels like taking a step backward.

tomekpiotrowski: Alright, I'll revert this.

Dunbaratu: Sorry for being harsh about it.  I just really want that feature to work some day (the help URL).  I know it's just a stub at the moment that doesn't do much yet.

Also, the consequences of this change, percolating through the rest of the code, were bloating the "files changed" count, and I was trying to narrow it down to which files I *really* needed to be looking at to see what changed for subdirectories.

tomekpiotrowski:
Rebased on top of develop-1.1 (KSP 1.1)

Dunbaratu:
@tomekpiotrowski is the "pending documentation" label still appropriate for this PR or are you done with that now?

tomekpiotrowski:
There's some documentation missing, primarily the Commands->File i/O page. It needs a few more hours of work.

doc/source/structures/volumes_and_files/path.rst
```diff
@@ -0,0 +1,134 @@
...
+
+    Returns a new path that points to this path's parent. This method will throw an exception if this path does not have a parent (its length is 0).
+
+.. method:: Path:CHANGEEXTENSION(extension)
+
+    :parameter extension: :struct:`String` new path extension
+    :return: :struct:`Path`
+
+    Will return a new path with the extension of the last segment of this path replaced (or added if there's none).
+
```
Dunbaratu: What's the rationale for making the extension different from the rest of the name, in the sense that the extension *can* be changed with a suffix method but the rest of the name cannot?

hvacengi: Wouldn't the way to change the rest of the path name be to do a chain of 
```
path:parent:combine("newname.ks").
```
Or are you specifically looking for a way to generate a path with the same folder and extension but a different file name?

Dunbaratu: In the Unix world (and therefore also Macs from OSX onward), the extension was never a separate special field.  The filename IS just one string including the extension. (From the way the filesystem "thinks" in most UNIX systems, the humans who choose the filename strings seem to like using ones with exactly one period character in them a lot of the time for some mysterious reason the filesystem doesn't understand or care about.)  The notion of an "extension" is a higher level logical abstraction consisting of that substring of the whole filename that comes after the lastmost period, if there is one.  Things like GUI file managers, and system libraries, apply this abstraction on top.  The filesystem itself has no such thing as "extensions".

For all I know the newer Windows filesystems may be the same way now too.  The notion of a file "extension" came from DOS and FAT originally.

So this is why it looks very arbitrary and weird to me for there to be a rule that you can change the extension but not the rest of the name, via the PATH suffixes.  The extension is nothing more than a subset of the filename.  Either you should be able to change it via PATH, or you shouldn't be able to.  Not this strange "it depends on which part of it you're trying to change".

doc/source/structures/volumes_and_files/volume.rst
```diff
@@ -34,27 +34,31 @@ Represents a :struct:`kOSProcessor` hard disk or the archive.
 
         * - :attr:`FILES`
           - :struct:`Lexicon`
-          - Lexicon of all files on the volume
+          - Lexicon of all files and directories on the volume
 
         * - :attr:`POWERREQUIREMENT`
           - :ref:`scalar <scalar>`
           - Amount of power consumed when this volume is set as the current volume
```
Dunbaratu: *This may be something to refactor later, not as part of this PR:*

Is it really true that the electric charge is only consumed for the current volume?  That would seem a bit arbitrary and weird to me, given that with the new subdirectories changes in this PR, it will now be possible to use a volume's files by full path, without ever switching to that volume.  The fact that a volume isn't the current one has nothing to do with whether or not the files on it are being accessed.  Not anymore.  Maybe we might have some sort of ability to have a script shut a volume off when not in use.  When it's off, then trying to open its files doesn't work until you turn it on.  When it's off it wouldn't drain charge.  When it's on it would (regardless of if it's the current volume or not).

doc/source/structures/volumes_and_files/path.rst
tomekpiotrowski: That's a good point.

To be honest IIRC I took a couple of suffixes straight from C#'s Path class, this was one of them. I thought it could be useful to expose that functionality to the user. I see no problem with adding a similar `CHANGENAME` suffix.

doc/source/structures/volumes_and_files/volume.rst
tomekpiotrowski: I tried to change as little as possible when it comes to electric charge consumption. I should have some time tomorrow to look into it and give you a better answer.

hvacengi: This is no longer true... ish.  I think I broke this actually when I changed the power logic.  We do have an option to have EC draw based on file system size, however it is defaulted to zero (and I don't anticipate us changing that).  But we still call the method in `ProcessElectricity`.  I'm gonna have to work on that logic anyways for #1363, so this change will flow into that fix easily.

doc/source/structures/volumes_and_files/path.rst
Dunbaratu: From the description in the RST file, I didn't understand how the ``combine`` suffix works.

If it really is just "return a new path with a new filename in the same directory as this path", then it gives me what I'm looking for.  I'm not looking for a suffix that ONLY changes the base substring of the name.  I see that as just as arbitrary as one that only changes the extension substring of the name.  I'm looking for how to change the whole name in one go, and treat it like it really is - as a single string - ignoring the "base dot extension" abstraction layer entirely and just changing the file's actual raw name, which includes the base, the period, and the extension in one string.

hvacengi: What does appear to be missing from that comparison though is a way to get just the file name.  the best logic I see right now is
```
set segments to path:segments.
set filename to segments[segments:length - 1].
set filename to filename:replace("boot", "noboot").
set newPath to path:parent:combine(filename).
```
An addition of some kind of file name suffix, or maybe we should look at adding a `last` suffix (and `first`) to collections.

tomekpiotrowski: See `path:name` :)

hvacengi: I feel dumb... I swear it wasn't there 10 minutes ago... never mind me

Dunbaratu: I'd be fine with `CHANGENAME` iff ``path:changename(newname)`` does exactly the same thing as ``path:parent:combine(newname)`` currently does, but not if it only works on the base name and not the extension.  (The thing @hvacengi was guessing I might have meant up above is definitely not what I wanted.)  We need to be fundamentally consistent in our terminology.  When we say "file name" elsewhere, we do include the extension (i.e. ``path:name`` returns "prog.ks" rather than just "prog", implying that when we say "name" we mean we're including the extension, not merely the base name.  Therefore ``:changename`` should be the same and the "name" it is changing should be the same "name" that ``:name`` is returning.

tomekpiotrowski: It was my intention to make it behave like `path:parent:combine(newname)`.

I assumed extension to be the part of the file name after last dot (if any). So file name includes extension.

hvacengi: I'd love it if we could get to a point where we always refer to files with the extension.  `run test.` should run a file named "test" not "test.ks" (or "test.ksm", that inference is part of the problem).

tomekpiotrowski: I think the behavior that was introduced by one of my previous PR is a good compromise between what you mention and staying backwards compatible.

That is when you create a file you need to specify the full name, no extensions are added by kOS. However when you access a file kOS tries to add '.ks' and '.ksm' to the file name if it doesn't find the original one. I've kept the same behavior in this PR.

src/kOS.Safe/Persistence/HarddiskFile.cs
```diff
@@ -1,22 +1,21 @@
-﻿using kOS.Safe.Persistence;
-
-namespace kOS.Safe.Encapsulation
+﻿namespace kOS.Safe.Persistence
 {
     [kOS.Safe.Utilities.KOSNomenclature("VolumeFile", KOSToCSharp = false)]
     public class HarddiskFile : VolumeFile
```
Dunbaratu: Should we consider "Harddisk" as two words ("hard disk") or one compound word ("harddisk")?  If we consider it two words, we should capitalize the "D", otherwise not.

tomekpiotrowski: I think it was `Harddisk` before and I just kept it that way throughout my changes. So I guess someone decided it was one word.

src/kOS/Function/Persistence.cs
```diff
@@ -6,9 +6,136 @@
...
+    public class FunctionCopyDeprecated : FunctionBase
+    {
+        public override void Execute(SharedObjects shared)
+        {
+            PopValueAssert(shared, true);
+            PopValueAssert(shared, true);
+            PopValueAssert(shared, true);
+            AssertArgBottomAndConsume(shared);
+
+            throw new KOSDeprecationException("1.0.0", "`COPY FILENAME FROM VOLUMEID.` syntax", "`COPYPATH(FROMPATH, TOPATH)`", string.Empty);
```
Dunbaratu: This isn't what "deprecation" is (maybe we should rename this exception to ``KOSObsoletionException``.)
"Deprecation" means it still works but it's frowned upon.  It's a warning that it's *going* to become obsoleted later on, but hasn't *yet*.

What an actual "deprecation" would do is throw a warning message on the screen but then redirect the call into the new technique so it still doesn't break the old scripts yet, but just annoys people using them enough to nag them to make the change.

Dunbaratu: I'm not against the idea of breaking this entirely eventually, but maybe that should wait for a big major version number update, so we can break it at the same time as we break everything else.

hvacengi: We should probably get in the habit of searching all references to this exception type and removing anything that's at least 2 "minor" releases in the past (if we get to a point where we start to increment the major version regularly we could do it then instead).

tomekpiotrowski:
I should have some time to work on this tomorrow and finish the documentation.

Dunbaratu:
great!  I'm using it a bit, trying to work from just the small bits of documentation I can see, and from looking at the code to see the function names (I gave up on trying to "black box" test it, because I can't do that without all the docs).

doc/source/structures/volumes_and_files/path.rst
tomekpiotrowski: I've added `PATH:CHANGENAME(name)`

tomekpiotrowski:
Ok, I think I'm done with documentation. I will review and test everything some more and upload updates if necessary, but it should be ready.

PR #1571, opened 2016-03-29
Fix quicksave

### LINKED ISSUE #1372: Quicksave/quickload functionality
While quicksave/quickload has not been documented or supported for a while, the bound variables still exist.  I would like to remove the existing broken logic and replace it with something that works, if possible.  I'd also like to move the logic out of `BindingsUniverse` and into `KUniverseValue`.  Ironically, if quickload worked properly now, we would be executing a quickload any time `debugdump()` was called (since quickload is a bound variable where "getting" the variable executes the quickload).

I think I have this solved, and I'll submit a PR later with the info.  For the time being, I don't plan on supporting arbitrarily named saving or loading (like `loadgame("checkpoint1")`).  This is mostly due to complications in figuring out valid names for save files, and getting that information accessible by the script.

hvacengi:
Fixes #1372

Gets quicksave and quickload to work again.  The functionality has been moved to `KUNIVERSE` (my catch-all solution to reducing binding inflation).  I implemented it such that you can both save and load the default quicksave (named "quicksave") as well as save to and from a specified file.  You can also get a list of existing saves.

Features:
- [x] Save default quicksave
- [x] Load default quicksave
- [x] Save an arbitrary quicksave file
- [x] Load an arbitrary quicksave file
- [x] List available quicksave files from kuniverse
- [x] Documentation

_Removed from plans:_
- ~~List available quicksave files using `list quicksaves in lst.` syntax~~
- ~~Print available quicksave files using `list quicksaves.` syntax.~~


hvacengi:
Removing "Not Ready" and "Pending Documentation" labels.  This should be ready for review.

PR #1605, opened 2016-04-18
Explain inter-vessel comms limitations

tomekpiotrowski:


doc/source/commands/communication.rst
```diff
@@ -6,6 +6,22 @@ Communication
...
+
+Whenever you switch to a vessel KSP simulates the vessel itself and anything that is less than 2.5 kilometers from it
+(including other vessel if there are any). Things that are outside of that range are not loaded into the simulation.
+KSP uses a simpler simulation method for them that only calculates their position and other most basic properties.
+KSP calls that "being on rails".
+
+This means kOS can't run two processors that are separated by more than 2.5km from each other at the same time.
+As a consequence it's impossible for them to communicate live. If you want to exchange messages between them
+you will have to switch from one vessel to the other. This can be troublesome if you plan on sending more
+than a few messages back and forth. Of course everything will work as expected when vessels are closer than 2.5km.
```
Dunbaratu: I think it's a bad idea to claim it's exactly 2.5km because it varies by situation (in atmosphere, landed, etc)  The bubble is actually more like 15km when in low Kerbin atmosphere, for example).  Maybe you can mention 2.5km in a parentheses comment, because as an example it's a number many people have heard of, but don't claim it's always exactly 2.5km because that's not true.

tomekpiotrowski: Oh, I didn't know that, I thought it always was 2.5km. I'll fix it.

Dunbaratu: It's often a challenge to make an explanation simultaneously simple and thruthful.  There's a great quote from Einstein (or at least often attributed to Einstein - maybe it's one of those fake quotes), that goes, "Everything should be made as simple as possible, but no simpler".  That's one of the problems I have with DasValdez's streams is that he doesn't mind making statements "simpler than possible" so they end up being misleading, and if anyone in the chat mentions an important qualifier that covers a case where what he said isn't really true, he chastises them for "confusing" the newbies with, you know, facts.

Anyway, what might be a good way to keep the explanation simple here without being false would be to just say something like "This is a greatly simplified explanation of what's really going on that's leaving out a few important qualifiers.  If you want to fully understand how this "on rails" system works, check the "loaddistance" documentation..."  With a link to the documentation page about loaddistance, which has the full description of the rails system and now messy it is.

hvacengi: I'd suggest referencing our documentation on load distance, since it discusses all of these complexities pretty thoroughly.

http://ksp-kos.github.io/KOS_DOC/structures/misc/loaddistance.html#vessel-load-distance

PR #1614, opened 2016-04-19
fix(comms): create VesselTarget using receiver's SharedObjects

### LINKED ISSUE #1611: Comms' message:sender isn't a proper fully correct vessel object
This was also discovered during my "let's kos with a kos dev" stream last night.

When you get a message that was sent to your vessel from another vessel, i.e. like so:
```
set msg to ship:messages:pop.
```
And then print out its sender field, it seems just fine, like so:
```
print msg:sender.
SHIP("Docker 1")
```
Until you try to use some of its fields and they don't work properly.
For example:
```
print msg:sender:position.
V(0,0,0)
```
Whereas if you use the sender vessel name to re-construct the vessel object from scratch, you get something correct like so:
```
print vessel(msg:sender:name):position.
V(-40.2349195, 4.1515919, -50.12249192)
```
They're the same vessel, but in the second case it is forced to reconstruct the vessel object from the name lookup again.  There is *something* about the vessel object that is embedded inside the message that isn't quite right.  Some of its fields work and some don't.  It's almost like it's a handle on dead data that would otherwise be orphaned if not for having that handle, or something like that.

I was trying to use comms to make a docking script, in which the "server" on the station would try to orient itself to point its port at the incoming ship that send it the docking request message, but as part of that I was trying to query the ``:position`` of the ``msg:sender`` and ran into this weird behaviour.  I had to use the vessel name constructor workaround shown above to make it work.

Oddly, the other suffixes I tried other than ``:position`` all seemed fine.

tomekpiotrowski:
Fixes #1611

### DIFF
--- a/src/kOS/Communication/MessageStructure.cs
+++ b/src/kOS/Communication/MessageStructure.cs
@@ -5,6 +5,7 @@ using kOS.Suffixed;
 using kOS.Safe;
 using kOS.Serialization;
 using kOS.Safe.Serialization;
+using kOS.Safe.Exceptions;
 
 namespace kOS.Communication
 {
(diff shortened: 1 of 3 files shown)


src/kOS/Communication/MessageStructure.cs
```diff
@@ -41,10 +42,22 @@ private void InitializeSuffixes()
...
+            Vessel vessel = FlightGlobals.Vessels.Find((v) => v.id.ToString().Equals(Message.Vessel));
+
+            if (vessel == null)
+            {
+                throw new KOSCommunicationException("Vessel does not exist");
+            }
+
+            return new VesselTarget(vessel, shared);
+        }
+
```
Dunbaratu: I think given that this can throw exception that breaks the execution, we need protection against that somehow.  I don't see a way for the script to detect if the vessel it's getting no longer exists.  Using the pattern used elsewhere, that would mean having a `HASSENDER` boolean suffix.

tomekpiotrowski: I could also make the method return `false`. User could then use `:ISTYPE` to check whether the result is a proper vessel.

Dunbaratu: That might be better

``if msg:sender:istype("Vessel") { stuff... }``

With the documentation for the vessel suffix saying "If the vessel stopped existing after the message was sent (i.e. it crashed or was recovered), then the type of the sender suffix won't be ``Vessel``.  So you can use the test (show the above example) to check if the vessel still exists.

That way we don't explicitly say the result is boolean false even though it is.  My reason for this is that I don't like the notion of a boolean that can never be true and only is of type boolean when it's false.  (One of the good features that was lost when going from old C to modern languages was the ability to treat *everything* as a boolean, which often allowed this kind of check to always work - null is false, non-null is true).

Dunbaratu: I'll implement this suggested change as I review and merge this (make it return boolean false, and update the docs).

Dunbaratu: In the end I decided to support both.  HASSENDER *and* making :SENDER return two different types.

PR #1618, opened 2016-04-21
Fixes #1610: The problem was caused by not clearing the stack when aborting with c…

### LINKED ISSUE #1610: Ctrl+C abort doesn't clear context
Found in @Dunbaratu's stream that aborting the current program using `[Ctrl]+[C]` does not properly clear out old variables.  This was made evident by an attempt to call a function throwing an exception about the function being from a different context, rather than complaining about an undefined variable.

Dunbaratu:
…ontrol-C.

Fixes #1610 

It was falsifying the notion of whether or not the system is at
global scope when it sees a "local" directive, because it
saw the leftover remains of the previous run on the stack.  That
error cascaded into other problems causing it to end up seeing
the old leftover version of the delegate that was still in the
global variable from last time.

### DIFF
--- a/src/kOS.Safe/Execution/CPU.cs
+++ b/src/kOS.Safe/Execution/CPU.cs
@@ -371,6 +371,7 @@ namespace kOS.Safe.Execution
                     shared.Screen.Print("Program aborted.");
                     shared.BindingMgr.UnBindAll();
                     PrintStatistics();
+                    stack.Clear();
                 }
                 else
                 {
@@ -381,6 +382,7 @@ namespace kOS.Safe.Execution
                         shared.Screen.Print("Program ended.");
                         shared.BindingMgr.UnBindAll();
                         PrintStatistics();
+                        stack.Clear();
                     }
                 }
             }
@@ -413,11 +415,13 @@ namespace kOS.Safe.Execution
         {
             if (userDelegate.ProgContext != currentContext)
             {
-                throw new KOSInvalidDelegateContextException(
-                    (currentContext == contexts[0] ? "the interpreter" : "a program"),
-                    (currentContext == contexts[0] ? "a program" : "the interpreter")
-                    );
-            }
+                string currentContextName = (currentContext == contexts[0] ? "the interpreter" : "a program");
+                string delegateContextName = (
+                    userDelegate.ProgContext == contexts[0] ? currentContextName = "the interpreter" : (
+                        currentContext == contexts[0] ? "a program" : "a different program from a previous run" ) );
+
+                throw new KOSInvalidDelegateContextException(currentContextName, delegateContextName);
+           }
         }
 
         /// <summary>


src/kOS.Safe/Execution/CPU.cs
```diff
@@ -413,11 +415,13 @@ public void AssertValidDelegateCall(IUserDelegate userDelegate)
...
-                    (currentContext == contexts[0] ? "a program" : "the interpreter")
-                    );
-            }
+                string currentContextName = (currentContext == contexts[0] ? "the interpreter" : "a program");
+                string delegateContextName = (
+                    userDelegate.ProgContext == contexts[0] ? currentContextName = "the interpreter" : (
+                        currentContext == contexts[0] ? "a program" : "a different program from a previous run" ) );
+
+                throw new KOSInvalidDelegateContextException(currentContextName, delegateContextName);
+           }
```
Dunbaratu: Not directly related to the error, but I noticed that the error message was printing out incorrectly before.  (In addition to the fact that it shouldn't have triggered an error in the first place, once it did, the error it showed was misleading.)

hvacengi: Do you mind if I try to pull this into a descreet `if` block?  I've never been a huge fan of the `?` `:` syntax, but it really gets tough to read when chaining multiple instances of it together.

Dunbaratu: For some reason I didn't see this line comment when you first made it. 

Go ahead and change it if you prefer.

I understand that it looks ugly, but it's not just a syntax difference from an``if`` block.  It's also a design pattern difference because essentially using ``?`` ``:`` is functional thinking while using ``if``/``else`` is procedural thinking.  I just wish it looked better.  It's algorithmically prettier, but syntactically uglier.

hvacengi: My issue is a matter of readability, multiple chains like this just take more attention to review.  Here is how I interpret what you have submitted:
```
                string delegateContextName;
                if (userDelegate.ProgContext == contexts[0])
                {
                    currentContextName = "the interpreter";
                    delegateContextName = "the interpreter";
                }
                else if (currentContext == contexts[0])
                {
                    delegateContextName = "a program";
                }
                else
                {
                    delegateContextName = "a different program from a previous run";
                }
```

Dunbaratu: Your re-write altered the logic.

Possible valid phrasings for the exception message are:

* Cannot call this lock or function from ``the interpreter`` when it was declared in ``a program``.
* Cannot call this lock or function from ``a program`` when it was declared in ``the interpreter``.
* Cannot call this lock or function from ``a program`` when it was declared in ``a different program from a previous run``.

What your first condition ends up printing the same string for both, which makes the nonsensical message:

* Cannot call this lock or function from ``the interpreter`` when it was declared in ``the interpreter``.

The only reason for the ugly additional message (the change I made from the red text you see above in github), is that previously it was making the false assumption that at least one of the two contexts must be the interpreter, therefore if one is a program than the other must be the interpreter.  This is false because they can be two different programs. (but can't be two different interpreters).

The truth table is this:

```
 Is currentContext | is userDelegate.ProgramContext ||  Resulting message
   equal to        |   equal to contexts[0] ?       ||
  contexts[0] ?    |                                ||
 ================  | ============================   || ===========================
                   |                                ||  n/a  (This case is impossible because
     yes           |        yes                     ||  this code is inside an if-check that only
                   |                                ||  is true when the two contexts are different.)
 ----------------  | ---------------------------    ||  ------------------------
                   |                                ||  currentContextName = "the interpreter"
     yes           |        no                      ||  delegateContextName = "a program"
                   |                                ||
 ----------------  | ---------------------------    ||  ------------------------
                   |                                ||  currentContextName = "a program"
     no            |        yes                     ||  delegateContextName = "the interpreter"
                   |                                ||
 ----------------  | ---------------------------    ||  ------------------------
                   |                                ||  currentContextName = "a program"
     no            |        no                      ||  delegateContextName = "a different program from a previous run"
                   |                                ||
```
Re-write that if/else logic however you like as long as it matches that truth table like the original expression did.  

The plain English version of the logic is basically "say 'interpreter' for the one that's context[0], and 'program' for the one that isn't, but also check for the case where they're both 'program' but different programs, and make sure the phrasing makes sense in that edge case".

hvacengi: That's kind of what I expected, however I'm not sure that's what the chain is doing.  Maybe I'm reading it wrong.  I read this portion:
```
string delegateContextName = (
                    userDelegate.ProgContext == contexts[0] ? currentContextName = "the interpreter" : \\ the rest of the chain
```
_(specifically the single `=` after currentContextName)_

to mean
```
if (userDelegate.ProgContext == contexts[0])
{
    delegateContextName = currentContextName = "the interpreter";
}
else // continue down the chain
```
Am I reading it wrong, or is it behaving differently than intended?

hvacengi: I've corrected the error per your truth-table and tested to verify.  I'll push it upstream in an hour or so

Dunbaratu: Good catch.  It was an error from cutting and pasting as I arranged the statement around.  The ``currentContextName =`` should have been deleted.

Issue #1620, opened 2016-04-24
vessel:controlpart suffix


AlchemistCH:
It could be useful to have the option get the control part of the vessel. 
For example, for docking it's common practice to set control from the docking port you are going to use (unless it's right on top of your cockpit) and docking autopilot/assist programs would benefit from knowing exactly what part it is (since those need the position, not only the axis of the port of choice)

PR #1621, opened 2016-04-24
added Vessel:CONTROLPART suffix

AlchemistCH:
The suffix returns the part currently set for the control reference of the vessel.
for #1620

### DIFF
--- a/src/kOS/Suffixed/VesselTarget.cs
+++ b/src/kOS/Suffixed/VesselTarget.cs
@@ -445,6 +445,7 @@ namespace kOS.Suffixed
             AddSuffix("LOADED", new Suffix<BooleanValue>(() => Vessel.loaded));
             AddSuffix("UNPACKED", new Suffix<BooleanValue>(() => !Vessel.packed));
             AddSuffix("ROOTPART", new Suffix<PartValue>(() => PartValueFactory.Construct(Vessel.rootPart, Shared)));
+            AddSuffix("CONTROLPART", new Suffix<PartValue>(() => PartValueFactory.Construct(Vessel.GetReferenceTransformPart(), Shared)));
             AddSuffix("DRYMASS", new Suffix<ScalarValue>(() => Vessel.GetDryMass(), "The Ship's mass when empty"));
             AddSuffix("WETMASS", new Suffix<ScalarValue>(() => Vessel.GetWetMass(), "The Ship's mass when full"));
             AddSuffix("RESOURCES", new Suffix<ListValue<AggregateResourceValue>>(() => AggregateResourceValue.FromVessel(Vessel, Shared), "The Aggregate resources from every part on the craft"));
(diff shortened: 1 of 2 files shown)


src/kOS/Suffixed/VesselTarget.cs
```diff
@@ -464,6 +465,13 @@ private void InitializeSuffixes()
             AddSuffix("MESSAGES", new NoArgsSuffix<MessageQueueStructure>(() => GetMessages()));
         }
 
+        public global::Part GetControlPart()
+        {
+            global::Part res = Vessel.GetReferenceTransformPart(); //this can actually be null if "control from here" has never been used
```
hvacengi: In a quick test, it looks like this comment is not true.  On freshly launched ship, with the root part set as the 2.5mx3.75m adapter, the suffix returns one of the command pods located on the ship as the value for `CONTROLPART`.  I don't doubt that the value can be null (the internal implementation appears to leave it null if `referenceTransformId == 0u`) but I think I'll revise the comment.  I wanted to give you a chance to comment as well before I simply made the sweeping change.

AlchemistCH: With crewed pods it may automatically do "control from here" on one of them if nothing is set. 
Try starting with a probe core, then put a crewed module upside down and put a kerbal in - the navball will be upside down (I've got this effect even with the cabin that has no control authority!).

I first made it without this addition and tested on couple crafts and it worked. Then I wrote a script that uses it, ran it on one thing controlled with just a probe core (on which I haven't changed the control point) and got the null reference

hvacengi: I've tried a few tests, and I can't seem to find a way to get the control part to actually show as the root part.  You must have found an odd combination that caused the null reference.  I'm going to merge it and just edit the comment so that we don't imply that it will always be null if you don't call "control from here".

AlchemistCH: Of course, you may change the comment as you see fit.
The craft that gave me the null reference had its probe core as the root part. Maybe if the root isn't a control module (or maybe a docking port), it always switches control to a core or a capsule, if there's one (and crewed modules may even override probe cores in the root)
Uncontrollable root part as the control reference is probably left only for uncontrollable debris

hvacengi: Oh, I also didn't try it on a vessel with no core or module, since we have historically been able to control those vessels anyways (purely as a bug) but I could see that being another point where it comes up.  I'm glad it came up in your testing, it might have been hard to track down once it was released.

AlchemistCH: Incidentally, that was the first craft I used this on after uploading initial the version here. The tests I did before showed no issues. Testing luck.
It seems a bit strange that the function just returns null instead of root, but maybe it's used somewhere as "if there's something better than root, switch control to that, otherwise control from root"

hvacengi: It seems to me that no matter what it _shouldn't_ want to give you a null, cause it should only be null if there is no "reference transform" and every vessel (controlled or not) should have a reference transform.  I'd have to dig through ILSpy to verify, but it very well could be that a ship with no "control" when it loads doesn't get the "reference transform" set (which might also be an interesting interaction with remote tech).

AlchemistCH: Must be some wonders with legacy code stemming from the times when you could have only one command module on the ship and it was always the root.

PR #1622, opened 2016-04-24
Separate pitch, yaw and roll gimbal enabling/locking

AlchemistCH:
One of the new features of 1.1 - ability to enable/disable pitch yaw and roll response of the gimbals separately. 
Added corresponding suffixes to the Gimbal part module.

### DIFF
--- a/src/kOS/Suffixed/PartModuleField/GimbalFields.cs
+++ b/src/kOS/Suffixed/PartModuleField/GimbalFields.cs
@@ -20,6 +20,18 @@ namespace kOS.Suffixed.PartModuleField
             {
                 gimbal.gimbalLock = value;
             }, "Is the Gimbal free to travel?"));
+            AddSuffix("PITCH", new SetSuffix<BooleanValue>(() => gimbal.enablePitch, value =>
+            {
+                gimbal.enablePitch = value;
+            }, "Does the Gimbal respond to pitch controls?"));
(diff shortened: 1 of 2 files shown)


hvacengi:
I'm going to tag issue #1587 here as it is the associated issue, but it won't fully fix it since that issue was focused heavily towards cooked steering.

src/kOS/Suffixed/PartModuleField/GimbalFields.cs
```diff
@@ -20,6 +20,18 @@ private void InitializeGimbalSuffixes()
...
+                gimbal.enablePitch = value;
+            }, "Does the Gimbal respond to pitch controls?"));
+            AddSuffix("YAW", new SetSuffix<BooleanValue>(() => gimbal.enableYaw, value =>
+            {
+                gimbal.enableYaw = value;
+            }, "Does the Gimbal respond to yaw controls?"));
+            AddSuffix("ROLL", new SetSuffix<BooleanValue>(() => gimbal.enableRoll, value =>
+            {
+                gimbal.enableRoll = value;
+            }, "Does the Gimbal respond to roll controls?"));
```
Dunbaratu: The indentation above looks really wonky.  The braces shouldn't be at the same level as the "A" in "AddSuffix" because they're a continuation of the AddSuffix call.  They should be indented further in.

Issue #1623, opened 2016-04-27
ActiveResource doesn't support :density


jberkman:
stage:resourcesLex claims it returns `Resource`s, which support `:density`, but they are actually `ActiveResource`s, which don't.

hvacengi:
Close via #1624

PR #1624, opened 2016-04-27
Add :density suffix to ActiveResourceValue

jberkman:
Resolves issue #1623


hvacengi:
Maybe it's because I'm reviewing this fairly late at night, but it looks like we don't document an `ActiveResource` type in the documentation, only `Resource` and `AggregateResource`.  But in the C# we only have the types of `ActiveResource` and `AggregateResource` and not plain `Resource`.  I suspect that we want to set the `KOSNomenclature` on `ActiveResource` to point to the name "Resource" at the same time, to avoid further confusion.

Dunbaratu:
I see no Sphinx documentation for this new suffix.

hvacengi:
Yeah, the issue is that we technically have 3 different resource structures, that didn't all have the same suffixes, but were documented as a single structure type.

PR #1626, opened 2016-04-29
More custom action groups

### LINKED ISSUE #1469: PANELS ON affects some things that aren't solar panels
I had a vessel that had radiator panels and a crew ladder, and solar panels on it.

When I do a `PANELS ON.` command, it toggles more than just the panels.  It also extends the radiator panels and the ladder.

AlchemistCH:
fixes #1469 and expands the case. 
And adds a few more groups.
Warning: some cases of quite complex (and sometimes a bit weird) inter-module interaction

The changes are:

**changed groups** 
`PANELS ON/OFF` now affects only deployable solar panels (#1469)

**new groups:**

`CHUTESSAFE` - all chutes that can be safely deployed in the current conditions
returns false if there are chutes that can be safely deployed, true otherwise
`CHUTESSAFE ON` deploys all chutes that are safe to deploy.
this code will gradually deploy all the chutes as the speed drops:
`WHEN (NOT CHUTESSAFE) THEN {CHUTESSAFE ON. RETURN (NOT CHUTES).}`

`RADIATORS`
returns true if all the radiators are working.
`RADIATORS ON/OFF` deploys/retracts all the deployable radiators and activates/deactivates all the fixed ones.

`LADDERS`
returns true if all the ladders are extended.
`LADDERS ON/OFF` extends/retracts all the extendable ladders.

`BAYS` - payload/service bays
returns true if at least one is open (since even one of them being open may be a critical flight issue).
`BAYS ON/OFF` opens/closes all the payload and service bays.

`DEPLOYDRILLS` - mining drills deployment
returns true if all the drills are deployed.
`DEPLOYDRILLS ON/OFF` - extends/retracts all the drills.

`DRILLS` - mining drills
returns true if at least one of them is actually mining
`DRILLS ON/OFF` - activates (will do anything only if the drill is deployed and has contact with something minable) or deactivates all the drills.

`FUELCELLS` (distinguished from other converters by converter/action names)
returns true if at least one is working (since it's common practice to have converters activated on individual basis).
`FUELCELLS ON/OFF` activates/deactivates all the fuel cells.

`ISRU` - ISRU converters (distinguished from other converters by converter/action names)
returns true if at least one is working (since it's common practice to have converters activated on individual basis).
`ISRU ON/OFF` activates/deactivates all the ISRU converters.

### DIFF
--- a/src/kOS/Utilities/VesselUtils.cs
+++ b/src/kOS/Utilities/VesselUtils.cs
@@ -231,29 +231,64 @@ namespace kOS.Utilities
             return "None";
         }
 
-        public static void LandingLegsCtrl(Vessel vessel, bool state)
+               public static KSPActionParam makeActionParam(bool state) // just use this to when you need to call action that requires bool input for direction
         {
-            // This appears to work on all legs in 0.22
-            vessel.rootPart.SendEvent(state ? "LowerLeg" : "RaiseLeg");
+            if (state){ return new KSPActionParam(0, KSPActionType.Activate); }
+            else { return new KSPActionParam(0, KSPActionType.Deactivate);}
         }
 
-        internal static object GetLandingLegStatus(Vessel vessel)
+        public static void LandingLegsCtrl(Vessel vessel, bool state)
+        {
+            vessel.rootPart.SendEvent(state ? "LowerLeg" : "RaiseLeg"); //legacy
+
+            foreach (var p in vessel.parts)
+            {
+                var geardeploy = p.FindModulesImplementing<ModuleWheels.ModuleWheelDeployment>();
+                foreach (var gd in geardeploy)
+                {
+                    var gear = p.Modules[gd.baseModuleIndex] as ModuleWheelBase;
+                    if ((gear != null) && (gear.wheelType == WheelType.LEG)) // identify leg
+                    {gd.ActionToggle(makeActionParam(state)); }
+                }
+            }
+        }
+
+        public static object GetLandingLegStatus(Vessel vessel)
         {
             var atLeastOneLeg = false; // No legs at all? Always return false
 
             foreach (var p in vessel.parts)
             {
-                if (!p.Modules.OfType<ModuleLandingLeg>().Any()) continue;
-                atLeastOneLeg = true;
-
-                var legs = p.FindModulesImplementing<ModuleLandingLeg>();
-
-                if (legs.Any(l => l.savedLegState != (int)(ModuleLandingLeg.LegStates.DEPLOYED)))
+                if (p.Modules.OfType<ModuleLandingLeg>().Any()) //Legacy
                 {
-                    return false;
+                    atLeastOneLeg = true;
+
+                    var legs = p.FindModulesImplementing<ModuleLandingLeg>();
+
+                    if (legs.Any(l => l.savedLegState != (int)(ModuleLandingLeg.LegStates.DEPLOYED)))
+                    {
+                        return false;
+                    }
+                }
+               var gearlist = p.FindModulesImplementing<ModuleWheelBase>();
+                if (gearlist.Count>0)
+                {
+                    foreach (var gear in gearlist)
+                    {
+                        if (gear.wheelType == WheelType.LEG)
+                        {
+                            atLeastOneLeg = true;
+                            var geardeploy = p.FindModulesImplementing<ModuleWheels.ModuleWheelDeployment>();
+                            foreach (var gd in geardeploy) {
+                                if ((p.Modules[gd.baseModuleIndex]==gear)&&(gd.fsm.CurrentState!=gd.st_deployed)) //state string is unreliable - may be just empty
+                                {
+                                    return false;
+                                }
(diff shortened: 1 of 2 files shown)


hvacengi:
Wow, this looks like a nice expansion of action groups and should help eliminate some helper functions.

hvacengi:
Could you add documentation for the new action groups as well?

AlchemistCH:
The situation with the current docs on the matter needs a bit of rework.
I think, there needs to be proper explanation on each of them in  Command Reference » Flight Control » Ship Systems  and references to that from the table in  General Topics » Catalog of Bound Variable Names -> BOOLEAN TOGGLE FIELDS

I'll try to make the docs for this.

AlchemistCH:
While testing something for documentation, found another issue - LEGS action group doesn't seem to work anymore. And they now use the same bunch of modules as aircraft gear.

AlchemistCH:
OK, fixed the legs. Left the legacy code there as well, since some mods may still use the old module for legs.

hvacengi:
If you think the docs are gonna be more than you want to bite off, I can help out and do it myself. Last night I was trying to speed run as many PRs as I could, and managed 3...

AlchemistCH:
I'll handle this. But given the scale beyond just documenting the current modification, I'm submitting the docs as a separate PR.

P.S. Remembered another thing that can be useful to control this way - air intakes. Now they're in.

src/kOS/Utilities/VesselUtils.cs
```diff
@@ -318,7 +389,312 @@ public static object GetSolarPanelStatus(Vessel vessel)
...
+        {
+            var atLeastOneRadiator = false; // No radiators at all? Always return false
+
+            foreach (var p in vessel.parts)
+            {
+                foreach (var c in p.FindModulesImplementing<ModuleActiveRadiator>())
+                {
+                    atLeastOneRadiator = true;
+
+                    if (! c.IsCooling)
```
hvacengi: Should we also check for `!p.ShieldedFromAirstream` so that this will not always return false if you have a radiator stowed inside of a cargo bay.  I believe that parts which are shielded will automatically not extend if given the command.

AlchemistCH: Depends on what we want with parts that don't want to be deployed...
But in case of radiators this code references both deployable (which won't deploy and work if stowed) and fixed (which don't seem to radiate when stowed, but still drain heat from adjacent parts if active - acting as active heat sinks) radiators

src/kOS/Utilities/VesselUtils.cs
```diff
@@ -231,29 +231,64 @@ public static object TryGetEncounter(Vessel vessel, SharedObjects sharedObj)
...
-        internal static object GetLandingLegStatus(Vessel vessel)
+        public static object GetLandingLegStatus(Vessel vessel)
         {
             var atLeastOneLeg = false; // No legs at all? Always return false
 
             foreach (var p in vessel.parts)
             {
-                if (!p.Modules.OfType<ModuleLandingLeg>().Any()) continue;
-                atLeastOneLeg = true;
+                if (p.Modules.OfType<ModuleLandingLeg>().Any()) //Legacy
```
AlchemistCH: Maybe we don't need that legacy code anymore. By this point most mods would likely have this updated to the new system. The stock had fully switched right away, including automatic conversion on existing crafts.

hvacengi: I think I'll leave the legacy code in for the next update, but I'm going to re-order them to ensure that we check for the new module first, just in case some one accidentally puts both on a part.

src/kOS/Utilities/VesselUtils.cs
hvacengi: My primary concern (and it applies to chutes too I think) is in the case of `toggle radiators.` where it will first check radiator status and then set it to the opposite value.  If you have a deployable radiator inside of a fairing or cargo bay, toggle will never be able to de-activate the other radiators because it will always read the current value as `false`.  I'll take a closer look at the interaction implied by the fixed radiators having an active state as well.

AlchemistCH: Why would anybody use toggle on these? Unlike stock action groups which work on toggle, these custom groups can be called (and will cause corresponding action) both ON and OFF no matter the displayed state. If some are activated and some are deactivated, ON will activate what's off and OFF will deactivate what's on

hvacengi: Yeah, I mostly don't like the syntax inconsistency of it.  Functionally, it's arguable if toggle really needs to exist for anything outside of stock action groups (even then, it's essentially `set ag1 to not ag1.`).  I'll leave the fix out for now, we'll see what we get for comments in release.

AlchemistCH: Let's hope I put enough documentation for people to actually find these groups (and get the difference from the stock ones). Because I really first noticed most of them after seeing the issue about panels group on GitHub - and then looking into the source code

PR #1635, opened 2016-05-01
NAVMODE -  NavBall / speed display modes

### LINKED ISSUE #1448: NavBall mode access
We currently have SAS modes controllable from the script, however the behavior of SAS is dependent on the navball mode (surface, orbit or target) to which there seems to be no access currently implemented.
We probably need to implement it, even if most people use the steering manager anyway.

AlchemistCH:
Added NAVMODE (navball mode, only accessible for the active vessel) 
(Fixes #1448 )

also slightly adjusted getting SASMODE due to still persisting RADIALIN/RADIALOUT inconsistency.

### DIFF
--- a/src/kOS/Binding/FlightControlManager.cs
+++ b/src/kOS/Binding/FlightControlManager.cs
@@ -1,4 +1,4 @@
-﻿using kOS.AddOns.RemoteTech;
+using kOS.AddOns.RemoteTech;
 using kOS.Safe.Binding;
 using kOS.Safe.Utilities;
 using kOS.Safe.Exceptions;
@@ -47,7 +47,9 @@ namespace kOS.Binding
             AddNewFlightParam("wheelsteering", Shared);
 
             shared.BindingMgr.AddSetter("SASMODE", value => SelectAutopilotMode((string)value));
-            shared.BindingMgr.AddGetter("SASMODE", () => currentVessel.Autopilot.Mode.ToString().ToUpper());
+            shared.BindingMgr.AddGetter("SASMODE", () => GetAutopilotModeName().ToUpper());
+            shared.BindingMgr.AddSetter("NAVMODE", value => SetNavMode((string)value));
+            shared.BindingMgr.AddGetter("NAVMODE", () => getNavMode().ToString().ToUpper());
         }
 
 
@@ -225,6 +227,15 @@ namespace kOS.Binding
             }
         }
 
+        public string GetAutopilotModeName()
+        {
+            // TODO: As of KSP 1.0.4, RadialIn and RadialOut are swapped.  Check if still true in future versions.
+            if (currentVessel.Autopilot.Mode == VesselAutopilot.AutopilotMode.RadialOut) { return "RadialIn"; }
+            if (currentVessel.Autopilot.Mode == VesselAutopilot.AutopilotMode.RadialIn) { return "RadialOut"; }
+
+            return currentVessel.Autopilot.Mode.ToString();
+        }
+
         public void SelectAutopilotMode(string autopilotMode)
         {
             // handle a null/empty value in case of an unset command or setting to empty string to clear.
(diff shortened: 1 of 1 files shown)


src/kOS/Binding/FlightControlManager.cs
```diff
@@ -225,6 +227,15 @@ public void SelectAutopilotMode(VesselAutopilot.AutopilotMode autopilotMode)
             }
         }
 
+        public string GetAutopilotModeName()
+        {
+            // TODO: As of KSP 1.0.4, RadialIn and RadialOut are swapped.  Check if still true in future versions.
```
hvacengi: I don't suppose you checked this value in KSP 1.1.2?  I seem to recall a mention in a change log about fixing the radial button, but that may have just meant that when they redid the Button UI they didn't reverse the two either.  The fix may have been to fix the enums, or it could have been to add back in the odd behavior.

src/kOS/Binding/FlightControlManager.cs
```diff
@@ -225,6 +227,15 @@ public void SelectAutopilotMode(VesselAutopilot.AutopilotMode autopilotMode)
...
+            if (currentVessel.Autopilot.Mode == VesselAutopilot.AutopilotMode.RadialOut) { return "RadialIn"; }
+            if (currentVessel.Autopilot.Mode == VesselAutopilot.AutopilotMode.RadialIn) { return "RadialOut"; }
+
+            return currentVessel.Autopilot.Mode.ToString();
```
hvacengi: Lets just return the upper string here, and then the suffix only needs to reference the method and won't need the anonymous method.  (I'll grab this in review unless you object)

src/kOS/Binding/FlightControlManager.cs
```diff
@@ -270,13 +281,64 @@ public void SelectAutopilotMode(string autopilotMode)
...
         }
 
+  
+        public FlightGlobals.SpeedDisplayModes getNavMode()
+        {
+            if (Shared.Vessel != FlightGlobals.ActiveVessel)
+            {
+                throw new kOS.Safe.Exceptions.KOSSituationallyInvalidException("NAVMODE can only be accessed for the Active Vessel");
+            }
+            return FlightGlobals.speedDisplayMode;
```
hvacengi: Same note as above for GetAutopilotModeName

src/kOS/Binding/FlightControlManager.cs
```diff
@@ -270,13 +281,64 @@ public void SelectAutopilotMode(string autopilotMode)
...
+        }   
+
+        public void SetNavMode(FlightGlobals.SpeedDisplayModes navMode)
+        {
+            FlightGlobals.SetSpeedMode(navMode);
+        }
+
+        public void SetNavMode(string navMode)
```
hvacengi: We can add a shim method that accepts `StringValue` and calls this `string` method, eliminating another anonymous method.

src/kOS/Binding/FlightControlManager.cs
AlchemistCH: I tested it exactly on 1.1.2. And these are still inverted.
Don't know what and where they fixed, but apparently not this.

src/kOS/Binding/FlightControlManager.cs
AlchemistCH: Of course, if it's better for the way these are represented internally (I don't have too much idea about this), since these functions are only used for these bindings.

AlchemistCH:
@hvacengi  since you are busy with other PRs, I fixed the anonymous functions in binders myself. 
Interestingly, `SelectAutopilotMode` already had the option of accepting `object`, yet the setter for `SASMODE` had explicit conversion - fixed this, too.

src/kOS/Binding/FlightControlManager.cs
```diff
@@ -201,6 +201,8 @@ public void SelectAutopilotMode(object autopilotMode)
             if (autopilotMode is Direction)
             {
                 //TODO: implment use of direction subclasses.
+                throw new Safe.Exceptions.KOSException(
```
hvacengi: This might as well throw an exception if the type is not `string` or `StringValue` as well, since setting it to `0` is illegal as well.  The exception should probably be a `KOSWrongControlValueTypeException` exception (see the `UpdateThrottle` method in the same file for an example).

AlchemistCH: Yes, I think there should be proper exceptions for wrong types on both these setters. I'll put t there.
But this particular branch looks interesting, as if somebody wanted to add the option to feed the Direction directly into SAS... 
Anyway, accidentally feeding `prograde` instead of `"prograde"` could be a way too likely mistake not to put a special message for trying to feed a direction.

hvacengi: Yeah, that was the goal a long time ago.  The problem is that the actual "direction" of prograde changes, even though the stored variable stays the same (sasmode uses `set` not `lock`).  Then we wanted to have a special way to identify these magic directions, so the bound variable `prograde` would include a flag that said "I was made to be a magic updating variable".  But that got complicated, and there isn't a good way to do "stability" the same way, nor anti target.  So we kind of gave up.

AlchemistCH: It may be possible to use the old-school stability assist (that doesn't reset itself when too far from the target rotation), but it wouldn't really help in case of typo, and there's cooked steering for just using a direction anyway.

PR #1636, opened 2016-05-02
Exclude Targeting input lock

### LINKED ISSUE #1634: Unable to set target via kOS.
When trying to set a target using kOS, using `set target to "vesselname"`, no target can be set, and a prompt from KSP (not kOS) is displayed in the upper center of the screen reading "Target selection locked." Targets can be set normally in KSP, and once set by hand, kOS functions normally.

Example, using a stock Velociteze with a KAL9000 unit, trying to target a probe titled "KSC Runway 09" using `set target to "KSC Runway 09"` resulted in the "Target selection locked" message. However, manually selecting "KSC Runway 09" as the target, and then using `print target` correctly resulted in SHIP("KSC Runway 09"). Additionally, once the target is set by hand, all properties of the target can be accessed by kOS; the issue appears to lie solely with setting the target within kOS itself.

hvacengi:
Fixes #1634

TermWindow.cs
* Filter `SetControlLock` to exclude ControlTypes.TARGETING`.

### DIFF
--- a/src/kOS/Screen/TermWindow.cs
+++ b/src/kOS/Screen/TermWindow.cs
@@ -250,6 +250,7 @@ namespace kOS.Screen
             cameraManager = CameraManager.Instance;
             cameraManager.enabled = false;
 
+            // Exclude the TARGETING ControlType so that we can set the target vessel with the terminal open.
             InputLockManager.SetControlLock(ControlTypes.All & ~ControlTypes.TARGETING, CONTROL_LOCKOUT);
 
             // Prevent editor keys from being pressed while typing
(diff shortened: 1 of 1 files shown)


hvacengi:
I'm labeling this as "Breaking" not so much because it breaks existing scripts, or even that it is kOS breaking anything.  Because KSP now uses an input lock to determine if you can set the current target, it means that it is possible that other mods may set an input lock of "ALL" which would still prevent us from setting the target.  All devs should be aware of this, and users should have a warning brought to their attention at least upon initial release so that we can try to find these mods and ask them to please exlude the "TARGETING" input.

PR #1640, opened 2016-05-07
Rt 17

tomekpiotrowski:


src/kOS/AddOns/RemoteTech/RemoteTechAntennaModuleFields.cs
```diff
@@ -118,28 +117,31 @@ protected new Structure GetKSPFieldValue(StringValue suffixName)
...
-                {
-                    return new Guid(MissionControlGuid);
+                    return api.GetActiveVesselGuid();
                 }
                 else
                 {
-                    var body = FlightGlobals.Bodies.FirstOrDefault(b => b.bodyName.Equals(targetString, StringComparison.InvariantCultureIgnoreCase));
+                    IEnumerable<string> groundStations = api.GetGroundStations();
+                    foreach (var groundStation in groundStations) {
+                        if (targetString.Equals(groundStation)) {
```
hvacengi: If switched to `Equals(groundStation, StringComparison.OrdinalIgnoreCase)` setting the ground station will no longer be case sensitive, which is preferred since most of kOS is case-insensitive.

I'll grab it in review, and I'll switch the other string comparisons in this section to us Oridinal which is faster than InvariantCulture

doc/source/addons/RemoteTech.rst
```diff
@@ -22,15 +22,24 @@ Antennas
...
 
-Acceptable values for `"target"` are: `"no-target"`, `"active-vessel"`, `"mission-control"`, a :struct:`Body`, a :struct:`Vessel`, or a string containing the name of a body or vessel.
+Acceptable values for `"target"` are:
+
+- `"no-target"`
+- `"active-vessel"`
+- a :struct:`Body`
+- a :struct:`Vessel`
+- a string containing the name of a body or vessel
+- a string containing the name of a ground station (case-sensitive)
```
hvacengi: Given my previous comment, I'll delete the "(case-sensitive)" portion.

src/kOS/AddOns/RemoteTech/RemoteTechAntennaModuleFields.cs
tomekpiotrowski: I've made them case-sensitive on purpose. Those names are case-sensitive in RemoteTech. We want to be able to differentiate between `Site` and `site` in kOS.

hvacengi:
Note to self: make sure this fixes #1679 when I review

PR #1644, opened 2016-05-11
Fix part mass modifications

### LINKED ISSUE #1115: Physics-less parts still report as 0 mass.
This was the correct thing to do in older versions of KSP but as of 1.0.1 or 1.0(?) it seems that the mass of a physics-less part is added to its parent.

### LINKED ISSUE #1643: Improper implementation of IPartMassModifier
Currently we don't handle `IPartMassModifier` correctly, which can result in ever increasing mass.  Anecdotal evidence implies that the issue is only showing up when multiple part modules modify mass (which is why it isn't an issue on the base parts).

I'll get a fix pulled together in a PR.

hvacengi:
Fixes #1643

Fixes #1115

kOSProcessor.cs
* Fix implementation of IPartMassModifier so that kOS does not change
the part's mass directly.
* Revise base mass/cost logic to allow for adding mass and cost when the
module is added to other parts.
* Pulled mass/cost multiplication factors out of constants and into the
part definition.  This way the part itself can define how much mass to
add (since the KAL9000's mass added is not in line with the initial
mass).
* Removed old logic for establishing the base cost based on additional
cost.

PartUtilities.cs
* Revise mass handling for parts so that physics-less parts are
accounted for
* Revise wet and dry mass calculation to use the new part mass
calculation.  This should make both the ship and part masses for each
correct.

### DIFF
--- a/doc/source/general/kospartmodule.rst
+++ b/doc/source/general/kospartmodule.rst
@@ -29,6 +29,12 @@ attached to the small disk shaped CPU part (KR-2402 b)::
 	    diskSpace = 5000
 	    ECPerBytePerSecond = 0
 	    ECPerInstruction = 0.000004
+        # Optional fields shown below with default value
+        baseDiskSpace = 0
+        diskSpaceCostFactor = 0.0244140625
+        baseModuleCost = 0
+        diskSpaceMassFactor = 0.0000048829
+        baseModuleMass = 0
     }
 
 If you add a section like that to the part.cfg, via directly editing it,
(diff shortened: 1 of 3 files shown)


src/kOS/Utilities/PartUtilities.cs
```diff
@@ -1,12 +1,23 @@
...
+using System;
 
 namespace kOS.Utilities
 {
     public static class PartUtilities
     {
         public static float CalculateCurrentMass(this Part part)
         {
-            return part.HasPhysics() ? part.mass + part.GetResourceMass() : 0;
+            if (part.HasPhysics())
```
tomekpiotrowski: We could substitute it with `part.HasPhysics() && part.rb != null ? part.rt.mass : 0` if I'm not mistaken.

doc/source/general/kospartmodule.rst
```diff
@@ -105,4 +112,64 @@ by a re-balancing mod by changing this value.  This value is
...
+--------------------
+
+   - **Type:** float
+   - **Default if omitted:** 0.0244140625
+   - **Effect:** How much additional cost is incurred per
+     byte of disk space added via the editor tweakable.
+
+When using the editor tweakable to increase storage, cost is added to the
+module.  That additional cost is found by multiplying the number of additional
+bytes by this factor.The default value is balanced for approximately 100
```
tomekpiotrowski: Missing space (or two :P)

Issue #1580, opened 2016-04-03
"volume out of range" error confusing, doesn't specify RemoteTech


Gaiiden:
I had my script show me this error:

>Volume is out of range
>(Cannot Show kOS Error Location - error might really be internal. See kOS devs.)

Which originally had me thinking it was some internal issue thanks to the second line. Took me a few tries at debugging to realize it was because I had lost my RemoteTech signal over the horizon while I was logging data to the Archive volume. I think if only the first line had popped up alone I still would have been slightly confused at first. Would be better if stated:

"Connection to KSC lost, volume is no longer accessible"

hvacengi:
The `(Cannot Show kOS Error Location - error might really be internal. See kOS devs.)` portion is related to the stack trace, rather than the error itself, which is probably more misleading than the message itself.

If the message was:
> Volume is out of range
> Error Trace:
>    (Cannot Show kOS Error Location - error might really be internal. See kOS devs.)

Would that have been more clear?

We can also update the exception text, but with the future communication update from Squad (not in version 1.1) we may be updating the text further.  My ideal message would actually be "Volume is out of communication range." because it is short, while still giving you the information that it is communication related.

Dunbaratu:
Maybe the message about the stack trace can be changed a bit.  Basically it comes up whenever the problem can't be traced to a line of your script so we can't show you where it came from.  It *assumes* this must be because it's an internal error that's not your fault, but maybe that's not always true.

Gaiiden:
I agree it would be better off being more generic in saying "communication", good call

PR #1648, opened 2016-05-12
Check volume range in ProcessElectricity

### LINKED ISSUE #1363: Start on archive yields "power starved" when powered up with no KSC connection.

hvacengi:
Fixes #1363

kOSProcessor.cs
* Add a check to the volume range before accessing the current volume in
Process electricity.
* This allows "start on archive" to properly work with RemoteTech, and
will simply skip the boot sequence if out of range.
* Add Archive property for storing the archive volume after creation.
* Log exception when boot volume is out of range (so that it is
displayed on the terminal too).
* Change CheckCanBoot logger calls to be shared.Logger, so that the
message displays on the terminal.

KOSVolumeOutOfRangeException
* Create new exception class to handle "volume out of range" exceptions.

RemoteTechVolumeManager.cs
* Use the new exception type.

kOS.Safe.csproj
* Add reference for new exception.

KSPLogger.cs
* Add null checks before attempting to create the code fragment message.
If called from the SafeHouse instance, Shared may be null.
* Remove call to base.Log(text) in Log(string text).  This forced the
logger to attempt to display the text on the screen as well as in the
log.  But the screen printing feature never got used, because the method
was only called on the SafeHouse instance (which has no Shared object
and therefore no Screen).

Logger.cs
* Add null checking to audible exceptions, because a call to the
SafeHouse instance may have a null Shared object.

### DIFF
--- a/src/kOS/KSPLogger.cs
+++ b/src/kOS/KSPLogger.cs
@@ -23,7 +23,6 @@ namespace kOS
 
         public override void Log(string text)
         {
-            base.Log(text);
             UnityEngine.Debug.Log(string.Format("{0} {1}", LOGGER_PREFIX, text));
         }
 
@@ -64,17 +63,20 @@ namespace kOS
             // print the call stack
             UnityEngine.Debug.Log(e);
             
-            // print a fragment of the code where the exception ocurred
-            int logContextLines = 16;
-            #if DEBUG
-            logContextLines = 999999; // in debug mode let's just dump everything because it's easier that way.
-            #endif
-            List<string> codeFragment = Shared.Cpu.GetCodeFragment(logContextLines);
-            var messageBuilder = new StringBuilder();
-            messageBuilder.AppendLine("Code Fragment");
-            foreach (string instruction in codeFragment)
-                messageBuilder.AppendLine(instruction);
-            UnityEngine.Debug.Log(messageBuilder.ToString());
+            if (Shared != null && Shared.Cpu != null)
+            {
+                // print a fragment of the code where the exception ocurred
+                int logContextLines = 16;
+#if DEBUG
+                logContextLines = 999999; // in debug mode let's just dump everything because it's easier that way.
+#endif
+                List<string> codeFragment = Shared.Cpu.GetCodeFragment(logContextLines);
+                var messageBuilder = new StringBuilder();
+                messageBuilder.AppendLine("Code Fragment");
+                foreach (string instruction in codeFragment)
+                    messageBuilder.AppendLine(instruction);
+                UnityEngine.Debug.Log(messageBuilder.ToString());
+            }
         }
         
         /// <summary>
--- a/src/kOS/Logger.cs
+++ b/src/kOS/Logger.cs
@@ -44,11 +44,8 @@ namespace kOS
 
                 message += "\n" + LINE_RULE + "           VERBOSE DESCRIPTION\n";
                 
-                message += ((KOSException)e).VerboseMessage;
+                message += ((KOSException)e).VerboseMessage + "\n";
                 
-                // Fallback if there was no verbose message defined:
-                if (message == String.Empty)
-                    message += e.Message;
                 message += LINE_RULE;
                 
                 // Take on the URL if there is one:
--- a/src/kOS/Logger.cs
+++ b/src/kOS/Logger.cs
@@ -58,7 +58,7 @@ namespace kOS
                 message += LINE_RULE;
             }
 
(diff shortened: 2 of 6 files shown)


src/kOS/KSPLogger.cs
```diff
@@ -23,7 +23,6 @@ public KSPLogger()
 
         public override void Log(string text)
         {
-            base.Log(text);
```
hvacengi: This forced the logger to attempt to display the text on the screen as well as in the log. But the screen printing feature never got used, because the method was only called on the `SafeHouse` instance (which has no Shared object and therefore no Screen).  I did a "find all references" to be sure.  Since it isn't being used, I removed the reference to avoid confusion regarding the differing behavior.

src/kOS/Module/kOSProcessor.cs
```diff
@@ -343,8 +345,8 @@ public void InitObjects()
             shared.Window.AttachTo(shared);
 
             // initialize archive
-            var archive = shared.Factory.CreateArchive();
-            shared.VolumeMgr.Add(archive);
+            Archive = shared.Factory.CreateArchive();
```
hvacengi: Storing the `Archive` volume allows us to bypass the range check in "getting" the archive when switching to it.  While slightly hacky, I didn't want to make a major change to the system with a pending sub-directories PR out there, and the alternative was to add a new method for returning the volume without a range check, or checking the range without the volume object itself.

src/kOS/Module/kOSProcessor.cs
```diff
@@ -817,10 +826,10 @@ public string BootFilename
...
         public bool CheckCanBoot()
         {
-            if (shared.VolumeMgr == null) { SafeHouse.Logger.Log("No volume mgr"); }
-            else if (!shared.VolumeMgr.CheckCurrentVolumeRange(shared.Vessel)) { SafeHouse.Logger.Log("Boot volume not in range"); }
-            else if (shared.VolumeMgr.CurrentVolume == null) { SafeHouse.Logger.Log("No current volume"); }
-            else if (shared.ScriptHandler == null) { SafeHouse.Logger.Log("No script handler"); }
+            if (shared.VolumeMgr == null) { shared.Logger.Log("No volume mgr"); }
+            else if (!shared.VolumeMgr.CheckCurrentVolumeRange(shared.Vessel)) { shared.Logger.Log(new Safe.Exceptions.KOSVolumeOutOfRangeException("Boot")); }
+            else if (shared.VolumeMgr.CurrentVolume == null) { shared.Logger.Log("No current volume"); }
+            else if (shared.ScriptHandler == null) { shared.Logger.Log("No script handler"); }
```
hvacengi: using `shared.Logger` instead of `SafeHouse`.  Ideally I'd like to see all of these turned into `KOSYouShouldNeverSeeThisException` exceptions because it will allow us to show an error on the terminal, instead of needing to pull out the debug menu or output log.

src/kOS/Logger.cs
```diff
@@ -44,11 +44,8 @@ public virtual void Log(Exception e)
 
                 message += "\n" + LINE_RULE + "           VERBOSE DESCRIPTION\n";
                 
-                message += ((KOSException)e).VerboseMessage;
+                message += ((KOSException)e).VerboseMessage + "\n";
                 
-                // Fallback if there was no verbose message defined:
-                if (message == String.Empty)
-                    message += e.Message;
```
hvacengi: This condition was literally impossible.  `message` was already set to `e.Message`, so the only way it is empty is if `e.Message` is empty, which doesn't do us any good at all.

tomekpiotrowski:
It seems to fix #1580 as well, I've added it to the OP.

PR #1657, opened 2016-05-28
New run command that takes variable filename args (not hardcoded at compile time)

### LINKED ISSUE #1213: Compiler treats "run progname" and "run progname.ks" as if they were two unique programs.
If you do this:
```
run progname.
```
and then later do this in the same program context run:
```
run progname.ks.
```
The current design fails to realize that those are two different alias names for really the same file.  It determines whether or not it's a unique program entirely from the name that existed inside the script, like so, in Compiler.cs, PreProcessRunStatement():
```
                    string subprogramName = node.Nodes[1].Token.Text;
```
So it treats them like independent programs,  building two different copies compiled into memory, and possibly this can cause confusion later on down the road.  Right now it *works*, by making it so the second instance sort of masks the first instance when it comes to duplicate function names, but probably needs a refactor to go along with better processing of file handles.

### LINKED ISSUE #1257: A way to allow a RUN statement to work on runtime-changing instructions.

Dunbaratu:
Fixes #1213 
Fixes #1257 

### Synopsis:
```
// Run command now works with filenames that vary at runtime, like so:
set progname to "0:/some_folder/some_prog.ks". // can use full paths
run progname.
cd("0:/some_folder").
set progname to "some_prog". // .ks is still optional.
run progname.
run "some_prog".   // literal string.  .ks suffix is still optional.
run some_prog. // bare-word still works.

// Note, the system recognizes that the above are 
// all the same file, really.  (Even though they are
// different means of getting to it), and thus the 
// "compile only once" logic means only the first
// one actually performs the compile (if these
// occurred inside a parent program, that is.  If these
// happened in the interpreter, they'd all be compiled
// each time, of course.)
```

### Overview of design of the change:

Previously: If you did a ``run prog1.`` and a ``run prog2.`` and a ``run prog3.`` in the program, it made 3 different hardcoded "compile if needed then run it" functions up in the header of the program - one for each different filename.  That's why the filenames had to be hardcoded at runtime.

The Changes.  To the code-reviewer:  Read this section first before looking at the code and it should help make the code changes make more sense:

- There is now one boilerplate hardcoded kRisc function to be called by all instances of ``run`` for all subprograms, and the file to be run is passed to it as an argument on the stack.
- The logic for "don't compile if already compiled" has been moved into the ``load()`` built-in function instead, because first we want to do the work to find which file it's talking about.  Having to pay attention to current working directory, following the directory path in the filename, etc, to obtain the fully-qualified path name for the file, etc, would be a PITA to do from the kRISC boilerplate (possible, by calling suffix methods in the subdirs PR, but a pain).  The ``load()`` built-in function already does all that now after the subdirs PR got merged, so it makes more sense for it to handle the "already compiled" check from there.
- Because of the above, the kRISC boilerplate function just calls ``load()`` unconditionally now, letting ``load()`` decide to abort early if the compile is already done.
- Because ``load()`` is also called from places where it SHOULD compile unconditionally regardless of whether the file exists already (i.e. from the ``COMPILE fname1 TO fname2`` command), it now has an additional argument - a boolean flag telling it whether or not you want it to use its abort-if-already-compiled logic.
- Because of the ``run once`` command needing to know whether or not the program was already compiled so it can know whether or not it should run the program or ignore it, ``load()`` also communicates a boolean flag back to the caller letting it know whether or not it actually did any work or if it just aborted early because it's already compiled.  Note, the ``load()`` function always weirdly had two return values on the stack, but the second one was a dummy zero before.  Now it is no longer a dummy and it actually carries this information.

### ~~"Needs documentation" tag because:~~

- ~~I still need to edit the docs to remove all the places where it claims you have to hardcode the run program name.  Other than that there shouldn't be any real documentation changes.  It's more about *removing* claims that are no longer true, than about adding new docs.~~


hvacengi:
Edited out the documentation requirement because I moved it to #1677

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -1042,6 +960,81 @@ private void TraverseScopeBranch(ParseNode node)
...
+
+            // If load() didn't claim the program was already loaded, or if we aren't operating
+            // in "once" mode, then jump to the part where we call the loaded program, else
+            // fall through to a dummy do-nothing return for the "run once, but it already ran" case:
+            Opcode branchFromOne = new OpcodeBranchIfFalse();
+            AddOpcode(branchFromOne);
+            AddOpcode(new OpcodePush("$runonce"));
+            Opcode branchFromTwo = new OpcodeBranchIfFalse();
+            AddOpcode(branchFromTwo);
+            AddOpcode(new OpcodePush(0));   // ---+-- The dummy do-nothing return.
```
hvacengi: At this point, the "already loaded" flag has been popped, but the "entry point" is still on the stack.  So if "already loaded" is true and "$runonce" is true, we push "0" and return, won't that leave "entry point" on the stack?

Dunbaratu: yup that's right.  good catch.

Dunbaratu: By the way, when you look for where the edit to fix this is in my next commit, you'll have to look a different file.  This is being moved into ProgramBuilder.cs.

hvacengi: Probably a better place for it given the use anyways.  It isn't really part of the compilation (though it makes sense to keep it there if it continues to need to be embedded.

doc/make.bat
```diff
@@ -1,242 +1,242 @@
...
+
+if "%1" == "pseudoxml" (
+    %SPHINXBUILD% -b pseudoxml %ALLSPHINXOPTS% %BUILDDIR%/pseudoxml
+    if errorlevel 1 exit /b 1
+    echo.
+    echo.Build finished. The pseudo-XML files are in %BUILDDIR%/pseudoxml.
+    goto end
+)
+
+:end
```
Dunbaratu: I don't understand how all these make changes got into my PR.  They've got nothing to do with anything I edited (not manually anyway).  I wonder if there's a CR/LF line ending issue that flagged the whole file as changed.

hvacengi: There is something funny with this file.  I keep my repository in dropbox and whenever I switch from work to home or vice versa I show make.bat as updated until I use:
```
git rm --cached -r .
git reset --hard
```
I've inspected the files and they are literally identical at that point, even line endings.  I just got into the habit of doing my adds as:
```
git add .
git reset *.bat
```

PR #1662, opened 2016-06-01
Issue #1661 Force program abort

### LINKED ISSUE #1661: Abort program on flight ending suffixes
Some new suffixes on the `kuniverse` structure don't abort the program after being called.  These methods will continue executing program code until the end of the physics tick, More likely than not, the user won't include code after calling these suffixes, since you don't expect to continue the program after calling them.

These methods include

* revert
* loading a quicksave
* launching a new vessel.

I have prepared a pull request and it will be submitted momentarily.

hvacengi:
Fixes #1661

ICpu.cs
* Expose "GetCurrentOpcode" on the ICpu interface

FakeCpu.cs
* Add "GetCurrentOpcode" (throws not implemented exception)

Misc.cs
* Remove casting ICpu to CPU.

KuniverseValue.cs
* Abort the current program when reverting the game
* Abort the current program when launching a new craft (because it will kill the program anyways)
* Abort the program when loading a quick save.

### DIFF
--- a/src/kOS.Safe.Test/Opcode/FakeCpu.cs
+++ b/src/kOS.Safe.Test/Opcode/FakeCpu.cs
@@ -249,5 +249,10 @@ namespace kOS.Safe.Test.Opcode
         {
             throw new NotImplementedException();
         }
+
+        public Compilation.Opcode GetCurrentOpcode()
+        {
+            throw new NotImplementedException();
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


Dunbaratu:
Is there any documentation to go with this?  It may seem obvious to us, but it is probably worth mentioning in the kOS docs for these commands.  Something along the lines of "When you do this, you stop the current program right here and that means if you've written any code after this point, It won't be reached."

PR #1667, opened 2016-06-07
Issue 103 External API

### LINKED ISSUE #103: Allow for external calls to execute code
We have had a request for an API to invoke a script from another module. For the first version of this feature I am going to propose a very small subset of features.

Limits: 
- One kOS module per craft
- Only the ability to invoke a script by name

The API might only have a few calls
bool:IsActive(<Vessel.Id>)
bool:ExecuteProgram(<Vessel.Id>,<FileName>) 

This might not work for everything we want to do long term but I believe getting something very simple working and going from there is a prudent course.

As far as packaging this up for consumption, Would you rather We
- Create a very small API dll you can reference and the kOS team updates?
- Distribute a .cs file with all of the guts in it for you to include in your project?
- publish the API as a doc and let you get to it however you like?

hvacengi:
Related to #103

This provides an avenue for other KSP mods to provide addon support.  The biggest "feature" is that kOS now has the ability to walk all of the loaded assemblies, and create a list of types to register with "manager" classes.  The existing Binding and Function managers have been ported to use this system as well.

Essentially manger classes define an attribute, inherited class, or interface that identifies classes that they will manage.  Then the AssemblyWalkAttribute will run through every type that's currently loaded (but not in the GAC) and see if it meets any of the multiple managers' criteria.  If it does, it registers the type with that manager.

The AddonManager will register classes decorated with the `kOSAddon` attribute and which inherit from `Suffixed.Addon`.

Of note, the 3rd party assembly **must** include the following line inside of AssemblyInfo.cs to ensure the proper loading order of the assemblies:
```
[assembly: KSPAssemblyDependency("kOS", 0, 20)]
```

Features
- [x] Walk assemblies to find classes by attribute, inheritance, or interface
- [x] Manager classes register each type for use during main runtime.
- [x] AddonManager allows Addon's to be loaded from other assemblies
- [x] Error messages in the log if a class is decorated by does not implement a feature as expected (nag messages)
- [x] Update kOS nomenclature attributes to work with the additional assemblies (adapt to use `AssembliyWalkAttribute`)
- [ ] Sample 3rd party addon to use as an example
- [x] Documentation for 3rd parties to use


src/kOS/AddOns/Addon Readme.md
```diff
@@ -0,0 +1,40 @@
...
+  is the string that will be used to access the Addon as `addons:identifier`.
+* Your Addon should override the abstract `Available` method, returning true
+  only when the addon's core functionality is available.  You might choose to
+  return false if another required mod is not installed, or if a required part
+  is not on the current vessel.
+* You may add additional kOS Structures, and then expose them using suffixes
+  on your central Addon class.
+* All new structures, including your Addon, must be decorated with a
+  `[KOSNomenclature]` attribute, identifying it's "kOS Name" (the value returned
+  by the `typename` suffix).
```
Dunbaratu: The document should probably mention the parameters to KOSNomenclature and not make it look like it's a bare paremeter-less attribute.   (i.e. "must be decorated with a `[KOSNomenclature("put the kos name here")]` attribute.....")

src/kOS.Safe/Utilities/AssemblyWalkAttribute.cs
```diff
@@ -0,0 +1,302 @@
...
+                var parameters = managerRegisterMethod.GetParameters();
+                if (parameters.Length != parameterTypes.Length)
+                {
+                    return false;
+                }
+                // We don't need to check the parameter types if there are no parameters.
+                if (parameters.Length == 0)
+                {
+                    return true;
+                }
```
Dunbaratu: There is no need for the above check whatsoever.  If parameters.Length == 0, then the loop below will execute zero iterations and fall through to the `return true;`.

Dunbaratu: As a rule of thumb, I don't like exception cases that say "avoid the loop when collection is empty", if the loop is already written as a "do the loop once per element of the collection".  "avoid it when it's empty" is sort of already built-in in that case.

src/kOS/AddOns/AddonManager.cs
```diff
@@ -0,0 +1,41 @@
...
+        private static readonly Dictionary<kOSAddonAttribute, Type> rawAttributes = new Dictionary<kOSAddonAttribute, Type>();
+        private readonly SharedObjects shared;
+
+        public AddonManager(SharedObjects sharedObj)
+        {
+            AllAddons = new Dictionary<string, Suffixed.Addon>(StringComparer.OrdinalIgnoreCase);
+            foreach (kOSAddonAttribute attr in rawAttributes.Keys)
+            {
+                AllAddons.Add(attr.Identifier, (Suffixed.Addon)Activator.CreateInstance(rawAttributes[attr], sharedObj));
+            }
```
Dunbaratu: As far as I can tell, `rawAttributes` is always still an empty collection during the constructor here, and isn't populated until later, so out of curiosity, under what circumstances is it possible for the above loop to ever execute any iterations?

src/kOS/AddOns/Addon Readme.md
hvacengi: Absolutely correct.  If for no other reason than consistency since that's what I did with kOSAddon above.

src/kOS.Safe/Utilities/AssemblyWalkAttribute.cs
hvacengi: Yeah... I'm not sure why I did that.  I don't normally do this either.  Probably started out with something more complicated and then shaved it down to this code but never removed the extra check.

src/kOS/AddOns/AddonManager.cs
hvacengi: `AddonManager` is not instantiated until the `kOSProcessor`'s `InitializeObjects` method is called.  `rawAttributes` is filled by `RegisterMethod` during the initial assembly walk at the very beginning when KSP launches.  So the order goes:

Dunbaratu: It's not that I didn't understand the order of things.  It's that I didn't notice that it was static (a prerequisite for it being possible for it to be populated prior to the constructor getting called.)

So this seems strange to me.  If you have two different instances of `AddonManager`, then they both share the same static instance of `rawAttributes` and both try to run this code.

If the intention was that there must only ever be one instance of AddonManager (so that scenario of having two of them I described doesn't happen), then it looks like an odd archetecture to store half the one-instance information in static members and store the other half of the one-instance information in dynamic members.

Dunbaratu: Is the intention that there be one instance of `AddonManager` per instance of `shared`?  Such that each contains basically the same list of `AllAddons`, but with the only difference being the member of `AllAddons` have been given different instances of `shared` to operate on?

Dunbaratu:
Should issue #103 get auto-closed when this gets merged?  It only sort of resolves half of what 103 was talking about.  (The half mentioned later in comments that wasn't even part of the original issue - that of having kOS scripts invoke addon code.  The original issue description talks about going the other way around - having addons invoke kOS scripts.  Not that I think that's a very good idea to implement, but it is what issue #103 was asking for.  I think I'd rather close issue #103 as a "wont fix" rather than claiming it got fixed by this PR, since this PR doesn't quite really do what 103 was asking for, and is instead doing something else orthogonal to 103.

Dunbaratu:
I'd be ready to merge this now, depending on that last question about issue 103, and the fact that we don't have a good example case of using it yet (how could we - it has to get merged to develop before other people start using it).  Code-wise it looks fine.

hvacengi:
Hm, I thought I posted a comment to this last night, but apparently didn't.  A re-read of #103 indeed makes it look like this is less related to that precise issue.  Telnet really is a closer solution to that problem.  Though using this new system, a mod will have access to the cpu the same way that the processor and the terminal do, so it would be possible for them to execute arbitrary code on the processor in much the same way.  When this and your `run` PR are both merged, we should look at whether or not there is a direct public route that the addon would be able to call a method like `shared.CPU.RunCode`

src/kOS/AddOns/AddonManager.cs
hvacengi: Yes that is correct, there is one instance of AddonManager per instance of shared, only to pass a different value of shared to each addon instance.

Issue #1678, opened 2016-06-23
Auto-migrate boot scripts


hvacengi:
When the Scripts directory was moved into the global "ships" directory and the extension was changed from .txt to .ks, we included a small utility that would automatically move existing scripts and change the file extensions.  Now that boot files are located in a separate folder, we should offer to move and strip the "boot" prefix.

_When implementing, the file "boot.ks" should not be renamed, it should just be moved._

dewiniaid:
Would it make more sense to allow both old-style and new-style boot scripts to exist, deprecating the old style with a plan to remove it in the next major release or two?

I say this only because of the potential havoc this may raise if someone is directly running a boot script from another script -- or they have their entire script folder under source control...

hvacengi:
I think that I've reached something of a compromise position after reviewing comments to PR #1686.  I don't think that we will be supporting selection of boot files from locations other than the boot directory in the editor, but they will still be valid from within scripts.  I've updated the code so that vessels already in flight will still use the existing root location as defined by the boot script currently.

Existing craft files will be a little smarter.  They will look to see if the existing saved file is located in the archive root.  If it doesn't find the file there, it will look for a file by the same name within the boot directory.  If it still doesn't find it, and if the file name starts with "boot" or "boot_" it will strip the prefix off and look again.  If you are using a file that doesn't exist in the boot directory, selecting another file **and** saving the craft file will prevent you from selecting that file again with that same craft.

I sympathize with the issue of running boot scripts from another script.  This is a breaking change in that regard, and it will be a pain.  But given the large all encompassing nature of this release, and how many changes we have to file name and path handling, I feel like simply forcing the change is the correct choice.

With regard to the source control, it should be able to recognize a file as being moved if it has the same name.  We should be able to make renaming the files optional, so that you can select "yes move my files and rename", "yes move my files but don't rename", and "no, I'll deal with it myself".

hvacengi:
I'm also adding a note here to revise the documentation to address the boot file change, since it's a decent change that users need to have a clear understanding of.

PR #1686, opened 2016-06-28
Fix boot file volumes

### LINKED ISSUE #1684: Bootfiles are broken now
`kosProcessor.BootFilePath` always returns the path of the boot file on the archive, even after the vessel was launched.

hvacengi:
Fixes #1684

kOSProcessor.cs
* Update null check in InitObjects to not throw a null ref when the boot file path is null
* Update ProcessorModeChanged to set the bootDirectoryPath based on Config.StartOnArchive

CPU.cs
* Update guard against boot file path being null.


Dunbaratu:
I volunteer to be tester on this.

hvacengi:
Hm, I just saw that this has conflicts.  I'll bring it current for your testing purposes.

One thing that we may want to discuss before finalizing this PR: do we want the local boot files to be stored in `1:/` (as they are in the current release, and current develop) or in `1:/boot/` (which would more closely mirror the archive structure)

Dunbaratu:
I'd rather see the copied to the local drive in the same structure as in the archive.  It's messed up when the same program that works when run from archive doesn't work when set as the boot file because the copy we made locally got a different path.

i.e. let's say you have `0:/boot/myauncher.ks`, which also runs 0:/boot/bootstep2.ks` when it's done.  Well, should it call it `run "boot/boostep2.ks".` or should it call it `run "bootstep2.ks".`?  If the answer to that question changes depending on whether it's being tested from the archive or has been copied locally, it's a mess to work with (this is the way it is now and it's been giving me some problems).

I think it would make much more sense to preserve the same folder layout on the local copy as you are using on the archive copy - and that means when someone selects the "boot/mylaunch.ks" file from the VAB rightclick menu, then it should be called "boot/mylaunch.ks" when it arrives on the local drive.  (i.e. mkdir the boot/ on local if it's not already there, then copy mylaunch.ks. into it.)

hvacengi:
That shouldn't be too hard to do.  I can see if I can pull that together rapidly, but that probably requires another pass over the docs to make sure it's corrected as well.

Want me to try to do it here (since I'm already directly interacting with "boot path") or separate issue/PR?

Dunbaratu:
The more complex issue is, when someone does `set core:bootfilename to blarg.`, is blarg the string "myprog" or the string "boot/myprog"?  If we don't hardcode the path, they have the ability to pick any path they like once they reset the boot file to something else other than the one that it was loaded with.  If we presume boot/, then we're saying from then on all programs you want to boot with must be in boot/ also, even if they're not the one you load in the VAB.  (i.e. consider a script that picks a new bootfile at each step of a mission, so if you save and come back ,you get it to start at the right step of the mission.  Some of those programs it would run might not be ones you'd normally boot from on the launchpad.

Dunbaratu:
You can try to edit the dos for it if you like - it would help when I get around to testing it so I know what it was *meant* to do.

hvacengi:
By it's nature, "boot file name" means the file name itself, not the full path.  I rather like the idea of segregating boot scripts into their own folder, but I'm not sure I have a strong and valid argument for it.  We've already had a request to allow boot files to be valid under the old name based system (#1678).  I don't like the idea of allowing the old prefix based system in the VAB using our current UI.  I only say that because the selector truncates longer names, and the boot prefix itself already makes it harder to distinguish files, but having the boot prefix and the folder prefix listed would really complicate things.  As the setup stands right now, it is much easier to find boot files now that the prefix has been removed.  When we start creating UI's using Unity's new tools, I'm all for giving universal choice in the VAB.

However, we could fairly feasibly expose the existing "boot path" (as a volume path, assuming that we want to keep "start on the archive" as a thing) to the user and allow them to edit it via script even though we don't have a UI to do it.

If we want to have further in depth conversation on this, it probably should be moved to issue #1678 or a new issue.

Dunbaratu:
> By it's nature, "boot file name" means the file name itself, not the full path.

Only if you start from the premise that the term "file name" contains the assumption that it must only be the lastmost term in the name.  I never heard the word "file name" used in that narrow of a definition until I starting having to use Microsoft's libraries.  It's not a universal convention.  In the way I was always using it, if you drew a Venn diagram, filenames that contain the full path and ones that do not are both contained entirely inside the circle for the term "file name".

Dunbaratu:
I could accept the following compromise:  the boot file name contains the presumption that the current working directory for referring to it is `boot/`.  If you don't give the full path, you get a file in `boot/` by default, but you can still give the full path to force it to be somewhere else, or use `../` to get out of the `boot/` directory.

hvacengi:
An attempt to support both the current working directory and an explicitly specified directory throws a major wrench in the system when it comes to running the boot file.  At that point, we probably need to change from `CPU.Boot` compiling the command `run [bootfile].` to actually compiling the and running the boot file.  And at that point I'd rather we created a single method to be called from the run function (the one called when you type `run file.` into the terminal) and from the CPU's Boot method, so that we don't continue to run into issues where we have to remember to update the logic in multiple locations.  At that point, it probably should be a separate issue as it's a major revision and it will delay the next release.

I'd rather we took the time to do it right and delayed the release, than released a half assed solution with the promise to fix it in the next update.  Either that or we revert to the previous convention of using the "boot" prefix until we can properly work with changing paths.

hvacengi:
I may have thought of a way that I can make this work and keep most people happy.  Let me work on it tonight and see if I can make it work.

Dunbaratu:
I don't really understand the description of the problem, but I'll just wait and see what you have after the changes you said you're still making.

hvacengi:
@Dunbaratu This should be ready for testing now.  Please see my comments on the most recent commit (9ff6c873df5728ae71444ff5c438ef6d158a8a6f) for details of what I was trying to fix.

src/kOS/Module/kOSProcessor.cs
```diff
@@ -818,19 +871,26 @@ private void ProcessorModeChanged()
...
-                break;
+                case ProcessorModes.READY:
+                    if (SafeHouse.Config.StartOnArchive)
+                    {
+                        shared.VolumeMgr.SwitchTo(Archive);
+                    }
+                    else
+                    {
+                        shared.VolumeMgr.SwitchTo(HardDisk);
+                    }
```
Dunbaratu: Is this just the same thing it did before and you just have an aversion to the simpler trinary syntax?

Dunbaratu: Okay that sounded a bit snarky.  Sorry about that.  I just wanted to verify it was your intention to not actually change the algorithm meaningfully, because this looks like it's the same exact thing just expressed differently.

src/kOS/Module/kOSProcessor.cs
```diff
@@ -298,15 +296,71 @@ private void InitUI()
...
+                                {
+                                    // only need to strip "_" here
+                                    name = name.Substring(1);
+                                    path = VolumePath.FromString(BootDirectoryName).Combine(name);
+                                    file = archive.Open(path) as VolumeFile;  // try to open the new path
+                                }
+                            }
+                        }
+                    }
+                }
```
Dunbaratu: This looks much easier to follow now.  Thanks.

So, to verify, is this the intended behavior? :

If boot filename can be found, as is, starting from a path that begins in the ``boot/`` directory, use that one first.  (Note, the reason I say  "begins in" that directory rather than saying the file "is in" that directory, is to clarify the meaning if the filename is given as something like ``../myfile.ks``, what the meaning of the ``..`` should end up being.  For the sake of correct documentation, this difference could matter.  Saying it "is not allowed to" have path information in the name would be quite different from saying it "is not necessary to" have path information in the filename.)

Else if boot filename can be found from a path that begins in the archive directory, use that one.  (again, "begins from" having the same meaning as above.)

Else if the filename starts with ``boot_*`` (and thus contains no path information like ``../`` or ``0:/``), then try looking for a file in the ``boot/`` directory that would match the name if the ``boot_`` prefix was stripped off.

Else it has failed to find the file.

Is that the intended behavior?  Because that's what I think it does from the above code path, if I'm understanding it correctly.

src/kOS/Module/kOSProcessor.cs
hvacengi: I changed the logic in the first commit, when I had to [here](https://github.com/KSP-KOS/KOS/pull/1686/commits/517510c62b5bc6433ed4ee4a3a4785c9cd4aaafd#diff-7be643b4883b5a4b8146fc3687dccf1cR823).  When I changed the higher level logic to use a `VolumePath` instead of `GlobalPath` I no longer needed to set the volume of the `GlobalPath` in this method, and I simply forgot to simplify it back to a one line expression.  Feel free to revert it.

Dunbaratu: Okay thanks.  I just wanted to make sure I wasn't reading it wrong since it looked like a non-change.

src/kOS/Module/kOSProcessor.cs
hvacengi: That's mostly correct.  Technically, including `0:/` in the path will break the boot file, as the `VolumePath` will simply parse that as another `/` separated segment.  

And with regard to the `boot[_]` prefix, the path can actually contain any additional information you desire, it only pays attention to the name of the file excluding the path information.  As this code is only called from within the editor, the only way to get a file name that is not either in the archive's root, or the archive's boot directory is to manually hack your craft file.  So a user can set `core:bootfilename` to `/boot/../libraries/orbit/../../boot/boot.ks` and this code would not interfere at all.  Adding `1:/` or `0:/` or `archive:/` would break the boot script.  If we want to make it valid to specify the volume of the boot script, we probably want to more drastically change the meaning of "start on archive".

In answer to your question though, yes, the behavior as you describe above is the intention.

You bring up a couple of potential documentation issues that I had not considered, and thus did not document.  Honestly, I don't know what happens if you use the `..` up directory shortcut on a volume root.  But other than the limitation of not specifying the volume, it should behave the same as any other path.

hvacengi:
Hm... speaking of documentation changes, I didn't look to see if I needed to revise anything when I made boot scripts default to "None" instead of `boot/boot.ks`, I probably should look into that.

(I made the change because of complications trying catch that `boot/boot.ks` doesn't exist in the instance of no boot scripts existing at all)

PR #1690, opened 2016-06-29
prevent crash by InvalidOperationException

bssthu:
Modify `ToolbarTypes.getType()` in `ToolbarWrapper.cs` .

Otherwise, toolbar button may crash if there is a broken mod.

### DIFF
--- a/src/kOS/Screen/ToolbarWrapper.cs
+++ b/src/kOS/Screen/ToolbarWrapper.cs
@@ -726,9 +726,16 @@ namespace kOS.Screen {
 		}
 
 		internal static Type getType(string name) {
-			return AssemblyLoader.loadedAssemblies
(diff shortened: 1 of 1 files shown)


hvacengi:
I don't see an issue associated with this.  Can you provide more information regarding the circumstances that led you to this PR?

As this is the recommended wrapper class published by @blizzy78, it may be a better decision to submit the pull request against https://github.com/blizzy78/ksp_toolbar as they would be in a better place to evaluate the modification, and a note at the top of the class definition instructs us not to edit the implementation.

blizzy78:
> it may be a better decision to submit the pull request against https://github.com/blizzy78/ksp_toolbar

Indeed.

bssthu:
Looks like the EXC would appear if a mod forgot to change namespace in ToolbarWrapper.cs.

Issue #1665, opened 2016-06-05
obt:meananomalyatepoch is unusable without a way to get the orbit's epoch.


dewiniaid:
I originally had assumed that `orb:meananomalyatepoch` is based on what the mean anomaly was at time=0.  However, some testing quickly revealed that to be untrue and it seemed to be based on the current time instead.

It turns out that neither of these are true:
- Outside of warp, it's (at least very close) to the mean anomaly at the current time.  (Presumably because this is the epoch time).
- On rails, it seems to be the time that warp was entered/the vessel was put on rails.

This makes sense, since this is precisely the value that'd be needed in a save; mean anomaly corresponds to MNA in savegames but there's also an EPH ("Epoch") parameter.

That said, there doesn't appear to be any way to get the current epoch that meananomalyatepoch is referring to which renders the suffix unreliable at best, useless at worst.

_current_ mean anomaly thus requires

```
degrees = 180/pi
eccentric_anomaly = arccos((ecc+cos(trueanomaly))/(1+ecc*cos(trueanomaly)))
mean_anomaly = degrees * (eccentric_anomaly/degrees - ecc*sin(eccentric_anomaly))
```

hvacengi:
Honestly, we had previously made a conscious decision to leave mean anomaly as an exercise for the user.  At the time, we assumed that the epoch did not move as much as it appears to.  We have discussed adding an epoch suffix on the orbit structure numerous times, it simply has not been very high on the priority list.  It probably should get bumped though, as without a definition for epoch you are correct the value of `meananomalyatepoch` is useless.

Dunbaratu:
handled by PR #2047

PR #1696, opened 2016-07-04
#1665 clarify Orbit:MEANANOMALYATEPOCH docs

sshilovsky:
for the sake of other unaware rocket scientists

### DIFF
--- a/doc/source/structures/orbits/orbit.rst
+++ b/doc/source/structures/orbits/orbit.rst
@@ -197,10 +197,10 @@ Structure
     :type: :ref:`scalar <scalar>`
     :access: Get only
 
-    `mean anomaly`_  in degrees. Even though orbital parameters are
+    `mean anomaly`_  in degrees at orbit epoch. Even though orbital parameters are
     traditionally done in radians, in keeping with the kOS standard
     of making everything into degrees, they are given as degrees by
-    kOS.
+    kOS. Orbit epoch is a complex term, refer to `#1665<https://github.com/KSP-KOS/KOS/issues/1665>`_ for details. 
 
 
 .. attribute:: Orbit:TRANSITION


doc/source/structures/orbits/orbit.rst
```diff
@@ -197,10 +197,10 @@ Structure
     :type: :ref:`scalar <scalar>`
     :access: Get only
 
-    `mean anomaly`_  in degrees. Even though orbital parameters are
+    `mean anomaly`_  in degrees at orbit epoch. Even though orbital parameters are
     traditionally done in radians, in keeping with the kOS standard
     of making everything into degrees, they are given as degrees by
-    kOS.
+    kOS. Orbit epoch is a complex term, refer to `#1665<https://github.com/KSP-KOS/KOS/issues/1665>`_ for details.
```
Dunbaratu: I'm not sure I like the idea of pointing people at the github issue about the problem, since it will also contain confusing information about what the mod *used* to be like...

sshilovsky: I think it's just better than nothing. Might be reasonable to just notice that orbit epoch is not the same that the universe epoch. A link to KSP docs could be also helpful and better than to github issue.

I think something has to be done here. I'd lost a couple hours trying to catch a bug related to this.

sshilovsky: There is probably no reason to describe the concept of epoch right now, but let #1665 fix to address that.

This PR is just to save others efforts before #1665 is fixed.

Dunbaratu:
@sshilovsky : The problem with the notion of putting in a temporary change to the docs until such a time as the #1665 is fixed.. is that the docs get published officially at the same time as the full releases do.  So if we assume #1665 is in the next release, there won't be any window of time during which the edit you talked about would help.

The edit we make should be the edit that makes the docs look like what we want them to look like after issue #1665 gets finished and closed.  By the time the documentation change becomes public, that's what the situation will be.

@hvacengi : I'm stalled on #1665 because @dewiniaid last said he was still going to mess with it a bit more, and I didn't want to go any further on it until I heard more about that.  If he's not planning on doing it, then I may take over testing the problem and seeing what can be done about it.  I do think we need something to make "find current position on orbit" more sane.  At the moment the "mean anomaly at epoch" value is useless without knowing the epoch timestamp, which keeps on changing a LOT as the game progresses.  I haven't figured out yet when and why it changes, bit it changes a lot.  The game seems to very frequently recalc the Euclidian parameters for the orbit object, and each time it does so it seems to move the epoch again.

Dunbaratu:
Oh, I didn't see that #1665 was moved to v1.0.1 rather than v1.0.0.  In that case there might be a need for a temp fix in the docs?

dewiniaid:
@dunbaratu err, I'm not an actual kOS developer and am not working on #1665, just a highly opinionated and technical user.  Apologies if anything implied otherwise.

Dunbaratu:
I misspoke.  I mean the person working on #1660 - @lamont-granquist

Dunbaratu:
Added to Milestone v1.1.0 because the issue this is connected to is in milestone V1.1.0

hvacengi:
It looks like we added the `epoch` suffix in #2047 including documentation.  As such, I'd like to close this PR as the stop-gap documentation is no longer needed.

PR #1701, opened 2016-07-08
First draft of changelog - need to see it rendered.

Dunbaratu:
There's one TODO in here because I wanted to link new features to docs (not to PRs or issues - for bugs I link to github issues, but for new features I think it should be linked to the docs)... but one of these I couldn't find the docs for and I think they were never made?

The doc links point to the public site http://ksp-kos.github.io/KOS_DOC/.
Those links won't actually work yet because the docs haven't been published (that's the URL for what I expect them to be when they are).

If we want to do a pre-release, we could just search/replace "https?//ksp-kos.github.io/KOS_DOC/" with some temp site where the new docs are rendered during pre-release, and then replace it with the right place later when the pre-release becomes a real release.

(Note, the ``s?`` in the above search string: ``https?://....etc....`` - some of the time when I cut-n-pasted from my browser I got a https link, other times I got a http link, so the URLs are a bit of a mix of the two, thus that's a regular expression that will catch them all: h t t p (s-maybe) : // (etc).  If you do the search/replace I refer to here, be sure to do it with regular-expression mode).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ worked on since the last release.
 
 ### NEW FEATURES
 
-* **Subdirectories:** (http://ksp-kos.github.io/KOS_DOC/commands/files.html)
+* **Subdirectories:** (http://hvacengi.github.io/KOS/commands/files.html)
   You are now able to store subdirectories ("folders") in your volumes,
   both in the archive and in local volumes.  To accomodate the new feature
   new versions of the file manipulation commands had to be made (please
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,166 @@
...
+  recompile the KSM files when using a new release of kOS or results will
+  be unpredictable.
+* New Subdirectories ability has deprecated several filename commands such
+  as ``delete``, ``copy``, and ``rename``.  They will still work, but will
+  complain with a message every time you use them, as we may be removing
+  them eventually.  The new commands ``deletepath``, ``copypath``, and
+  ``movepath`` described below are meant to replace them.
+* When using a RemoteTech antenna that requires directional aiming,
+  in the past you could aim it at mission control with
+  ``SETFIELD("target", "Mission Control")`` and now you have to
```
tomekpiotrowski: That should actually be reversed, it used to be `mission-control`, now it's `Mission Control`. That's because `mission-control` was kind of a reserved keyword before, now we support multiple ground stations so we have to reference them by name.

Dunbaratu: Okay, I swapped them - thanks for looking it over and finding the mistake.

Dunbaratu:
Didn't this content already get merged in?  Is there a reason this is showing up as an open PR?

PR #1723, opened 2016-07-13
Fixes #1713 anon functions implicit returns.

### LINKED ISSUE #1713: Empty anonymous user function causes stack overflow
I have been playing around with the anonymous user functions in the 1.0.0 pre-release. Absolutely love that feature!

I tried creating an empty function with `{}`. Calling that function results in a stack overflow. `{ return. }` works. I would expect both forms to be equivalent.

Here's a test script:
````
print "Variant 1:".
function foo {
}
local f is foo@.
print f().

print "Variant 2:".
local g is { return. }.
print g().

print "Variant 3 -- this does not work:".
local h is {}.
print h().

````

Output:
````
run test.
Variant 1:
0
Variant 2:
0
Variant 3 -- this does not work:
Stack overflow!!
At 1:/test, line 13
print h().
````

### PR DESCRIPTION
**Pull request by Dunbaratu**
Now anonymous functions follow the same logic as is
used in named functions, for handling the implicit
returns when an explicit return is missing.

Note, this system is a bit wonky at the moment, but
does work (I'm imitating what happens with named functions,
which is a bit in need of a refactor, as I described
in separate issue #1722.)

Fixes #1713

### DIFF
--- a/src/kOS.Safe/Compilation/KS/Compiler.cs
+++ b/src/kOS.Safe/Compilation/KS/Compiler.cs
@@ -1267,10 +1267,18 @@ namespace kOS.Safe.Compilation.KS
                 Opcode skipPastFunctionBody = AddOpcode(new OpcodeBranchJump());
                 string functionStartLabel = GetNextLabel(false);
                 
+                needImplicitReturn = true;
                 nextBraceIsFunction = true;
                 VisitNode(node.Nodes[0]); // the braces of the anonymous function and its contents get compiled in-line here.
                 nextBraceIsFunction = false;
-
+                if (needImplicitReturn)
+                    // needImplicitReturn is unconditionally true here, but it's being used anyway so we'll find this block
+                    // of code later when we search for "all the places using needImplicitReturn" and perform a refactor
+                    // of the logic for adding implicit returns.
+                {
+                    AddOpcode(new OpcodePush(0)); // Functions must push a dummy return val when making implicit returns. Locks already leave an expr atop the stack.
+                    AddOpcode(new OpcodeReturn(0));
+                }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/kOS.Safe/Compilation/KS/Compiler.cs [hunk 1]
```diff
@@ -1267,10 +1267,18 @@ private void VisitExpr(ParseNode node)
...
                 
+                needImplicitReturn = true;
                 nextBraceIsFunction = true;
                 VisitNode(node.Nodes[0]); // the braces of the anonymous function and its contents get compiled in-line here.
                 nextBraceIsFunction = false;
-
+                if (needImplicitReturn)
+                    // needImplicitReturn is unconditionally true here, but it's being used anyway so we'll find this block
+                    // of code later when we search for "all the places using needImplicitReturn" and perform a refactor
+                    // of the logic for adding implicit returns.
```
hvacengi: Just to confirm, this essentially means that if a function already has a return, this currently just serves as a safety net return statement.  Kind of like this:
```
set h to {
    return 1 + 1. // normal return
    return 0. // safety net that never gets called
}
```

Dunbaratu: Yes.  The reason for the check "needImplicitReturn" is that my original plan was to detect the lack of a return statement and only do this when it's missing.  But then I realized it would take a lot more work than I'd originally anticipated to do that check properly (it has to walk the path of all the if/else conditions and loops and see if there's a case where it is missing even though it is present in other cases).  At the time I wanted to get functions working at all, so I hardcoded it to true for the time being, with the idea it could be revisited later on if we care enough to spend that effort.

My reason for still including it here even though I know it's hardcoded to true here too, is so that if we ever do decide to make this smarter in the future, this bit of code will stick out in a search as being obviously the same sort of thing and therefore get noticed as needing to be changed as well as the compiler's main function code.

Issue #1732, opened 2016-07-16
.ksm file optimization (experiments and results)


chippydip:
Sorry for the long post, I wanted to brain dump this stuff in case anyone else is interested.
#1712 got me thinking about .ksm files in general and why they are so gosh-darn big. In my experience they are always larger than a minified text version, and often bigger than the text version with whitespace and comments. This has always seemed ridiculous to me so I decided to start investigating a couple days ago and I now have numbers and suggestions!

All my testing so far as been done with [boot.ks](http://pastebin.com/82UqzPXk) which I've heavily hand-optimized in the past. This may not be totally representative of scripts in general, but being able to compete with a hand-optimized and minified version seems like as good a place as any to start:
- as-is the file is 2403 bytes
- minified it's 1014 bytes
- compiled with KOS 1.0 it's 2593 bytes (yup, bigger than the source file)

I started by writing a disassembler for ksm files so I could see what was going on under the hood. The output below is from the summary section of this tool (I also dump the full symbol table and instruction listings).
## Starting Point

```
File Size: 2593 bytes

Arguments: 147 in 922 bytes (6.3 avg)
  Strings: 818 bytes
  Label: 26 in 182 bytes
  Magic: Args 33 Ops 33

Code: 365 ops w/ 324 args in 1013 bytes
  PUSH                      x150       450
  CALL                      x31        155
  PUSHSCOPE                 x14         70
  LABELRESET                x19         57
  RETURN                    x17         51
  PUSHDELEGATERELOCATELATER x8          40
  POPSCOPE                  x13         39
  BRANCHFALSE               x11         33

Debug: 600 bytes
```
## First Pass

The debug section takes a ton of space. It's like bundling a .pdb and functionally isn't required (this is no different than minimizing a text file to be all on one line). I also noticed that there were 33 separate copies of the "magic" stack marking value in the arguments table. Apparently `KOSArgMarkerType` doesn't have an Equals/GetHashCode override so a new value is added for pretty much every function call in the code.
- remove debug info: 590 bytes
- fix duplicate stack markers: 32 bytes

```
File Size: 1971 bytes

Arguments: 115 in 890 bytes (7.7 avg)
  Strings: 818 bytes
  Label: 26 in 182 bytes
  Magic: Args 1 Ops 33

Code: 365 ops w/ 324 args in 1013 bytes
  PUSH                      x150       450
  CALL                      x31        155
  PUSHSCOPE                 x14         70
  LABELRESET                x19         57
[...]

PR #1733, opened 2016-07-16
.ksm file optimizations

chippydip:
Simple fixes from https://github.com/KSP-KOS/KOS/issues/1732
- Added Equals and GetHashCode to KOSArgMarkerType so this will only get added to the arguments dictionary once (just like PseudoNull).
- Switched argument indexing from using byte offsets to use ordinals instead so the section can be larger before switching to 2 bytes per opcode argument reference.
- Reset previousLabel after generating argument offsets and before generating the actual code.

### DIFF
--- a/src/kOS.Safe/Compilation/CompiledObject.cs
+++ b/src/kOS.Safe/Compilation/CompiledObject.cs
@@ -556,7 +556,6 @@ namespace kOS.Safe.Compilation
             bool sectionEnded = false;
             while (reader.BaseStream.Position < reader.BaseStream.Length && !(sectionEnded))
             {
-                int offsetLocation = (int)(reader.BaseStream.Position) - startPos;
                 
                 byte argTypeId = reader.ReadByte();
                 Type argCSharpType;


Dunbaratu:
I've been reading through some old PR's to see if any can be merged in, or if they are hopelessly too old.
Your PR looks like it's still merge-able, @chippydip.

I missed it before because it came at a time when we had a lot of things to get out and it would have increased testing burden to suddenly break all compiled KSM's right then.  After putting it off, I never got back to it (sorry).

I've looked over the changes and all of them look good, but they would also require some changes to this file too:

https://github.com/KSP-KOS/KOS/blob/develop/src/kOS.Safe/Compilation/CompiledObject-doc.md

(It describes the .ksm format in detail, and your PR would change how arguments are stored, which would make some of what that document says become wrong (i.e. the arg being 20 meaning it's the 20th argument in the pack, rather than the argument at byte position 20).)

I like these changes, and if you're still around (I know this is old), would you consider adding edits to that file as part of this PR so they can be accepted as an atomic unit along with the code changes?

chippydip:
Sure thing! I'll see if I can carve out some time in the next couple days to make the doc changes.

PR #1739, opened 2016-07-18
Add boot script migration

### LINKED ISSUE #1678: Auto-migrate boot scripts

hvacengi:
Fixes #1678

Bootstrapper.cs
* Re-enabled the legacy script folder migration popup
* Add popup window for migrating boot scripts
* Options for migrating with the same names, or with stripping "boot"
* Default option to copy the files rather than move them

volumes.rst
* Document migration to the new boot directory and explain the only broken use case (boot files starting on the archive).

### DIFF
--- a/doc/source/general/volumes.rst
+++ b/doc/source/general/volumes.rst
@@ -185,7 +185,7 @@ migration.
         the archive root, and finally it will check the boot directory again
         after stripping :code:`boot` or :code:`boot_` from the beginning of the name.
         Vessels in flight will continue to work with the existing structure, so
-        long as :attr:`CONFIG:ARCH` is set to false.  If:attr:`CONFIG:ARCH` is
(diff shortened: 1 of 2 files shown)


Dunbaratu:
What would be pretty awesome would be if the dialog box actually showed you a list of the filenames it's talking about moving.  (put them under the buttons, so the buttons never get pushed down past the bottom edge of the window)

Dunbaratu:
on second thought, nah.  it's a one-off dialog box.  Not worth putting the time into it to do all that,.

hvacengi:
We could do that.  Or we could offer an option to pop open a window listing all the files.

PR #1744, opened 2016-07-19
Swap MoI components

### LINKED ISSUE #1743: Steering using wrong MoI
Apparently a side effect of fixing the Moment of Inertia in KSP 1.1.x is that the `y` component now represents yaw, while the `z` component represents roll.  Previously these were reversed.  This caused issues in @Dunbaratu's stream last night.  Interestingly, the issue hides for most normal alignments because part of the fix for KSP 1.1.x was to have better roll values.  As a result, the difference between roll and yaw is not as great in the straight forward alignment.  But changing the control reference to be off-center causes a major issue with the roll calculation.  I'll have to update the steering manager to use the correct values (pull request will be later today).

hvacengi:
Fixes #1743

SteeringManager.cs
* `.y` now is yaw, and `.z` is now roll
* Remove transform method that is no longer used.


Dunbaratu:
I tried this with the specific craft I was having such trouble with during the stream and it seems to have fixed it.
merging now.

PR #1748, opened 2016-07-20
Fix compile command not working with other volumes

### LINKED ISSUE #1741: Can't COMPILE to a different volume than source file
On kOS 1.0.0-pre1, when I execute `compile "0:/sidewinder.ks" to "1:/sidewinder.ksm".`, the output file sidewinder.ksm is created on the Archive instead of on volume 1 like I expected.

hvacengi:
Fixes #1741

Misc.cs
* Modify LoadFunction so that it will get the volume of the outPath before saving the compiled file

### DIFF
--- a/src/kOS/Function/Misc.cs
+++ b/src/kOS/Function/Misc.cs
@@ -322,7 +322,9 @@ namespace kOS.Function
                 // or to a file to save:
                 if (justCompiling)
                 {
-                    List<CodePart> compileParts = shared.ScriptHandler.Compile(path, 1, fileContent.String, String.Empty, options);
+                    // since we've already read the file content, use the volume from outPath instead of the source path
+                    volume = shared.VolumeMgr.GetVolumeFromPath(outPath);
+                    List<CodePart> compileParts = shared.ScriptHandler.Compile(path, 1, fileContent.String, string.Empty, options);
                     VolumeFile written = volume.SaveFile(outPath, new FileContent(compileParts));
                     if (written == null)
                     {


src/kOS/Function/Misc.cs
```diff
@@ -322,7 +322,9 @@ public override void Execute(SharedObjects shared)
                 // or to a file to save:
                 if (justCompiling)
                 {
-                    List<CodePart> compileParts = shared.ScriptHandler.Compile(path, 1, fileContent.String, String.Empty, options);
+                    // since we've already read the file content, use the volume from outPath instead of the source path
+                    volume = shared.VolumeMgr.GetVolumeFromPath(outPath);
+                    List<CodePart> compileParts = shared.ScriptHandler.Compile(path, 1, fileContent.String, string.Empty, options);
                     VolumeFile written = volume.SaveFile(outPath, new FileContent(compileParts));
```
Dunbaratu: Hmm.  The need to do this makes me think we should have a more universal API that really does work properly with a global path.  I mean, the path *contains* "0:" or "1:" in it, so our API method should know which volume it's meant to be saved onto, rather than having to be told which volume it has to be saved to.

hvacengi: Agreed.  It probably should mean that `VolumeManager` allows us to save the file instead of the volume itself.  But trying to do that seemed beyond this fix (and apparently you agree since you just merged)

Dunbaratu:
Yeah it's not this PR's problem, so I merged.

But the reason I brought it up is the semantics of what it should do when told to save a global path onto a volume that's wrong for that global path... I'd say it should fail rather that write it to the wrong volume, which is what it does now.

"please write a file called``0:/myfile.txt`` onto local volume 1" doesn't make semantic sense, and shouldn't have been allowed to work.

PR #1753, opened 2016-07-22
Issue 1752 steering issues again

### LINKED ISSUE #1752: latest SteeringManager commit is not good
the following commit https://github.com/KSP-KOS/KOS/commit/f197c32f01b24ee5fe974905316f95395ffeb1a4 for the y-z Roll Swap is currently breaking my launches. 

When I tested a launch the roll lever on the left down corner was going nuts (flickering left-richt) and my rockets (I tested a few) where flying spirals. 

Reverting the above commit repaired the steering manager for me.

They are all normally build (no offcenter or rotated root-parts)

hvacengi:
Fixes #1752

Rather than figure out exactly what's wrong with the stock Moment of Inertia calculation, I just reverted us back to our old calculation which appears to behave much better in my testing tonight.

Using this pull request, a normal straight facing reference frame should result in the exact same Moment of Inertia as what stock shows.  But when you use "control from" it differs from stock (in a good way, since the stock value seems to be messed up.

My best test case was to take a long and slender-ish rocket, and stick docking ports out the side.  Using the natural up reference frame the stock and kOS calculations matched.  But when controlling from the side mounted docking port, stock was reporting that the roll direction still had a very low moment of inertia (which should have essentially been swapped with the yaw direction).  Using the kOS calculation this appears to perform correctly.

**This should be tested with as many ships/applications as possible, by as many users as possible, to confirm I got it right this time**

### DIFF
--- a/src/kOS/Control/SteeringManager.cs
+++ b/src/kOS/Control/SteeringManager.cs
@@ -529,7 +529,9 @@ namespace kOS.Control
                 angularAcceleration = new Vector3d(angularAcceleration.x, angularAcceleration.z, angularAcceleration.y);
             }
             
-            momentOfInertia = shared.Vessel.MOI;
+            // TODO: This is a temporary fix for the Moment of Inertial  This can be removed after the next KSP release.
+            //momentOfInertia = shared.Vessel.MOI;
+            momentOfInertia = FindMoI();
             adjustTorque = Vector3d.zero;
             measuredTorque = Vector3d.Scale(momentOfInertia, angularAcceleration);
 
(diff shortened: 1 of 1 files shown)


src/kOS/Control/SteeringManager.cs
```diff
@@ -529,7 +529,9 @@ public void UpdateStateVectors()
                 angularAcceleration = new Vector3d(angularAcceleration.x, angularAcceleration.z, angularAcceleration.y);
             }
             
-            momentOfInertia = shared.Vessel.MOI;
+            // TODO: This is a temporary fix for the Moment of Inertial  This can be removed after the next KSP release.
+            //momentOfInertia = shared.Vessel.MOI;
```
Dunbaratu: It might be a good idea to re-phrase this comment now that we don't know if the "next" release will have a better MoI or not.

hvacengi: good point

src/kOS/Control/SteeringManager.cs
```diff
@@ -606,6 +608,69 @@ public void UpdateTorque()
             if (controlTorque.z < minTorque) controlTorque.z = minTorque;
         }
 
+        #region TEMPORARY MOI CALCULATION
+        // TODO: This is a temporary fix for the Moment of Inertial  This can be removed after the next KSP release.
```
Dunbaratu: Here too.

Issue #1755, opened 2016-07-23
diskSpaceMassFactor 1000x too large?


cfe316:
In the documentation it is listed that diskSpaceMassFactor is set so that 4096bytes ~ 0.02kg. I believe the factor is more like 4096byes ~ 0.02 tons. When turning up the storage of the Compomax Radial Tubeless my ship's mass increased by 0.88 tons.
(Is this meant to model rope memory? :D )
This link suggests that the mass setting is in megagrams, not kilograms:
https://github.com/KSP-KOS/KOS/issues/1303

Additionally here is are two screenshots. I'm running 1.1.3 on Mac 10.9.5 with the kOS-v1.0.0-pre-1 (https://github.com/KSP-KOS/KOS/commit/7cbf846b862f7a2e0e38319289a48c3e3c14e323).

![withlittlemass](https://cloud.githubusercontent.com/assets/9857723/17075429/08e0123e-5061-11e6-9950-503b5bf82893.png)
![withlotsofmass](https://cloud.githubusercontent.com/assets/9857723/17075430/08e2851e-5061-11e6-8c26-dfe934a873ad.png)

Dunbaratu:
I'm not entirely happy with the entire disk size game balance, period, but figured there wasn't much point in addressing it until other systems were in place.  Now that those other systems are in place, maybe we can come back to it and balance the weights a bit better.

PR #1760, opened 2016-07-28
Quick fix for #1755

Efratror:
I know Dunbaratu isn't happy with the current disk size game balance, but with a new release at hand I tried to build a quick fix so the MassFactor is correct with the docs
- updated the diskSpaceMassFactor so it represents Kg instead of tons
-   added a float "additionalMassGui" for displaying the additional mass in kG, also added a guiUnit and guiFormat

### DIFF
--- a/src/kOS/Module/kOSProcessor.cs
+++ b/src/kOS/Module/kOSProcessor.cs
@@ -84,14 +84,14 @@ namespace kOS.Module
         [KSPField(isPersistant = true, guiName = "CPU/Disk Upgrade Cost", guiActive = false, guiActiveEditor = true)]
         public float additionalCost = 0F;
 
-        [KSPField(isPersistant = false, guiName = "CPU/Disk Upgrade Mass", guiActive = false, guiActiveEditor = true)]
-        public float additionalMass = 0F;
+        [KSPField(isPersistant = false, guiName = "CPU/Disk Upgrade Mass", guiActive = false, guiActiveEditor = true, guiUnits = "Kg", guiFormat = "0.00")]
+        public float additionalMassGui = 0F;
 
         [KSPField(isPersistant = false, guiActive = false, guiActiveEditor = false)]
         public float diskSpaceCostFactor = 0.0244140625F; //implies approx 100funds for 4096bytes of diskSpace
 
         [KSPField(isPersistant = false, guiActive = false, guiActiveEditor = false)]
-        public float diskSpaceMassFactor = 0.0000048829F;  //implies approx 0.020kg for 4096bytes of diskSpace
+        public float diskSpaceMassFactor = 0.0000000048829F;  //implies approx 0.020kg for 4096bytes of diskSpace
 
         [KSPField(isPersistant = true, guiActive = false)]
         public int MaxPartId = 100;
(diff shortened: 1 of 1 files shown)


Dunbaratu:
@hvacengi 
We need to talk about this subject to decide whether or not to merge this as-is or do more fiddling.

hvacengi:
I'm inclined to not make a change that we'll only adjust again in short order.  So if we merge the PR, with the mass change we should make sure it's the mass we want to end with.  More importantly, I would rather we made the changes in the part config files themselves anyways, so that we can balance the mass based on the part itself.  Right now all parts have the same mass added, which is a bit of an issue for the KAL9000 since changing the storage value jumps the mass from 0.0005T to 1.25T.

src/kOS/Module/kOSProcessor.cs
```diff
@@ -240,11 +240,14 @@ public ModifierChangeWhen GetModuleCostChangeWhen()
             return ModifierChangeWhen.FIXED;
         }
 
+        public float additionalMass = 0F;
```
hvacengi: Defining this field should remain at the top of the file where it was previously instead of moving it.

src/kOS/Module/kOSProcessor.cs
```diff
@@ -240,11 +240,14 @@ public ModifierChangeWhen GetModuleCostChangeWhen()
...
         private void UpdateCostAndMass()
         {
+            
             float spaceDelta = diskSpace - baseDiskSpace;
             additionalCost = (float)System.Math.Round(spaceDelta * diskSpaceCostFactor, 0);
             additionalMass = spaceDelta * diskSpaceMassFactor;
+            additionalMassGui = spaceDelta * diskSpaceMassFactor * 1000;
```
hvacengi: I'd change this to be
```
additionalMassGui = additionalMass * 1000;
```
just to be more concise, and prevent a later change from only modifying one of the assignments.

Efratror: Your totally wright. Thought of it after i posted the PR. Was waiting for your reaction to see if you wished to see it changed.

src/kOS/Module/kOSProcessor.cs
Efratror: How do you mean by "remain at the top" isn't it so if i put it at it's original place KSP will think it's the variable it needs for that field?

Efratror:
So there is still a debate going on about the wright use of additional mass, i updated this quickly when kOS V1.0.0 was in pre-release so it could be merged before the official full release. But since it released already there is no need for a quick fix and can be done properly.

hvacengi:
I'd leave the PR open, if we can come to some kind of consensus on the changes for all parts, you could just push that to this existing PR.

It also means that you have a way to keep us talking about it so that it actually gets done.

src/kOS/Module/kOSProcessor.cs
Efratror: Fair enough.

Efratror:
There seems to be a problem with editing the current PR, it looks like the merge i'v done has overwritten something. I will close this PR and make a new one with the changes you proposed @hvacengi

Dunbaratu:
@hvacengi - there was some discussion above ages ago about this, but I looked it over and I approve of the change.  Are you okay with me merging it?

hvacengi:
It looks like my recommended changes may have been implemented in another branch, but I don't see them currently.  If you could verify at least the organizational structure change (field placement in the file) then it should be fine to merge.

Dunbaratu:
yes the change was made (moving that line up).

Dunbaratu:
Due to how confused and messed up the PR became for this one, I'm going to "merge" it by just re-typing the changes I see from the diffs page in my own branch.  (for example, The top of this PR lists the branch this came from as being from "unknown depositor", which kind of makes it impossible for me to know how I'm supposed to fetch it.)

Dunbaratu:
I also moved the ``public float additionalMass = 0F;`` to being ``public float AdditionalMass { get; set; }`` instead.  Now that that's no longer the name of the KSPField, it became possible for us to use our standard and make it a Property instead.  (when it was a KSPfield it had to be a field because SQUAD.)

Dunbaratu:
Closing because I manually copied the same edits into my own branch to get around the messed up PR.  (this PR says it has conflicts specifically because I did that).

PR #1763, opened 2016-07-30
Fixes #1762 - Skip param compiler check for anon delegates

### LINKED ISSUE #1762: Anon-delegates break with optional params.
Repro case:
```
global foo is { parameter a, b is 0. }.
global bar is { parameter c. }.
print "Program Runs".
```

Error:
```
 works when all mandatory parameters come before all optional parameters.
__________________________________________
           VERBOSE DESCRIPTION
While most commands in kOS work anywhere you
put them, there are a few exceptions in which
a command is only meaningful in some limited
places of the code.  This is one of those cases.
All defaultable parameters must come at the
end of the list of all parameters for a
program or a function.
For example, this is legal:
   declare x, y is 0.
but this is not:
   declare x is 0, y.
because when x had a default then all other
parameters that came after it had to have one.
__________________________________________
More Information at:
http://ksp-kos.github.io/KOS_DOC/language/variables/index.html#DECLARE_PARAMETER
__________________________________________
```

The following examples _do_ work:
```
global foo is { parameter a, b is 0. }.
print "Program Runs".
```

```
function foo { parameter a, b is 0. }
function bar { parameter c. }
print "Program Runs".
```

gisikw:
Fixes #1762

### DIFF
--- a/src/kOS.Safe/Compilation/KS/Compiler.cs
+++ b/src/kOS.Safe/Compilation/KS/Compiler.cs
@@ -2266,7 +2266,9 @@ namespace kOS.Safe.Compilation.KS
             for (int i = node.Nodes.Count - 1 ; i >= 0 ; --i)
             {
                 ParseNode child = node.Nodes[i];
-                if (child.Token.Type != TokenType.declare_function_clause) // functions nested in functions don't count
+
+                // functions nested in functions don't count, nor do anonymous delegates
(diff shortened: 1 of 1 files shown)


Dunbaratu:
I assigned this to @hvacengi because @gisikw was basically just implementing exactly the fix I suggested to him in the issue for this, which means if I review it, I'm basically just reviewing my own edit, and that doesn't count as a proper review.

PR #1769, opened 2016-08-02
Fix for RT loosing control

### LINKED ISSUE #1731: RemoteTech is breaking the steering manager on vessel-switch, without a scene reload
Hi,

I have ha rocket with sevreal smaller sattelites on it. The mothership launches the sats, waits a few seconds and then it switches to the released sat via KUniverse:ACTIVEVESSEL(my_sat).

After that the new Sat has no control with the seeting-manager. 

[output_log.txt](https://github.com/KSP-KOS/KOS/files/367173/output_log.txt)
[KSP-log.txt](https://github.com/KSP-KOS/KOS/files/367174/KSP-log.txt)

hvacengi:
fixes #1731

kOSVesselModule.cs
* Set up the VesselModule to check if RT is available in fixed update,
and toggle event subscriptions accordingly.
* Pull out a few methods into stand alone methods for the purpose of
logging and profiling.  Leaving the stubs for future testing.

### DIFF
--- a/src/kOS/Module/kOSVesselModule.cs
+++ b/src/kOS/Module/kOSVesselModule.cs
@@ -18,6 +18,8 @@ namespace kOS.Module
     {
         private bool initialized = false;
         private Vessel parentVessel;
+        private bool hasRemoteTech = false;
+
         public Guid ID
         {
             get
@@ -115,6 +117,7 @@ namespace kOS.Module
                 {
                     UpdateParameterState();
                 }
+                CheckSwapEvents();
             }
         }
 
@@ -210,11 +213,11 @@ namespace kOS.Module
         {
             if (RemoteTechHook.IsAvailable(parentVessel.id))
             {
-                RemoteTechHook.Instance.AddSanctionedPilot(parentVessel.id, UpdateAutopilot);
+                HookRemoteTechPilot();
             }
             else
             {
-                parentVessel.OnPreAutopilotUpdate += UpdateAutopilot;
+                HookStockPilot();
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/kOS/Module/kOSVesselModule.cs
```diff
@@ -223,16 +226,55 @@ private void HookEvents()
...
+                UnHookRemoteTechPilot();
             }
             else
             {
-                parentVessel.OnPreAutopilotUpdate -= UpdateAutopilot;
+                UnHookStockPilot();
             }
         }
 
+        private void CheckSwapEvents()
```
Dunbaratu: I think it's a good idea to mention the reason we're doing this in a comment for this method.  (i.e. comments that are explaining what you're doing might be just a way to excuse badly written code, but commends that are explaning WHY you're doing it, other other hand, are good).

(i.e. mention the race condition here.  Point out that we have no guarantee whether RemoteTech wakes up before or after we do, and therefore `RemoteTechIs.Aavailable` may be false the first time we check this, but then become true a few moments later.)

Dunbaratu: (I edited the comment above several times, trying to change how I was describing it.  I'm done doing that now.  Sorry about all the edits while you might have been trying to respond to it.)

Dunbaratu: @hvacengi : After my tests, I'm ready to merge this but I just wanted your feedback first on that \<summary\> comment I have above - is what it says true?  I'd like to add it to the method if it's truthful.

hvacengi:
```
/// <summary>
/// A race condition exists when KSP can load the kOS module onto the vessel
/// before it loaded the RemoteTech module.  That makes it so that kOS may not
/// see the existence of RT yet when kOS is first initialized.<br/>
/// <br/>
/// This fixes that case by continually re-querying for RT post-loading, and
/// re-initializing kOS's RT-related behaviors if it seems that the RT module
/// now exists when it didn't before (or visa versa).
/// </summary>
```
I changed a couple of tenses above because it read a little funny to me identifying the race condition in past tense, when the condition still exists we just catch it better.

I was fully planning on doing comments tonight before you merge.  But if you think this summary is enough, then I'm good with it.  That gives me more time to focus on merging PR's instead.

PR #1772, opened 2016-08-03
fix for #1771 Added a Floor

### LINKED ISSUE #1771: time:clock has to many decimals
Some where along the changes of KSP i guess they added decibels for seconds so if you do `print time:clock.` it prints something like this: 4:23:3.1151365138212 instead of 4:23:03

Efratror:
Added a Floor so that the seconds are displayed without decimals.

There is one thing i couldn't fix
if the time is say: 01:01:01 then it's printed as 1:1:1. maybe someone else could have a look at it?

Fixes #1771

### DIFF
--- a/src/kOS/Suffixed/Timespan.cs
+++ b/src/kOS/Suffixed/Timespan.cs
@@ -46,7 +46,7 @@ namespace kOS.Suffixed
             AddSuffix("MINUTE", new Suffix<ScalarValue>(CalculateMinute));
             AddSuffix("SECOND", new Suffix<ScalarValue>(CalculateSecond));
             AddSuffix("SECONDS", new Suffix<ScalarValue>(() => span));
-            AddSuffix("CLOCK", new Suffix<StringValue>(() => string.Format("{0:00}:{1:00}:{2:00}", CalculateHour(), CalculateMinute(), CalculateSecond())));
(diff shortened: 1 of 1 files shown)


hvacengi:
The issue with the formatting is probably because our value is not an integer itself anymore, so the numeric string format function doesn't know what to do (it can't even infer decently, since `ScalarValue` can be `int` or `double` and supports casting to either).  I suspect that if you change line 49 to cast as integers it will format correctly again:

```
AddSuffix("CLOCK", new Suffix<StringValue>(() => string.Format("{0:00}:{1:00}:{2:00}", (int)CalculateHour(), (int)CalculateMinute(), (int)CalculateSecond())));
```

Efratror:
Fixes the string format error.

Efratror:
But what i can't understand is that the `ScalarValue` is return as an int so it should be an int already.

hvacengi:
`String.Format` uses the value in brackets to format known types of data.  So the internal definition of `Format` reads `"{0:00}"` and says "If the first parameter is an integer, I'll pad a single digit number with a zero prefix".  But when we pass it a `ScalarValue` it doesn't know that the parameter is an integer.  In fact `ScalarValue` will convert into either a `double` or an `int` using implicit casting, so it can't really do anything to reliably infer that the value is an integer.  So `Format` doesn't recognize the `:00` portion as meaning anything, and ignores it.

Efratror:
Thats explains a lot, thanks.

PR #1802, opened 2016-09-04
Fixes #1565 - moved the "invalid target" logic.

### LINKED ISSUE #1565: Vessel(name) always fails to find the current cpu vessel.
If ship:name is "foo", and you try to call ``vessel("foo").``, you get an exception that there is no such vessel called "foo", when clearly there is one - your own.

It seems that the candidate vessels that the ``Vessel(...)`` constructor function chooses from is the list of all vessels *other than* the current vessel.

I think it might be using the same list of vessels that ``LIST TARGETS`` uses, and scanning through that list (I haven't looked at the code to know for sure - I'm just reporting this bug and moving on).  It might make sense why LIST TARGETS doesn't list the current vessel - you can't target your own ship.  But you should be able to access it by its name.

(Note, as an aside, while it might make sense why the current vessel isn't a candidate among the LIST TARGETS, that does also raise the issue that if that's the case then we are still missing a means to actually *really* list all the vessels.  Also, I never liked that LIST TARGETS lists vessels.  By name it sounds like it should list all potential targets, meaning vessels, bodies, and docking ports, but that may be a different issue to raise separately.)

Dunbaratu:
The problem was caused by having the "invalid target" logic at too "deep" a level of the code, such that not only was it preventing you from selecting your own ship as a your target, but it was also preventing you from EVER using your own ship in the Vessel() constructor at all.

I had to move the logic that prevents invalid targets a bit higher up, getting it out of the low level constructor so it's possible to construct a Vessel() of your own ship.

### DIFF
--- a/src/kOS.Safe/Exceptions/KOSInvalidTargetException.cs
+++ b/src/kOS.Safe/Exceptions/KOSInvalidTargetException.cs
@@ -1,0 +1,14 @@
+﻿using System;
+
+namespace kOS.Safe.Exceptions
+{
+    /// <summary>
+    /// Description of KOSInvalidTargetException.
+    /// </summary>
+    public class KOSInvalidTargetException : KOSException
+    {
+        public KOSInvalidTargetException(string msg) : base(msg)
+        {
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


hvacengi:
Adding reference to the issue itself: #1565

Dunbaratu:
So does having the big phrase "Fixes #1565" up in the title of the thing not work for associating?  It has to be down in the body?  That would explain why I keep not understanding why it doesn't work.

hvacengi:
No, it associated.  It was auto closed, but when I was reviewing I didn't see the link in the commit title.  My bad, I didn't look close enough.

I've just gotten into the habit of making a list of all the "Fixes" like you did with the docs PR the other day.

PR #1808, opened 2016-09-05
Fix 1755

Efratror:
New PR see PR #1760 for futher details.


Efratror:
Can't get the hang of the inner workings of Github I see :( sorry for the fix for string format error. it's the same as in PR 1772.

PR #1809, opened 2016-09-06
Try to fix RemoteTech steering again

### LINKED ISSUE #1806: More issues with RT and steering
Continuing the never ending saga of troubleshooting Remote Tech and our steering code.  The previously fixed race condition did not solve a separate race condition where RT does not initialize steering computers until a vessel is unpacked, which can result in a one tick delay between kOS registering for steering events and RT re-initializing the list of handlers.  There is no way to query the list of handlers directly, nor is there a way to be notified when the handler list is cleared.  Since RT checks the handler list internally when we attempt to register, it appears that the easiest answer will be to continually register for the events, relying on the internal list check to prevent duplication.  We could introduce some kind of counter to restrict the updates to once every 25 ticks, which should work out to once per game second.

I have a solution prepared that I'm testing and will submit a PR soon.

hvacengi:
Fixes #1806

RT seems to clear the sanctioned pilots list when the vessel unpacks.  There isn't a way for us to subscribe to an update to tell us when things get cleared, and there is no way to directly check the list.  Since RT checks the list of sanctioned pilots when we attempt to register, and only adds if our handler isn't already registered, we can simply keep calling the method to add a sanctioned pilot.  To cut down on the potential performance hit, I limited these calls to once every 25 ticks.

kOSVesselModule.cs
* Update some logging to simplify references
* Add counter to refresh RT sactioned pilot once every 25 ticks (1s real time in theory)

### DIFF
--- a/src/kOS/Module/kOSVesselModule.cs
+++ b/src/kOS/Module/kOSVesselModule.cs
@@ -282,7 +282,7 @@ namespace kOS.Module
         private void UnHookRemoteTechPilot()
         {
             RemoteTechHook.Instance.RemoveSanctionedPilot(parentVessel.id, HandleRemoteTechSanctionedPilot);
-            counterRemoteTechRefresh = 0;
+            counterRemoteTechRefresh = RemoteTechRehookPeriod - 2; // make sure it starts out ready to trigger soon
         }
 
         private void HookStockPilot()
--- a/src/kOS/Module/kOSVesselModule.cs
+++ b/src/kOS/Module/kOSVesselModule.cs
@@ -19,6 +19,7 @@ namespace kOS.Module
         private bool initialized = false;
         private Vessel parentVessel;
         private bool hasRemoteTech = false;
+        private int counterRemoteTechRefresh = 0;
 
         public Guid ID
         {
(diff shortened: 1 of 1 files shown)


src/kOS/Module/kOSVesselModule.cs
```diff
@@ -19,6 +19,7 @@ public class kOSVesselModule : VesselModule
         private bool initialized = false;
         private Vessel parentVessel;
         private bool hasRemoteTech = false;
+        private int counterRemoteTechRefresh = 0;
```
Dunbaratu: Can we begin this at 24 instead of at zero?  That way it only waits 1 update before trying to steer for the very first time.  Right now, I've had problems where the fact that steering is disabled for 1 second after a staging event during liftoff causes the aero effects to start deflecting the vessel from prograde in ways that feel a bit dangerously close to the flipover point.

Dunbaratu: I'm going to make this change and try running with it like that during my stream, which should be starting in an hour.

hvacengi: Interesting, I haven't observed any issues like that on my end.  Something on your system must trigger the race more often than on mine.

We may need to tweak the value a little bit to ensure that we don't randomly find out that one tick is good enough for you, but two ticks might be common as well.  If 24 works for you, go for it.  If 23 or 22 doesn't give you the same aero effect issues, it might be a little more conservative of a choice.

Dunbaratu: Okay I'll try it with 23 then.  (I've also moved the hardcoded `25` to a `private const int` identifier instead just for readability and possible fiddling later if the physics rate ever changes in future releases.

hvacengi:
@Dunbaratu I just merged your pull request and copied the same check in the UnHook code, just in case.  If you're OK with that we should be ready to merge.

Dunbaratu:
great.  yeah let's push the button.

Dunbaratu:
I'm not really in a position to do it myself right now (I have some temp junk currently checked out into my local working copy).  If you want to push the button and merge, go ahead.

hvacengi:
I will do so

PR #1811, opened 2016-09-09
Fixes #1801 and Fixes #1784.

### LINKED ISSUE #1784: Anonymous functions break with `lock steering`
I have a function framework consisting of

-a boot file, which copies the start script from the archive:
http://pastebin.com/jyg22Bj8 in /boot/mission.ks
-a mission file based on the crafts name:
http://pastebin.com/SvPEQDDt in /Mission/StationCore.mission.ks
-and some librarys which supplys all the functions:
http://pastebin.com/sKpzDnP0 /Mission/mission_runner.ks
http://pastebin.com/rYbbNzyn /Mission/Functions/Utility.ks

I am using this setup for some time now. But suddenly, it failed with the error:
koserr1
and:
koserr2

Relevant lines are in mission.ks @ 77-78
I try to get the target file name, download it into local volume and run it. Which worked fine,
but now i keep getting this error message. I added the print line @77 just in case it was using
the wrong name, but as seen in the screenshots, the name matches up. When i list the files,
the file is there but i cannot run it, not even manually.

At the point where it started crashing i was changing things in Utility.ks. The only thing i noted
was that the file grew bigger then the 40kb mark imposed by the smaller core. However i was
using the bigger KAL-9000 core for the whole time to avoid this issue. I even increased the
volume size in the VAB but the same error occured, so the file size might be unrelated. I did
not change anything in mission.ks or in StationCore.mission.ks at the time of the first crash.

Even after restarting the error persists. The file gets copied into the volume but i cannot run it.

### LINKED ISSUE #1801: Can't reference named functions inside anonymous functions
Was trying to set up a named-function workaround for managing steering within anonymous functions (which at the moment are necessary to prevent collisions from higher scope).

```
local fn is {
  function foo { print "foo". }
  foo().
}.
fn().
```

Results in `Undefined Variable Name 'foo'.` on the `foo()` function call. Referring to named functions from an outer scope works as expected:
```
function foo { print "foo". }`
local fn is {
  foo().
}.
fn()
```
`"foo"`

Dunbaratu:
The problem in issue #1801 was that the logic in the Compiler's IterateuserFunctions() that does a tree walk through everywhere a ``function`` statement might exist to preform function preprocessing on it, was failing to descend into the ``instruction_block`` of an anon function.

This was because the ``instruction_block`` of an anon function is a kind of ``expr`` instead of being a kind of ``instruction``, and the case statement wasn't set up to descend into ``expr``'s.

It turns out that, as a bonus, this also fixed issue #1784 as well, for the same reason.

Fixes #1784
Fixes #1801

### DIFF
--- a/kerboscript_tests/user_functions/functest27.ks
+++ b/kerboscript_tests/user_functions/functest27.ks
@@ -2,7 +2,7 @@ print "running a test to see if you can call".
 print "a function from inside a KSM file.".
 
 print "Ensuring functest27_b.ksm is deleted:".
-log "" to "functest27_b.ksm". delete "functest27_b.ksm".
+log "" to "functest27_b.ksm". deletepath("functest27_b.ksm").
 
 print "Trying to compile functest27_b.ks to KSM file:".
 compile functest27_b.
--- a/kerboscript_tests/user_functions/functest28.ks
+++ b/kerboscript_tests/user_functions/functest28.ks
@@ -1,0 +1,42 @@
+// Testing issue 1801 - functions nested in anon functions.
+
+// global named function called from anon function:
+function foo_works1 { print "Global named func called from anon works.". }
+local fn_works1 is { foo_works1(). }.
+fn_works1().
+
+// anon function nested in anon function:
+local fn_works2 is {
+  local foo_works2 is {
+    print "Anon func nested in anon func works.".
+  }.
+  foo_works2().
+}.
+fn_works2().
+
+// nanmed function nested in named function:
+function fn_works3 {
+  function foo_works3 {
+    print "named func nested in named func works.".
+  }
+  foo_works3().
+}
+fn_works3().
+
+// named function inside anything that is not a function
+if true {
+  function func_inside_if {
+    print "named func nested in generic braces works.".
+  }
+  func_inside_if().
+}
+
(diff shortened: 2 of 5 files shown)


hvacengi:
@Dunbaratu I'm testing out this PR to merge.  I noticed that you've tagged #1784 as being fixed as well but that there wasn't a function test for it.  So I added the following script, and I still get the error: "The given key was not present in the dictionary."

```
// functest33.ks
// Testing issue 1784 - lock steering nested in anon functions.

function testNamed {
    lock steering to up.
}

set testAnon to {
    lock steering to up.
}.

testNamed().
testAnon().
```

Am I missing something in the testing of that error?

Dunbaratu:
I'll give your test case a try and see what I get.  I was sure I tested that case but maybe I didn't.

Dunbaratu:
Okay now I know why my test didn't catch the problem and yours did.

Although this crashes:
```
set testAnon to {
  lock steering to up.
}
testAnon().
```
This works fine, and doesn't crash:
```
local testAnon is {
  lock steering to up.
}
testAnon().
```

The error does not surface when using a declaration statement like ``local`` or ``global``, and thus my tests didn't catch it.

Issue #1787, opened 2016-08-24
Add running mode to kosprocessor:mode


Gotbread:
I'm looking for a way to tell if a cpu has crashed or is currently running. The kosprocessor has a
field [mode](url=https://ksp-kos.github.io/KOS/structures/vessels/kosprocessor.html#structure:KOSPROCESSOR) but this does not tell you if the cpu is running or idle.

A use case for this is a launch vessel with multible cores. The launcher is programmed to launch immediately. If one of the probe cores crash, the vessel launches with a faulty script. Depending on the kind of error the script might not have a chance to run (to send a heartbeat or so). As the generic launcher does not know the number of probes in advance, and whether they should be running or not, a more elegant solution would be to get information like "running", or "crashed" from this field.

Since a core is technically "ready" after a crash this field could be set depending whether the last instruction resulted in an error or not.

Since most of my launches are unsupervisored, this would greatly help if i came back forget a syntax error.

PR #1820, opened 2016-09-17
Alternate way to " Fixes #1805 "

### LINKED ISSUE #1785: First wait after boot not actually waiting
My boot file includes ```ADDONS:RT:HASCONNECTION(SHIP)```  and when I dont have a connection my frame rate suffers. I will add logs to this once i get home.

### LINKED ISSUE #1805: Allow any opcode or function to "wait"

Dunbaratu:
Fixes #1805 
Fixes #1785 

After some discussion with @hvacengi during the review of PR #1807 I came up with this alternate way to do it that might avoid some of the problem cases that PR mentioned.

(Also it streamlines some of the messiness in how the CPU was detecting "am I in a trigger or main code right now", and I think the new way is cleaner.)

### DIFF
--- a/src/kOS.Safe/Compilation/Opcode.cs
+++ b/src/kOS.Safe/Compilation/Opcode.cs
@@ -74,7 +74,7 @@ namespace kOS.Safe.Compilation
         ADDTRIGGER     = 0x53,
         REMOVETRIGGER  = 0x54,
         WAIT           = 0x55,
-        ENDWAIT        = 0x56,
+        // Removing ENDWAIT, 0x56 may be reused in a future version
         GETMETHOD      = 0x57,
         STORELOCAL     = 0x58,
         STOREGLOBAL    = 0x59,
@@ -241,6 +241,9 @@ namespace kOS.Safe.Compilation
 
         public bool AbortProgram { get; set; }
         public bool AbortContext { get; set; }
+        public bool IsYielding { get; protected set; } // allows the cpu to know if the opcode is yielding (waiting) to know if it should call Execute or CheckYield
+
+        public Func<ICpu, bool> YieldFinishedDelegate { get; protected set; }
         
         public virtual void Execute(ICpu cpu)
         {
@@ -256,6 +259,7 @@ namespace kOS.Safe.Compilation
             DeltaInstructionPointer = 1;
             AbortProgram = false;
             AbortContext = false;
+            IsYielding = false;
         }
         
         /// <summary>
@@ -460,6 +464,47 @@ namespace kOS.Safe.Compilation
         protected Structure PopStructureAssertEncapsulated(ICpu cpu, bool barewordOkay = false)
         {
             return Structure.FromPrimitiveWithAssert(PopValueAssert(cpu, barewordOkay));
+        }
+
+        public void YieldProgram(ICpu cpu, Func<ICpu, bool> isFinishedDelegate)
+        {
+            if (isFinishedDelegate != null)
+            {
+                YieldFinishedDelegate = isFinishedDelegate;
+                BeginYield();
+                cpu.StartWait(0);
+            }
+        }
+
+        public void CheckYield(ICpu cpu)
+        {
+            if (YieldFinishedDelegate != null)
+            {
+                if (YieldFinishedDelegate.Invoke(cpu))
+                {
+                    EndYield();
+                }
+                else
+                {
+                    cpu.StartWait(0);
+                }
+            }
+            else
+            {
+                throw new KOSException("Error checking yielded Opcode: YieldFinishedDelegate is null");
+            }
+        }
+
+        protected virtual void BeginYield()
+        {
+            DeltaInstructionPointer = 0;
+            IsYielding = true;
+        }
+
+        protected virtual void EndYield()
+        {
+            DeltaInstructionPointer = 1;
+            IsYielding = false;
         }
     }
 
@@ -2136,37 +2181,24 @@ namespace kOS.Safe.Compilation
             // remaining time on the wait.  There can be more than one place where
             // a wait was pending.  So instead track the pending time inside the
             // opcode itself, by having the opcode store its timestamp when it is
-            // meant to end, and checking it each time.
+            // meant to end, and then yield.  The end time is then checked inside
+            // of CheckWaitTime
             if (endTime < 0) // initial time being executed.
             {
                 double arg = Convert.ToDouble(cpu.PopValue());
                 endTime = cpu.StartWait(arg);
-                DeltaInstructionPointer = 0; // stay here next time.
-            }
-            else if (cpu.SessionTime < endTime) // It's in the midst of an already running wait, and it's not expired yet.
-            {
-                cpu.StartWait(0); // kick back to wait mode again.
-                DeltaInstructionPointer = 0; // stay here next time, to test this again.
-            }
-            else // It was in an already running wait, and the wait just expired.
-            {
-                endTime = -1; // reset in case this is called again in a loop.
-                DeltaInstructionPointer = 1; // okay now move on.
+                YieldProgram(cpu, CheckWaitTime);
             }
         }
-    }
 
-    // TODO: Does this Opcode even need to exist?  What's the point?  You can't execute it while you're stuck inside
-    // a wait so.... when was anyone ever expecting it to happen?  Discuss if it can just be removed.  Nothing in
-    // the compiler currently builds any code that uses it:
-    public class OpcodeEndWait : Opcode
-    {
-        protected override string Name { get { return "endwait"; } }
-        public override ByteCode Code { get { return ByteCode.ENDWAIT; } }
-
-        public override void Execute(ICpu cpu)
+        public bool CheckWaitTime(ICpu cpu)
         {
-            cpu.EndWait();
+            if (cpu.SessionTime > endTime)
+            {
+                endTime = -1; // reset in case this is called again in a loop.
+                return true;
+            }
+            return false;
         }
     }
 
--- a/src/kOS.Safe/Compilation/Opcode.cs
+++ b/src/kOS.Safe/Compilation/Opcode.cs
@@ -215,6 +215,11 @@ namespace kOS.Safe.Compilation
             "+-------------------------------------------------------------------------+\n";
 
         public int Id { get { return id; } }
+        /// <summary>
+        /// How far to jump to the next opcode.  1 is typical (just go to the next opcode),
+        /// but in the case of jump and branch statements, it can be other numbers.  This will
+        /// be ignored if the CPU has been put into a yield state with CPU.YieldProgram().
+        /// </summary>
         public int DeltaInstructionPointer { get; protected set; }
         public int MLIndex { get; set; } // index into the Machine Language code file for the COMPILE command.
         
@@ -241,9 +246,6 @@ namespace kOS.Safe.Compilation
 
         public bool AbortProgram { get; set; }
         public bool AbortContext { get; set; }
-        public bool IsYielding { get; protected set; } // allows the cpu to know if the opcode is yielding (waiting) to know if it should call Execute or CheckYield
-
-        public Func<ICpu, bool> YieldFinishedDelegate { get; protected set; }
         
         public virtual void Execute(ICpu cpu)
         {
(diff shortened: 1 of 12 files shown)


Dunbaratu:
(Changed the title around hoping github will auto-associate it).

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -215,6 +215,11 @@ public abstract class Opcode
             "+-------------------------------------------------------------------------+\n";
 
         public int Id { get { return id; } }
+        /// <summary>
+        /// How far to jump to the next opcode.  1 is typical (just go to the next opcode),
+        /// but in the case of jump and branch statements, it can be other numbers.  This will
+        /// be ignored if the CPU has been put into a yield state with CPU.YieldProgram().
+        /// </summary>
         public int DeltaInstructionPointer { get; protected set; }
```
Dunbaratu: Added the above comment to clarify the new technique will NOT be to have OpcodeWait set its DeltaInstructionPointer to 0, but rather to have the CPU itself go ahead and move the IP to the next instruction, but then not execute it yet until the wait period is over.  I didn't like the way it used to spin on the OpcodeWait, re-executing it each update, instead of just not executing anything until the time is over.

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -2127,46 +2133,11 @@ public class OpcodeWait : Opcode
...
-    public class OpcodeEndWait : Opcode
-    {
-        protected override string Name { get { return "endwait"; } }
-        public override ByteCode Code { get { return ByteCode.ENDWAIT; } }
 
         public override void Execute(ICpu cpu)
         {
-            cpu.EndWait();
+            double arg = Convert.ToDouble(cpu.PopValue());
+            cpu.YieldProgram(new YieldFinishedGameTimer(arg));
```
Dunbaratu: With the new technique, OpcodeWait got super simple because the "wait for timer to expire" style of wait is now a class and you just make a new instance of it, passing in the duration of the wait.  All the thinking is now happening in the class ``YieldFinishedGameTimer``.

src/kOS.Safe/Execution/CPU.cs
```diff
@@ -15,15 +15,18 @@ namespace kOS.Safe.Execution
...
     public class CPU : ICpu
     {
-        private enum Status
+        private enum Section
         {
-            Running = 1,
-            Waiting = 2
+            Main = 1,
+            Trigger = 2
         }
```
Dunbaratu: I realized that ``Cpu.Status`` was only ever used in one and one place only and that was to handle the waits the old way.  The new way made it moot.

The new enum, "Section" is just to remember which code section we are in, trigger or main - I tried using "context" but that word is overused in kOS for many different things and it would have been yet another meaning for it.

src/kOS.Safe/Execution/CPU.cs
```diff
@@ -15,15 +15,18 @@ namespace kOS.Safe.Execution
...
 
         private readonly IStack stack;
         private readonly VariableScope globalVariables;
-        private Status currentStatus;
+        private Section currentRunSection;
+        private List<YieldFinishedDetector> triggerYields;
+        private List<YieldFinishedDetector> mainYields;
```
Dunbaratu: These two lists contain the list of Pending Yields that haven't stated they're finished yet.  When a yiedler says it's finished, the CPU will remove it from the list.  Only when the list is empty again will the CPU resume.  This takes the place of the notion of Status = running or Status = waiting.

src/kOS.Safe/Execution/CPU.cs
```diff
@@ -54,12 +57,6 @@ public int InstructionPointer
             set { currentContext.InstructionPointer = value; }
         }
         
-        /// <summary>
-        /// Where was the instruction pointer at the time this update
-        /// first started?
-        /// </summary>
-        int firstInstructionPointerInUpdate;
-
```
Dunbaratu: this was only ever used as part of the old way to detect if we were still in trigger functions or if we'd popped back to mainline code, and it wasn't really working right anyway.

src/kOS.Safe/Execution/CPU.cs
```diff
@@ -409,6 +409,70 @@ public void BreakExecution(bool manual)
...
+            // Query the yield trackers and remove all the ones that claim they are finished.
+            // Always treat them as unfinished if this is the same fixed time stamp as the
+            // one in which the yield got Begin()'ed regardless of what their IsFinished() claims,
+            // because all waits will always wait at least one tick, according to all our docs.
+            yieldTrackers.RemoveAll((t) => t.creationTimeStamp != currentTime && t.IsFinished());
+            
+            // If any are still present, that means not all yielders in this context (main or trigger)
+            // are finished and we should still return that we are waiting:
+            return yieldTrackers.Count > 0;
+        }
```
Dunbaratu: Most of the above should be explained already by the comments.

src/kOS.Safe/Execution/CPU.cs
```diff
@@ -1021,27 +1085,6 @@ public void RemoveTrigger(int triggerFunctionPointer)
...
-                currentStatus = Status.Waiting;
-            return currentTime + waitTime;
-        }
-
-        public void EndWait()
-        {
-            if (currentStatus == Status.Waiting)
-                currentStatus = Status.Running;
-        }
-
```
Dunbaratu: These are part of a way even older than the current develop branch way of doing this.  StartWait, for example, was never called from anywhere, and EndWait was only called internally from inside CPU, despite being public.  Anyway both have been replaced with new logic.

src/kOS.Safe/Execution/CPU.cs
```diff
@@ -1052,8 +1095,6 @@ public void KOSFixedUpdate(double deltaTime)
             instructionsSoFarInUpdate = 0;
             var numMainlineInstructions = 0;
 
-            firstInstructionPointerInUpdate = InstructionPointer;
-
```
Dunbaratu: This was never accurate anyway as if you had a trigger that lasted longer than 1 update, on the second update's worth of running it, this would mark an instruction in the middle of the trigger, and the system would never go back to mainline mode until it hit this instruction, which it falsely believed to be the mainline instruction it had first branched into the triggers from (it was in the first update the trigger fired in, but if it's a long trigger lasting more than 1 update, then in the second update, it's no longer pointing at the mainline code and that meant it would fail to realize it hit mainline code for at least one whole extra update after it already had.)

src/kOS.Safe/Execution/CPU.cs
```diff
@@ -1074,19 +1115,16 @@ public void KOSFixedUpdate(double deltaTime)
                 {
                     ProcessTriggers();
 
-                    if (currentStatus == Status.Running || currentStatus == Status.Waiting)
```
Dunbaratu: This statement was literally irrelevant anyway.  currentStatus was an enum, and it was an enum where these were the *only* two values defined for the enum.  It was impossible for this condition to ever be false.

src/kOS.Safe/Execution/CPU.cs
```diff
@@ -1074,19 +1115,16 @@ public void KOSFixedUpdate(double deltaTime)
...
-                            executionWatch.Stop();
-                        }
+                        executionWatch.Start();
+                        instructionWatch.Start();
+                    }
+                    ContinueExecution(showStatistics);
+                    numMainlineInstructions = instructionsSoFarInUpdate;
+                    if (showStatistics)
+                    {
+                        executionWatch.Stop();
```
Dunbaratu: The above is just git diff's messy way of showing that I out-dented this code when I removed the wrapping if-condition around it.

src/kOS.Safe/Execution/CPU.cs
```diff
@@ -1205,39 +1243,38 @@ private void ContinueExecution(bool doProfiling)
...
             {
-                if (InstructionPointer == firstInstructionPointerInUpdate &&
-                    ! stack.HasTriggerContexts()) // This additional check is important.
-                                                 // Hypothetically it could be possible for trigger code to hit the same
-                                                 // IP as where mainline code was interrupted, if both trigger code and
-                                                 // mainline code make use of the same library function, and the
-                                                 // trigger interruption occurred when mainline code was in that
-                                                 // same library function.  This ensures that if so, that doesn't count.
-                                                 // It only counts when all the triggers have popped off the call stack.
+                if (! stack.HasTriggerContexts())
```
Dunbaratu: With my new changes to ``stack.HasTriggerContexts()``, it's no longer a super expensive walk of the stack, so it's fine to just query it every time instead of trying to protect around it with the former messy logic.

src/kOS.Safe/Execution/CPU.cs
```diff
@@ -1205,39 +1243,38 @@ private void ContinueExecution(bool doProfiling)
...
                 {
-                    reachedMainLine = true;
+                    currentRunSection = Section.Main;
+                }
+                
+                if (IsYielding())
+                {
+                    executeNext = false;
+                }
```
Dunbaratu: stop whenever anything an opcode did seems to have left us in a waiting state.

src/kOS.Safe/Execution/Stack.cs
```diff
@@ -203,22 +215,16 @@ public List<int> GetCallTrace()
...
         {
-            for (int index = stackPointer + 1; index < stack.Count; ++index)
-            {
-                SubroutineContext contextRecord = stack[index] as SubroutineContext;
-                if (contextRecord != null)
-                    if (contextRecord.IsTrigger)
-                        return true;
-            }
-            return false;
+            return triggerContextCount > 0;
```
Dunbaratu: The new triggerContextCount is what allowed the expense of this walk to be ripped out, which in turn allows us to just call ``stack.HasTriggerContext()`` willy-nilly whenever we feel like without it slowing things down.

src/kOS.Safe/Compilation/Opcode.cs
hvacengi: The only concern I have with this is in the case of a opcode or function that may not know the delta instruction pointer until _after_ the yielded operation finishes.  For example, running a program that has already been compiled will have a different delta instruction than a newly compiled program.

src/kOS.Safe/Execution/CPU.cs
hvacengi: Should we maybe include an enum for "interpreter".  It might allow us to address #1787 at the same time.  I guess that really isn't what this enum is used for though, so maybe it's a bad idea.  Just a thought that popped into my head.

src/kOS.Safe/Execution/CPU.cs
hvacengi: Interesting idea.  By using a list of detectors, a function or opcode could theoretically chain a few wait commands together.  Not sure where we would use it, but it remains an option.

src/kOS.Safe/Execution/CPU.cs
Dunbaratu: Yeah that one is tricky because the interpreter ALSO has a separate main and trigger.  (you can type a "when" command into the interpreter and it *will* trigger later on provided you're in the interpreter and not in the middle of a program.)

In fact that's why I didn't call it "context".  We already have a thing called "context" and it is that - the difference between being in the interpreter or in a program is the ProgramContext.

src/kOS.Safe/Compilation/Opcode.cs
hvacengi: After I posted it I looked back and realized that changing the pointer is still very possible.  Since the IP changes when the cpu begins yielding, a function or opcode is free to change the instruction pointer again just before returning true in `YieldFinishedDetector.IsFinished`

PR #1823, opened 2016-09-22
Documentation edits

jacktose:
Some copyediting and formatting of the docs. Nothing heavy. I tried to be conservative because I'm new at this and I'm not sure of style or other conventions here.

Fixes [1824](https://github.com/KSP-KOS/KOS/issues/1824).

### DIFF
--- a/doc/source/language/delegates.rst
+++ b/doc/source/language/delegates.rst
@@ -3,9 +3,10 @@
 Delegates (function references)
 ===============================
 
-.. versionadded:: 0.19.0
-   The delegate feature described on this page did not exist
-   prior to kOS 0.19.0.
+.. note::
+    .. versionadded:: 0.19.0
+        The delegate feature described on this page did not exist
+        prior to kOS 0.19.0.
 
 .. contents:: Contents
     :local:
@@ -317,7 +318,7 @@ Note that you can combine the two lines above that looked like this::
 
 into just this::
 
-    local draw_a_to_b is draw_a_to_b@:bind(shipA, shipB).
+    local draw_a_to_b is draw_ship_to_ship@:bind(shipA, shipB).
 
 When you use the at-sign(``@``), you are returning an object of type
 :struct:`KOSDelegate` that can be used in-line right in the expression,
@@ -369,31 +370,31 @@ normally be in :ref:`scope <scope>`.
 Kinds of Delegate (no suffixes)
 ===============================
 
-Under the hood, kOS handles several different kinds of 'functions' and
+Under the hood, kOS handles several different kinds of "functions" and
 methods that aren't actually implemented the same way.  A ``KOSDelegate``
 attempts to hide the details of these differences from the user, but
 one difference in particular still stands out.  In kOS version 0.19.0,
 you cannot reliably make a delegate of a suffix just yet.  (*This is
 intended as a future feature though.  It's been put off because it
-involves decisions that impact the future of the language and once made,
-can't be changed easily.*)
+involves decisions that impact the future of the language and which, once
+made, can't be changed easily.*)
 
 - You **can** make a delegate of a :ref:`user function <user_functions>`
-  implemented in Kerboscript code.::
+  implemented in Kerboscript code::
     
     function mysquarefunc { parameter a. return a*a. }
     set x to mysquarefunc@.
     set y to x:call(5). // y is now 25.
 
 - You **can** make a delegate of a built-in function provided by kOS
-  itself, provided it isn't a structure suffix.::
+  itself, provided it isn't a structure suffix::
 
     set r to round@.
     set s to sqrt@.
     print "square root of 7, to the nearest 2 places is: " + r:call(s:call(7), 2).
 
 - You **cannot** make a delegate of a suffix of a structure (*yet?*)
-  in Kerboscript.::
+  in Kerboscript::
 
     //
     // WON'T WORK, WILL GIVE ERROR:
@@ -402,5 +403,5 @@ can't be changed easily.*)
     print "altpos at altitude 1000 is " + altpos:call(1000).
 
   However, if you like you can make your own user function that is a
-  wrapper around a structure suffix call, and make a delegate of THAT.
+  wrapper around a structure suffix call, and make a delegate of *that*.
 
--- a/doc/source/language/delegates.rst
+++ b/doc/source/language/delegates.rst
@@ -380,21 +380,27 @@ involves decisions that impact the future of the language and which, once
 made, can't be changed easily.*)
 
 - You **can** make a delegate of a :ref:`user function <user_functions>`
-  implemented in Kerboscript code::
+  implemented in Kerboscript code.
+    
+    ::
     
     function mysquarefunc { parameter a. return a*a. }
     set x to mysquarefunc@.
     set y to x:call(5). // y is now 25.
 
 - You **can** make a delegate of a built-in function provided by kOS
-  itself, provided it isn't a structure suffix::
+  itself, provided it isn't a structure suffix.
+
+    ::
 
     set r to round@.
     set s to sqrt@.
     print "square root of 7, to the nearest 2 places is: " + r:call(s:call(7), 2).
 
 - You **cannot** make a delegate of a suffix of a structure (*yet?*)
-  in Kerboscript::
+  in Kerboscript.
+
+    ::
 
     //
     // WON'T WORK, WILL GIVE ERROR:
@@ -403,5 +409,5 @@ made, can't be changed easily.*)
     print "altpos at altitude 1000 is " + altpos:call(1000).
 
   However, if you like you can make your own user function that is a
-  wrapper around a structure suffix call, and make a delegate of *that*.
+  wrapper around a structure suffix call, and make a delegate of **that**.
 
--- a/doc/source/language/features.rst
+++ b/doc/source/language/features.rst
@@ -68,7 +68,9 @@ Structures are variables that contain more than one piece of information. For ex
 
 .. note::
     .. versionadded:: 0.19.0
-        As of kOS version 0.19.0, in fact ALL values a script can see are now a kind of :struct:`Structure <structure>`, even basic primitive types such as Boolean and Scalar.
+        As of kOS version 0.19.0, in fact ALL values a script can see are
+        now a kind of :struct:`Structure <structure>`, even basic primitive
+        types such as Boolean and Scalar.
 
 
 Many structures also let you set a specific component of them, for example::
--- a/doc/source/language/features.rst
+++ b/doc/source/language/features.rst
@@ -66,10 +66,9 @@ Structures are variables that contain more than one piece of information. For ex
     PRINT "The Mun's periapsis altitude is: " + MUN:PERIAPSIS.
     PRINT "The ship's surface velocity is: " + SHIP:VELOCITY:SURFACE.
 
-.. versionadded:: 0.19.0
-    As of kOS version 0.19.0, in fact ALL values a script can see are
-    now a kind of :struct:`Structure <structure>`, even basic primitive
-    types such as Boolean and Scalar.
+.. note::
+    .. versionadded:: 0.19.0
+        As of kOS version 0.19.0, in fact ALL values a script can see are now a kind of :struct:`Structure <structure>`, even basic primitive types such as Boolean and Scalar.
 
 
 Many structures also let you set a specific component of them, for example::
@@ -166,7 +165,7 @@ programming.
 But if you are an experienced programmer you might not like this
 behavior, and there are good arguments for why you might want to
 disable it.  If you wish to do so, a syntax exists to do so called
-:ref:``NOLAZYGLOBAL``.
+:ref:`@LAZYGLOBAL OFF <lazyglobal>`.
 
 .. _feature functions:
 
--- a/doc/source/language/syntax.rst
+++ b/doc/source/language/syntax.rst
@@ -176,13 +176,13 @@ Help for the new user - What is a Function?
     In programming terminology, there is a commonly used feature of
     many programming languages that works as follows:
 
-    1. Create a chunk of program instructions that you don't intend to execute YET.
-    2. Later, when executing other parts of the program, do the following:
+    - 1. Create a chunk of program instructions that you don't intend to execute YET.
+    - 2. Later, when executing other parts of the program, do the following:
 
-       A. Remember the current location in the program.
-       B. Jump to the previously created chunk of code from (1) above.
-       C. Run the instructions there.
-       D. Return to where you remembered from (2.A) and continue from there.
+        - A. Remember the current location in the program.
+        - B. Jump to the previously created chunk of code from (1) above.
+        - C. Run the instructions there.
+        - D. Return to where you remembered from (2.A) and continue from there.
 
     This feature goes by many different names, with slightly different
     precise meanings: *Subroutines*, *Procedures*, *Functions*, etc.
--- a/doc/source/language/syntax.rst
+++ b/doc/source/language/syntax.rst
@@ -176,12 +176,13 @@ Help for the new user - What is a Function?
     In programming terminology, there is a commonly used feature of
     many programming languages that works as follows:
 
-    - 1. Create a chunk of program instructions that you don't intend to execute YET.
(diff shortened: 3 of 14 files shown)


doc/source/math/ref_frame.rst
```diff
@@ -19,14 +19,14 @@ documentation:
...
-the axes is **left-handed**.  This means that if you imagine opening your
+the axes is **left-handed**. This means that if you imagine opening your
 palm and pointing your fingers down the x-axis, then curling your fingers
 in the direction of the y-axis, then sticking your thumb up, that the
-direction your thumb would have to be pointing to accomplish this task
-is the direction of the z-axis **if and only if** you used your left
-hand to perform those steps.  (If you do those steps with your right
-hand, you get a z-axis in the opposite direction and that is known as
-a **right handed** coordinate system).
+direction in which your thumb must be pointing is the direction of the
```
Dunbaratu: "direction in which your thumb **would** be pointing" instead of "**must** be pointing" might be a better way to say it, as it matches with how things are phrased up above.

doc/source/math/scalar.rst
```diff
@@ -68,19 +71,19 @@ Or "What the heck does 'attempted to push NaN onto the stack' mean?".
 kOS does not have a numeric type designed to deal with 
 *imaginary numbers* or *complex numbers*.  Therefore, for
 example, if you attempted to perform ``sqrt(-4)``, you would get
-a "NaN error", rather than the irrational number ``2i``.  NaN means
-"Not a Number" and it means the system is incapable of storing the
+a "NaN error", rather than the irrational number ``2i``.  "NaN" stands
```
Dunbaratu: On re-reading this to review your edit, I noticed a factual error (not merely a formatting or phrasing disagreement) in my original text which needs to be fixed as well:  It claims that ``2i`` is an *irrational* number.  It's not.  It's an *imaginary* number.

doc/source/language/delegates.rst
```diff
@@ -402,5 +403,5 @@ can't be changed easily.*)
     print "altpos at altitude 1000 is " + altpos:call(1000).
 
   However, if you like you can make your own user function that is a
-  wrapper around a structure suffix call, and make a delegate of THAT.
+  wrapper around a structure suffix call, and make a delegate of *that*.
```
hvacengi: +1 to bold in place of italics.  I have no idea if there is any kind of a standard, but I treat bold as emphasis, and italics as a change in tone or other subtle variation that comes out in spoken but not written language.

doc/source/language/delegates.rst
```diff
@@ -369,31 +370,31 @@ normally be in :ref:`scope <scope>`.
...
 you cannot reliably make a delegate of a suffix just yet.  (*This is
 intended as a future feature though.  It's been put off because it
-involves decisions that impact the future of the language and once made,
-can't be changed easily.*)
+involves decisions that impact the future of the language and which, once
+made, can't be changed easily.*)
 
 - You **can** make a delegate of a :ref:`user function <user_functions>`
-  implemented in Kerboscript code.::
+  implemented in Kerboscript code::
```
hvacengi: The way I try to get around the colon limitation is to put the double colon on the next line and that seems to suppress the compiled html having a colon.  A quick compile test shows that it appears to work in this instance as well:
```
- You **can** make a delegate of a :ref:`user function <user_functions>`
  implemented in Kerboscript code.
  ::
```

doc/source/language/variables.rst
```diff
@@ -234,8 +235,9 @@ expression, the expression will not get executed if the calling
 function had an argument present in that position.  The expression
 only gets executed if the system needed to pad a missing argument.
 
-.. versionadded:: 0.18.3
-   Optional Parameters were added as a new feature in kOS 0.18.3
+.. note::
+    .. versionadded:: 0.18.3
+        Optional Parameters were added as a new feature in kOS 0.18.3
```
hvacengi: Actually, `.. versionadded::` appears to be behaving as expected.  It's only configured to italicize and add the text "New in version" before the version number.  I have long suspected that we could simply modify the configuration to use a similar style as `.. note::` (though different so that it stands out).  It has been on my list to see if that's actually possible.

I do feel like there is probably a point at which we can simply remove these blocks, for instance there have been 6 releases and 6 months since v0.18.3 was released.  Maybe the note can be considered no longer necessary.

doc/source/math/basic.rst
```diff
@@ -63,11 +63,22 @@ constants about the universe that you may find handy in your math operations.
...
 .. global:: Constant:PI
 
-    Ratio of circumference of a circle to its diameter
+    Ratio of circumference of a circle to its diameter, 3.14159265...
+    
+    ::
+
+        SET diameter to 10.
+        PRINT "circumference is:".
+        PRINT constant:pi * diameter.
```
hvacengi: 1st off, I don't think it's a good idea for us to be exclusionary and imply that if a person doesn't understand the definition of pi they shouldn't use kOS.  KSP itself caters to all kinds of people, including some surprisingly young children.  It isn't hard to believe that some one might get their start with kOS only using wait commands to stage or throttle.  But that doesn't mean it's our job to explain pi, just that we should be careful about the phrasing.

On topic however: if it's important enough to mention in the text explanation I'm not sure that we can argue against providing a code example as well.  At least the example is _doing_ something with the value.

Remove the example or leave it, I don't really mind either way, just wanted to add my two cents.

doc/source/math/basic.rst
```diff
@@ -63,11 +63,22 @@ constants about the universe that you may find handy in your math operations.
...
 
 .. global:: Constant:C
 
     Speed of light in a vacuum, in meters per second.
+    
+    ::
+
+        PRINT "Light goes this fast:".
+        PRINT constant:c.
```
hvacengi: Speaking of an example not doing anything...

I agree, this example is nothing more than showing how to access a suffix.  If you're set on providing an example, maybe you could perform some math with it.

```
local speed is ship:velocity:orbit:mag.
local speedOfLight is constant:c.
local ratio is speed / speedOfLight.
print "Current speed is " + ratio + " times the speed of light.".
```
While still relatively meaningless, at least it gives the reader an example that does something.

jacktose:
I'm working on incorporating these responses.

How do I update the pull request? Just by updating the branch?

hvacengi:
Yup, just push any new commits to the same branch and the pull request will automatically update.

doc/source/language/features.rst
```diff
@@ -66,10 +66,9 @@ Structures are variables that contain more than one piece of information. For ex
     PRINT "The Mun's periapsis altitude is: " + MUN:PERIAPSIS.
     PRINT "The ship's surface velocity is: " + SHIP:VELOCITY:SURFACE.
 
-.. versionadded:: 0.19.0
-    As of kOS version 0.19.0, in fact ALL values a script can see are
-    now a kind of :struct:`Structure <structure>`, even basic primitive
-    types such as Boolean and Scalar.
+.. note::
+    .. versionadded:: 0.19.0
+        As of kOS version 0.19.0, in fact ALL values a script can see are now a kind of :struct:`Structure <structure>`, even basic primitive types such as Boolean and Scalar.
```
jacktose: That was probably an accident while doing the .. note thing. Switching it back.

Related: is there a convention for line breaks? There's some of everything.

doc/source/language/delegates.rst
jacktose: Yeah, the :: on its own line is the rst standard for that.

doc/source/language/delegates.rst
jacktose: @hvacengi There are too many standards. AP, Chicago, Strunk & White, etc. I don't know that kOS documentation is a formal enough project to officially adopt a style guide, so I'm just going with some commonsense guesses. One source I like, especially for modern, online text is *Butterick's Practical Typography*. [Here it is on bold and italic.][bpt]

BPT (and most sources) say italic is emphasis and bold is stronger emphasis. BPT notes that italic is often poorly rendered or hard to read on-screen, so it may be better to go straight to bold for online text. All-caps is sometimes described as an emphasis option, but also has the unpleasant “shouting” connotation.

In any case, it seems that bold is a good consensus choice here. I'll be happy to change it.

[bpt]: http://practicaltypography.com/bold-or-italic.html "Bold or italic | Butterick's Practical Typography"

doc/source/language/syntax.rst
```diff
@@ -176,13 +176,13 @@ Help for the new user - What is a Function?
...
+    2. Later, when executing other parts of the program, do the following:
 
-       - 2.A. Remember the current location in the program.
-       - 2.B. Jump to the previously created chunk of code from (1) above.
-       - 2.C. Run the instructions there.
-       - 2.D. Return to where you remembered from (2.A) and continue from there.
+       A. Remember the current location in the program.
+       B. Jump to the previously created chunk of code from (1) above.
+       C. Run the instructions there.
+       D. Return to where you remembered from (2.A) and continue from there.
```
jacktose: Fair enough. I could fix the OL rendering in CSS, but I don't know sphinx well enough. In the meantime, bullets seems like a good workaround.

doc/source/language/variables.rst
jacktose: It's not too hard to find all the `.. note::`s and change them all back if `versionadded` gets tweaked later. Can I suggest leaving them in `note`s so that they look right for now? Same idea as the bulleted lists as a workaround.

doc/source/math/basic.rst
jacktose: I'm with the “helpfulness for beginners” argument here, but mostly I thought it was weird that some of these had examples and some didn't. I don't think it hurts to have examples for each one.

doc/source/math/basic.rst
jacktose: Yeah, that one was lazy. I'll incorporate @hvacengi's suggestion.

doc/source/math/direction.rst
```diff
@@ -211,31 +213,30 @@ Structure
     
     :struct: Gives a `Direction` with the opposite rotation around its axes.
     
-
-.. note:: **The difference between a :struct:`Direction` and a ``Vector``**
-
-    ``Vector`` and a :struct:`Direction` can be represented with the exact same amount of information: a tuple of 3 floating point numbers. So you might wonder why it is that a ``Vector`` can hold information about the magnitude of the line segment, while a :struct:`Direction` cannot, given that both have the same amount of information. The answer is that a :struct:`Direction` does contain one thing a ``Vector`` does not. A :struct:`Direction` knows which way is "up", while a ``Vector`` does not. If you tell **kOS** to ``LOCK STEERING`` to a ``Vector``, it will be able to point the nose of the vessel in the correct direction, but won't know which way you want the roof of the craft rotated to. This works fine for axial symmetrical rockets but can be a problem for airplanes.
-
```
jacktose: Sure thing.
I figured the *Vectors and Directions* section was its natural title, but if you think it needs a `note`, no problem.

doc/source/math/geocoordinates.rst
```diff
@@ -107,40 +107,40 @@ Structure
...
-                                    // surface directly underneath the
-                                    // current ship
+    // Make spot into a location on the surface directly underneath the current ship:
+    SET spot TO SHIP:GEOPOSITION.
                                     
-    SET spot TO LATLNG(spot:LAT,spot:LNG+5). // Make spot into a new
-                                             // location 5  degrees east
-                                             // of the old one
+    // Make spot into a new location 5 degrees east of the old one:
+    SET spot TO LATLNG(spot:LAT,spot:LNG+5).
```
jacktose: I've always understood that full-line comments are preferable to to inline comments, and especially so with line wrapping because they get broken into so many little chunks and take up a lot of vertical space.

If there's a convention here that I'm bucking, it was unintentional and I'll revert it.

doc/source/math/ref_frame.rst
```diff
@@ -19,14 +19,14 @@ documentation:
     system.  kOS inherits this behavior from KSP.
 
 In all the reference frames mentioned below, the orientation of 
-the axes is **left-handed**.  This means that if you imagine opening your
+the axes is **left-handed**. This means that if you imagine opening your
```
jacktose: I also learned double-spacing in school and fought for it valiantly until a coworker finally convinced me with [this info][gg], or a version of that story. Realizing that almost all modern writing uses single spaces, I made myself switch. It wasn't nearly as hard as I thought.
Then I found [the rebuttal][r], and now I care both more and less. I use the single space by modern convention, but I've stopped evangelizing either way.

In any case, I don't even remember changing that intentionally and it makes no difference in source. Reverted.

[gg]: http://www.quickanddirtytips.com/education/grammar/two-spaces-after-a-period
[r]: http://www.heracliteanriver.com/?p=324

doc/source/math/ref_frame.rst
jacktose: You're right about consistency, but it still seemed awkward to me. I'm trying a different rephrasing. Hopefully it will be simpler overall.

doc/source/math/scalar.rst
jacktose: Good catch! I should have been reading more closely.

doc/source/language/features.rst
jacktose: That was probably an accident while doing the .. note thing. Switching it back.

Related: is there a convention for line breaks? There's some of everything.

jacktose:
Okay, pushed most of those changes.

doc/source/language/features.rst
```diff
@@ -166,7 +165,7 @@ programming.
 But if you are an experienced programmer you might not like this
 behavior, and there are good arguments for why you might want to
 disable it.  If you wish to do so, a syntax exists to do so called
-:ref:``NOLAZYGLOBAL``.
+:ref:`@LAZYGLOBAL OFF <lazyglobal>`.
```
jacktose: Fun fact: That was the first thing that I noticed, which sent me off on this wonderful journey of learning about sphinx and RST.

doc/source/math/direction.rst
Dunbaratu: Hmm - I don't care so much about the ``..note`` markup as the loss of the title phrase "the difference between".  If that could be incorporated into the normal formatting of the titles for the sections that would be fine.

doc/source/math/geocoordinates.rst
Dunbaratu:
```
line 1 // comment A
line 2
line 3
```
Meaning: "Line 1 is explained with comment A, while line 2 and line 3 are not commented."
```
// comment A:
line 1
line 2
line 3
```
Meaning: "comment A refers to the next upcoming block of code.  Maybe it's just referring to line 1, or maybe it's referring to the work being done collectively by lines 1, 2, and 3.  You don't know.

If there's enough room to allow for it, I prefer to put the comment at the end if it refers to just that one line only, and put it as a header on its own line only when it's referring to a multi-line chunk of code, or if it's just too long to fit (in which case I'll try to separate with vertical space after the next line to make it clear that it only refers to the next one single line - but doing this too much makes the code take lots of screen space to just say a few lines because of all the extra blank lines this requires.

But really the reason for this is because I was deliberately *avoiding* consistency in comment usage from one example to the next.  I wanted to make sure there were examples of different valid syntax formats in use throughout the examples in the docs.  When *every single example you ever see* in the manual always formats things the same way, and it's the manual that is explaining the syntax to you, it muddies the explanation of how the syntax really works.

jacktose:
That should do it.

Only problem is that I'm not somewhere where I can build it and check the results.

jacktose:
Okay, built and spot-checked. Looks good to me. Have at it.

doc/source/math/geocoordinates.rst
```diff
@@ -119,7 +119,7 @@ Example Usage
     PRINT spot:LNG.                 // Print 20
     
     PRINT spot:DISTANCE.            // Print distance from vessel to x
-    
+                                    // (same altitude is presumed)
```
Dunbaratu: This is false.
It does not presume the spot is at the same altitude as the vessel.  It presumes the spot is at the terrain surface.

jacktose: Oh, huh! The deletion was a mistake, but I'll re-do it.

jacktose: Re-removed in the commit below.

PR #1829, opened 2016-10-01
Fixes #1790 timewarp enhancements

### LINKED ISSUE #1790: No way to detect if warp is really ended.
WAIT UNTIL WARP = 0. will detect when the game's warp factor has been commanded to 0, but it is not useful for determining when the warp has actually finished. The game takes a while (sometimes minutes of game time for long warps) to actually stop warping.

This is pretty important because many things like :MEANANOMALYATEPOCH, positionat(), velocityat() etc, are completely wrong during warps, not to mention doing a normal WAIT <seconds> that is shorter than the time the game takes to stop warping will not wait at all.

I've been working on a workaround, but I haven't found one yet, unfortunately.

Dunbaratu:
Okay I think this is working now.  Give it a go.


Dunbaratu:
Okay,  NOW it's ready (I just edited the documentation several times upon re-reading it this morning.)

PR #1830, opened 2016-10-02
Gives some keyboard input abilities.

Dunbaratu:
Note I didn't add the ability to get a whole string at a time
yet, but that's an uglier problem and what's here is good
by itself as a useful feature, so it's worth looking at it
as a separate PR without line input.

In the end I decided to go with this method and abandon the
"everything is a file" approach to key input, because the way
the file reading libraries in kOS got made just aren't
compatible with this style of thinking.

I also abandoned the notion of handling input on top of
the comms system because I realized that would prevent it
from working in a command-line only way (it would have
tied user input to the Vessel and PartModule objects and
thus prevented it from being in kOS.Safe).

### DIFF
--- a/src/kOS.Safe/Screen/IInterpreter.cs
+++ b/src/kOS.Safe/Screen/IInterpreter.cs
@@ -9,7 +9,7 @@ namespace kOS.Safe.Screen
         string GetCommandHistoryAbsolute(int absoluteIndex);
         void SetInputLock(bool isLocked);
         bool IsAtStartOfCommand();
-        bool isWaitingForCommand();
+        bool IsWaitingForCommand();
         void Reset();
     }
 }
--- a/src/kOS.Safe/Screen/IInterpreter.cs
+++ b/src/kOS.Safe/Screen/IInterpreter.cs
@@ -9,6 +9,7 @@ namespace kOS.Safe.Screen
         string GetCommandHistoryAbsolute(int absoluteIndex);
         void SetInputLock(bool isLocked);
         bool IsAtStartOfCommand();
+        bool isWaitingForCommand();
         void Reset();
     }
 }
--- a/src/kOS/Screen/Interpreter.cs
+++ b/src/kOS/Screen/Interpreter.cs
@@ -157,6 +157,12 @@ namespace kOS.Screen
             }
         }
 
+        public bool isWaitingForCommand()
+        {
+            IProgramContext context = ((CPU)Shared.Cpu).GetInterpreterContext();
+            return context.Program[context.InstructionPointer] is OpcodeEOF;
+        }
+
         public void SetInputLock(bool isLocked)
         {
             locked = isLocked;
--- a/src/kOS/Screen/Interpreter.cs
+++ b/src/kOS/Screen/Interpreter.cs
@@ -157,7 +157,7 @@ namespace kOS.Screen
             }
         }
 
-        public bool isWaitingForCommand()
+        public bool IsWaitingForCommand()
         {
             IProgramContext context = ((CPU)Shared.Cpu).GetInterpreterContext();
             return context.Program[context.InstructionPointer] is OpcodeEOF;
--- a/src/kOS/Screen/TermWindow.cs
+++ b/src/kOS/Screen/TermWindow.cs
@@ -295,6 +295,8 @@ namespace kOS.Screen
         void OnGUI()
         {
             if (!IsOpen) return;
+            
+            ProcessUnconsumedInput();
 
             if (isLocked) ProcessKeyEvents();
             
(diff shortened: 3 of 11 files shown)


src/kOS/Screen/TermWindow.cs
```diff
@@ -625,21 +625,36 @@ public void ProcessOneInputChar(char ch, TelnetSingletonServer whichTelnet)
...
+                }
+                else 
+                {
+                    shared.Screen.CharInputQueue.Enqueue(ch);
+                }
             }
         }
 
         void SpecialKey(char key)
         {
```
Dunbaratu: One consequence of putting the hook here is that the "cooking" of the input into more vanilla ascii has already happened in TermWindow.ProcessOneInputChar() at this point.  (Therefore people don't get UnicodeCommand.DELETELEFT.  Instead they get `0x0008` (control-H, the more vanilla ASCII way to say backspace).)  I'm not sure if this is good or bad.

Dunbaratu:
I didn't associate this PR with an issue yet because I know there's been a few times in the past the issue was raised and I was going to do a search to find all of them and close them down after this is merged.

src/kOS/Screen/TermWindow.cs
```diff
@@ -625,21 +625,36 @@ public void ProcessOneInputChar(char ch, TelnetSingletonServer whichTelnet)
...
+            if (shared != null)
             {
-                shared.Interpreter.Type(ch);
-                if (IsOpen && keyClickEnabled)
-                    shared.SoundMaker.BeginSound("click");
+                if (shared.Interpreter != null && shared.Interpreter.isWaitingForCommand())
+                {
+                    shared.Interpreter.Type(ch);
+                    if (IsOpen && keyClickEnabled)
+                         shared.SoundMaker.BeginSound("click");
```
Dunbaratu: Ah darn.  I just realized the way this is implemented, the click only happens when sending the key to the interpreter to process.  This clicker gets skipped when sending the key to a user program to process in the else clause below.

I'm pretty sure very few people are using the key clicker, but for the few who are, I should probably change this.  the key click is supposed to be a feature of the terminal itself, not of which program is running inside it.

src/kOS.Safe/Screen/IInterpreter.cs
```diff
@@ -9,6 +9,7 @@ public interface IInterpreter : IScreenBuffer
         string GetCommandHistoryAbsolute(int absoluteIndex);
         void SetInputLock(bool isLocked);
         bool IsAtStartOfCommand();
+        bool isWaitingForCommand();
```
Dunbaratu: Arrrrg, the "is" should be capitalized here.  I just came off of doing some Java programming and had a different set of coding standards in my head.  (Typically in most Java coding standards, you capitalize types and lowercase members.)

src/kOS/Screen/Interpreter.cs
```diff
@@ -157,6 +157,12 @@ protected void CompileCommand(string commandText)
             }
         }
 
+        public bool isWaitingForCommand()
+        {
+            IProgramContext context = ((CPU)Shared.Cpu).GetInterpreterContext();
+            return context.Program[context.InstructionPointer] is OpcodeEOF;
+        }
+
```
Dunbaratu: I just pushed a new commit that fixed the above so it does the more elegant solution mentioned.

hvacengi:
Haven't had a chance to track down the cause yet, but it appears that terminal input is broken if executed from a boot script.  At first boot (loading the vessel) the program appears to keep waiting for input, never seeming to recognize any keystrokes.  Breaking execution (Ctrl+C) and rebooting allows you to type at the terminal, but it appears to be an interactive terminal and again the input structure doesn't recognize keystrokes (though typing giberish will give an error message).  If I break execution again and manually run the boot script, I don't see any of the above issues.

For reference, the script I am using for the test is attached below:
```
wait 1.
print "Press any key to begin input test...".
global input is terminal:input:getchar().
global done is false.
print "Input will be echoed back to you.  Press q to quit".
until done {
    if (terminal:input:haschar) {
        set input to terminal:input:getchar().
        if input = "q" {
            set done to true.
        }
        else {
            print "Input read was: " + input.
        }
    }
}
```

hvacengi:
I've got the input in boot scripts fixed now.

But I then found out that some of those special characters (light "right arrow") crash the UI due to an array out of range exception.  Working on that now.

hvacengi:
Fixed the special characters now too.  Merging shortly.

Dunbaratu:
Can you post a branch with the fixes you've got so far?  I have time over the next couple of hours to look at this.

Dunbaratu:
Although I am curious what you changed from what I had that fixed the boot issue.  I don't see it as a separate commit in the PR record so I don't know what you changed.

hvacengi:
https://github.com/KSP-KOS/KOS/commit/63a7240f7bfa648c2c2f952707a10eb9be847059

PR #1837, opened 2016-10-10
KSP 1.2 Support

### LINKED ISSUE #1835: Support for KSP 1.2
Simple placeholder issue for compiling against KSP 1.2 and making necessary modifications to build.

Notable TODO notes following my first push:
- [x] TransferManager.cs https://github.com/hvacengi/KOS/blob/experimentals/src/kOS/Execution/TransferManager.cs#L53
- [ ] ActiveResourceValue.cs https://github.com/hvacengi/KOS/blob/experimentals/src/kOS/Suffixed/ActiveResourceValue.cs#L30
- [ ] ScienceDataValue.cs https://github.com/hvacengi/KOS/blob/experimentals/src/kOS/Suffixed/ScienceDataValue.cs#L39
- [x] Utils.cs https://github.com/hvacengi/KOS/blob/experimentals/src/kOS/Utilities/Utils.cs#L317
- [x] VesselUtils.cs https://github.com/hvacengi/KOS/blob/experimentals/src/kOS/Utilities/VesselUtils.cs#L13

hvacengi:
Partial work towards #1835 

Code modifications to compile and run using KSP 1.2.  No new feature integration is included, simply modifications necessary to build, test, and run with KSP version 1.2.


src/kOS/Suffixed/ActiveResourceValue.cs
```diff
@@ -7,28 +7,55 @@ namespace kOS.Suffixed
...
-            AddSuffix("AMOUNT", new Suffix<ScalarValue>(() => activeResource.amount, "The resources currently available"));
-            AddSuffix("CAPACITY", new Suffix<ScalarValue>(() => activeResource.maxAmount, "The total storage capacity currently available"));
-            AddSuffix("DENSITY", new Suffix<ScalarValue>(() => activeResource.info.density, "The density of this resource"));
-            AddSuffix("PARTS", new Suffix<ListValue<PartValue>>(() => PartValueFactory.ConstructGeneric(activeResource.parts, shared), "The containers for this resource"));
+            AddSuffix("NAME", new Suffix<StringValue>(GetName, "The name of the resource (eg LiguidFuel, ElectricCharge)"));
+            AddSuffix("AMOUNT", new Suffix<ScalarValue>(GetAmount, "The resources currently available"));
+            AddSuffix("CAPACITY", new Suffix<ScalarValue>(GetCapacity, "The total storage capacity currently available"));
+            AddSuffix("DENSITY", new Suffix<ScalarValue>(GetDensity, "The density of this resource"));
+            // TODO: IMPORTANT fix this before release, though it isn't a documented suffix and there may be value to unifying the resource classes
```
Dunbaratu: Signposting this TODO with a comment to make it stand out.

src/kOS/Suffixed/ActiveResourceValue.cs
```diff
@@ -7,28 +7,55 @@ namespace kOS.Suffixed
...
+
+        public ScalarValue GetCapacity()
+        {
+            vessel.GetConnectedResourceTotals(resourceId, out amount, out capacity, true);
+            return capacity;
+        }
+
+        public ScalarValue GetDensity()
+        {
+            return PartResourceLibrary.Instance.resourceDefinitions[resourceId].density;
```
Dunbaratu: So if I understand the gist of this change: you're getting the values live on the fly now from the vessel instead of from buffered data in the activeResource object.  I assume this fixes a problem with the activeResource data being stale in some way?

src/kOS/Suffixed/StageValues.cs
```diff
@@ -31,81 +35,80 @@ private void InitializeSuffixes()
...
+        public void CreatePartSet()
+        {
+            // The following replicates the logic in KSP.UI.Screens.ResourceDisplay.CreateResourceList
+            // We're creating the set every time because it doesn't pay attention to the various events
+            // that would tell us that the old partset is no longer valid.
+            partHash.Clear();
+            int vstgComp = shared.Vessel.currentStage - 2;
+            var parts = shared.Vessel.Parts;
+            global::Part part;
+            for (int i = parts.Count; i-- > 0;)
```
Dunbaratu: I think this would be a lot clearer what your intent was: 
``for (int i = parts.Count - 1 ; i >= 0 ; --i)``
That makes it a lot more straightforward the this is a loop that goes from n-1 through 0.  Right now that behavior is a bit obfuscated behind the side-effecting of the boolean check.  The initializer says ``parts.Count`` despite the fact that the first iteration will actually be using ``parts.Count-1`` due to that side-effect.  The check says ``>0`` despite the fact that zero will be included within the iterations (because it checks the value of ``i`` prior to the decrement, ``>0`` *really* ends up meaning more like ``>=0``.)

src/kOS/Utilities/Utils.cs
```diff
@@ -275,5 +275,60 @@ public static CelestialBody KOSExtensionGetParentBody(this CelestialBody body)
...
+        /// <summary>
+        /// Determines if a given case insensitive name corresponds with a defined resource, and
+        /// outputs the case sensitive name for easy access from Squad's lists.
+        /// </summary>
+        /// <param name="insensitiveName">case insensitive name</param>
+        /// <param name="fixedName">output case sensitive name</param>
+        /// <returns>true if a matching resource definition is found</returns>
+        public static bool IsResource(string insensitiveName, out string fixedName)
+        {
+            // eliminate enumerators, use index based access
```
Dunbaratu: This comment looks like the opposite of what you did.  It looks like you commented out the index-based access and kept the enumerator version.

Or is this meant to be a TODO comment but it doesn't say "TODO"?

src/kOS/Utilities/Utils.cs
```diff
@@ -275,5 +275,60 @@ public static CelestialBody KOSExtensionGetParentBody(this CelestialBody body)
...
+        /// <summary>
+        /// Determines if a given case insensitive name corresponds with a defined resource, and
+        /// outputs the resource id for easy access from Squad's lists.
+        /// </summary>
+        /// <param name="insensitiveName">case insensitive name</param>
+        /// <param name="foundId">output id of the found resource, zero if none found</param>
+        /// <returns>true if a matching resource definition is found</returns>
+        public static bool IsResource(string insensitiveName, out int foundId)
+        {
+            // eliminate enumerators, use index based access
```
Dunbaratu: Same comment as above.

src/kOS/Utilities/VesselUtils.cs
```diff
@@ -10,9 +10,11 @@ namespace kOS.Utilities
 {
     public static class VesselUtils
     {
+        // TODO: Can we just rememove this method?  It doesn't appear to be referenced anywhere, and the test is more complicated now.
```
Dunbaratu: If you can't find it being called by anything, then sure.

src/kOS/Suffixed/StageValues.cs
```diff
@@ -6,17 +6,21 @@
...
+using kOS.Module;
 
 namespace kOS.Suffixed
 {
     [kOS.Safe.Utilities.KOSNomenclature("Stage")]
     public class StageValues : Structure
     {
         private readonly SharedObjects shared;
+        private HashSet<global::Part> partHash = new HashSet<global::Part>();
+        private PartSet partSet;
```
Dunbaratu: Is this a new thing from the base game API?  If so I like it.  I could think of a few other places we could use it - like mapping a name tag to a PartSet in KOSNameTag to make managing the data a bit cleaner.

src/kOS/Function/PrintList.cs
```diff
@@ -189,8 +189,10 @@ private kList GetResourceList(SharedObjects shared)
             foreach (Part part in shared.Vessel.Parts)
             {
                 string stageStr = part.inverseStage.ToString();
-                foreach (PartResource resource in part.Resources)
+                PartResource resource;
+                for (int i = 0; i < part.Resources.Count; ++i)
```
hvacengi: Squad removed a lot of enumerators.  Then they put some back in... I didn't go back and find every time that enumerators got re-added however.

Essentially thee idea is that frequent use of enumerators is expensive, because the enumerator needs to allocate a new object when it's used.  For most of our stuff, it isn't a big deal because we don't use them multiple times per tick and because we compile using MS's compiler.  The Mono GC is still a bit of an issue too.  But for resources it's a really big deal for Squad, since they may need to call the parts walk many times per tick (at least once per engine).  Using a traditional for loop only allocates the integer, assuming the list is already constructed.

src/kOS/Module/kOSLightModule.cs
```diff
@@ -28,23 +28,24 @@ internal class kOSLightModule : PartModule
...
         private double lastTime = double.MaxValue;
+        private Color lastColor = new Color(0, 0, 0, 0);
 
         private readonly Color powerOffColor = new Color(0, 0, 0, 0);
 
         private ModuleLight lightModule;
-        private Light[] lights;
-        private Renderer[] renderers;
+        private List<Light> lights;
+        private List<Renderer> renderers;
```
hvacengi: In this particular case, I see no benefit.  I don't know why Squad did it, but they did.

src/kOS/Function/PrintList.cs
Dunbaratu: I actually prefer the loop indexing myself because iterators seem like unneeded overhead most of the time.  I was just wondering if the change was required because the iterator no longer existed, or if it was done just for performance reasons.

src/kOS/Suffixed/ActiveResourceValue.cs
hvacengi: Well, what it really fixes is the fact that KSP no longer has a class for `ActiveResource`, and this is the recommended way around it.  The ID lookup uses a dictionary, so it's actually pretty fast.  But yes, in theory we should be avoiding stale data.  The only thing `ActiveResourceValue` is used for is stage values anyways, so I thought about just merging the function into another class if I could do so easily.

src/kOS/Module/kOSLightModule.cs
Dunbaratu: I thought this was in our own code, not SQUAD's.  Or are we hooking into something in the base PartModule here?

src/kOS/Suffixed/StageValues.cs
hvacengi: Yeah, this is their new more generic fix for removing active resources.  I like it too.  I'm not sure I see how we can integrate it with the name tag, but if you've got an idea I'm all for it.

src/kOS/Module/kOSVesselModule.cs
```diff
@@ -67,7 +68,7 @@ public override void OnAwake()
         /// KSP after Awake.  All parts should be added to the vessel now, so it is safe to walk the
         /// parts tree to find the attached kOSProcessor modules.
         /// </summary>
-        public void Start()
+        protected override void OnStart()
```
Dunbaratu: It might be an effect of the fact that Unity views C# as an "also" supported language rather than having been written explicitly for it and it alone.  Using its inheritence rules for base functionality in the API might have made it hard to implement things for other languages.

src/kOS/Utilities/Utils.cs
hvacengi: Oh yeah!  I tried to convert away from an enumerator at first, but then found out it doesn't support indexes.  Then I expected them to finish the enumerator purge so I left the code ready.  It never got purged apparently.  I'll remove the comment.

src/kOS/Module/kOSLightModule.cs
hvacengi: It's our own definition for the field, but Squad changed the type of the methods we use to populate the field (I have no idea why Animations didn't get changed too, see the next line).  We could have continued to re-cast back to an array, but at that point it's a coin flip which is better on our end.

src/kOS/Suffixed/StageValues.cs
hvacengi: Oh, yeah.  I only did that method because I literally grabbed it from Squad's code.

Dunbaratu:
I'm okay with the things I see here (maybe except for re-writing that one for loop header to look more normal and remove the boolean check's side effect increment).  But I don't want to merge because the PR still has the "not ready" label.

I'd be ready to merge it tomorrow if you decide to remove the label.

hvacengi:
I was gonna push an update a little later tonight addressing your couple of comments, getting travis fixed, and updating the .version file so we can tag a pre-release with it.

Does this mean you've been running the branch and haven't seen any issues, or just via code review?

Dunbaratu:
I have started running it a bit.  I won't be getting much further with that this afternoon because I've decided to try Shenzhen I/O on the stream and I made arrangements for it to be in this time slot (can't move it).

PR #1842, opened 2016-10-14
Fixed forum links

### LINKED ISSUE #1841: Readme.md forum links broken
Not sure if it's just me and hopefully it's temporary, but I'm guessing not. KSP forum links now appear to need the full link.
e.g:
http://forum.kerbalspaceprogram.com/index.php?/topic/61827-113-kos-scriptable-autopilot-system-v101-20161011
instead of 
http://forum.kerbalspaceprogram.com/index.php?/topic/61834-/
Which is going to be a bummer if they need changing every time the title changes with an update.

Going to look through the forum and see if i can find anything on it there.

TDW89:
Fixes #1841 
Added an extra `-` to the end of the ksp forum links as for some reason they need to have something after the first one to go to the thread instead of the forum home page.
The `-` can be swapped out for anything else I just thought that it looked neater.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,9 +5,9 @@ kOS is a scriptable autopilot Mod for Kerbal Space Program. It allows you write 
 
 Documentation : http://ksp-kos.github.io/KOS_DOC/
 
-Addon release thread: http://forum.kerbalspaceprogram.com/index.php?/topic/61827-/
+Addon release thread: http://forum.kerbalspaceprogram.com/index.php?/topic/61827--/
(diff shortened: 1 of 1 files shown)


hvacengi:
Ironically when I fixed those links last time (right after they updated the forum software) I was complaining because the required the dash after the thread number.  Whatever they're using to parse their url's is apparently very rigid.  I absolutely hate that SEO has to get in the way of a perfectly legible/predictable URL.

PR #1843, opened 2016-10-16
Difficulty settings

hvacengi:
Related to #1835

Migrates Config values to use KSP's new CustomParameters and embed some settings into the dificulty settings menu.


src/kOS/Module/kOSCustomParameters.cs
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        public override bool Enabled(MemberInfo member, GameParameters parameters)
+        {
+            if (member.Name == "migrated" || member.Name == "version")
+            {
+                return false;
+            }
+            return base.Enabled(member, parameters);
+        }
```
Dunbaratu: What does this do (sadly I don't know the stock API for how to use this new class type)?  Are you saying if the setting is one that was populated by migrating it from the older config that this means the user can't edit it post-migration?

Dunbaratu: grrrr. dangit github.  I deleted the above comment - I see how it works now.  But github kept the comment in its review anyway.  Ignore it.

Dunbaratu:
In testing this, I decided I don't like how the KOS settings are *only* editable during the initial creation of a new career.  Like you can't change IPU now without starting a new career.  Is there a setting we can fiddle with to make these appear on the in-game settings menu and not just on the "starting new career" menu?

src/kOS/Module/kOSCustomParameters.cs
```diff
@@ -0,0 +1,187 @@
...
+        public bool enableRTIntegration = true;
+
+        [GameParameters.CustomParameterUI("Start on the archive")]
+        public bool startOnArchive = false;
+
+        [GameParameters.CustomParameterUI("Obey hide UI toggle")]
+        public bool obeyHideUi = true;
+
+        [GameParameters.CustomParameterUI("Enable safe mode")]
+        public bool enableSafeMode = true;
```
hvacengi: We can definitely do that with tooltips.  I think that it's another parameter on the attribute.

hvacengi:
You can change the settings by bringing up the menu, selecting "difficulty" and then picking the "kos" tab.  The difficulty menu is not hidden when in game.  We have the option of specifically setting some settings as read only any time the settings are displayed, and in the connectivity pull request I use this feature to make it so that you cannot change the "connectivity manager" during flight.

hvacengi:
Or is editing difficulty settings blocked in "hard" mode?

Dunbaratu:
Can you show me a screenshot of where you're changing the settings post-start-of-career?  I can't seem to find it.

hvacengi:
After YouTube finishes processing the video you should be able to find it here: https://youtu.be/rDSkA5WYbOc

Dunbaratu:
Ohhh.  I never realized that "Difficulty Options; Custom" was a clickable button.  I thought it was just a passive label and never tried clicking it.

Dunbaratu:
I'm going to take a stab at writing those tooltip popups, and maybe changing the wording of the initial migration message to help direct people to where the new place to find the migrated settings is.

doc/source/general/settingsWindows.rst
```diff
@@ -0,0 +1,82 @@
...
+    .. versionadded:: v1.0.2
+        Previous versions of kOS kept all settings accessible from the App
+        Launcher Window.  KSP version 1.2.0 introduced a new way to
+        store settings within the save file itself, and most settings were
+        migrated to this system/window.
+
+This settings window is accessible when you first start a new game by clicking
+on "Difficulty Options", or in an existing game by clicking on "Difficulty
+Options" from the in game settings menu (accessed by pressing the :kbd:`Escape`
+key, and then clicking "Settings" from the pop up window).
```
Dunbaratu: I want to add a sentence here to make it absolutely clear that these setttings aren't about difficulty at all and that this is just the only place where KSP supports mods adding their own settings in the UI.

doc/source/structures/misc/config.rst
```diff
@@ -6,11 +6,35 @@ Configuration of kOS
...
+            store settings within the save file itself, and most settings were
+            migrated to this system.
+
+    .. note::
+        If your save file has not yet migrated to the new settings storage
+        system and an old config file is present, you will be prompted with a
+        dialog box offering to migrate the old settings or use the defaults.
+        You may also choose to prevent further attempts to migrate settings,
+        which will instruct kOS to set the ``InstructionsPerUpdate`` to a
+        negative value.
```
Dunbaratu: Our docs never explained anywhere that setting it to -1 is just how we record the fact that the settings are now defunct.  I think it makes more sense to mention that (that the settings are flagged as "do not use"), than mentioning the mechanism by which we do it and not explaining what that means.

Issue #1844, opened 2016-10-16
Support for more Unicode characters


khaiduk:
I created font_sml.png with support for more characters. It is able to handle (with only minor changes in source code) box drawing charaters and some national characters.

![All characters](https://cloud.githubusercontent.com/assets/2662789/19418340/32d5c198-93c2-11e6-9247-0e7252eda90a.png)

Some characters are from MrOnak's font_sml.png (https://github.com/KSP-KOS/KOS/issues/159).
Improvments can be done, such as moving box-drawing characters to Box Drawing Unicode block, but for it would require better solution for holding font data - maybe something like Unifont hex format.

Here is some live example:
![Sample usage of extended font_sml.png](https://cloud.githubusercontent.com/assets/2662789/19418313/a9dc31ec-93c1-11e6-82e0-ff8b79861c66.png)

Dunbaratu:
My reason for not having done this earlier is the conflict between codepage 437 and Unicode.  Those box-drawing characters being at that position of the code is not true in Unicode, where they are much higher up.  If I did a thing like this and released it then people would start writing scripts that assume things like ``char(129)`` and ``char(130)`` are box drawing chars, and this assumption would break if we ever got the ability to just use a normal monospace unicode font.  The only reason we have to use this image texture template method is because KSP ships without a monospace unicode font and Unity didn't allow you to add your own fonts unless you had the full paid version.

But there's a possibility of changing that in the future.  There's an alternate way the entire terminal could be implemented where it would use a proper unicode font.   The hope that we may do that some day is sort of why we haven't supported this yet - because the char() codes would be all different once it's fully supported.

But I realize that putting off support for a thing now, in favor of a far-off feature that may come 'some day' is really annoying, so let's try to come up with a possible solution here that would be future-compatible.

All we'd have to do to be able to use this idea as a temporary stopgap would be to change how the system reads the font image PNG file so it doesn't presume all the pictures are consecutive adjacent codes.  If we offered the ability for the image file to contain "skips" or a mapping of some sort (i.e. the image at position 133 isn't really char(133) but some other number.) then we could support this now but not break future compatibility.

How about a secondary file alongside the PNG that contains this information?  Maybe some cheap text table format that just says things like:
```
139 9472
197 9532
```
etc to mean "position 139 is really unicode 9472" and so on.

khaiduk:
[...]

PR #1845, opened 2016-10-16
Added support for more Unicode characters

khaiduk:
Related to https://github.com/KSP-KOS/KOS/issues/1844

Updated font_sml.png:
Added charcters from Latin-1 Supplement and Latin Extended-A Unicode blocks and some box-drawing characters in C1 control codes range (0x80 - 0x9F) that can be accessed with char() function. 
Removed unused MrOnak's arrow characters (kOS strips characters below 0x20) and booting animation (not used, replaced by Latin-1 characters).
Added replacement character at 0x02

Adjustmented TermWindow.cs to support bigger font_sml.png.


hvacengi:
I want to thank you for submitting this pull request and taking the time to come up with a solution.  Per the conversation in #1844 @Dunbaratu has implemented a fix that moves away from adding additional glyphs to the texture and instead provides unicode support through the system fonts.  I'm closing this PR in anticipation of merging #1948 later this week.

Thanks again for the submission!

PR #1850, opened 2016-10-20
CommNet support for KSP 1.2.0

hvacengi:
An overhaul of the connectivity system to accommodate more abstraction and support CommNet along with RemoteTech for communication enforcement.

I suggest that any reviewers read through the commit messages as I explained pretty thoroughly what I was changing and why.

Essentially there is now a `ConnectivityManager` class which exposes static methods to access instances of `IConnectivityManager` which enforce connection.  All other changes were made to centralize all of this logic so that it didn't need to be spread across many classes, nor did it need to be tied to `SharedObjects`.

This is built on top of #1843 because I'm further extending the settings system.

A few important notes:
- I included a hack to make things work with "Require Signal for Control" enabled with CommNet.  I have not done substantial testing to make sure that this doesn't break control lockout somewhere else in KSP.  In fact, if we would like we can move this logic into it's own "addon" dll to make it easy to opt in/out of.  Or we could remove the hack and just say that kOS doesn't support that setting.
- At least one of the two ships you are testing communication between must have a relay antenna.  This is a limitation on the part of KSP, where they do not calculate connections between unloaded vessels that have no relay.  I ran into some fringe cases where it works one direction but not the other, so I set up this logic to try reversing the connection if no connection is shown initially.


hvacengi:
@Dunbaratu I've changed this label from "Not Ready" to "Pending Documentation".  I will work on revising the docs tonight, but it should be pretty straight forward as I only need to document two new bound variables (`homeconnection` and `controlconnection`), the 3 options for Connectivity Managers, and the new elements on the settings window.  If you want to start your code review before I have the docs finished, feel free.  But if you'd rather wait until the docs are done to review I'm fine with that too.

Dunbaratu:
I might be able to start looking at it tomorrow, with or without the docs being done.

src/kOS/Module/kOSVesselModule.cs
```diff
@@ -237,6 +237,15 @@ private void AddDefaultParameters()
...
         {
             ConnectivityManager.AddAutopilotHook(parentVessel, UpdateAutopilot);
+
+            // HACK: The following events and their methods are a hack to work around KSP's limitation that
+            // blocks our autopilot from having control of the vessel if out of signal and the setting
+            // "Require Signal for Control" is enabled.  It's very hacky, and my have unexpected results.
+            if (Vessel.vesselType != VesselType.Unknown && Vessel.vesselType != VesselType.SpaceObject)
+            {
+                TimingManager.FixedUpdateAdd(TimingManager.TimingStage.ObscenelyEarly, cacheControllable);
+                TimingManager.FixedUpdateAdd(TimingManager.TimingStage.BetterLateThanNever, resetControllable);
```
Dunbaratu: I assumed these looked like names you'd made up yourself.  Then I pulled this into the IDE and found out they're part of stock.  Are these seriously the actual names they use in stock for these timings?  i.e. "FashionablyLate"?  Oh well.

src/kOS/Module/kOSVesselModule.cs
```diff
@@ -245,8 +254,51 @@ private void HookEvents()
...
+                isControllableField.SetValue(parentVessel, true);
+                workAroundControllable = true;
+            }
+            else
+            {
+                workAroundControllable = false;
+            }
+        }
+        #endregion
+
```
Dunbaratu: Wow you're right this *IS* a nasty hack.  It depends on knowing exactly when stock chooses to run its checks for seeing if the vessel is controllable, so you can change it only when it's about to check, and then immediately put it back the way it was afterward.  If they ever do anything that fiddles with those timings it breaks.  

This looks VERY fragile.  I really hope they get this issue fixed for us in stock so we don't need to do something so invasive and can get rid of this as soon as possible.  But yeah, for now we sort of have no choice and have to do it.

src/kOS/Binding/BindingsUniverse.cs
```diff
@@ -6,16 +6,21 @@
...
     {
+        private HomeConnection homeConnection;
+        private ControlConnection controlConnection;
+
         public override void AddTo(SharedObjects shared)
         {
             shared.BindingMgr.AddGetter("KUNIVERSE", () => new KUniverseValue(shared));
-
+            shared.BindingMgr.AddGetter("HOMECONNECTION", () => homeConnection ?? (homeConnection = new HomeConnection(shared)));
+            shared.BindingMgr.AddGetter("CONTROLCONNECTION", () => controlConnection ?? (controlConnection = new ControlConnection(shared)));
```
Dunbaratu: So are these meant to be inter-vessel connections you can send messages on in scripts?  I guess I'll have to wait until I get to the later commit where the docs are.

doc/source/commands/communication.rst
```diff
@@ -143,3 +143,68 @@ The receiving CPU will use :attr:`CORE:MESSAGES` to access its message queue::
...
+becomes available, you will be prompted with a dialog box to select the manager
+you want to use.
+
+.. object:: Connectivity Managers
+
+    By default kOS supports the following list of managers. If a manager is not
+    currently available (because the required mod isn't installed, or the system
+    is disabled in the settings) it will not be shown in lists of available managers.
+
+    StockConnectivityManager
```
Dunbaratu: Again, same comment as in the code review about using the word "stock" to mean "not CommNet", when "CommNet" is a stock feature - its a bit confusing.

doc/source/commands/communication.rst
```diff
@@ -143,3 +143,68 @@ The receiving CPU will use :attr:`CORE:MESSAGES` to access its message queue::
...
+    CommNetConnectivityManager
+      This manager will use KSP's stock CommNet implementation to monitor
+      connections.  CommNet has limitations on updating connections for vessels
+      which are not the active vessel.  The best way to ensure that a connection
+      is updated is to include a relay antenna on one or both of the vessels.
+    RemoteTechConnectivityManager
+      This manager will use the RemoteTech mod to monitor connections. It will
+      only be available if RemoteTech is installed.  You can access more
+      detailed information and methods using the :ref:`RemoteTech Addon<remotetech>`
+
```
Dunbaratu: Perhaps somewhere we need to mention the fact that the current RemoteTech implementation doesn't just ignore CommNet, but in fact *requires that you disable* CommNet because they step on each other's toes.  Just selecting RemoteTechConnectivityManager alone doesn't make things work if you happened to have picked "hard mode" when starting a career and just accepting the default "hard mode" options.  (The default turns on parts of CommNet, and thus steps on RT's toes.  To use the current RT that hasn't yet been fully integrated into CommNet, you have to go in to the options and turn all of CommNet *off*.  If they're both on, it messes up, and having CommNet turn on when you didn't realize it was on is easy for a player to do - because it comes on by default with hard mode.  Granted, this is technically an entirely RT issue, not kOS, but if we're going to say that we "work" with RT it's going to affect our users too so maybe we should mention it as well in our docs.)

I'd like to pro-actively stop the complaints that might otherwise come in about how "our" mod is broken.

doc/source/commands/communication.rst
```diff
@@ -143,3 +143,68 @@ The receiving CPU will use :attr:`CORE:MESSAGES` to access its message queue::
...
+monitor the basic network status using a single unified system, regardless of
+which connectivity manager is selected.
+
+.. global:: HOMECONNECTION
+
+    Returns a :struct:`Connection` representing the :ref:`CPU Vessel's<cpu vessel>`
+    communication line to a network "home" node.  This home node may be the KSC
+    ground station, or other ground stations added by the CommNet settings or
+    RemoteTech.
+
```
Dunbaratu: It would help a lot to explain to users how you intended for them to use this - maybe by giving a small example that makes the idea concrete in their heads.  Do you want people to be able to send a message "from home" to the vessel in the same way as it could have been sent from vessel to vessel, as laying the groundwork for having a "home mainframe" later?

src/kOS/Communication/CommNetConnectivityManager.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            if (!IsEnabled)
+                return true;
+
+            // IsConnectedHome is only set to true on the active vessel, so we have to manually
+            // call the FindHome method to evaluate the home connection.
+
+            // WARNING: In stock this will only work for vessels with a relay antenna installed.
+            // This is a limitation put in place to improve performance in the stock game, and
+            // there isn't a very good way around it.
```
hvacengi: That is correct.  I have it documented, but I don't do it in a big "note" box, which may be a good decision.  I agree it's silly.  At least I was able to work around it a little for inter-vessel communication by trying the path both ways.

I can update the documentation.

src/kOS/Communication/ConnectivityManager.cs
```diff
@@ -1,9 +1,146 @@
...
+                {
+                    return t;
+                }
+            }
+            return null;
+        }
+
+        public static IConnectivityManager CreateManagerObject()
+        {
+            var t = GetSelectedManagerType();
```
hvacengi: Yeah, I tend to start with `var` mostly so I'm not strongly typed in case I make a change.  I didn't swap it back because the method mentions the word "Type" but I can strongly type it instead.

src/kOS/Communication/StockConnectivityManager.cs
```diff
@@ -1,14 +1,50 @@
 ﻿using System;
+using CommNet;
+using kOS.Module;
 
 namespace kOS.Communication
 {
-    public class StockConnectivityManager : ConnectivityManager
+    public class StockConnectivityManager : IConnectivityManager
```
hvacengi: Maybe `IgnoreConnectivityManager`?  I kind of thought about changing it, but couldn't come up with a good name.  So I left it for others to comment on.  I'm in favor of changing the name, I'm just not sure what a good choice would be.

src/kOS/Communication/StockConnectivityManager.cs
```diff
@@ -46,5 +54,18 @@ public bool HasConnection(Vessel vessel1, Vessel vessel2)
         {
             return true;
         }
+
+        public void AddAutopilotHook(Vessel vessel, FlightInputCallback hook)
+        {
+            // removing the callback if not already added doesn't throw an error
+            // but adding it a 2nd time will result in 2 calls.  Remove to be safe.
+            vessel.OnPreAutopilotUpdate -= hook;
+            vessel.OnPreAutopilotUpdate += hook;
```
hvacengi: The RT manager does not subscribe to the vessel event because RT surpresses all control modifications made inside of the autopilot events in favor of it's own `AddSanctionedPilot` method.  See https://github.com/KSP-KOS/KOS/pull/1850/files#diff-e76af280dcf349ffd6916450b3aa238fR79

src/kOS/Module/kOSProcessor.cs
```diff
@@ -405,19 +401,17 @@ private IEnumerable<VolumePath> BootDirectoryFiles()
...
             shared = new SharedObjects();
-            CreateFactory();
 
             shared.Vessel = vessel;
             shared.Processor = this;
             shared.KSPPart = part;
             shared.UpdateHandler = new UpdateHandler();
             shared.BindingMgr = new BindingManager(shared);
-            shared.Interpreter = shared.Factory.CreateInterpreter(shared);
+            shared.Interpreter = new Screen.ConnectivityInterpreter(shared);
```
hvacengi: Correct.  The only reason I didn't just merge the `ConnectivityInterpreter` with the `Interpreter` was to keep all of this logic separated.  Really, there isn't a major reason to keep the inheritance tree anymore.

src/kOS/Module/kOSVesselModule.cs
hvacengi: Yeah... Ironically, the underlying names that they actually use are `Timing0` and `Timing5`.  Just one of those random things Squad has done that looks hacky.

doc/source/commands/communication.rst
hvacengi: This is a complicated thing to document well.  Because RemoteTech will technically allow you to leave CommNet active, though RT will disable CommNet by default.  So at that point we are in an awkward position: we can't really support both at the same time, but they can each restrict our interaction.

A big note saying "hey, don't do something stupid" may indeed be called for.  At least then we can point to that note if someone complains.

doc/source/commands/communication.rst
hvacengi: I think that yes it lays the framework for our desired mainframe.  But for now the concrete usage is to determine if the archive is accessible.  I'll update accordingly.

src/kOS/Communication/StockConnectivityManager.cs
Dunbaratu: `PlaceholderConnectivityManager` ?

hvacengi: I kind of like `PermitAll`, or maybe `AllwaysOn` or `AllwaysTrue`.

Dunbaratu:
When you make the adjustments to documentation and the naming convention of "stock" connectivity manager, let me know and I'll push the button on the merge.  (I often don't see it when you update a PR or make a new PR against a PR branch so I might not see it if you don't make an announcement here.)

Dunbaratu:
I'll be away tonight and tomorrow night but will be back online for an hour or so late at night both nights and can do it then.

hvacengi:
I'll ping you as soon as I push those updates.  Should be shortly.

hvacengi:
@Dunbaratu should be good to go pending your final review

Dunbaratu:
Just got back now.  I'll have a look tonight.

doc/source/commands/communication.rst
```diff
@@ -198,7 +212,15 @@ which connectivity manager is selected.
     Returns a :struct:`Connection` representing the :ref:`CPU Vessel's<cpu vessel>`
     communication line to a network "home" node.  This home node may be the KSC
     ground station, or other ground stations added by the CommNet settings or
-    RemoteTech.
+    RemoteTech.  Functionally, this connection may be used to determing if the
```
Dunbaratu: "determing" - I'll correct this small typo as I merge.

doc/source/commands/communication.rst
```diff
@@ -170,23 +170,37 @@ you want to use.
...
-      which are not the active vessel.  The best way to ensure that a connection
-      is updated is to include a relay antenna on one or both of the vessels.
+      connections.  It will only be available if CommNet is enabled in the KSP
+      difficulty settings.
+
+      .. note::
+
+          CommNet has limitations on updating connections for vessels
+          which are not the active vessel.  The best way to ensure that a connection
+          is updated is to include a relay antenna on one or both of the vessels.
```
Dunbaratu: "a relay antenna" -> would be more  clear if it was phrased as "one of the kinds of antenna that can act as a relay".  Phrasing it that way makes it very clear that we are drawing a distinction between two types of antenna, (as opposed to "relay antenna versus no antenna")

If you don't mind, I'll make the edit as I merge (I hope such a small change is okay with you.  If not we can back it out.  I'd rather not go back and forth again and again multiple times on this and just hurry up and get it merged.)

hvacengi:
Thanks for making both of those edits.  Both look fine to me.

Issue #1849, opened 2016-10-20
[1.0.2-pre] ContractConfigurator kills the Addon-wrappers


GER-Space:
The next release of ContractConfigurator will not play nice with kOS wrapper functions. I got the described errors from the  IR-wrapper about every tick. 

The needed Change is described here:
https://github.com/blizzy78/ksp_toolbar/pull/39

https://github.com/blizzy78/ksp_toolbar/pull/39/files

The issue first showed up with the toolbar users, but kOS is also affected.

GER-Space:
the Error in the log is:

` 10/21/2016 11:44:43 AM,kOS-IRWrapper,Attempting to Grab IR Types...

(Filename: C:/buildslave/unity/build/artifacts/generated/common/runtime/UnityEngineDebugBindings.gen.cpp Line: 42)

NotSupportedException: The invoked member is not supported in a dynamic module.
  at System.Reflection.Emit.AssemblyBuilder.GetExportedTypes () [0x00000] in <filename unknown>:0
  at kOS.AddOns.InfernalRobotics.IRWrapper+<>c.<InitWrapper>b__39_0 (.LoadedAssembly a) [0x00000] in <filename unknown>:0
  at System.Linq.Enumerable+<CreateSelectIterator>c__Iterator10`2[AssemblyLoader+LoadedAssembly,System.Type[]].MoveNext () [0x00000] in <filename unknown>:0
  at System.Linq.Enumerable+<CreateSelectManyIterator>c__Iterator12`2[System.Type[],System.Type].MoveNext () [0x00000] in <filename unknown>:0
  at System.Linq.Enumerable.First[Type] (IEnumerable`1 source, System.Func`2 predicate, Fallback fallback) [0x00000] in <filename unknown>:0
  at System.Linq.Enumerable.FirstOrDefault[Type] (IEnumerable`1 source, System.Func`2 predicate) [0x00000] in <filename unknown>:0
  at kOS.AddOns.InfernalRobotics.IRWrapper.InitWrapper () [0x00000] in <filename unknown>:0
  at kOS.AddOns.InfernalRobotics.IRHandler.FixedUpdate () [0x00000] in <filename unknown>:0

(Filename:  Line: -1)

10/21/2016 11:44:43 AM,kOS-IRWrapper,Attempting to Grab IR Types...

(Filename: C:/buildslave/unity/build/artifacts/generated/common/runtime/UnityEngineDebugBindings.gen.cpp Line: 42)
`

[...]

PR #1851, opened 2016-10-21
Fix for issue #1849

GER-Space:
This fixes the errors I see, when ContractConfigurator is installed. #1849


src/kOS/AddOns/KerbalAlarmClock/KACWrapper.cs
```diff
@@ -55,6 +55,17 @@ public class KACWrapper
 
         public static Boolean NeedUpgrade { get; private set; }
 
+        internal static Type GetType(string name)
```
hvacengi: This doesn't really need to be `internal`, doing so in the `IRWrapper` makes a little sense because it is consistent with other wrappers, but it could simply be private for `KACWrapper`.  Not a big enough deal to require a change alone, but I'm going to request another change anyways so I thought I might as well comment.

PR #1858, opened 2016-10-28
Broken harddisk file load and gzip ksm files

### LINKED ISSUE #1857: 4 character files break harddisk loading from save game
Originally reported here: https://www.reddit.com/r/Kos/comments/59jisr/a_missionrunmode_file_with_any_four_characters/

Essentially the issue is rooted in how kOS handles storing/reading binary data.  Because we store both ASCII and binary the same way in the KSP save's `ConfigNode`, without any way to differentiate between the two types, we currently just "try" to decode each file as if it was a compressed binary.  If there is a `FormatException` during that decoding process, we fall back to treating it like ASCII.  This has the benefit of being agnostic to the "Use Compressed Persistence" setting, without needing to store information in the save about what the setting was when it was saved.  Regardless of the value of this setting however, all binary and KSM files are stored in the save file using compression (they are not compressed when saving to the archive).

But there are a few issues with this system.  First, because we only catch `FormatException` (which is potentially thrown by `FromBase64String`) it is possible for there to be an exception when reading the gzip stream which is not directly caught.  This actually happens when reading a 4 byte stream because the `FromBase64String` succeeds, and the gzip stream fails to match it's "magic id" at the beginning of the file.  The second issue is that an exception loading any single file from the save will crash the entire processor's `OnLoad` method, loosing any additional files that have already been loaded and preventing other files from loading.  The new `HardDisk` object is not even associated with the processor, and a new blank disc is created when `InitObjects` is called in `OnStart` (the call to `InitObjects` in `OnLoad` is skipped due to the exception).  The third issue is that this system relies on an exception being thrown in order to determine the state of content being processed.  While throwing exceptions based on errors and catching those errors to default to another implementation is perfectly valid in some cases, it results in slow performance (exceptions are expensive) and should be avoided if we can easily determine the correct state without the exception.

A separate concern might be that we are treating each of these files as if they are ASCII, while we support the use of Unicode text elsewhere.  In the future we may need to address this.

I am working on a PR with a solution for this problem already.

hvacengi:
Fixes #1857 

To fix the underlying problem, I modified the `Decode` method in `PersistenceExtensions.cs` to catch any `Exception` type, which should prevent a failed decode from destroying a valid text file, and protect the loading of other files.  But to fix the issue where we're using an exception to determine valid state I needed to modify how files are stored in the save file.  For the sake of backwards compatibility, the current node key of "line" remains valid for loading and we will load any node containing a "line" essentially the same way we do now, with the above fix.  But when saving files the code now uses keys of "ascii" and "binary" based on the type of data being stored.  This way the decoding logic is able to determine if the file needs to be Base64 decoded and decompressed without needing to try to perform either operation.

If "Use Compressed Persistence" is enabled, ASCII files are still saved using the gzip compression but they get saved using the "binary" key instead.  Because the actual `FileContent` object is agnostic of the data it contains (the "category" is determined based on the actual byte content at any given time) we don't need any special handling when dealing with zipped ASCII data, it simply is unzipped and passed into the constructor.

As an added bonus, I figured out that we can add gzip compression directly to ksm files for a very noticable space savings without changing anything in the format itself.  Since I was looking at it, I fixed that at the same time.

### DIFF
--- a/src/kOS.Safe/Compilation/CompiledObject-doc.md
+++ b/src/kOS.Safe/Compilation/CompiledObject-doc.md
@@ -23,7 +23,7 @@ that loads the file to reconstitute all the "missing" redundant data.
 * **kRISC** The name for @marianoapp's overhaul of kOS that created
 a virtual computer running a virtual assembly code.  For the context
 here, when this document mentions kRISC, it's referring to the
-slightly more high-level way of representing the kRISC code, as 
+slightly more high-level way of representing the kRISC code, as
 C# Collections of the C# type kos.Compilation.Opcode.  For example,
 if you are thinking in terms of List&lt;Opcode&gt;, then you're thinking
 of what this document is calling *kRISC*
@@ -58,12 +58,12 @@ and comments and long variable names, and lets users run code
 on their meager small local volumes with them crunched down into
 the smallest possible form.
 
-Getting rid of wastage from comments and spacing is simple. 
+Getting rid of wastage from comments and spacing is simple.
 Because we're storing the compiled format, they're already gone.
 
 But getting rid of wastage from long variable names is a bit more
 complex.  Because the "CPU" runs the code in a namespace-aware way,
-opcodes refer to values by their variable names (not by their 
+opcodes refer to values by their variable names (not by their
 offset from a stack position or a heap address), and therefore
 variable names cannot be entirely removed from the compiled
 file.  Instead the format tries to "mash down" all the names into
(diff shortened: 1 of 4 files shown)


Dunbaratu:
Okay mm.  the gzip and gunzip programs dislike working with files that don't have one of the standard gzip suffixes (that means ending in .gz. typically.)  It will do it, but you have to force it with some obscure command line args.

I think that's okay, but I just wanted to make note of it.  If people try to unzip the file themselves with a tool outside of kOS they might need to know about this.  (more importantly if they gzip a KSM file they need to know to change the extension to KSM instead of the default .ksm.gz that gzip will give it by default).

PR #1859, opened 2016-10-29
Introducing the SKID chip.  ( Fixes #1832 - sound system.)

### LINKED ISSUE #1832: Procedural simple sound system
(What I'd like to see is a feature that will work similarly to the old Commodore 64 SID chip and similar systems).

For making it so users can implement their own sounds emitting from scripts, it has been suggested that they be able to drop their own WAV or MP3 files into the archive and we play them like we play our own internal sounds.  The problem I have with this suggestion is that such files are big, and the disk size is limited.  We'd have to either abandon the idea of old-school disk size limits, or break the 4th wall again by having access to those files live "outside" the normal system.

I'd like to try implementing a means for people to make simple computery sounds from their scripts by letting them say things like "I'd like a sawtooth wave, for 0.5 seconds at volume 0.2 please...".

It may be possible to improve upon the system with attack/sustain/decay envelopes, and so on later, but even just getting a simple "make it beep at the frequency you say so" would be enough of a start to let people start adding that to their user interfaces for things.

Dunbaratu:
Issue #1832, implementing a sound system.

This is the end result of what I've been working on on-stream in #Creative mode in Twitch.  It can still be improved later on by adding more wave forms if we feel like it, and fancier effects (like the ADSR really should vary with frequency because IRL when an instrument is vibrating the air faster, its notes tend to die down sooner due to the higher energy it takes to keep the vibration going.)

But I feel like this is the appropriate "level" of complexity for the kinds of hardware we're trying to pretend to be.

As always, with any new feature, start from reading the RST files to get the gist of what it does and how.


doc/source/general/skid.rst
```diff
@@ -0,0 +1,506 @@
...
+   characters, but they are cumbersome to type on most keyboards.)
+3. Mandatory: The last character is a digit indicating which octave
+   number (0 through 7) this note is in.  (4 is the "middle" octave
+   that starts with "middle C" on a piano keyboard.)
+
+Examples:  ``"C4"`` is middle C.  ``"C#4"`` is the C-sharp just one half
+step above middle C.  ``"Db4"`` is the D-flat that is in fact the same
+thing as ``"C#4"``.  ``"B3"`` is the B that is just to the left of
+middle C (Note the octave numbering starts with C at the bottom of the
+octave, not A, thus why A3 is adjacent to C4.)
```
Dunbaratu: Dangit, I proofread this once over but missed this.  Where it says "A3 is adjacent" it should say "B3 is adjacent".

doc/source/general/skid.rst
```diff
@@ -0,0 +1,506 @@
...
+The SKID chip contains an interior lookup system that makes it possible,
+anywhere the chip expects you to give a frequency in Hertz, to instead
+specify its frequency using the more familiar
+`"letter notes". <https://en.wikipedia.org/wiki/Scientific_pitch_notation>`_
+
+To do this, you use a string in the following format:
+
+1. Mandatory: First character is the note letter, one of 
+   "C","D","E","F","G","A", "B", or "R"(to mean "rest").
+2. Optonal: Followed by an optional character, "#" or "b" for "sharp" or
```
teleksterling: Also a small typo here for 'optional'.

Dunbaratu:
I may try to add a small change to support "slide" notes.  (Like a NOTE() constructor that gives both a start and end frequency, as in "start this note at 440hz, end it at 510hz, taking 0.5 seconds to do so".)  In trying to make such effects in kerboscript code I ran into the fact that the rounded-off integer-ized updating made it sound really bad.  This is because whenever you change the NOTE() in Kerboscript, it starts playing a *new* wave from the start of a new period, regardless of where the note was before within it's period at the moment it was swapped out for the new note.  That causes disjoint jumps in the sound wave as it changes sharply between two adjacent samples.  When you only change notes no faster than 5 times a second or so, you don't care.  But when it's meant to be a smooth transition that changes frequency 25 times a second, you start to hear those jumps that are happening 25 times a second as a secondary raspberry buzz effect underlying the main sound.

But if I have the C# code (i.e. the 'chip hardware') do the changes itself and it was told to make it a smooth slide between notes, then it knows not to start a new note over, and instead just smoothly adjust the rate at which it's playing the wave samples.  Each frequency change picks up with the sample value right where the previous one left off, eliminating those jumps.

Dunbaratu:
I hope you don't mind that I am pushing a new change to the PR branch while you were reviewing.  I implemented the sliding note feature talked about in the above comment.

hvacengi:
I don't mind.  I haven't made any major changes yet anyways, or finished my full code walk (though I've made it most of the way through now).

src/kOS/Sound/VoiceValue.cs
```diff
@@ -0,0 +1,149 @@
...
+                    float durationPortion = (now - noteStartTimeStamp) / (noteEndTimeStamp - noteStartTimeStamp);
+                    float newFreq = note.Frequency + durationPortion*noteFreqTotalChange;
+                    voice.ChangeFrequency(newFreq);
+                }
+                return;
+            }
+            
+            // Increment to next note and start playing it:
+            ++noteNum;
+            if (noteNum > song.Count())
```
hvacengi: This needs to be `>=` since if `noteNum == song.Count()` it will cause an index out of range exception.

src/kOS/Screen/TermWindow.cs
```diff
@@ -133,6 +134,8 @@ public void Awake()
 
             GameEvents.onHideUI.Add (OnHideUI);
 			GameEvents.onShowUI.Add (OnShowUI);
```
hvacengi: Some how this section previously used tabs instead of spaces, so the new portion that follows also uses tabs, I'll swap to spaces.

src/kOS/Sound/Voice.cs
```diff
@@ -0,0 +1,197 @@
+﻿using System;
+using UnityEngine;
+using kOS.Safe.Sound;
+
+
+namespace kOS.Sound
+{
+	/// <summary>
```
hvacengi: This whole file seems to use tabs instead of spaces, I'll swap to spaces.

Dunbaratu: I swear I told it to always use spaces instead of tabs.  I recently upgraded my copy of SharpDevelop.  Maybe in so doing I lost that setting.

src/kOS/Sound/VoiceValue.cs
Dunbaratu: It looks like you're right, but what's weird is that I've used this code to make looping music and it worked just fine.  Let me look into why.

src/kOS/Sound/VoiceValue.cs
```diff
@@ -0,0 +1,149 @@
...
+            else
+            {
+                curNote = song[noteNum] as NoteValue;
+                if (curNote != null)
+                {
+                    noteStartTimeStamp = now;
+                    noteEndTimeStamp = now + tempo*curNote.Duration;
+                    noteFreqTotalChange = curNote.EndFrequency - curNote.Frequency;
+                    voice.BeginProceduralSound(curNote.Frequency, tempo*curNote.KeyDownLength, curNote.Volume);
+                }
```
Dunbaratu: Ah this is why it worked despite the index off by one error.  When it hits the end the exception aborts the Update(), but only just at the point where the update didn't have any work left to do anyway so aborting the Update() didn't make any noticable difference unless I looked at the output log.  The only thing it did is delay the restart of the song (when looping) until the next update after this one (when noteNum was +1'ed one more time).  I wasn't able to hear that 1/25th of a second extra delay so didn't see any obvious sign of error.

src/kOS/Screen/TermWindow.cs
```diff
@@ -178,6 +181,11 @@ private void LoadFontArray()
             }
         }
         
+        public kOS.Safe.Sound.ISoundMaker GetSoundMaker()
```
Dunbaratu: My intent was to expose `GetSoundMaker()` to `kOS.Safe` so it could be called from kOS.Safe without breaking (it would be just a dummy do-nothing if Unity's not there, but it wouldn't break the compile).  But the way `SharedObjects` vs `SafeSharedObects` works, I can't do that interface abstraction at the level of `SharedObjects`, I have to do it on one of the items *in* `SharedObjects`, and `ITermWindow` made more sense than any of the others did.

But in the end I didn't end up needing to expose anything about it after all, so it didn't end up in `ITermWindow`.  But I prefer it the way it is now because it leaves open the chance of putting it there later if we need to.

src/kOS/Sound/VoiceValue.cs
hvacengi: Yeah, I test using the debug instance of KSP, which pops up a red error message any time that there's an error, it really helps me find fringe cases like this.  In fact I have some other error messages that have popped up that I want to fix eventually too.

Dunbaratu:
Are you making the changes (importantly, the off-by-one error you caught with '<' instead of '<='), or do you want me to do it and push to the branch?

hvacengi:
I'm making changes myself.  You don't need to worry about it.

And I think I've decided that if we enable procedural sound waves as one of the things that addons can add, we'll do it in a future release.  There is no sense in complicating the review of this pull request itself with feature addition.

PR #1865, opened 2016-11-16
Fix Blizzy config lookup

### LINKED ISSUE #1863: Blizzy toolbar support broken
When migrating to the new difficulty settings, I didn't test with Blizzy's toolbar installed.  So I missed that the config property didn't get updated to support `kOSCustomParameters`.  If the blizzy's toolbar is installed, this causes a crash during the toolbar initialization and breaks both Blizzy support and kOS stock tool bar support.

The offending line is here: https://github.com/KSP-KOS/KOS/blob/develop/src/kOS/Suffixed/Config.cs#L30

hvacengi:
Fixes #1863

Config.cs
* Fix UseBlizzyToolbarOnly to use kOSCustomParameters instead of the
config properties.


Dunbaratu:
I didn't test beyond "it compiles".  I don't use Blizzy toolbar, but eyeballing the one line of code change I can't see any way it could possibly break anything that wasn't already broken.

PR #1870, opened 2016-11-19
Fix #1866: Permit read locks on archive file write

### LINKED ISSUE #1866: Log/writeln can't write when file is open for reading
Was playing around with an idea for exporting data from a craft into a live dashboard, ideally through writing to file. However, if a file is open for reading, kOS crashes on write.

*Unix env:* `tail -f Ships/Script/test.log` (echo new lines as they're written)
*kOS:* `open("0:/test.log"):writeln("Test").`

*Result:* Sharing violation on path %GAMEDIR%/Ships/Script/test.log

Not sure how solvable this is, and I'm sure it's low-priority, but I believe it's the only way to get data *out* from kOS, aside from trying to write some scraper around a telnet connection.

gisikw:
Prevents C# from insisting on a write-only lock when writing to the file
system. This means external programs (a simple `tail -f $FILE` for example) can
receive data from a running kOS program without causing errors.

### DIFF
--- a/src/kOS.Safe/Persistence/ArchiveFile.cs
+++ b/src/kOS.Safe/Persistence/ArchiveFile.cs
@@ -33,7 +33,7 @@ namespace kOS.Safe.Persistence
             }
 
             byte[] bytes = Archive.ConvertToWindowsNewlines(content);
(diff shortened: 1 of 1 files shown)


Dunbaratu:
Ah thank you.  Saves me having to look up what the flag option was to change this.

PR #1872, opened 2016-11-20
default brightness 70%, and added config option for it.

### LINKED ISSUE #1864: Note To Self:  Fiddle with terminal brightness a bit before next release.
I've been meaning to fix this for a while and keep forgetting to - it's probably just a tiny one or two line change too.

The current default brightness on the slider is too dim for the terminal.  This is probably because I put it at the halfway point of a linear scale when perception of brightness is actually a non-linear phenomenon.  But at any rate the default brightness is kind of dim and everyone always adjusts it up as the first thing they do on loading it.

Also, let's try to make the default brightness a persisting slider in the preferences panel.

Dunbaratu:
Fixes #1864

### DIFF
--- a/src/kOS/Module/kOSCustomParameters.cs
+++ b/src/kOS/Module/kOSCustomParameters.cs
@@ -104,6 +104,11 @@ namespace kOS.Module
                                          "button on it instead of both it and the stock toolbar.")]
         public bool useBlizzyToolbarOnly = false;
 
+        [GameParameters.CustomFloatParameterUI("Default terminal brightness", minValue = 0f, maxValue = 1f, stepCount = 50, displayFormat = "N2",
+                                         toolTip = "Brightness of a kOS terminal when it appears for the first time in a scene.\n" +
+                                         "(You must reload the scene to see the effect of any changes to this slider.)")]
+        public float terminalBrightness = 0.7f;
+
         [GameParameters.CustomParameterUI("Debug each opcode",
                                          toolTip = "(For mod developers) Spams the Unity log file with a message for every time\n" +
                                          "an opcode is executed in the virtual machine.  Very laggy.")]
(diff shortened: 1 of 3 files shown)


src/kOS/Module/kOSCustomParameters.cs
```diff
@@ -104,6 +104,11 @@ public int InstructionsPerUpdate
                                          "button on it instead of both it and the stock toolbar.")]
         public bool useBlizzyToolbarOnly = false;
 
+        [GameParameters.CustomFloatParameterUI("Default terminal brightness", minValue = 0f, maxValue = 1f, stepCount = 50, displayFormat = "N2",
+                                         toolTip = "Brightness of a kOS terminal when it appears for the first time in a scene.\n" +
+                                         "(You must reload the scene to see the effect of any changes to this slider.)")]
```
Dunbaratu: I chose to do it this way because previously if you chose to adjust the brightness GUI slider on a single terminal window (or use ``set terminal:brightness to something.`` in a script which is the same thing), then closed the terminal window and re-opened it, the setting you chose for that one terminal persisted through the closing and opening of it.  It only reset to the default of 50% when you first instantiate the ``Shared`` object for the CPU, not when you open or close the window again and again.  Thus to preserve that behaviour and not wipe the different individual settings of different terminals whenever you change this ``config`` setting, I kept the fact that it only gets looked at during initializing the ``Shared`` member of the CPU.

Dunbaratu: I also considered making the GUI widget here more UI friendly by making it go from 0% to 100% rather than from 0.0 to 1.0, but then it would have different semantics from the script's settings and I figured making them agree is less bad than making the GUI slider look weird because it goes from 0 to 1.

PR #1874, opened 2016-11-20
Fixes #1873 : Seems to be working.

### LINKED ISSUE #1873: Get velocity of the ground at a geoposition.
Right now you can get a geopositions' XYZ pos with a ``:position`` suffix.
For completeness we should add a ``:velocity`` suffix for the XYZ velocity of the ground at that location.

It should be a ``kOS.Suffixed.OrbitableVelocity`` not just a single vector.  (just like ``ship:velocity``) so you have to specify if you meant ``:surface`` or ``:obt``.

The useful one is probably ``:obt``, but ``:surface`` might be useful in cases where ``SHIP`` isn't close to the geoposition right now.  (Because using the ``:surface`` suffix will give us the velocity in the frame of reference of surface rotation **where the ship is**, not where the geoposition is, so it won't always be zero like you might expect at first glance.  It would only be zero if the ship is currently sitting right at the geoposition.)

This information can be derived in a script by knowing the planet's rotational period and the radius to the center at the geoposition (sea level radius plus geoposition's altitude), and some vector operations to find out which way is the eastward unit vector at that spot, but that's a lot of extra steps to get around the fact that we don't provide it.

Dunbaratu:
Test case examples:

```
// Should (if on kerbin near KSC) draw a white vector straight east
// at magnitude 174.94 m :
set vd1 to vecdraw(
  v(0,0,0),
  ship:geoposition:velocity:orbit,
  white,"velocity here",1,true).

// Should (if on kerbin near KSC) draw a nearly zero vector because
// surface-relative velocity of the ground right under me is negligable:
set vd1_s to vecdraw(
  v(0,0,0),
  ship:geoposition:velocity:surface,
  white,"srf velocity here",1,true).

// Should (if on kerbin near KSC) draw a surface vector like vd1 above,
// but longer in length because it's the rotational effect with a radius
// that's 200,000m bigger:
set vd2 to vecdraw(
  v(0,0,0),
  ship:geoposition:altitudevelocity(200000):orbit,
  green,"geoposition velocity higher up above me",1,true).

// Should (if on kerbin near KSC) draw a vector angled upward by
// 30 degreees because it's the velocity of the ground 30 longitude
// degrees west of the current position:
set vd3 to vecdraw(
  v(0,0,0),
  latlng(ship:latitude, ship:longitude-30),
  red,"velocity 30 degrees west of here",1,true).

// Should (if on kerbin near KSC) draw a vector angled upward by
// 30 degreees like above,  but this time it should be a much
// shorter vector because it's how fast the surface of the
// planet is moving much higher north, near the pole:
set vd4 to vecdraw(
  v(0,0,0),
  latlng(ship:latitude-80, ship:longitude-30),
  yellow,"velocity 30 deg west and 80 deg north of here",1,true).
```

Fixes #1873

### DIFF
--- a/doc/source/math/geocoordinates.rst
+++ b/doc/source/math/geocoordinates.rst
@@ -70,6 +70,14 @@ Structure
           - `Vector` (3D Ship-Raw coords)
 	  - :ref:`scalar <scalar>` (altitude above sea level)
           - Position of a point above (or below) the surface point, by giving the altitude number.
+        * - :attr:`VELOCITY`
+          - :struct:`OrbitableVelocity`
+	  - none
+          - Velocity of the surface at this point (due to the rotation of the planet/moon).
+        * - :attr:`ALTITUDEVELOCITY`
+          - :struct:`OrbitableVelocity`
+	  - :ref:`scalar <scalar>` (altitude above sea level)
+          - Velocity of a point above (or below) the surface point, by giving the altitude number.
 
 .. note::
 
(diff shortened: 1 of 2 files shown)


Dunbaratu:
I just realized the docs were missing from the commit, and added them.  It should be ready now.

PR #1878, opened 2016-11-23
A GUI toolkit.

WazWaz:
I'd like a lot more input from others before I'd even _want_ this in kOS of course, but this is presented here for discussion.


Dunbaratu:
While this is cool, I'm not sure how I feel about leaving the telnet control option out in the cold like this.
As long as the gui is nothing more than fields, checkboxes, buttons, and sliders, it should be possible to make a "text-ified" variant of it that works on the terminal window too, and abstract away the difference for people.  (So you can just select if you want the text or the graphical version of things when invoking the gui).

But I'm not sure how much it matters.  I don't really see a lot of people using the telnet option much.  Up until now we haven't introduced anything that makes a big difference.  I'm a bit torn on whether or not introducing this is good.  It's a very cool option I've been wanting for a while, but I've always entertained the notion that it could be do-able entirely on the terminal window itself.

WazWaz:
I was considering the terminal approach, but it takes up a lot of screen realestate (especially with all the settings gadgets on it now). In contrast, having a boot program that starts up a simple custom GUI means kOS users could even develop entire "mods within a mod" for non-kOS users (hmm... ROM modules in Parts.......).

As for Telnet, I see these users as an entirely different breed from those who would want to use/write a GUI. How horrible would Windows have been if it had a TTY renderer. I think ultimately people will write useful functions (like "launch to orbit" and "execute node" and "land" that I'm sure we've _all_ written), then bind them to a GUI while still being easily able to execute them from the command line, in the same way that MacOSX is Unix underneath.

One thing I forgot to mention is that I had no luck working out how ScopeLost works. It almost never seemed to be called. So I settled for the idea that programs really should manually call Hide() on the window and when power is lost that windows close themselves.

Dunbaratu:
What I mean is that "here's a list - pick something from it", and "here's a checkbox you could select" are concepts that are not gui-specific in the slightest.  People have coded TUI toolkits on top of things like curses that do all that.  But that sort of thing is sparse and hard to find in the C# world.

But what using the existing gui toolkit buys us is that we don't have to code all the stuff to deal with the user input and updating the visual look based on what the user did, and all that jazz.  The fact that it's on a GUI isn't much of an advantage.  The fact that if we use the Unity GUI system then the work has been done already by other people, however, is, so I might begrudgingly accept doing it that way even though it helps drive the artificial wedge between interactive and command line usage by making it possible to write scripts that *cannot be* used from a remote connection.

WazWaz:
Note that while it uses the older Unity IMGUI, it's presented as object-based so should in theory port easily to UGUI should the need arise. Ideally, that abstraction should be kept however it goes forward.

Dunbaratu:
I suppose I shouldn't be the only one deciding things.  I'm just one person and kOS is a team effort that nobody really singly owns.  So since it seems the GUI idea is gaining traction it's better to include it rather than keep it as a messy separate fork.  (I admit that by adding sound I moved things a bit away from the "remote text allows complete control" direction myself, but sound was a gui-only **output** thing so it wasn't quite the same situation as having a gui-only **input** method would be.)

In the end, really to be honest, the problem is only a problem because I've been dragging my feet on the text facilities so they're not up to snuff to compete against the easier to implement GUI stuff (easier because, again, the toolkit is right there ready to be used in Unity and not something we'd have to make from scratch.)  I mean, I never even implemented a "type a line of text" input to a program yet, and that's really quite basic.

So yeah, I shouldn't be objecting to it.  It's only a problem of my own making by dragging my feet on implementing the textual alternative.  That's no reason to nay-say someone else getting the task done quicker and simpler and (let's be honest) in a way most users probably would prefer anyway.

I would at least like to maybe put in a little effort on trying to make the UI *look* a bit more KOS-like, but that can be done later by manipulating the UI skin settings.  It's not a reason to block progress.

Dunbaratu:
Incidentally, @WazWaz : Can you add an issue to associate with this Pull Request?  Our preferred organizational way to deal with major feature enhancements like this is to separate the implementation (PR) from the high level feature description (Issue).  Even in cases like this where you already implemented it, it's still nice to submit both, to have the pairing to fit together.

WazWaz:
I'd like to leave it in a branch for a little longer as I'm still discovering "deep" changes (like the scope thing). When I think it's at least at "beta", that would be a good time to pull to main.

Issue: will do.

Dunbaratu:
Fair enough.  Let me tag this PR with the label "Not Ready" then, to remind us not to merge it.

I don't know if you have the permissions to alter labels.  If you don't then let us know when you do consider it ready and we can remove the "Not Ready" label.

WazWaz:
The "Not Ready" can come off. I believe this is now good for release. Of course, I'm happy to be told "do X, Y, and Z first" (eg. I've made sure it's documented, but does it need more examples, tests, or anything?).

I'm assuming the big red crosses above are because things unrelated to this branch are causing test failures.

doc/source/structures/misc/gui.rst
```diff
@@ -0,0 +1,304 @@
...
+
+    The `BG` attribute comes in 8 flavours: `BG`, `BG_FOCUSED`, `BG_ACTIVE`, `BG_ON`, `BG_FOCUSED_ON`, `BG_ACTIVE_ON`, and `BG_HOVER_ON`.
+
+    ============= =====================================================
+    FOCUSED       The widget has keyboard focus.
+    ACTIVE        The widget is hot (eg. button is being held down).
+    HOVER         The widget is under the mouse.
+    ON            The widget is on (eg. button is PRESSED).
+    ============= =====================================================
+
```
Dunbaratu: I'm guessing there's some files missing (i.e. the ``git add`` command wasn't run for them).  I only see one PNG file for the button, and I don't see any of these suffixes corresponding to any ``AddSuffix`` calls within the C# code (that C# file is probably missing from the PR).

WazWaz: BG_* - It's just trying to shorthand the documentation. There are 8 attributes, but it looked cluttered and confusing to list them all in the normal section, so I put it in a note where they could be properly explained.

I'm not sure which PNG files you mean. There should be some in the documentation, and 3 in the GFX directory:

* commDelay - the "busy" icon
* popupmenu_pb_hover - the yellow color seen on PopupMenus.
* radiobutton - wrong name I just realized, it's the icon for the left side of PopupMenus.

WazWaz:
BTW, someone messaged me about the possibility of using Delegates (eg. ONPRESSED), so I'm also looking into that. I missed that addition to KSP - neat!

hvacengi:
@WazWaz our code convention is to keep one class per file (there are some exceptions that come from pre-existing code, and functions are an unwritten exception).  It looks like the Widgets.cs file contains all of you types of widgets.  Could you break those classes out into there own files, maybe create a new Widgets sub-folder?

WazWaz:
Sure, I'll break it up. It was kept together because it's a tightly coupled hierarchy of classes.

Still looking at Delegates, but I can't work out how to actually call the delegate successfully from C#. I'm trying `onPressed.Call(new Structure[0]);` but it asserts.

hvacengi:
@Dunbaratu was talking about looking into implementing events to go with the ability to invoke script code from inside C#, but I don't know that he's done a lot on it in the last little bit.

doc/source/structures/misc/gui.rst
Dunbaratu: If there are 8 different suffixes, this was not clear from the docs.  (Remember that the C# code that implements what this document is talking about is missing from your PR (I saw no calls to ``AddSuffix`` within the diff output listed on the PR page, which probably means the C# file that implements them wasn't ``git add``-ed to the branch yet.  So I only see the docs and not the implementation.  (Which is a pretty good test of how understandable the docs are, as that's the situation we should expect a typical user of kOS to be in).

As to the URL you linked, It looks like you have added more to the PR than was there when I looked at it a few hours ago.  That file was definitely missing before.  I'll go have a look now.

WazWaz:
Using delegates would allow users to much more easily write self-contained widgets, such as a TabWidget, which is really just a HBOXLAYOUT with specially-styled RADIOBUTTONs on top of a STACK box, with the "ONPRESSED" delegate for the radiobuttons set to show the corresponding tab in the stack.

I pushed non-working code I used above (to show it here and because it was in the way of the one-class-per-file change. I'll of course revert it since it doesn't work.

doc/source/structures/misc/gui.rst
WazWaz: They're in src/kOS/Suffixed/Widgets.cs in the original PR (src/kOS/Suffixed/Widget/Button.cs now) - note that github elides it because it's a long file, so CTRL-F won't find it.

Dunbaratu:
Minor naming convention points:

The leading underscore naming convention for privates isn't used in kOS code.  Instead it's:

If private: start with lowercase letter.
If public or protected: start with uppercase letter.

I come from Java where this is not the convention (the convention there is start-uppercase means it's a type name and start-lowercase means its an instance name), but it's the convention used in the kOS code so we should keep it consistent with itself.

Also: only private members should be raw fields unless there's a very strong justification otherwise.  everything public or protected should be a method (call like a function) or a Property (Those get/set things that are called like it was a field, but under the hood it actually can run code that does stuff before returning the value or setting the value.)  The rationale for this is that it gives us the freedom to change quite a lot about the underlying implementation while preserving its appearance on the surface so the way other classes are calling it doesn't have to change.

(I tried writing the above comments on specific lines of code in the github PR screen, but you've been changing the code while I was typing so instead of commenting on an obsolete picture of the code I just made the general comments here instead.)

Dunbaratu:
If you're wondering why I was looking specifically at the background image stuff, it's because I was thinking about what could be done to the artwork look and feel to make the widgets look more thematic, and looking at what support for it you already had it looks like it can be done as an optional thing, but I might prefer making it the default look (things like a brushed metal plate for the window background, and toggle switches you throw for the toggle buttons, etc.)

WazWaz:
I'm a terrible artist, but yes, if a user comes up with a nice "skin", it might even be worth adding a skin concept (GUI:SKIN) rather than expecting every widget to do it manually. Unity GUISkin would be sufficient, especially if CustomStyles were supported, so that for example, a skin could specify the look for a TabWidget or other concept that was actually just built up from primitive widgets with different styles.

hvacengi:
> Minor naming convention points:

@WazWaz I think our style guide looks pretty up to date.  While it isn't gospel truth it at least helps you get an idea of the general framework. https://github.com/KSP-KOS/KOS/blob/develop/STYLEGUIDE.md

WazWaz:
I reconsidered what @hvacengi said about a "bg config" structure. I still didn't do that, but moving all the style attributes to their own structure cleans up the structs (and docs) nicely, and the concept of a SKIN comes for free.

I'm unsure of the namespace consequences of these GUI elements. I think they should affect users (because only functions should have that problem, of which there is only GUI(width)), and the C# class is "WidgetSkin", not just "Skin", but does the kOS struct name (Skin) have any consequence?

hvacengi:
Looks like you're most recent commits are missing the changes to kOS.csproj to account for WidgetStyle.cs and WidgetSkin.cs

WazWaz:
I committed them in ffa8408, but accidentally also committed my local "install to KSP" changes, which I've now reverted.

Dunbaratu:
@WazWaz - can you remove the "Not Ready" label on this PR to state officially that you're okay with merging it now? (Do you have the privileges to change labels in this github? If not then just post a comment telling us to remove the label.)

In the slack channel, you said yesterday "I've merged and pushed.  Just one minor conflict in SharedObjects" - but in this PR I don't see it claiming anything has changed since Dec 13th.  Are you sure your push made it all the way to here?

hvacengi:
Last commit was yesterday: https://github.com/KSP-KOS/KOS/commit/5e05b1e95fe735d1d9f36534d3f2490f42094215

It looks like he's a member of the "contrib" team, which only has read access to the repository.  I assume that applies to labels and such as well.

Dunbaratu:
In that case, github's presentation is totally bogus.

It's indented thusly:

* WazWaz added some commits on Dec 13, 2016
	* @WazWaz 	Stop sphinxbuild help spamming stdout on version test. 			172b020
	* @WazWaz 	Use Object.Instantiate to copy Skin. Util function is suspect, and is	42839fb
	* @WazWaz 	Merge branch 'develop' of https://github.com/KSP-KOS/KOS into gui	5e05b1e

Implying all three of those commits are underneath the heading "some commits on Dec 13th".  If they're not all part of that, they shouldn't be indented that way.

WazWaz:
The merge commit from yesterday appear to be included in "Commits" on this PR, but no, I can't edit labels.

But I'll say here I believe this is Ready.

src/kOS/Screen/GUIWindow.cs
```diff
@@ -0,0 +1,294 @@
...
+            if (!IsOpen) return;
+            
+            try
+            {
+                if (FlightResultsDialog.isDisplaying) return;
+                if (uiGloballyHidden && kOS.Safe.Utilities.SafeHouse.Config.ObeyHideUI) return;
+            }
+            catch(NullReferenceException)
+            {
+            }
```
hvacengi: It looks like this was copied from TermWindow.cs, and was part of the initial commit by Nivekk.  That being said, it would probably be worthwhile to look into this behavior.  It may not be an issue at all anymore, or it may be possible to detect the null reference without throwing an exception (which is expensive, and particularly an issue inside of GUI update methods).  At the very least, catching the excepting on log a message of some kind (I'd prefer an error message) so that we can find out if this exception is ever being thrown.  I've checked and it looks like internally `FlightResultDialog.isDisplaying` includes a null check for the private `Instance` field.

src/kOS/Screen/GUIWindow.cs
```diff
@@ -0,0 +1,294 @@
...
+        public void SetY(float y)
+        {
+            var r = WindowRect;
+            if (y < 0) r.y = UnityEngine.Screen.height + y - r.height + 1;
+            else r.y = y;
+            WindowRect = r;
+        }
+
+        public PopupMenu currentPopup;
+
```
hvacengi: While not "gospel", the referenence guidlines state:
>Classes member should be alphabetized, and grouped into sections (Fields, Constructors, Properties, Events, Methods, Private interface implementations, Nested types)

I'm not one to get too hung up on it, cause it isn't a huge issue (particularly the alphabetizing).  But if most of the fields/properties are declared at the top, it does make it easier to read the file if that location is consistent.  I would probably move them if I were doing the review.

src/kOS/SharedObjects.cs
```diff
@@ -12,15 +12,27 @@ public class SharedObjects : Safe.SafeSharedObjects
         public ProcessorManager ProcessorMgr { get; set; }
         public Part KSPPart { get; set; }
         public TermWindow Window { get; set; }
+        public List<KOSManagedWindow> ManagedWindows { get; }
```
hvacengi: The `TermWindow` would also technically qualify.  This list should also be cleared, and the contained windows destroyed with `UnityEngine.Object.Destroy`, in the `DestroyObjects` method, unless they have no reference to `SharedObjets` to create a circular reference.

src/kOS/Screen/GUIWindow.cs
WazWaz: Yes, this was from TermWindow.

src/kOS/SharedObjects.cs
WazWaz: Ideally yes, but I didn't want to tread on the code that references SharedObjects.Window. That change should be an Issue.

Dunbaratu: I'm compiling a list of all these "should be addressed, but shouldn't prevent the GUI from being merged" kinds of issues and I added them to the issues list.  This is one of them, here: https://github.com/KSP-KOS/KOS/issues/1952

PR #1883, opened 2016-11-27
Fixes #1882 with a TONUMBER string suffix

### LINKED ISSUE #1882: Need a means to parse string numbers into actual numbers
While this can be done in kerboscript code, it's a bit clunky to have to do so.

We need to provide scripts with some means to take as input strings like so:
```
"1234"
"1234.567"
"1.32e8"
"1.32e-6"
```
And return their number result as a ``Scalar`` type (either int or float depending on what's necessary).

The ability needs to have at least these two abilities:
- Not care about the difference between int and float values (to make it fit into the way kerboscript does ``Scalars``
- Have a means to detect when the string isn't valid for a number, that doesn't just crash the script.

And it would be nice if it also had this, but it's not strictly necessary:
- The ability to work with different international formats (mostly whether the decimal point is a period or a comma).

I propose that this be implemented via a suffix on the ``StringValue`` type.

I also propose we try to make it as permissive as possible (i.e. don't complain about the format unless we cannot correctly guess what was meant.  As in, make it legal if the value has trailing or leading spaces since it's obvious that "      123" or "123          " both mean 123, and things like that.  Because this is likely to be used in cases of user input and the user will often type a space and not "see" it.)

Dunbaratu:
Pretty self explanatory.  It does what the issue #1882 says.

### DIFF
--- a/doc/source/structures/misc/string.rst
+++ b/doc/source/structures/misc/string.rst
@@ -12,9 +12,9 @@ Unlike other structures, strings are created with a special syntax::
 
     // Create a new string
     SET s TO "Hello, Strings!".
-    
 
-Strings are immutable. This means, once a string has been created, it 
+
+Strings are immutable. This means, once a string has been created, it
 can not be directly modified. However, new strings can be created out
 of existing strings. For example::
 
@@ -70,7 +70,7 @@ In the first example, both sides of the ``<`` operator are
 :ref:`scalars <scalar>`, so the comparison is done numerically,
 and 1234 is much bigger than 99.
 
-In the second example, one side of the ``<`` operator is a 
+In the second example, one side of the ``<`` operator is a
 string, so the other side is converted from the :ref:`scalar <scalar>`
 ``99`` into the :ref:`string <string>` ``"99"`` to perform the
 comparison, and then the string comparison looks one character at
@@ -89,7 +89,7 @@ will let you choose case-sensitivity when you prefer.
 At the moment the only way to force a case-sensitive comparison is
 to look at the characters one at a time and obtain their numerical
 ordinal Unicode value with the :func:`unchar(a)` function.
-	
+
 Structure
 ---------
 
@@ -105,10 +105,10 @@ Structure
 
         * - :meth:`CONTAINS(string)`
           - :ref:`boolean <boolean>`
-          - True if the given string is contained within this string  
+          - True if the given string is contained within this string
         * - :meth:`ENDSWITH(string)`
           - :ref:`boolean <boolean>`
-          - True if this string ends with the given string 
+          - True if this string ends with the given string
         * - :meth:`FIND(string)`
           - :struct:`Scalar`
           - Returns the index of the first occurrence of the given string in this string (starting from 0)
@@ -150,7 +150,7 @@ Structure
           - Breaks this string up into a list of smaller strings on each occurrence of the given separator
         * - :meth:`STARTSWITH(string)`
           - :ref:`boolean <boolean>`
-          - True if this string starts with the given string 
+          - True if this string starts with the given string
         * - :meth:`SUBSTRING(start, count)`
           - :struct:`String`
           - Returns a new string with the given count of characters from this string starting from the given start position
@@ -172,13 +172,16 @@ Structure
         * - :meth:`TONUMBER(defaultIfError)`
           - :struct:`Scalar`
           - Parse the string into a number that can be used for mathematics.
+        * - :meth:`TOSCALAR(defaultIfError)`
+          - :struct:`Scalar`
+          - Alias for :meth:`TONUMBER`
 
 
 .. method:: String:CONTAINS(string)
 
     :parameter string: :struct:`String` to look for
     :type: :ref:`boolean <boolean>`
-    
+
     True if the given string is contained within this string.
 
 .. method:: String:ENDSWITH(string)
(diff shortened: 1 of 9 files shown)


src/kOS.Safe/Encapsulation/StringValue.cs
```diff
@@ -145,6 +145,61 @@ public string TrimStart()
...
+        /// <summary>
+        /// Parse the string into a number
+        /// </summary>
+        /// <param name="defaultIfError">If the string parse fails, return this value instead.  Note that if
+        /// this optional value is left off, a KOSexception will be thrown on parsing errors instead.</param>
+        /// <returns></returns>
+        public ScalarValue ToScalar(ScalarValue defaultIfError = null)
+        {
+            char[] decimalMarkers = new char[] {'.', ','}; // depends on culture
+            char[] scientificNotationMarkers = new Char[] {'e', 'E'};
```
Dunbaratu: They could be combined together now, true.  (Originally the algorithm was more complex than what it looks like now and it necessitated looking at them separately.  I didn't notice I could combine them when I simplified it.)

src/kOS.Safe/Encapsulation/StringValue.cs
```diff
@@ -145,6 +145,61 @@ public string TrimStart()
...
+            string trimmed = internalString.Trim();
+            bool needsDoubleParse =
+                (trimmed.IndexOfAny(decimalMarkers) >= 0) || (trimmed.IndexOfAny(scientificNotationMarkers) >= 0);
+            
+            if (needsDoubleParse)
+            {
+                double val;
+                if (double.TryParse(trimmed, out val))
+                    return (ScalarValue) ScalarValue.FromPrimitiveWithAssert(val);
```
Dunbaratu: Okay.  I can change it.

src/kOS.Safe/Encapsulation/StringValue.cs
```diff
@@ -216,6 +271,7 @@ private void StringInitializeSuffixes()
             AddSuffix("TRIM",       new NoArgsSuffix<StringValue>(() => Trim()));
             AddSuffix("TRIMEND",    new NoArgsSuffix<StringValue>(() => TrimEnd()));
             AddSuffix("TRIMSTART",  new NoArgsSuffix<StringValue>(() => TrimStart()));
+            AddSuffix("TONUMBER",   new VarArgsSuffix<ScalarValue, Structure>(ToScalarVarArgsWrapper));
```
Dunbaratu: *shrug* : It's all the same to me - we could do it either way.  I could change it.

doc/source/structures/misc/string.rst
```diff
@@ -326,6 +329,48 @@ Structure
...
+    never expect to get in normal circumstances so you can use it as a
+    test to see if the string was formatted well.
+
+    The argument ``defaultIfError`` is optional.  If it is left off, then
+    when there is a problem in the format of the string, you will get
+    an error that stops the script instead of returning a value.
+
+    The valid understood format allows an optional leading sign,
+    a decimal point with fractional part, and scientific notation
+    using "e" as in "1.23e3" for "1230" or "1.23e-3" for "0.00123".
```
Dunbaratu: Well, TONUMBER can be changed to strip all spaces not just leading and trailing (so "1.23 e - 3" would work, but that would also make "1   2" get parsed as 12 instead of as an error, if you're okay with that.

Dunbaratu: Hmm.  How about we make a single static method in ``kOS.Safe.Utilities.kOSMath`` that does the work, and then call it from both here and from ``kOS.Safe.Compilation.KS.Compiler`` So it forces the two to be identical because they both use the same routine?

hvacengi: I like the idea of the single method, but I think the compiler problem is with how it's recognized.  Right now the tpg is:
```
E            -> @"\be\b";
...
INTEGER      -> @"[0-9]+";
DOUBLE       -> @"[0-9]*\.[0-9]+";
...
sci_number         -> number (E PLUSMINUS? INTEGER)?;
```
So the scientific number requires a word boundary after the "e".  It's fine to not have the boundary before the "e" because it just parses as a double or integer (they don't need a word boundary following).  But if we remove the word boundary condition from "E" it will return true on anything with an e at the beginning of a word.  The regex I included above would short circuit the parsing of "e", and technically allow `1.92e-12 e 12` to be valid, which might be a problem... but we could add some optional word boundaries around the "e" to make that entire step unnecessary.

Dunbaratu: How about removing the extra step?  We just make DOUBLE parse so it allows the 'e' inside it, as follows:
```
INTEGER   ->   @"[0-9]+(e[+-]?[0-9]+)?";   // allows whole-number sci notation like "123e4"
DOUBLE    ->    @"[0-9]*\.[0-9]+(e[+-]?[0-9]+)?";   // allows sci notation with a fraction like 1.23e4"
...

// Maybe merge these two together in the syntax everywhere since one is just the same as the other now?
number   -> (DOUBLE | INTEGER);
sci_number   -> number;

```
Then VisitSciNumber() would just call TryParse for either int or double depending on which it is.  NOTICE:  the rule above would allow "123e3" to be called an INTEGER even though I don't int.TryParse  cab;t understands scientific notation, so we'd need to check that case and divert such "integers" over into being treated as doubles just to make the TryParse work.

(My reason for proposing it this way is that the regular expression for DOUBLE gets very ugly if we want to support the logic "a period is mandatory *except when* there is scientific notation, then it's optional" in it.  It's easier to just temporarily treat the period-less case as INTEGER and clean it up later in Compiler.cs.)

PR #1884, opened 2016-12-01
Issue 1880 memory leaks and errors

### LINKED ISSUE #1880: Leaking kOSTermWindow objects badly
I noticed using Kerbal Object Inspector that even time a craft is reverted, loaded, etc., 2 more objects named "kOSTermWindow" get created.

The cause for two seems to be that InitObjects() is called in both OnStart and in OnLoad.

The cause for them not being deleted is that InitObjects() calls DontDestroyOnLoad() of the object.

Some basic testing showed that everything still seems to work fine if the call to InitObjects from OnLoad and the DontDestroyOnLoad are removed.

hvacengi:
Fixes #1880

This work isn't done, but I wanted to get the pull request up so that others can see the changes that I have pending, and maybe @Dunbaratu can get a head start on reviewing the "easy" part of the PR.  I have revisions to about 20 other files as well, but those changes are a little more in depth.

I tried to be fairly verbose in my commit messages, since it didn't make sense to comment all the changes citing code that "doesn't exist" anymore.


hvacengi:
I doubt this is every optimization that we have available to us, or even if it's all of the low hanging fruit.  But I tried to take a stab at some of the obvious spots where we could reduce our memory footprint.  In my testing, it seems to have dramatically improved memory usage over long game sessions (greater than 24hr).

It's probably too much to review and get in v1.0.2, but if there happens to be some time to review it great.  Otherwise, we can plan on it being a big feature for v1.0.3 (along with threaded compiling and the GUI classes)

src/kOS/Suffixed/VectorRenderer.cs
```diff
@@ -74,7 +75,13 @@ public void ScopeLost()
...
             // then as far as C# thinks, I wouldn't be orphaned because
             // UpdateHandler is holding a reference to me.)
-            SetShow(false);
+
+            // Store the information about lost scope to be checked in the KOSUpdate method
+            // because Unity complains if we try to set visual element visibility from
+            // outside of the main thread.  That isn't a problem most of the time, but the
+            // finalizer method called by the GC (see Variable.cs ~Variable method) is
+            // apparently called from another thread, or at least Unity thinks it is.
+            scopeLost = true;
```
Dunbaratu: Is this duplicate information given that scopelost should be true if LinkCount <= 0?  Can we just check for that?  (and I'm not a fan of having the same spelling ``ScopeLost`` and ``scopeLost`` being things differing in time (present tense "scope is being lost right now" versus past tense "scope has already been lost").  If we keep this, can we call this ``scopeWasLost`` or something like that  (It remembers the fact that ``ScopeLost()`` has happened in the past, not right now.).

src/kOS/Screen/TermWindow.cs
```diff
@@ -849,32 +849,37 @@ void TerminalGui(int windowId)
...
+                bool blinkOn = cursorBlinkTime < 0.5f &&
+                               screen.CursorRowShow < screen.RowCount &&
+                               IsPowered &&
+                               ShowCursor;
+
+                if (blinkOn)
+                {
+                    ShowCharacterByAscii((char)1, screen.CursorColumnShow, screen.CursorRowShow, reversingScreen,
+                                         charWidth, charHeight, screen.Brightness);
+                }
```
Dunbaratu: I stared at this a while trying to figure out what you changed by deleting all those lines and adding these replacement lines that don't seem that different.. and then remembered that git's diff tool keeps showing mere indentation changes this way.  It seems to be capable of detecting mere indentations and showing them as such only when they last just a couple of lines.  When it's a whole block, it can't detect it.

src/kOS.Safe/Compilation/Opcode.cs
```diff
@@ -744,7 +744,7 @@ public class OpcodeSetMember : Opcode
         public override void Execute(ICpu cpu)
         {
             Structure value = cpu.PopStructureEncapsulated();         // new value to set it to
-            string suffixName = cpu.PopStack().ToString().ToUpper();  // name of suffix being set
+            string suffixName = cpu.PopStack().ToString();            // name of suffix being set
```
hvacengi: The case sensitivity change is being done for the sake of memory allocation more than it is for actual execution time.  I actually doubt that the time is faster, at best it could probably be equal time if the case matches.  The big reason for doing it is to cut down on the number of local scope orphan strings we allocate.  Every time that you call `ToUpper` another string is allocated, and since we discard it right away it gets tossed to the GC.  kOS by it's nature is already fairly garbage heavy anything that we don't store in a variable is essentially thrown to the GC.  So I'm just trying to minimize the number of objects we throw away, thereby reducing how frequently the GC needs to run.

In the grand scheme of things, it probably isn't a huge change either way.  And our string comparisons by far are not the biggest bottle neck,

You'd be surprised though.  Because of how often we throw away objects, we are constantly recreating suffix lists, and the list results of those suffixes.  That's why I implemented caching of the `ship` variable elsewhere, and caching the value of `parts`.  It's also why we defer making the suffix list on strings, scalars, and booleans, since there is a good chance you won't even try to access a suffix of those values instead performing math on them.

TLDR: I don't think it is faster, but it's more memory efficient.

src/kOS.Safe/Encapsulation/Structure.cs
```diff
@@ -255,7 +255,7 @@ public virtual StringValue GetKOSInheritance()
 
         public override string ToString()
         {
-            return "Structure ";
+            return KOSNomenclature.GetKOSName(GetType()) + "()"; // print as the KOSNomenclature string name
```
hvacengi: That's a valid point, but then I'd like to figure out another way to indicate that the string value is a structure identifier.  Maybe steal the format structure from `VesselEta`: `"ETA: Apoapsis={0} Periapsis={1} Transition={2}""`.  We might also need to audit other types for the same "has a constructor function" limitation.

src/kOS.Safe/Execution/Stack.cs
```diff
@@ -165,35 +165,42 @@ public void Clear()
...
+                        builder.AppendFormat("          ScopeId={0}, ParentScopeId={1}, ParentSkipLevels={2} IsClosure={3}",
+                                             dict.ScopeId, dict.ParentScopeId, dict.ParentSkipLevels, dict.IsClosure);
                         builder.AppendLine();
+                        // Dump the local variable context stored here on the stack:
+                        foreach (string varName in dict.Variables.Keys)
+                        {
+                            var value = dict.Variables[varName].Value;
+                            builder.AppendFormat("            local var {0} is {1} with value = {2}", varName, KOSNomenclature.GetKOSName(value.GetType()), dict.Variables[varName].Value);
+                            builder.AppendLine();
+                        }
```
hvacengi: It wouldn't be so bad if they didn't mix the delete and add sections.  If it looked like the whole original block was deleted, and then the new block was added, at least you could more easily confirm the similarity.

src/kOS/Module/kOSProcessor.cs
```diff
@@ -201,7 +199,9 @@ public IScreenBuffer GetScreen()
         // TODO - later refactor making this kOS.Safer so it can work on ITermWindow, which also means moving all of UserIO's classes too.
         public Screen.TermWindow GetWindow()
         {
-            return shared.Window;
+            if (shared.Window != null)
+                return shared.Window;
+            return GetComponent<Screen.TermWindow>();
```
hvacengi: In testing I had a single instance where the new shared `DestroyObjects` had been called before the last attempt to call `GetWindow`, but it is possible that later changes removed the error case.  The problem with an error that you only see a few times is that it's hard to say for sure that you eliminated every chance for it to appear.  Especially since we don't control when Unity calls various methods.

src/kOS/Module/kOSProcessor.cs
```diff
@@ -498,18 +502,19 @@ private void InitProcessorTracking()
                     return (a.part.uid() < b.part.uid()) ? -1 : (a.part.uid() > b.part.uid()) ? 1 : 0;
                 });
             }
-            GameEvents.onPartDestroyed.Add(OnDestroyingMyHardware);
         }
 
-        private void OnDestroyingMyHardware(Part p)
+        public void OnDestroy()
```
hvacengi: Yup, very reliable.  That's why I added the super-verbose log the followed, so that I could confirm that it was happening when it was supposed to.

src/kOS/SharedObjects.cs
```diff
@@ -19,14 +19,20 @@ public class SharedObjects : Safe.SafeSharedObjects
...
+            if (Window != null) { UnityEngine.Object.Destroy(Window); }
+            if (SoundMaker != null) { SoundMaker.StopAllVoices(); }
+            var props = typeof(SharedObjects).GetProperties();
+            foreach (var prop in props)
             {
-                BindingMgr.Dispose();
+                if (!prop.PropertyType.IsValueType)
+                {
+                    prop.SetValue(this, null, null);
+                }
```
hvacengi: Honestly, I think that every time we attempt to access a property on shared it should be null checked.  But you're right, the bigger issue is that we could potentially null check before another thread sets the value to null.  Since I'm only setting the reference to null, you can easily get around this by storing the reference locally in the other thread object.  The calls to `Dispose` (which existed before my change) are actually a bigger deal because they potentially release unmanaged resources and set internal properties to null.

src/kOS/Suffixed/Part/PartValueFactory.cs
```diff
@@ -20,21 +20,28 @@ public static ListValue<PartValue> ConstructGeneric(IEnumerable<global::Part> pa
...
+                }
+                else if (part.Modules[i] is ModuleDockingNode)
+                {
+                    ret = new DockingPortValue((ModuleDockingNode)part.Modules[i], shared);
+                }
+                else if (part.Modules[i] is ModuleEnviroSensor)
+                {
+                    ret = new SensorValue(part, (ModuleEnviroSensor)part.Modules[i], shared);
+                }
+            }
```
hvacengi: While there is no searchable index of `part.Modules`, the `PartModule` class does support a `name` property which is a `string` and should be the name of the type.  I didn't do any testing to see if type checking or string comparison is faster.  But it may be worth while.

I would prefer using LINQ just because they have some clever optimizations (partial ordering of lists to minimize comparisons when you find a match part way through, that kind of thing).  But it is allocation heavy, and the Mono GC that Unity uses doesn't handle it very well.

src/kOS/Suffixed/VesselTarget.cs
```diff
@@ -236,14 +269,41 @@ public ScalarValue GetDistance()
...
+        public int LinkCount
+        {
+            get
+            {
+                return linkCount; }
+            set
+            {
+                linkCount = value;
+                if (linkCount <= 0)
+                    ScopeLost();
```
hvacengi: I've been running with this code in my local copy for a couple weeks now, so I'm fairly confident that nothing ugly can happen.  All that `VesselTarget` does when it looses scope is unhook events, and I did take the time to confirm that there is no issue trying to unsubscribe from an event you aren't subscribed to.

src/kOS/Suffixed/VectorRenderer.cs
hvacengi: I hadn't really thought about that, but yes, we could modify `ScopeLost` to do nothing, and check `LinkCount` instead.  It's more clean than adding another variable.

src/kOS/AddOns/ActionGroupsExtended/ActionGroupsExtendedAPI.cs
```diff
@@ -26,6 +26,10 @@ public ActionGroupsExtendedAPI()
         /// <returns>Is AGX installed?</returns>
         public bool Installed() 
         {
+            if (calledType == null)
+            {
+                return false;
+            }
             try //try-catch is required as the below code returns a NullRef if AGX is not present.
```
hvacengi: I wasn't sure if there was still a risk for exception on the reflection call to `InvokeMember`.  I suppose that's unlikely to be a null reference, but I still like wrapping this call with `try` so that if the member name ever changes we don't pass the exception upstream.  In our other reflection API's we usually store the member's info itself, which ensures that we know the member/method exists.

Net result: yeah, we should edit the comment so it isn't so specific, but I'd like to keep the `try`.

Dunbaratu: That seems fine to me.  Just remove the comment because we don't know that it's true anymore.

src/kOS.Safe/Compilation/Opcode.cs
Dunbaratu: Alright, but then have we ensured that we changed the comparisons *everywhere*?  Because this is a universal change (i..e ALL GetMember calls everywhere no longer convert case), while each of the places you've edited ``if (foo == "STRING")`` to use OrdinalIgnoreCase are individual single cases.  Is there a danger we missed an individual single case.to correspond to this universal change?

It just makes me a little nervous because I doubt we have test cases of literally every hardcoded keyword we ever made.

Dunbaratu: Although, there's a part of me that wants to ILSpy on the contents of the actual string compare and see what it does when you tell it to use ``OrdinalIgnoreCase``.  I have a nagging suspicion it may end up generating a temp case converted string in its head anyway, which would actually be more GC usage then if it's happening on every comparison.

src/kOS.Safe/Encapsulation/Structure.cs
Dunbaratu: How about when we don't implement a class-specific ToString we'll just put a colon and empty quotes there with no further data?  ``GetKOSName(GetType()) + ": \"\""`` ?

src/kOS.Safe/Compilation/Opcode.cs
hvacengi: Looking at the reference source (which may or may not be how Mono does it) It looks like we should be in the clear.  If the strings are both ASCII, OrdinalIgnoreCase calls an unmanaged method without doing a string conversion.  Which means even if it does allocate the additional string, it won't hit the GC.  If they aren't both ASCII, you can see the different implementations at these two links:

https://referencesource.microsoft.com/#mscorlib/system/string.cs,786821813d8a6340

https://referencesource.microsoft.com/#mscorlib/system/string.cs,8711fff131bc4d0e

I'll look at ILSpy a little later

src/kOS.Safe/Encapsulation/Structure.cs
hvacengi: I like that idea.

src/kOS.Safe/Compilation/Opcode.cs
hvacengi: I've walked the code searching every override for `GetSuffix` and `SetSuffix` and all have already been converted to case insensitive checks.  We should have no issues with replacing the logic.

src/kOS/Suffixed/VectorRenderer.cs
hvacengi: I just figured out why I didn't do that initially: A local VectorRenderer added to a list, for which the VectorRenderer scope is lost but the list's scope is not lost.  Because scope is only modified by variable assignment, you can iterate through the list setting `show` to `true`, re-enabling the visibility.  If we constantly check the link count, this would eliminate the current workaround.  I'd be happy to remove this once we fix collection scope tracking though.  I'll add a todo note stating as much.

hvacengi:
@Dunbaratu I just pushed the edits based on your review comments, as well as a couple more fixes.  Do you think you'll have some time to look at this for final merge this weekend?  If so, I'll stop thinking about other potential fixes.  But If it will be a bit before you can get to it I'll think about testing a few more options as well.

Dunbaratu:
I just noticed this message now.  Yes, I should have a chance to look at this tonight.

Issue #1890, opened 2016-12-05
Unlock all in trigger breaks labeling


hvacengi:
Originally posted here: http://forum.kerbalspaceprogram.com/index.php?/topic/61827-113-kos-scriptable-autopilot-system-v101-20161011/&do=findComment&comment=2875242

Running the code supplied produces a relabeling error (followed by an argument mismatch error which masks the underlying issue).  I've managed to simplify the offending code to:
```
function hello {
    print "hello world".
}
on ag1 { 
    unlock all.
}
```
The issue appears to be because pre-processing triggers actually calls `VisitNode` on the children of the trigger node (see [here](https://github.com/KSP-KOS/KOS/blob/develop/src/kOS.Safe/Compilation/KS/Compiler.cs#L486) and [here](https://github.com/KSP-KOS/KOS/blob/develop/src/kOS.Safe/Compilation/KS/Compiler.cs#L522)).  As a result, the `unlock all.` is visited before `PreProcessUserFunctionStatement` determines if the function name "hello" represents a lock or a function, and the default value is to report as a lock (because the `IsFunction` property isn't explicitly set yet, it returns `(default)bool` which is `false`).

I think that the answer would be to update `IndentifyUserFunctions` (which is called before pre-processing statements and triggers) so that it sets the `IsFunction` property, instead of doing it from inside `PreProcessUserFunctionStatement`.

PR #1891, opened 2016-12-05
Issue 1890 trigger unlock error

hvacengi:
Fixe #1890

Set the `IsFunction` flag during `IdentifyUserFunctions` to prevent errors when ulocking inside of triggers.  This is because the trigger code will Visit nodes during preprocessing before the flag is set via function pre-processing.

### DIFF
--- a/src/kOS.Safe/Compilation/KS/Compiler.cs
+++ b/src/kOS.Safe/Compilation/KS/Compiler.cs
@@ -568,7 +568,8 @@ namespace kOS.Safe.Compilation.KS
             ParseNode bodyNode;
             
             ParseNode lastSubNode = node.Nodes[node.Nodes.Count-1];
-            if (IsLockStatement(node))
+            bool isLock = IsLockStatement(node);
+            if (isLock)
             {
                 funcIdentifier = lastSubNode.Nodes[1].Token.Text;
                 bodyNode = lastSubNode.Nodes[3];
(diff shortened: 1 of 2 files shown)


src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -585,6 +586,7 @@ private void IdentifyUserFunctions(ParseNode node)
                 context.UserFunctions.GetUserFunction(funcIdentifier, storageType == StorageModifier.GLOBAL ? (Int16)0 : GetContainingScopeId(node), node);
             int expressionHash = ConcatenateNodes(bodyNode).GetHashCode();
             userFuncObject.GetUserFunctionOpcodes(expressionHash);
+            userFuncObject.IsFunction = !isLock;
```
Dunbaratu: Is it possible for userFuncObject to already be IsFunction before this is called, but not be a lock (i.e. it used `define function....`), in which case this line of code would turn its flag *off* when it had previously been on.

In any case, I'd be more comfortable with a one-way flip here:

```
if (isLock)
    userFuncObject.IsFunction = true;
```

That way this line can make the flag become true, but can't make it become false if it was already true for other reasons.

hvacengi: All I did was replicate the logic in `PreProcessUserFunctionStatement`, which uses the same inverted logic (I suspect as a legacy from some old system).  As far as I can tell that is the only other time that `IsFunction` is even set.

I'd be OK with some kind of a flip if that makes it read cleaner to you, but your example would still need a `!` in front of `isLock`.  We could instead store the variable returned by `IsDefineFunctionStatement` and set `IsFunction` based on that.

Dunbaratu: I guess keep it for now.  Now that I think of it, that *is* the meaning of IsFunction.
It means "a real function as opposed to being the kind of pseudo-function that comes from being a lock".

PR #1902, opened 2016-12-15
Fix Blizzy toolbar support

### LINKED ISSUE #1901: Blizzy toolbar support still broken
So apparently the Blizzy toolbar support is still broken in the current version.  The issue is a null reference when trying to access the `kOSCustomParameters` on the main menu.  Because a game isn't loaded, the custom parameters throw a null reference trying to access the current game.

hvacengi:
Fixes #1901

kOSCustomParameters.cs
* Protect Instance property getter from null reference if there is no
current game

KOSToolbarWindow.cs
* Add null check when testing for useBlizzyToolbarOnly
* Stop destroying the blizzy button in GoAway, blizzy doesn't require it
to be removed on scene change or game load, and we were destroying it
whenever leaving the main menu
* Fix issue with null reference and location of toolbar window when
UseBlizzyToolbarOnly is true

### DIFF
--- a/src/kOS/Module/kOSCustomParameters.cs
+++ b/src/kOS/Module/kOSCustomParameters.cs
@@ -14,7 +14,10 @@ namespace kOS.Module
             {
                 if (instance == null)
                 {
-                    instance = HighLogic.CurrentGame.Parameters.CustomParams<kOSCustomParameters>();
+                    if (HighLogic.CurrentGame != null)
+                    {
+                        instance = HighLogic.CurrentGame.Parameters.CustomParams<kOSCustomParameters>();
+                    }
                 }
                 return instance;
             }
--- a/src/kOS/Screen/KOSToolbarWindow.cs
+++ b/src/kOS/Screen/KOSToolbarWindow.cs
@@ -132,10 +132,9 @@ namespace kOS.Screen
         {
             if (!ApplicationLauncher.Ready) return;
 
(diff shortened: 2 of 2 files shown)


Dunbaratu:
How certain are you that this is good.  The code looks fine from reading it through, but I'm not keen to install blizzy to test.

hvacengi:
I'm fairly confident in it.  I sent a version to some forum users to confirm that it worked for their purposes too.  There is potential for a null reference to propagate to other calls to `Instance` but as far as I could tell this toolbar button is the only time that we check `Instance` before the game has loaded.

Maybe we could discuss where to locate the pop-up window when called using Blizzy, but in order to do that correctly we'd need to change the inheritance to get a callback.  I didn't want to do that.  I mostly wanted to stop the errors.

I ran code with this installed for a while before I made the PR.  Assuming you don't see any issues, it should be OK to merge.

Dunbaratu:
Okay.  By the way, do you think it's the error this user is seeing from the KSP forums?

http://forum.kerbalspaceprogram.com/index.php?/topic/61827-122-kos-scriptable-autopilot-system-v103-20161207/&do=findComment&comment=3013893

Dunbaratu:
In trying to test this I'm having a hard time making the original problem occur in the first place.
Blizzy toolbar works fine even without the patch.

But at any rate, after the patch is applied nothing seems to break, so I'm willing to accept it.

Issue #1913, opened 2016-12-29
SKID chip's play feature gets voices out of sync under physics lag


Dunbaratu:
If you are trying to play songs on two or more voices at once that are meant to be in sync with each other (i.e. a rhythm on one voice and a melody on another), and the game experiences a bit of physics lag (i..e the current time indicator in the upper-left of the game window occasionally shows as yellow instead of green), this will cause the two voices to become out of sync with each other and make the songs sound awful.

The problem is that each song list is implemented as simply:
Every physics tick:
   If prev note's start time + duration > current time, then it expired, so stop it and start next note now.

And so if you miss the physics tick when the current note was supposed to end and the next one was supposed to start, the next note gets its start time delayed until you do get a physics tick .  The result is that if this keeps happening, that voice's notes get pushed further and further into the future.  If one voice's note hits the end during a "missed physics tick" but the other voice's note didn't (because it had a longer duration and wasn't supposed to turn off yet anyway so the missed physics tick doesn't harm it), then one voice's notes got pushed into the future a bit when the other voice's notes didn't get pushed into the future.  This is what causes the mismatch.

To force the problem to exist to create a realiable test case that proves it will require coming up with a reliable way to force physics lag in a predictable way.  I am going to make two PR's for this: one to make a diagnostic tool that forces physics lag when and where you want it, to make a test case that reliably re-creates the problem, and then a second PR that actually solves it.

Dunbaratu:
This is an example program that can reliably trigger the problem (run it on a simple craft sitting on the launchpad in a low-lag environment with not much going on, so the only sources of physics lag come from the script itself choosing to cause lag on purpose.)

**To run this test you will first need to have PR #1914 installed, to get the `debugfreezegame()` built-in function call to exist.**


```
// Testing what happens when a song is run during system
// lag where physics ticks aren't coming fast enough,
// so a note's end-mark gets skipped over:

// Rhythm line, repeat this pattern, that takes 2 seconds per repetition:
set song1 to list(
  note(200,0.5,0.20),  // boop and take a half a second.
  note(200,0.5,0.20),  // boop and take a half a second.
  note(200,0.5,0.20),  // boop and take a half a second.
  note(0,0.5)          // rest lasting half second
  ).
set v1 to getvoice(1).
set v1:attack to 0.0.
set v1:decay to 0.0.
set v1:sustain to 1.0.
set v1:release to 0.1.
set v1:wave to "triangle".
set v1:volume to 0.8.
[...]

PR #1914, opened 2016-12-29
Added a debugfreezegame built-in function,

Dunbaratu:
For use when trying to force laggy situations when testing.
I did not make this specific to having to compile in
DEBUG mode. It's there all the time.  I figure it might be
useful for the users to have access to as well.

This is here partly to make the example code in issue #1913 work,
to prove the bug exists in a deterministic test.

(However, it's not documented, on purpose for now, in the
Sphinx docs, until we get a consensus on whether or not
to make it public.)

### DIFF
--- a/src/kOS/Function/Misc.cs
+++ b/src/kOS/Function/Misc.cs
@@ -451,6 +451,31 @@ namespace kOS.Function
             ReturnValue = shared.Cpu.DumpVariables();
         }
     }
+    
+    [Function("debugfreezegame")]
+    /// <summary>
+    /// Deliberately cause physics lag by making the main game thread sleep.
+    /// Clearly not something there's a good reason to do *except* when
+    /// trying to test a script that only fails when the game is lagging.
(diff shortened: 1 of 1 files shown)


src/kOS/Function/Misc.cs
```diff
@@ -451,6 +451,31 @@ public override void Execute(SharedObjects shared)
...
+    /// (Warning: Thread.Sleep() is incapable of being as precise as it may
+    /// seem.  It takes milliseconds, but often can only sleep in 15 or 30 ms
+    /// increments, rounding the sleep time up to the nearest increment chunk.)
+    /// </summary>
+    public class DebugFreezeGame : FunctionBase
+    {
+        public override void Execute(SharedObjects shared)
+        {
+            // How many milliseconds of extra sleep to cause?
+            int ms = Convert.ToInt32(PopValueAssert(shared));
```
hvacengi: I'm going switch this to use the `Function`'s `GetInt` method instead, just for the sake of consistency.

Issue #1916, opened 2017-01-01
Getting parts by regexp/glob on part tags


zegkljan:
We have the ``:PARTSTAGGED()`` method to get parts with the specified tag. However, I would find it useful if there was a similar method that would get parts with tags matching a given pattern. I hope I'm not the only one who would find it useful.

For example, I have parts with tags ``s0:part-a``, ``s0:part-b`` and ``s1:part-a``. Using that method, let's call it e.g. ``:PARTSTAGGEDPATTERN()``, I could get all the ``s0``-prefixed parts as ``:PARTSTAGGEDPATTERN("^s0.*")``, or to get all ``part-a``s, I could do ``:PARTSTAGGEDPATTERN(".*part-a")`` and so on.

It could either be classical regexp, or a shell-like glob pattern, but regexp would probably be more useful and universal.

hvacengi:
The easiest way to do this would be to simply pass things through the .net/mono regular expression engine.  Anything else introduces issues where we have to maintain our own library.  It does present potential issues with our normally case insensitive language however.

It shouldn't be a concern, but I felt it should still be noted that the Mono implementation that Unity/KSP uses has a very poor performing regular expression engine.  But as long as you aren't trying to match super long patterns or strings that shouldn't be a problem.  (This is why it takes so long to compile a program).

zegkljan:
The case insensitivity can easily be preserved by using case insensitive matching.

I already made a fork and will try to implement it. It should be fairly easy. The only thing that is not so straightforward is, as usual, naming, i.e. what the suffix should be. ``PARTSTAGGEDPATTERN`` is descriptive, but a kind of ugly.

zegkljan:
I took the liberty of extending my original proposal also to ``PARTS{NAMED,TITLED,DUBBED}`` since if one is pattern-searchable, why not the others.

hvacengi:
We've toyed around with the idea of being able to call ks code from the C# back end, allowing users to script a sort or select method.  This kind of thing would be a good thing to include with that.  Is it possible you could extend it to provide a suffix on the `StringValue` class that lets you test for a match there too?  Something like:
```
if "1234":matchespattern("\d.") {
}
```

zegkljan:
Sure.

PR #1918, opened 2017-01-07
Partstagged regexp

zegkljan:
I've succesfully implemented the stuff discussed in #1916. The reason for [Not Ready] is that the names of the new suffixes should be discussed and maybe better could be found. Other than that it should be working without any problems (tested on Ubuntu 16.10 under Mono 4.6.2 but that shouldn't be an issue).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@ packages/
 [Ee]xported[Oo]bj/
 *.userprefs
 
+# VisualStudio Code
+.vscode/
+
 # Unity
 [Ll]ibrary/
 [Tt]emp/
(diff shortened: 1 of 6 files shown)


zegkljan:
Well, I didn't find any better-sounding suffixes (though I did not do the thinking particularly hard). I'm removing the [Not Ready] from the PR title and let you others decide whether it is ok or not. Merge if you want to, discuss if you don't.

hvacengi:
I'm going to start my review of this tomorrow.  One comment that came up in dev chat was the possibility of swapping the phrase "pattern" with "regex" so that the pattern recognition is explicit.  But I'll look at that when I get into the code.  It could be that simply aliasing the suffixes makes sense.

PR #1928, opened 2017-01-15
Fixed math in VECTORANGLE docs

JonDahm:
The divisor in the equation should be the products of the vectors' magnitudes, not the magnitude of the dot product of the vectors (which is a meaningless statement anyways). The division is done to normalize the vectors.

I built the changes on my own machine using the instructions on [the Contribute page](http://ksp-kos.github.io/KOS_DOC/contribute.html#how-to-edit-this-documentation) and verified that they built properly and looked fine.

### DIFF
--- a/doc/source/math/vector.rst
+++ b/doc/source/math/vector.rst
@@ -259,13 +259,13 @@ Method / Operator                                                         Return
             \frac{
                 \vec{v_1}\cdot\vec{v_2}
             }{
-                \left|\vec{v_1}\cdot\vec{v_2}\right|
+                \left|\vec{v_1}\right|\cdot\left|\vec{v_2}\right|
             }
         \right)
 
     or in **KerboScript**::
 
-        arccos( (VDOT(v1,v2) / VDOT(v1,v2):MAG ) )
+        arccos( (VDOT(v1,v2) / (v1:MAG * v2:MAG) ) )
 
 .. function:: VXCL(v1,v2)
 


Dunbaratu:
Yup, you're right.  And thanks for the doc update (Especially for reading the instructions about how you have to edit the .rst. file, not the .html file.  We get a lot of well meaning helpers who don't see that and submit pull requests against the (autogenerated) HTML docs.

JonDahm:
I'm glad I could help! Let me know if there's anything I should change before you accept the pull request. I went ahead and also looked at where the math was actually done in the plugin to see if the same error was there. It looks sound. I'm assuming Vector3d is a Unity class? If so, I'm going to have to trust that their Vector3d.Angle() method is implemented properly.

Dunbaratu:
``Vector3`` is a Unity Class, but ``Vector3d`` on the other hand, is a class made by KSP.

The "d" in ``Vector3d`` doesn't mean "dimension" as in 3-D glasses.  It means "double".  Unity tries to do everything using just single-precision floats, but they have a lot of error at the large magnitudes that KSP needs (for things like "give me a vector pointing from this planet to that planet, measured in meters").   So KSP made the ``Vector3d`` class, which tries to mimic everything ``Vector3`` does, but do it using doubles instead of floats.  (There's one or two API methods that KSP missed that only exist for ``Vector3``, not ``Vector3d``, but most of them are implemented for both.)

PR #1938, opened 2017-01-22
Fixes #1937 string iterator suffix

### LINKED ISSUE #1937: Characters in a String should be iterable with FOR
Apparently this used to work once upon a time in kOS a few years ago:
```
SET str TO "abcde".

FOR c IN str {
  PRINT c.
}
```
In fact the documenation for strings in `string.rst` explicitly says this works (that's where the above example was pulled from).

But when you try doing it now you get an error that `StringValue` doesn't have an `ITERATOR` suffix on it.

I suspect it used to work and something changed but  I don't know what changed.  At any rate we should add an ITERATOR suffix to stringvalue.

Dunbaratu:
Fixes #1937 by adding the ITERATOR suffix to StringValue.

In the process of testing this I found that the :RESET method of that iterator was throwing errors.   In trying to fix that and comparing it to our other collection iterators that I assumed worked, I found out that their :RESETs didn't work either and have had the same problem for a long time now.

See the commit log messages for further explanation - but basically the problem with RESET cannot be fixed.  We're using that `yield return` magic for making iterators, and the kinds of iterators that this creates don't have Reset().

(I considered adding the "breaking" label because I'm removing a previously existing suffix, but I didn't because technically this suffix was already broken before and threw errors when used - this will just change which kind of error message people will see.)

### DIFF
--- a/doc/source/structures/misc/string.rst
+++ b/doc/source/structures/misc/string.rst
@@ -22,13 +22,55 @@ of existing strings. For example::
     SET s TO "Hello, Strings!".
     SET t TO s:REPLACE("Hello", "Goodbye").
 
-Strings are iterable. This scripts prints the string's characters one per line::
+ACCESSING INDIVIDUAL CHARACTERS
+-------------------------------
 
-  SET str TO "abcde".
+There's two main ways to access the individual characters
+of a string - using an iterator or using index numbers:
+
+Using an Iterator (FOR)
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Strings can be treated a little bit like iterable lists
+of characters. This allows them to be used in FOR loops
+as in the example below:
+
+  LOCAL str is "abcde".
 
   FOR c IN str {
-    PRINT c.
+    PRINT c.  // prints "a" the first time, then "b", etc.
   }
+
+The reason you can use Strings with the FOR loop like this is
+because you can obatain an :struct:`Iterator` of a string with the
+:attr:`ITERATOR` suffix mentioned below.  (Any type that 
+implements the ITERATOR suffix can do this.)
+
+Using an Index ( [i] )
+~~~~~~~~~~~~~~~~~~~~~~
+
+Strings can also be treated a little bit like lists in that
+they allow you to use the square-brackets operator `[`..`]`
+to choose one character by its index number (numbers start
+counting at zero).  Here's an example that does the same thing
+as the FOR loop above, but using index notation::
+
+  LOCAL str is "abcde".
+  local index is 0.
+  until index = str:LENGTH {
+    print str[index].
+    set index to index + 1.
+  }
+
+Be aware that despite being able to read the characters this way,
+you cannot set them this way.  The following will give
+an error::
+
+  LOCAL str is "abcde".
+
+  // The following line gives an error because you can't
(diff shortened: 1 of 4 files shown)


Dunbaratu:
Trying to understand how yield return makes an IEnumerator was a pain to figure out.  It's not that the concept of yield returns itself is hard (return a value but the next time you're called pick up here where you left off instead of at the start of the method).  That all made sense.  What was hard to find was the fact that the yield keyword also wraps the return value inside an Enumerator and returns the Enumerator instead.  The apparent mismatch between the return type in the method body and the return type in the method declaration is baffling if you don't know the keyword has that effect.  (It's weird because that means one class in the .net library has been given its own special keyword in the C# language.  That doesn't feel right because it violates the wall of separation between compiler and library.)

doc/source/structures/misc/string.rst
```diff
@@ -127,6 +169,9 @@ Structure
         * - :meth:`INSERT(index, string)`
           - :struct:`String`
           - Returns a new string with the given string inserted at the given index into this string
+        * - :attr:`ITERATOR`
```
hvacengi: In all of our other structures, we don't duplicate suffixes from inherited structure types.  But this table doesn't direct the user to go look for `Enumerator`.  I think I will add that reference and delete the `ITERATOR` documentation in this file if that's OK with you.  It also fixes an error since Sphinx is complaining that it's defined twice.

doc/source/structures/misc/string.rst
```diff
@@ -22,14 +22,56 @@ of existing strings. For example::
...
+
+There's two main ways to access the individual characters
+of a string - using an iterator or using index numbers:
+
+Using an Iterator (FOR)
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Strings can be treated a little bit like iterable lists
+of characters. This allows them to be used in FOR loops
+as in the example below:
```
hvacengi: Need to add a `:` for the code example to work.  I'll handle it in my review edits.

doc/source/structures/misc/string.rst
```diff
@@ -22,14 +22,56 @@ of existing strings. For example::
...
+  LOCAL str is "abcde".
 
   FOR c IN str {
-    PRINT c.
+    PRINT c.  // prints "a" the first time, then "b", etc.
   }
 
+The reason you can use Strings with the FOR loop like this is
+because you can obatain an :struct:`Iterator` of a string with the
+:attr:`ITERATOR` suffix mentioned below.  (Any type that
```
hvacengi: Because I'll be pointing to `Enumerator` below, I'll edit this to match.

hvacengi:
I have committed the changes I mentioned above to my local branch.  When you get a chance to look at my comments, if you agree with them I can just push them upstream.

doc/source/structures/misc/string.rst
Dunbaratu: Does string inherit from Enumerator?

hvacengi: Oh hey, you're right.  It is literally just an iterator suffix, not the rest of the enumerator stuff... I take it all back.  I'll just remove the enumerator portion of that attribute definition below instead.  Ignore my craziness.

PR #1941, opened 2017-01-27
Issue #1053: Compile script on separate thread

### LINKED ISSUE #1053: Split long compiles across muiltiple updates
Currently when executing a `run program.`, if `program` is anything other than a very quick little script, you can noticeably see the universe freeze for a bit while it finishes the compile.  This happens because the entire compilation must fit inside one pass of FixedUpdate(), and the system thinks of it as only taking one single instruction out of the IPU's.  Until the compile finishes, FixedUpdate() doesn't return, and the entire KSP game is frozen.
### Implementation idea:

To fix this, compilation should probably use the Unity Coroutine system.  When the compile starts, we begin a co-routine to execute the compile inside of, and the compiler checks itself with a stopwatch to yield after it's been too long, choosing to yield until the next FixedUpdate() to pick up and continue where it left off.

In the meantime, the main FixedUpdate() for that CPU that's doing the compile should do nothing until the compile coroutine is done.  It should know to bypass its work until the compile says its done.  Otherwise the program would end up continuing on past the `run` command while the `run` is still working, and that would break everything.

hvacengi:
Fixes #1053

This modifies how kOS compiles scripts both for `load` and `run` so that the regex parsing occurs on a separate thread.  This allows KSP, and even kOS triggers, to continue execution while waiting for the compilation to finish.  This should be a big boost to imersion for players, as it will eliminate the very noticiable lag when compiling large files.

Marked as "not ready" for:

- [ ] Walk documentation to find where threading should be mentioned, and where mentions of compile freezes should be removed.
- [ ] Figure out why the text cursor keeps blinking when a boot file is executed (not technically related, but in the same area)
- [ ] I added some comments in the code, but I probably should add some more.


src/kOS.Safe/Execution/YiedFinishedThreadedDetector.cs
```diff
@@ -0,0 +1,89 @@
...
+                if (childException == null)
+                {
+                    ThreadFinsh();
+                }
+                else
+                {
+                    // If it died due to a compile error, then we won't really be able to switch to program context
+                    // as was implied by calling Cpu.SwitchToProgramContext() up above.  The CPU needs to be
+                    // told that it's still in interpreter context, or else it fails to advance the interpreter's
+                    // instruction pointer and it will just try the "call run()" instruction again:
```
Dunbaratu: This comment is confusing because it refers to supposedly calling Cpu.SwitchToProgramContext() "up above" but there's no such thing happening "above" this point in the file.  It's as if this comment might have been cut and pasted from somewhere else, and now here the "up above" doesn't make sense anymore.  At any rate, I don't know what it refers to.

src/kOS.Safe/Execution/YiedFinishedThreadedDetector.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        /// <summary>
+        /// This method is executed before starting the child thread.  It is called from the main thread and is not required
+        /// to be thread safe with respect to KSP.
+        /// </summary>
+        /// <param name="shared"></param>
+        public abstract void ThreadInitialize(SafeSharedObjects shared);
+
+        /// <summary>
+        /// WARNING: THIS METHOD EXECUTES ON A SEPARATE THREAD.  TAKE CARE TO ENSURE THAT IMPLEMENTATIONS ARE THREAD SAFE
```
Dunbaratu: I'd add a sentence in the comment reminding the programmer how the Unity API calls should be avoided if you want to ensure your code is thread-safe.  (A Unity programmer should already know this, but we could get people (like me) for whom working on the kOS C# code was their first exposure to Unity.

src/kOS/Function/Misc.cs
```diff
@@ -198,32 +181,20 @@ public override void Execute(SharedObjects shared)
...
-                    try
-                    {
-                        codeParts = shared.ScriptHandler.Compile(path, 1, content.String, "program", options);
-                    }
-                    catch (Exception)
-                    {
-                        // If it died due to a compile error, then we won't really be able to switch to program context
-                        // as was implied by calling Cpu.SwitchToProgramContext() up above.  The CPU needs to be
-                        // told that it's still in interpreter context, or else it fails to advance the interpreter's
-                        // instruction pointer and it will just try the "call run()" instruction again:
```
Dunbaratu: Oh now I see - The comment was cut and pasted from here.  Yeah it will need to be edited due to the fact that its not located here anymore.

hvacengi:
@Dunbaratu I'm removing the Pending Documentation label, and I'm merging with upstream.   You should probably compile these docs to review, since much of the update is in the formatting and I'd like some feedback to make sure it looks good.

doc/source/commands/runprogram.rst
```diff
@@ -0,0 +1,226 @@
...
+        Warning detailing issues with case sensitive host file systems.
+
+.. note::
+    If you attempt to run the same program twice from within another script,
+    the previously compiled version will be executed without attempting to
+    recompile even if the original source script has been modified.  When a
+    program has finished executing and returns to the main terminal input, the
+    memory containing the programs is released.  This means that every time
+    that a script file is run from the terminal it is recompiled, even if the script
+    file has not changed.
```
Dunbaratu: I think this phrase:

``When a program has finished executing and returns``

needs to be prefaced by some sort of introductory phrase along the lines of "however", or "on the other hand", or "except when"...  This makes it more clear that we're not trying to clarify the previous situation, but to instead compare and contrast it with what happens in an opposite case.

Without that clause it sounds like a further detail or clarification about the same situation, which is confusing to someone trying to pick apart the sentence and understand what it's saying who doesn't already know how things work.

This seems to contradict itself and is confusing:
"This is how it works.  When doing foo, this (contradictory explanation) is how it works".

But this is much clearer:
"This is how it works.  *except when* doing foo, where *instead* this (contradictory explanation) is how it works".

doc/source/general/compiling.rst
```diff
@@ -18,7 +18,33 @@ When you try to run your script, the first thing that kOS does is transform your
 
 This process of transforming your script into Machine Language, or "ML" is called "Compiling".
 
-The ":ref:`RUN`" and ":ref:`RUNPATH`" commands do this silently, without telling you. This is why you may have noticed the universe slightly stutter for a moment when you first run your program. Compiling is hard work for the universe to do.
+The :key:`RUN` and :func:`RUNPATH` commands do this silently, without telling
+you.  The :key:`COMPILE` command explicitly compiles the file and saves it for
+future use.
+
```
Dunbaratu: I'll probably want a link :ref: target here, like ``.. _threaded_compile:`` so it can be linked to in the change log to come with release.

doc/source/general/compiling.rst
```diff
@@ -18,7 +18,33 @@ When you try to run your script, the first thing that kOS does is transform your
...
+.. note::
+    .. versionchanged:: 1.1.0
+        kOS now performs compilation on a parallel thread (thought to run in a
+        parallel universe).  This means that the universe no longer needs to
+        freeze in order to perform the compilation.  As with the execution of
+        other triggers, compiling from mainline code will not interupt or pause
+        triggers but compiling within a trigger will result in both the mainline
+        and trigger code pausing until the compilation is complete.
```
Dunbaratu: I don't think the high-level effect of the threaded compile is that clear here - that from the kOS computer's point of view, it doesn't *look* multi-threaded.  The relevant high-level change here is that the compile takes time to finish, when before it happened instantly as far as the simulated universe was concerned and it froze the universe in order to accomplish this.  I think that fact should get top billing in this paragraph.  The fact that under the hood we used threading to do this is an implementation detail the script program never really sees.

Let me give it a go and then you can decide if you like yours better, or mine, or maybe a mix of the two:

```
        kOS no longer freezes the Kerbal Space Program universe when it compiles programs.  Now 
        only your kOS program waits for the compile to finish, rather than the entire KSP game.
        When you issue a RUN or RUNPATH command that requires the program to be compiled,
        the KSP universe will continue moving forward in time while the kOS CPU waits for the compile
        to complete.  The only effect this will have on programs you run is that if you run a program
        you have to be aware that it may be several game-seconds later before the program actually
        starts.  During that time the ship may have moved a bit, or fell a short distance, etc.

        (To programmers familiar with Unity: Under the hood, this change was
        accomplished by moving the compile to a separate thread so kOS's
        FixedUpdate can abort early and allow the rest of KSP's universe to
        keep moving.  kOS's FixedUpdate continues aborting early, not
        executing any of its CPU instructions, until the compile thread is done,
        at which point it resumes execution where it left off.)
```

doc/source/commands/files.rst
```diff
@@ -406,196 +415,10 @@ Example::
...
-- 1: If no path information was present in the filename, then assume the
-  file is in the current directory (that's pretty much standard for all
-  filename commands).
-
-- 2: Assume if no filename extension such as ".ks" or ".ksm" was given,
-  and there is no file found that lacks an extension in the way the
-  filename was given, then first try to find a file with the ".ksm"
-  extension appended to it, and if that file is not found then try
-  to find a file with the ".ks" extension appended to it.
+You may run saved script files using the various :ref:`Run Command<running>`.
```
hvacengi: I don't mind adding the basic example and that closing statement.  That's short enough and not detailed enough to cause issues with maintenance.

But I would really prefer not to provide any qualified statements like talking about file names or contents.  I like those details being on the dedicated page.  That way they're all in one place if we change them in the future.  It isn't a big enough detail that I'll argue about it, but thought I should at lest mention it.  It may be a good place for a `.. seealso::` clause to call special attention to the link.

doc/source/commands/runprogram.rst
```diff
@@ -0,0 +1,226 @@
...
+================
+
+.. contents:: Contents
+    :local:
+    :depth: 2
+
+KerboScript supports running executing code saved in files on the local or
+archive disks. The terms "program" and "script" are generally used
+interchangably when describing this behavior. Programs can either be run using
+one of the :ref:`runfunctions` or the :ref:`run`.  Supported file types are:
```
hvacengi: When you compile the docs these references expand to match the referenced heading.  So `:ref:'run'` will turn into a link reading "Run Keyword".  It's just sphinx/rest shorthand.

Dunbaratu: So it's not possible to link to :ref:`run` without it adding the word "keyword"?  I can think of cases where I'd want the link but not that added word.

"try typing [run] here..." -> turns into "try typing [run keyword] here." if I make [run] a link.

doc/source/commands/runprogram.rst
hvacengi: How about:
>If you attempt to run the same program twice from within another script, the previously compiled version will be executed without attempting to recompile even if the original source script has been modified. **However once the** program has finished executing and returns to the main terminal input, the memory containing the programs is released. This means that every time that a script file is run from the terminal it is recompiled, even if the script file has not changed.

Dunbaratu: That's good.  I'm okay with that.

doc/source/general/compiling.rst
hvacengi: Do you want to add it now, or when writing the changelog?  I've got no problem adding it.

Dunbaratu: I guess it can wait.  Sphinx may complain about how it's orphaned and nothing is targeting it yet.

doc/source/general/compiling.rst
hvacengi: It sounds like you mostly want to make sure we clarify that the universe continues to execute during compilation, instead of essentially being instant in the frozen universe.  Obviously the two of us are going to have our own style/tendency in the way we write the text, so see if this accurately conveys your meaning, without re-writing the whole paragraph (modification in bold):

>kOS now performs compilation on a parallel thread (thought to run in a parallel universe). This means that the universe no longer needs to freeze in order to perform the compilation. **It also means that the process now takes time (possibly several seconds), while the universe continues moving until the compilation is complete.** As with the execution of other triggers, compiling from mainline code will not interrupt or pause triggers but compiling within a trigger will result in both the mainline and trigger code pausing until the compilation is complete.  **There may be many seconds between starting the compilation and executing the code, so you shouldn't assume that the physical properties (mass, position, velocity, etc.) will remain constant during compilation.**

(As I went through and kept adding your thoughts, my revised version got much more wordy than the original one sentence solution I had prepared, it may now be a big enough change that selecting your phrasing or mine might not matter much in the long run)

Given the tone of this document page, I didn't want to break the 4th wall by talking about the underlying system.  I'd be inclined to add your programmer note as a comment in the source instead of in the user docs.  If you think that the underlying implementation is important enough to note though, that's fine.  It probably means we should move this block of text out of a `.. note::` box though and under it's own heading, since it's more than just side note.

doc/source/commands/runprogram.rst
hvacengi:
```
:ref:`run`
```
will always be replaced with the heading title (I liked that choice so that it would automatically change if I revised the title).  But you can always modify the display label like this 
```
:ref:`my label here<run>`
```
That system will actually work for any of the `:role:` declarations, including structures and functions.
```
:func:`max(1,2)<max>`
:key:`RUN ONCE<RUN>`
```

doc/source/general/compiling.rst
hvacengi: Meh, Sphinx is already complaining about a broken link in the GUI documentation, so that shouldn't hold us up.  If we're in that file anyways making changes, we might as well add the label.

doc/source/general/compiling.rst
Dunbaratu: > Given the tone of this document page, I didn't want to break the 4th wall by talking about the underlying system.

Mentioning the threading is *already* breaking the forth-wall.  In the gameworld side of that forth wall, it doesn't look threaded.  My edit was mostly under the assumption you *did* want to break the 4th wall here (because you mentioned the threading that's invisible to game side of the wall), and if you're doing to do that then we should be clear that the threading you're talking about isn't a kOS in-gameworld thing.

I guess my problem is this:
```
print "before".
compile "foo.ks" to foo.ksm".
print "after".
```
Saying "performs compilation on a parallel thread" would give the impression that the above example will print ``after`` right away, BEFORE the compile command is done.  Because we claimed the compile happens in parallel.  I wanted it to be clear that in the not-breaking-4th-wall way of looking at things, compiles still happen in serial just like before.

doc/source/commands/runprogram.rst
Dunbaratu: Okay as long as there's an escape out of it then I'm okay with it.

doc/source/general/compiling.rst
hvacengi: Hm, I see what you're saying.  It never explicitly states that the mainline code is paused during the compilation process.  Let me think about that a little bit to see if I can restructure to make a short but clear description.

PR #1948, opened 2017-02-07
System fonts for gui terminal and text edit popup

### LINKED ISSUE #1356: Rescaled text/terminal window
I suppose this is a suggestion for much further down the line post KSP 1.1 patch, but I think it would be nice to be able to scale up the font size, or just scale up the whole terminal window itself, making the text easier to read on screenshots, youtube videos and streams. When I record in 1440p+ the font looks pretty small, and 720p playback on youtube is basically unreadable.

### LINKED ISSUE #1844: Support for more Unicode characters

Dunbaratu:
Fixes (or Makes irrelevant) the following issues:
Fixes #1844 Fixes #1356

This Pull Request allows the user to pick from whatever monospaced fonts exist on their OS to render the in-game gui terminal in that font (and the text edit popup as well).

Archetecturally, it consists of these parts:

- ``kOS.Module.AssetManager`` : A class with one single instance per game, that handles the lazy-loading of fonts (and maybe other things later which is why it's called Asset Manager not Font Manager) for use anywhere in the mod.
- Changes to the rendering of the content of ``TermWindow`` so it prints lines of text in the new font, instead of using our old ``Texture2D`` - based characters.  Important: **This should allow printing any Unicode character that the user's chosen font happens to support.**
- Edits to the ``CONFIG:`` system to name a font to use globally for all terminals, and consequent GUI widgets for the ``KOSToolbarWIndow` to allow the user to pick this.
- Edits to the ``CONFIG:`` system to allow the user to pick a default font size globally for all terminals.
- Moved ``CONFIG:BRIGHTNESS`` back to the global config section, and away from the per-game-save section.  It's a user interface preference, not a game setting.  Added the ability for the ``KOSToolbarWindow to set ``float`` sliders so it can work with this setting type.
- Re-purposed ``TERMINAL:CHARHEIGHT`` to act as the per-terminal separate font height setting.  (it inherits from ``CONFIG:DEFAULTFONTSIZE`` on first launching, but then can change per terminal.)
- Obsoleted ``TERMINAL:CHARWIDTH`` since when using a font we can't set height and width of characters separately - the font decides how wide characters are for a given height.  Also removed the associated adjustment widget from the terminal display.
- Obsoleted ``font_small.png`` file, and the code that used to process and use it.

### Breaking
This PR is marked as "Breaking" because of the things mentioned above that got obsoleted.

### "Not Ready"
The only reason this PR is marked as "Not Ready" is because the Sphinx documentation page that shows screenshots of the ``KOSToolbarWindow`` needs to be updated to match the new things that are on it.  (And that is where the documentation explaining the font choices would naturally go). 

I am leaving the screenshot part of the documentation incomplete for now because the look of that window could still change subject to PR review, so I don't want to take the screenshots of it until after the PR has been reviewed and accepted (accepted in every way other than the need to edit that one docs page, of course)


Dunbaratu:
I found a fairly consistent to reproduce bug that started with this PR.  Have you done any work with merging this, @hvacengi?  If not then I'm going to try to debug it and push a fix to this PR's branch.

The bug is this:  In a stock (except for kOS) install, with this PR in place, do these steps:

1 -  Launch a vessel with a kOS unit onto the launchpad.

2 -  Open the KOSToolBarMenu from Flight View of the vessel.

2 -  In the KOSToolBarMenu, Select a different font using the font selection button (make sure it's a monospace font that's accepted.  It doesn't trigger the bug if you pick a proportional font that gets rejected.)

3 -  Go to space center view from the ESC menu.

4 -  Reload the vessel back into flight view again by clicking on the vessel's icon you can see at the launchpad.

5 -  Try clicking the kOS button again on the toolbar.  Now the KOSToolBarMenu refuses to show up.  It does not appear ever again for the rest of the game unless you save the game and relaunch KSP.


This only seems to happen when you've used the font selector.  I don't know yet what causes it.

Dunbaratu:
Hmm.  I found out that ``KOSToolbarWindow.OnDestroy()`` is being called by Unity during the work that happens during "Scene Change : From FLIGHT to SPACECENTER".  And since KOSToolbarWindow is now a load-once-at-startup object rather than one that gets made and destroyed over and over like it used to be, that means once it's destroyed it's gone for good.

So that just leaves the question.... why is OnDestroy() getting called during scene change when the font picker has been previously invoked, but not when it hasn't?  More digging to do...

Dunbaratu:
Okay, I fixed it.  It was a simpler fix than I thought it would be (see commit log message).

src/kOS/Module/AssetManager.cs
```diff
@@ -144,6 +139,22 @@ public Font GetSystemFontByNameAndSize(string name, int size, bool checkMono, bo
...
+                // them loaded at once.  Relying on the cleanup that automatically happens when you
+                // orphan the Font takes too long to avoid this bug.  As long as it *temporarily* had
+                // too much font data in memory, it breaks some of the other fonts that were loaded but
+                // hadn't be excercised yet.  (i.e nothing drawn in Arial with default GUI.skin can be
+                // seen anymore).
+                // This is a bug that during our user community testing didn't seem to happen to everybody,
+                // but it happened to some users depending on kinds of fonts they had on their OS and kinds
+                // of graphics cards (it seems to be related to the texture size of the texture being sent to
+                // the graphics card).  At any rate, explicitly doing this instead of waiting for the
+                // cleanup that is epxected to "magically" happen later on its own seems to cure the problem:
```
Dunbaratu: Actually some comments by Sarbian indicate that Unity *never* cleans up the data inside the various C# wrapper classes upon orphaning and finalizing the C# class.  In other words orphaning the C# Font object will still leave behind the underlying font information it has in its C++ world.  Unity never does anything upon destructing a C# class instance.  When the C# instance contains some links to C++ data inside them, the C++ data never clears until there's a scene change.  That surprises me greatly, given how often Unity themselves show C# examples all over their documentation that has you creating new C# instances of things that it shows you just letting get orphaned and go away.  It never shows you explicitly destroying those things in those examples.

hvacengi: So, calling `Destroy` won't actually clean up the underlying unmanaged resources?  That's not all that useful.  I hope that the resources aren't reallocated if we create another instance of a wrapper around the same thing, otherwise that's pretty much a memory leak by design.

Dunbaratu: No, the other way around.  NOT calling Destroy will fail to clean up the Unmanaged resources, which is why I had to add the explicit call just before that instance of ``Font`` was about to get orphaned (The variable being assigned to a new instance of ``Font`` on the next iteration of this loop).

I would have assumed that there's a ``~Font()`` somewhere that performs the Destroy for you.  What Sarbian was saying is that there isn't.  This seems like wrong design to me but apparently that's how Unity works.  It's API has constructors that create unmanaged objects but doesn't pair those up with destructors that clear them.

Dunbaratu: So in other words where I was presuming (in the code comment above) that the problem was caused by creating all the fonts faster than the GC would clear them out, actually the problem would have occurred no matter how fast the GC gets around to freeing my orphaned ``Font``s, because even if it *did* call all the ``~Font()`` destructors for them fast while in the midst of that loop that still wouldn't have really cleared the underlying Font data inside Unity that caused the problem.  Whatever data structure it is in Unity that has an upper maximum to how much font information you're allowed to have alive at once, that is part of what doesn't clear when ``~Font()`` is called.

If I didn't explicitly ``Destroy()`` it, then it would have lived not just up to the next GC pass, as my comment had said, but it would actually have lived all the way until the next scene change.

src/kOS/Screen/TermWindow.cs
```diff
@@ -259,6 +223,26 @@ public override void Open()
...
+            {
+                fontSize = newSize;
+                tryFontNames[0] = newName;
+                font = AssetManager.Instance.GetSystemFontByNameAndSize(tryFontNames, fontSize, false);
+
+                terminalLetterSkin.label.font = font;
+                terminalLetterSkin.label.fontSize = fontSize;
+
+                // TODO - There will probably be other measurements that have to get recalculated
+                // here too.
```
hvacengi: I'm removing this todo, it doesn't look like there is anything that needs to be added here for now.

src/kOS/Screen/TermWindow.cs
```diff
@@ -823,15 +819,14 @@ void TerminalGui(int windowId)
             screen.Brightness = GUI.VerticalSlider(brightnessRect, screen.Brightness, 1f, 0f);
             GUI.DrawTexture(brightnessButtonRect, brightnessButtonImage);
 
-            int charWidth = screen.CharacterPixelWidth;
+            // int charWidth = screen.CharacterPixelWidth;
```
hvacengi: deleting commented old block.

src/kOS/Screen/TermWindow.cs
```diff
@@ -963,11 +961,20 @@ protected void CheckResizeDrag()
             }
         }
         
-        void ShowCharacterByAscii(char ch, int x, int y, bool reversingScreen, int charWidth, int charHeight, float brightness)
+        void DrawCursorAt(char ch, int x, int y, bool reversingScreen, int charWidth, int charHeight, float brightness)
```
hvacengi: I want to say that if we printed whole lines at a time instead of individual characters we could save some time/memory.  Except our OnGui method is far cheaper than the actual code execution with the PR, so I don't really want to waste time with it.

src/kOS/Screen/KOSToolbarWindow.cs
```diff
@@ -459,6 +500,41 @@ public void DrawWindow(int windowID)
...
+                    kOS.Screen.ListPickerDialog.CloseAction onClose = delegate() { fontPicker = null; };
+
+                    fontPicker.Summon(windowRect.x, windowRect.y + windowRect.height, 300,
+                        key.Name, "(Only fonts detected as monospaced are shown.)",
+                        key.Value.ToString(), AssetManager.Instance.GetSystemFontNames(), onChange, onClose
+                        );
+                }
+                else
+                {
+                    fontPicker.Close();
```
Dunbaratu: Okay, agreed.

src/kOS/Screen/TermWindow.cs
Dunbaratu: I believe this is only still being used to draw the block cursor only.
For drawing the rest of the screen, I already changed it to draw one screen line at a time using GUI.Label(), or so I thought.

hvacengi: You're right.  I saw the method for drawing a character and didn't go back and see that indeed the main display is drawn line by line.   My bad for not doing enough research before commenting.

Dunbaratu: I didn't remove it because it sounded like the thing that could still be used for something important later.  It was complex enough that I don't want to forget how we did it.

Issue #1063, opened 2015-06-07
unable to differentiate similar-named part modules


Gaiiden:
As seen in the attached screenshot, I have three part modules all named ModuleScienceExperiment, and I am only able to call the first one on the list. Given how you can get a List of part module names, perhaps it would be better to just reference them by index rather than a string? Or at least in addition to a string?

![n0f1jnt](https://cloud.githubusercontent.com/assets/7699577/8026530/dd64b52c-0d47-11e5-872e-8696c594882c.png)

hvacengi:
Honestly, I've always thought it was a little odd that modules and `allmodules` return strings instead of the modules themselves.  This is an excellent reason why (non unique name identifiers).  My preference for a fix, even though it would be a breaking change, would be to leave `modules` returning the string list, and `allmodules` return a module list.  Alternatively, we could add a new suffix of `getmoduleindex(idx)` or whatever similar syntax.  There isn't even a work around for this by iterating through the list, since the only way to get the module itself is the `getmodule` suffix.

@Dunbaratu @erendrake (or anyone else) do you have a preference before I commit to a major breaking change?

hvacengi:
OK, I just went to open the code and look at this, it looks like we do have a `GETMODULEBYINDEX` suffix already.  It just is not documented.  We will have to add it to the documentation.  The proper syntax should be 
```
print ship:partstitled("avionics package")[0]:getmodulebyindex(7):allevents.
print ship:partstitled("avionics package")[0]:getmodulebyindex(8):allevents.
print ship:partstitled("avionics package")[0]:getmodulebyindex(9):allevents.
```

I still would like to consider creating a list of modules themselves, but this at least helps.  It still would mean you'd need to iterate through the list of names to create a list of indexes (no guarantee they're always in the same order) and then access them by index, which is pretty messy when we could just provide the list directly.

amedee:
I was also looking at the problem of modules with the same name and found issue #510 with pull request #511. I noticed that `GETMODULEBYINDEX` isn't documented and wanted to add an issue and maybe a pull request, but first I searched for `GETMODULEBYINDEX` and got here.

So what I propose is, I will try to figure out how the documentation works, add the missing bit, and send in a pull request. Will that be ok?

hvacengi:
[...]

PR #1962, opened 2017-02-24
Document GETMODULEBYINDEX suffix

amedee:
See #510 and #1063


hvacengi:
Thanks for the pull request.  At first blush it looks like it should compile just fine.  There shouldn't be an issue getting this compiled into our next release.

PR #1965, opened 2017-02-25
Remove some empty functions in VesselModule

soulsource:
This greatly reduces garbage creation, as even an empty OnGUI() is causing 384 kB of garbage every time it's called, and VesselModules are active for every vessel in the game (also those that are not loaded, see
http://forum.kerbalspaceprogram.com/index.php?/topic/156959-vessel-modules-seem-to-run-for-inactive-vessels-now/
With this simple change I got rid of about 50 kB garbage per frame in my current save game.

### DIFF
--- a/src/kOS/Module/kOSVesselModule.cs
+++ b/src/kOS/Module/kOSVesselModule.cs
@@ -149,24 +149,24 @@ namespace kOS.Module
         /// <summary>
         /// Update is called once per UI tick.
         /// </summary>
-        public void Update()
-        {
-        }
+        //public void Update()
+        //{
+        //}
 
         /// <summary>
         /// LastUpdate is called at the end of the update frame.  Useful if something needs to be
         /// done after the physics and UI updates have been completed, but before scene or gui rendering.
         /// </summary>
-        public void LastUpdate()
-        {
-        }
+        //public void LastUpdate()
+        //{
+        //}
 
         /// <summary>
         /// OnGUI is where any vessel specific interface rendering should be done.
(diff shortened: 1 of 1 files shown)


Dunbaratu:
Yes, this is a good idea.  I like that you just commented them out instead of deleting.  I approve, because it's clear from those header comments that part of the reason for having these there was to document their existence so we could fill them in later if we like.  But until we do feel like filling them in, we shouldn't be having a lot of do-nothing hooks getting called all the time.

Dunbaratu:
I decided to re-implement your change here in my own branch instead of pulling your change, because of how you made your change right in your `develop` branch instead of making a separate branch for it.

If this is the only PR you ever will submit to KOS, it would be no big deal to just pull your change as-is, but if you ever might give more PR's later, then the fact that you did it right in the develop branch would make it a bit hard to organize on our end.

(Our usual pattern is: you make a branch from `develop`, with something like:
```
git checkout develop
git checkout -b issue_9999_remove_unused_hooks
# makes edits and tests them...
git commit -a
git push origin issue_9999_remove_unused_hooks
# Now goes onto github and makes a PR requesting to merge soulsource:issue_9999_remove_unused_hooks into KSP-KOS:develop
```
)
Thus the branches remain for posterity if we want to go back and look at them.

Anyway, thanks for the fix.  It was a good catch.

soulsource:
Hi!
The problem was that I wasn't really aware of GitFlow at that point, thanks for your clarifications. I'll make feature-branches in the future.

PR #1967, opened 2017-03-01
Changed some places scalarCompare into stringCompare

edogrigqv2:
I was glimpsing through the code and noticed a small mistake.


hvacengi:
Thank you for the pull request.  That does appear to be a mistake.  I never really understood the support for min/max of strings anyways, but since some one appears to be using it we should definitely get that fixed.

Dunbaratu:
This is obviously from a cut & paste error.

If (foo) {  } else if (foo) {  }

If the condition foo is the same for both cases, then the else-if can never happen.

Prior to this PR, the else-if was the same condition as the if.

I think I'm just going to cheat and hit the "merge" button without bothering to pull this locally.

Issue #1964, opened 2017-02-25
Sending messages from KSC to vessel


zegkljan:
kOS processors can send and receive messages. But there is basically no way of sending data to vessels from the KSC (imagine sending instructions). The only way (that I'm aware of) of emulating this is that the vessel checks the archive (when it has connection to KSC) and looks for specific files it can copy over and run them.

I thought of a new mechanism of sending data to the vessel(s) from KSC, without the vessel being an active participant, just a receiver. I imagine it could work like this: when flying a vessel, the kOS UI would provide a "console" (or something like it) where the player could type arbitrary string* and then hit "send" and the message would be sent to the active vessel (or maybe, if multiple vessels are loaded, the player could choose from them) as a proper ``Message``. The vessel would then receive it in exactly the same way it would when sent from another vessel/processor, except the ``SENDER`` would be some constant representing KSC.

*Or bool, or number or maybe an arbitrary serializable kOS object, i.e. anything that can be the content of a ``Message``. The language could be something like JSON but with kOS data structures, i.e. a subset of kOS.

PR #1972, opened 2017-03-08
Ksc messages [NOT READY]

zegkljan:
So, I got a little bored (well, not really) and just went ahead and implemented what I proposed in #1964.

If you want to see it in "action", you can view a few screenshots here: http://imgur.com/a/dUyR0

I added a single new button to the main kOS window (the one that has the list of loaded processors) that allows to open an "uplink channel" to the currently active vessel. When opened a window very similar to the one used to edit source files appears. In this window the player can type whatever she wants to. Then she can hit the Send button and the content of the window is sent as a string with a proper delay (to home since we are sending from KSC).

Now, there are some things that are kind of hackish. Before this, the ``SENDER`` of a message was either ``Vessel`` that corresponded to the sender or it was ``False``. Now I needed to distinguish one extra case - the sender is the KSC. I achieved this by setting the ``SENDER`` to ``True`` when this is the case, but ``HASSENDER`` still returns false. In other words, the new "sender table" looks like this:

- ``HASSENDER`` is true, ``SENDER`` is a vessel: ``SENDER`` contains the vessel that sent the message
- ``HASSENDER`` is false, ``SENDER`` is false: the message was sent from a vessel that no longer exists
- ``HASSENDER`` is false, ``SENDER`` is true: the message was sent from KSC

In order to do that I modified the ``GetVesselTarget()`` method of ``MessageStructure``.

The other hackish thing is the way I represent the actual sender when the message is built. The ``Message.Create()`` needs a ``VesselTarget`` to be the sender. Therefore I created a new class ``KscTarget`` that is a singleton. I also made the method ``GetGuid()`` in ``VesselTarget`` virtual so that ``KscTarget`` can override it. When doing the ``GetVesselTarget()`` in ``MessageStructure``, if no vessel is found (via ``GetVessel()``), the singleton ``KscTarget``'s guid is checked for equality with the message's vessel.

Be warned that this is effectively my first touch on KSP mods and C# as well (apart from the regexp matching on part tags I have already done) so feel free to bash my head because all the mistakes I have done (which I'm not aware of).

The reason of the [NOT READY] thing is that I have no idea whether I have done it in a good way, and also because the documentation is not updated.

### DIFF
--- a/src/.vscode/tasks.json
+++ b/src/.vscode/tasks.json
@@ -1,0 +1,26 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
+    "version": "0.1.0",
+    "windows": {
+        "command": "msbuild"
+    },
+    "linux": {
+        "command": "xbuild"
+    },
+    "args": [
+        // Ask msbuild to generate full paths for file names.
+        "/property:GenerateFullPaths=true"
+    ],
+    "taskSelector": "/t:",
+    "showOutput": "silent",
+    "tasks": [
+        {
+            "taskName": "build",
+            // Show the output window only if unrecognized errors occur.
+            "showOutput": "silent",
+            // Use the standard MS compiler pattern to detect errors, warnings and infos
+            "problemMatcher": "$msCompile"
+        }
+    ]
+}
--- a/src/kOS/Communication/KscTarget.cs
+++ b/src/kOS/Communication/KscTarget.cs
@@ -1,0 +1,25 @@
+using System;
+using kOS.Suffixed;
+
+namespace kOS.Communication
+{
+    [Safe.Utilities.KOSNomenclature("KscTarget", KOSToCSharp = false)]
+    public class KscTarget : VesselTarget {
+        private static readonly KscTarget kscTarget = new KscTarget();
+
+        public static KscTarget Instance
+        {
+            get {
+                return kscTarget;
+            }
+        }
+        
+        private KscTarget()
+        {}
+
+        public override Guid GetGuid()
+        {
+            return Guid.Empty;
+        }
+    }
+}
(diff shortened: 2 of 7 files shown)


Dunbaratu:
This looks like a good idea.  I haven't had time to look at the code yet because I had my head buried in the GUI system and the fonts changes.  But it looks good and we should get around to looking at it.

As I understand it, it provides a way for you to send a message using the inter-vessel comms system from the KSC to a vessel.  Presumably the message has to be a string.

zegkljan:
Yes, exactly - you open a dedicated window where you write whatever you want and then hit Send and (after a delay, if any) the vessel receives the message with that string as the content.

I thought a little bit about sending more than strings but that would need one of two things:

- literals for lexicons and lists that are currently not in the language (or at least I think so)
- ability to run kOS code outside the vessel (in the KSC) that would do the message sending - that would probably be quite complicated as a load of variables that are usually there would not be, only some things would be allowed, etc.

Anyway, I decided to go the simplest way and just send strings for now. It is the receiver's business to parse it. In the future, if the lexicon and list literals show up, it should be easy to integrate - e.g. a checkbox would be added to the window that would tell whether I want to parse it or just send as string and then upon sending the message the typed stuff would be parsed.

hvacengi:
I'm moving the `[Not Ready]` title to a `Pending Documentation` label.  We can start working on reviewing the code while you figure out the documentation side.  Plus then we know that if it looks like we can handle the documentation easily enough we can do it during the code review.

zegkljan:
Everything should be documented. The only exception is the ``/_images/general/controlPanelWindow.png`` which should to be redone to reflect other changes too. For the time being I added a small TODO just above that picture in ``/general/settingsWindows.rst``.

Dunbaratu:
Thanks.  I'm digging through some older issues and PR's so it may be a day or two before I look at this.

sebseb7:
give the space center its own cpu and gui

zegkljan:
@sebseb7 Yes, that would be awesome. However, it would be also a lot of work (a lot of things in kerboscript assumes that there is a vessel to control, e.g. the whole SHIP variable) which for me is too much, as I don't know C# very much and also I know next to nothing about how kOS is internally organized. I thought this could be a first step towards what might be a KSC-side CPU.

PR #1976, opened 2017-03-22
Choose ip addr

### LINKED ISSUE #668: Telnet server binds to "wrong" interface
I have more than one IPv4 interface on my PC, and the kOS telnet server insists on attaching to my storage network interface rather than the one shared with the rest of the PCs on my network, including the laptop that I'd like to telnet from.

Is there a way to either hardcode the correct IP in a config file, or have the telnet server attach to all interfaces?

### LINKED ISSUE #1935: Telnet server needs to allow the user to choose which "real" IP address is the right one.
The telnet server can be run in one of two modes: "loopback" or "real" IP address.

The problem with the "real" IP address mode is that it just presumes it can ask the OS for "what is my IP" and assume that of the several IP addresses it gets back, the canonical "most official" answer is the first response in the list and it should use that.  Most modern computers now have at least two "real" addresses even if they have only one network adapter, because they have both an ipv6 and an ipv4 address.

We need to make it possible for a user to pick which address they mean when disabling loopback mode.  Just now I tried testing it and it tried assigning itself the ipv6 address instead of the ipv4 one.

Dunbaratu:
Fixes #1935 
Fixes #668

Now you will be allowed to choose which of the IP addresses on your computer is the one you want it to use for the telnet server.  This replaces the previous CONFIG:LOOPBACK boolean with a new CONFIG:IPADDRESS string that lets you pick the IP address (i.e. set it to "127.0.0.1" to get the old loopback behaviour.)

It also contains the changes to the ``KOSToolBarMenu` to show a pick list of the IP addresses of the local machine so you can pick from the list instead of typing it in.

### Do not merge until after PR #1948 !!!

Because this change depends on the existence of the `ListPickerDialog` introduced with PR #1948 , this should not be merged until after PR #1948 is.  Please note that the majority of the diffs you see in the "Files Changed" are there purely because PR #1948 isn't merged yet.  This change *actually* only changed 2 source files and some RST docs - the rest is carryover from #1948.

### DIFF
--- a/src/kOS/Screen/TermWindow.cs
+++ b/src/kOS/Screen/TermWindow.cs
@@ -1031,6 +1031,7 @@ namespace kOS.Screen
         {
             server.DisconnectFromProcessor();
             telnets.Remove(server);
+            prevTelnetScreens.Remove(server);
         }
         
         public void DetachAllTelnets()
--- a/src/kOS/Screen/TermWindow.cs
+++ b/src/kOS/Screen/TermWindow.cs
@@ -89,7 +89,7 @@ namespace kOS.Screen
         private KOSTextEditPopup popupEditor;
 
         // data stored per telnet client attached:
-        private readonly List<TelnetSingletonServer> telnets; // support exists for more than one telnet client to be attached to the same terminal, thus this is a list.
+        private volatile List<TelnetSingletonServer> telnets; // support exists for more than one telnet client to be attached to the same terminal, thus this is a list.
         private readonly Dictionary<TelnetSingletonServer, IScreenSnapShot> prevTelnetScreens;
         
         private ExpectNextChar inputExpected = ExpectNextChar.NORMAL;
@@ -549,11 +549,20 @@ namespace kOS.Screen
         /// </summary>
         private void ProcessTelnetInput()
         {
-            foreach (var telnet in telnets)
+            // It's possible to close and remove telnets from the list during the processing
+            // of input (if the detach signal Ctrl-D is sent).  Therefore we have to
+            // make this temp copy to prorect against the C# error "Collection was Modified"
+            // during the foreach loop:
+            TelnetSingletonServer[] tempTelnetList = new TelnetSingletonServer[telnets.Count()];
+            telnets.CopyTo(tempTelnetList);
+            foreach (TelnetSingletonServer telnet in tempTelnetList)
             {
-                while (telnet.InputWaiting())
+                if (telnet.ConnectedProcessor != null)
                 {
-                    ProcessOneInputChar(telnet.ReadChar(), telnet);
+                    while (telnet.InputWaiting())
+                    {
+                        ProcessOneInputChar(telnet.ReadChar(), telnet);
+                    }
                 }
             }
         }
@@ -633,7 +642,7 @@ namespace kOS.Screen
                             if (whichTelnet == null)
                                 Close();
                             else
-                                whichTelnet.DisconnectFromProcessor();
+                                DetachTelnet(whichTelnet);
                         }
                         break;
                         
--- a/src/kOS/UserIO/TelnetMainServer.cs
+++ b/src/kOS/UserIO/TelnetMainServer.cs
@@ -27,6 +27,7 @@ namespace kOS.UserIO
     {
         private TcpListener server = null;
         private IPAddress bindAddr;
+        private string bindAddrName = ""; // used to compare new address values to old
         private Int32 port;
         private readonly List<TelnetSingletonServer> telnets;
         private bool isListening;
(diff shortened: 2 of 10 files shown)


src/kOS/UserIO/TelnetMainServer.cs
```diff
@@ -143,18 +170,26 @@ private void SetPermanentRealIPPermission(bool newValue)
         
         public void SetConfigEnable(bool newVal)
         {
+            if (bindAddr == null) // when nothing has been set yet, this will be the case.
```
hvacengi: `bindAddr` is only set as part of `StartListening` or by selecting an IP address from the toolbar window.  This means that this line will prevent the server from being started unless you manually select an IP address first.  My test case was to open the game and try and toggle telnet on, but every time I clicked the button it would just clear itself.  After I manually selected the loop back address (which the window claimed was currently selected anyways) the toggle worked properly.  My recommendation is that we add `SetBindAddrFromString(SafeHouse.Config.TelnetIPAddrString)` into the `Start` method.  That presents other issues in that `SafeHouse.Config` may still be null (both `Bootstrapper` and `TelnetMainServer` are `StartInstantly`).  I would either move `TelnetMainServer` to start at the `MainMenu` or move `Bootstrapper`'s logic into `Awake`.  I tested the `MainMenu` solution and it worked fine in my use.

hvacengi: Don't make this change yourself.  I just had a thought about another way to implement it that would be a little cleaner.  I can do it as part of my review.

src/kOS/Screen/TermWindow.cs
```diff
@@ -89,7 +89,7 @@ public class TermWindow : KOSManagedWindow , ITermWindow
         private KOSTextEditPopup popupEditor;
 
         // data stored per telnet client attached:
-        private readonly List<TelnetSingletonServer> telnets; // support exists for more than one telnet client to be attached to the same terminal, thus this is a list.
+        private volatile List<TelnetSingletonServer> telnets; // support exists for more than one telnet client to be attached to the same terminal, thus this is a list.
```
Dunbaratu: The writing thread in the telnet does end up trying to edit this list, when it detects that the user hit ctrl-D to detach the terminal from the CPU.  (not directly, but as it tries to shut everything down about the association with the terminal, one of the methods it's calling ends up deleting the telnet from this list.)

That's why this was changed to volatile to remind C# it can be thread-accessed.

src/kOS/Screen/TermWindow.cs
```diff
@@ -549,11 +549,20 @@ public int GetFontSize()
         /// </summary>
         private void ProcessTelnetInput()
         {
-            foreach (var telnet in telnets)
+            // It's possible to close and remove telnets from the list during the processing
+            // of input (if the detach signal Ctrl-D is sent).  Therefore we have to
+            // make this temp copy to prorect against the C# error "Collection was Modified"
+            // during the foreach loop:
```
Dunbaratu: You'll see this comment in several places in this commit.  Anywhere that something fitting the following pattern was used:
```
foreach (var telnet in telnets)
{
   do something that could result in detaching the telnet from the telnets list.
}
```
The telnet list can have one of the telnets removed from it during the right kinds of keypresses, or when stopping the listener, so it's not safe to ``foreach`` through it while that can happen.  This problem was probably always there but not noticed because I wasn't running the debug EXE of Kerbal, so I'd only notice if I looked at the output log. (where the debug version shows the exception right away on screen as it happens so it's blatant).  The reason I never saw any major failures before was that this was always happening just as things were detaching and going away so most of the bad state of data caused by the exception aborting the work partway through didn't matter.  (I was constructing a new instance anyway, not re-using the same one like I am now.)

src/kOS/Screen/TermWindow.cs
```diff
@@ -633,7 +642,7 @@ public void ProcessOneInputChar(char ch, TelnetSingletonServer whichTelnet, bool
                             if (whichTelnet == null)
                                 Close();
                             else
-                                whichTelnet.DisconnectFromProcessor();
+                                DetachTelnet(whichTelnet);
```
Dunbaratu: There's really no difference here, other than me thinking "well, if I have this public method to do this, I should probably be using it internally too instead of explicitly doing the same code manually."

src/kOS/UserIO/TelnetSingletonServer.cs
```diff
@@ -170,15 +170,26 @@ public TelnetSingletonServer(TelnetMainServer mainServer, TcpClient client, int
             rawStream = client.GetStream();
             lock (keepAliveAccess)
                 gotSomeRecentTraffic = true;
-            inThread = new Thread(DoInThread);
-            outThread = new Thread(DoOutThread);
             outQueue = new Queue<char>();
             inQueue = new Queue<char>();
+            SpawnWelcomeMenu();
+            inThread = new Thread(DoInThread);
+            outThread = new Thread(DoOutThread);
```
Dunbaratu: This is the core change that in turn caused all the other changes.

The SpawnWelcomeMenu() is now happening before the threads are forked off, back in the main thread.
Previously, it was done inside ``outThread``'s ``ContinuousChecks()``.

To make this work, it was necessary to move from a design of "throw TelnetWelcomeMenu away and make a new one each time you go back to the menu screen", to one of "re-use the same TelnetWelcomeMenu and just reset it."

That change is the primary reason for most of the rest of what's in this commit.

src/kOS/UserIO/TelnetSingletonServer.cs
```diff
@@ -170,15 +170,26 @@ public TelnetSingletonServer(TelnetMainServer mainServer, TcpClient client, int
...
 
+        ~TelnetSingletonServer()
+        {
+            if (welcomeMenu != null)
+            {
+                welcomeMenu.Detach();
+                GameObject.Destroy(welcomeMenu);
+                welcomeMenu = null;
+            }
+        }
```
Dunbaratu: Since we no longer get rid of ``welcomeMenu`` and re-make a new one each time, the final point where we really do get rid of it has to wait until we're all done with this telnet singleton client.

src/kOS/UserIO/TelnetSingletonServer.cs
```diff
@@ -539,6 +556,7 @@ private void DoOutThread()
                 ContinuousChecks();
 
                 sb.Remove(0,sb.Length); // clear the whole thing.
+                Console.WriteLine("outThread, before lock section");
```
Dunbaratu: ah crud.  I forgot to include the magic word "eraseme" here so I didn't notice this debug line being left behind.  I'll push a new commit in a moment that has it removed.

src/kOS/UserIO/TelnetSingletonServer.cs
```diff
@@ -604,7 +620,7 @@ private void SpawnWelcomeMenu()
             var gObj = new GameObject( "TelnetWelcomeMenu_" + MySpawnOrder, typeof(TelnetWelcomeMenu) );
             Object.DontDestroyOnLoad(gObj);
             welcomeMenu = (TelnetWelcomeMenu)gObj.GetComponent(typeof(TelnetWelcomeMenu));
-            welcomeMenu.Setup(this);
+            welcomeMenu.Attach(this);
```
Dunbaratu: Same method, new name.  Since I now call it more than once in the life of the ``welcomeMenu`` it felt wrong to call it "Setup".

src/kOS/UserIO/TelnetWelcomeMenu.cs
```diff
@@ -25,7 +25,9 @@ public class TelnetWelcomeMenu : MonoBehaviour
         private readonly StringBuilder localMenuBuffer;
         private bool firstTime;
         private volatile bool forceMenuReprint;
-        
+
+        public bool IsAttached { get { return telnetServer != null; } }
+
         // Because this will be created as a Unity Gameobject, it has to have a parameterless constructor.
         // Actual setup args will be in the Setup() method below.
```
Dunbaratu: ahh - should rename that comment's use of the word "Setup".

Issue #1982, opened 2017-04-09
Error messages won't stop and program won't finish breaking, when two CPU's try to steer the same ship.


Dunbaratu:
(This was part of issue #1980, but I discovered that it's a second unrelated bug that still persists after the docking freeze happens, so I'm making it a separate issue.)

It is correct to throw an error when two CPU's try to use lock steering on the same vessel, but in this instance that error causes the program to get stuck trying to re-run the same instruction forever, bombing with nullref again and again and never quite breaking from the program (like an error message should cause to happen.)

To re-create, perform the exact same steps as in issue #1980.

After the merge of the two vessels, in addition to the (entirely expected) error message complaining that both kOS units are now trying to control the same vessel, you will also see the (unexpected) spamming of a null reference exception over and over and over, apparently once per FixedUpdate.

That exception looks like this:
```
NullReferenceException: Object reference not set to an instance of an object
  at kOS.Safe.Execution.CPU.SkipCurrentInstructionId () [0x00000] in <filename unknown>:0 
  at kOS.Safe.Execution.CPU.KOSFixedUpdate (Double deltaTime) [0x00000] in <filename unknown>:0 
  at kOS.Safe.UpdateHandler.UpdateFixedObservers (Double deltaTime) [0x00000] in <filename unknown>:0 
  at kOS.Module.kOSProcessor.UpdateFixedObservers () [0x00000] in <filename unknown>:0 
  at kOS.Module.kOSProcessor.FixedUpdate () [0x00000] in <filename unknown>:0 
 ```
(This exception is also dumping the program instructions to the log each time, in the same way that errors that are the script's fault normally do.)

I split this into a separate issue because it turns out that solving the cause of the freezing didn't solve this problem, so they're two different bugs.

Dunbaratu:
Closing because merging #1983 was supposed to auto-close this issue too, but it didn't.  I don't know if Github will let you associate one PR with two issues like I was trying to do.

PR #1983, opened 2017-04-09
Adds "uniques heap" for VesselTarget instances.  Fixes #1980 and #1982

### LINKED ISSUE #1980: Very long freeze when vessels dock by script.
## To re-create the problem

1. You don't need any other mods.  You can start from an installation of KSP 1.2.2 that only has the kOS current ``develop`` branch installed on it.
2. At the bottom of this issue there's an attached saved game ZIP file.  Extract it into your saved games folder, and then launch KSP and you should have a saved game called "VesselMergeBug - BeforeDock".
3. Open that saved game, go to the tracking center, and fly the ship called ``client_craft``.
4. Turn off the SAS on ``client_craft`` (I accidentally left it on in the saved game and it will mess with the steering if you leave it on).
5. Open both ships' terminals.  (``client craft`` is near another vessel called ``server_craft`` that is a bit south of it.  They should be within about 90 meters of each other so you should be able to have both windows open at the same time.)
6. (Note, given where the saved game is starting from, you may have to time warp a very small amount to get signal to reach the vessels before you do the next step - I saved the game at a point just before they're about to pass over a ground station, to give the largest possible time window to talk to them and fiddle about.).
7. On both the ``server_craft`` and the ``client_craft`` terminals, hit **control-C** to stop the ``just_obey_nodes`` script (which is the boot file so it will auto-run when you load the scene) and dump you back to the interpreter for both.
8. On the ``server_craft`` terminal, do ``run station_dock_server.``
9. On the ``client_craft`` terminal, do ``run station_dock_client.``
10. Wait for them to finish docking to each other, and pay attention to what happens at the moment of vessel merging.

## What happens

When the vessels merge together, the game freezes for about 6 or 7 seconds.  It's longer than the normal amount of time the stock game takes to merge vessels together.  (I've seen it take a VERY long time - 30 to 60 seconds or so, in some more complex instances before I narrowed down the example to a smaller scenario in a newly started game.)

(( EDIT: Several paragraphs of text that used to appear here in the issue have been moved to a new issue, issue #1982 ))

## It only happens if the script is in control of the docking.

If you don't use the scripts, and instead fly the vessels under manual control to do the docking, the bug does not happen.  The two vessels merge together nicely without the long freeze.

## It is something that changed recently

If you perform the exact same steps above, but this time use the publicly released kOS 1.0.3 ZIP file from CKAN instead of using the current `develop` branch, the bug does not occur.  The ships merge together quickly (under either manual or script control), and you only get the 1 (expected) error message from the fact that two kOS cores are both trying to control the lock steering at the same time.

(Saved game ZIP is here:)
[JustBeforeDockingBug.zip](https://github.com/KSP-KOS/KOS/files/904789/JustBeforeDockingBug.zip)

Dunbaratu:
Fixes #1980 and #1982.

Problem #1982 was simpler so I'll explain it first:

When the SteeringManager complains about the CPU controlling
an already-controlled craft, it throws an exception which
makes the CPU try to abort execution.  But as part of trying to
abort that execution it calls PopContext(), which in turn
calls the SteeringManager to ask it to unbind the steering,
where the SteeringManager again throws an exception because
of the dual CPU control.  This second exception makes the CPU
abort partway through finishing the work of breaking the
execution, so it never quite fully terminates the program, which
is why it keeps trying to run after some of the running context
information has been wiped away (thus the null refs in
ExecuteInstruction).

The fix:  The CPU now advertises when it's in a PopContext,
so the SteeringManager can choose not to throw that second
exception in that case.

Problem #1980 required a slight refactor:

This problem was caused by the fact that VesselTarget now
inserts callback hooks into GameEvents, which it didn't do
before, and those callback hooks prevent it from getting
garbage collected.   As long as GameEvents has a
reference to a delegate that needs the VesselTarget object
to keep existing as part of its closure, the VesselTarget
can't ever get orphaned.  It was depending on being a
KOSScopeObserver to discover when kOS orphans it, so it
can remove its hooks from GameEvents, but KOSScopeObserver
can't really reliably catch all cases like that, and relies
on a C# destructor ( ~Variable ) to call ScopeLost().  But
the C# destructor never happens if the thing never orphans,
so it's a catch-22.  The orphaning event that tells
VesselTarget to unhook itself won't happen because VesselTarget
hasn't unhooked itself.

Thus the freeze at docking was caused by the fact that
every time the docking script had referred to the target vessel
during its main loop, it was creating a new instance of the
VesselTarget wrapper around it, and the old instances
never truly orphaned away like they should.  By the time
the docking happened there were thousands of thse extra
instances in memory, and upon docking ALL their callback
hooks got fired off at once.

The Fix:

Since VesselTarget is just a wrapper around a KSP Vessel,
and all the information contained therein is meant to
just inherit the info from the Vessel, and all the
Setter Suffixes are meant to pass the change on to the
KSP Vessel itself, there is no practical difference between
a kerboscript having two different VesselTarget instances
wrapping the same Vessel versus having just one VesselTarget
instance for it and then just having two references to it.

Threfore we track whether or not a VesselTarget wrapper
around a Vessel already exists, and if it does then we just
hand out a reference to the existing one instead of constructing
a new one.  That prevents us from having thousands of copies
of the same VesselTarget in memory when they don't orphan.

It should also have the nice side effect of reducing
execution burden by not re-running the Parts query of
the vessel each time it's referrenced in the script, as
it had to do before when each mention of TARGET made a
brand new VesselTarget.

The reason so many files got touched is because all instances
of calling the VesselTarget constructor had to be edited
to use a factory method instead, that does what the
above paragraphs describe.

### DIFF
--- a/src/kOS/Sound/SawtoothSoundWave.cs
+++ b/src/kOS/Sound/SawtoothSoundWave.cs
@@ -14,9 +14,6 @@ namespace kOS.Sound
                 
         public override float SampleFunction(float t)
         {
-            // if (t >= 0.5f)   //eraseme
-            //    t -= 0.5f;    //eraseme
-            // return 2*t - 1;  //eraseme
             return 1 - 2*t;
         }
 
--- a/src/kOS.Safe.Test/Opcode/FakeCpu.cs
+++ b/src/kOS.Safe.Test/Opcode/FakeCpu.cs
@@ -7,6 +7,7 @@ namespace kOS.Safe.Test.Opcode
     public class FakeCpu : ICpu
     {
         private readonly Stack<object> fakeStack;
+        public bool IsPoppingContext { get { return false; } }
 
         public FakeCpu()
         {
--- a/src/kOS.Safe/Execution/CPU.cs
+++ b/src/kOS.Safe/Execution/CPU.cs
@@ -61,6 +61,14 @@ namespace kOS.Safe.Execution
         
         public List<string> ProfileResult { get; private set; }
 
+        /// <summary>
+        /// It's quite bad to abort the PopContext activity partway through while the CPU is
+        /// trying to clean up from a program crash or break, so this advertises when that's the case
+        /// so other parts of the system can decide not to use exceptions when in this fragile state:
+        /// </summary>
+        /// <value><c>true</c> if this CPU is popping context; otherwise, <c>false</c>.</value>
+        public bool IsPoppingContext { get; private set; }
+
         public CPU(SafeSharedObjects shared)
         {
             this.shared = shared;
@@ -186,6 +194,7 @@ namespace kOS.Safe.Execution
         private void PopContext()
         {
             SafeHouse.Logger.Log("Popping context " + contexts.Count);
+            IsPoppingContext = true;
             if (contexts.Any())
             {
                 // remove the last context
@@ -211,6 +220,7 @@ namespace kOS.Safe.Execution
                     shared.Interpreter.SetInputLock(false);
                 }
             }
+            IsPoppingContext = false;
         }
 
         /// <summary>
--- a/src/kOS.Safe/Execution/ICpu.cs
+++ b/src/kOS.Safe/Execution/ICpu.cs
@@ -59,5 +59,7 @@ namespace kOS.Safe.Execution
 
         List<string> GetCodeFragment(int contextLines);
         void RunProgram(List<Opcode> program);
+
+        bool IsPoppingContext { get; }
     }
 }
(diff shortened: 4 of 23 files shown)


src/kOS/Sound/SawtoothSoundWave.cs
```diff
@@ -14,9 +14,6 @@ public SawtoothSoundWave(): base()
                 
         public override float SampleFunction(float t)
         {
-            // if (t >= 0.5f)   //eraseme
-            //    t -= 0.5f;    //eraseme
-            // return 2*t - 1;  //eraseme
```
Dunbaratu: These aren't related to this PR, but I noticed them when I did a search for all my "// eraseme" comments to remove them at the end of my testing.  They should have been removed a while ago.

PR #1993, opened 2017-04-21
Fixes #1985   stock :deploy Reflection removed

### LINKED ISSUE #1985: Surface sample from a command seat is available before any facility upgrades.
In career mode facilities are required to be upgraded in order to collect surface samples, which is not respected by kOS.

### How to reproduce:

In a craft with Command Seat manned by a kerbal:

    set kerbal to ship:partsnamed("kerbalEVA")[0].
    print kerbal:allmodules.
        LIST of 15 items:
        [0] = "KerbalEVA"
        [1] = "ModuleScienceExperiment" // <- these are of interest
        [2] = "ModuleScienceExperiment" // <- these are of interest
        [3] = "ModuleScienceContainer"
        ...
    print kerbal:getmodulebyindex(1). // <- checking first
        ModuleScienceExperiment, containing:
        LIST of 2 items:
        [0] = "(callable) deploy, is KSPAction"
        [1] = "(callable) discard sample, is KSPAction" // <- surface sample
    print kerbal:getmodulebyindex(2). // <- checking second
        ModuleScienceExperiment, containing:
        LIST of 3 items:
        [0] = "(callable) eva report, is KSPEvent" // <- not surface sample
        [1] = "(callable) eva report, is KSPAction"
        [2] = "(callable) discard report, is KSPAction"
    
    set sample to kerbal:getmodulebyindex(1). // <- surface sample
    sample:deploy.
    print sample:data[0]:dataamount.
        21 // <- oops!

Dunbaratu:
Reflection was being used to bypass access
permissions in a naughty way.  (We were calling
a non-public method).  By not using the public
method instead, the code bypassed some of the
checks that KSP normally performs for seeing if
an experiment is allowed.

I added some better checks to avoid calling it
in cases where we shouldn't, but also replaced
that use of reflection with a call to the proper
public-facing method as well so that if we don't
catch everything, at least we're not bypassing
KSP's checking.  (The reason I didn't just rely
on KSP's checking alone is that it sometimes fails
silently when it doesn't like what you're doing,
and I wanted to let the user see a message.)

The same sort of naughty private-access
reflection is used in DMScienceExperiment too,
but there I didn't change it because that
had to use reflection to avoid the code blowing
up when DMagic science is missing.  Also, I didn't
write that code and don't use DMagic science so it
would have been hard to test it if I tried to
fix that part.

### DIFF
--- a/src/kOS/Suffixed/PartModuleField/StockScienceExperimentFields.cs
+++ b/src/kOS/Suffixed/PartModuleField/StockScienceExperimentFields.cs
@@ -42,19 +42,15 @@ namespace kOS.Suffixed.PartModuleField
         protected virtual void Deploy()
         {
             ThrowIfNotCPUVessel();
-
-            var gatherDataMethod = module.GetType().GetMethod("gatherData",
-                BindingFlags.NonPublic | BindingFlags.Instance);
-
-            object result = gatherDataMethod.Invoke(module, new object[] { false });
-
-            IEnumerator e = result as IEnumerator;
-
-            module.StartCoroutine(e);
+            module.DeployExperiment();
         }
             
         public override bool Inoperable()
         {
+            BaseEvent deployAction = module.Events["DeployExperiment"];
+            if (deployAction == null || ! deployAction.active)
+                return true;
+
             return module.Inoperable;
         }
 


hvacengi:
Moving to the v1.1.1 milestone because it looks like this change is now showing dialog boxes to confirm the experiments, rather than confirming the experiments without user mouse interaction.

Dunbaratu:
(To record a bit more:  The problem is that the only **public** method I could see that SQUAD provided in their API to do this is tied to the UI strongly and pops up the dialog box. Any methods that bypass the dialog box all seem to be private and also bypass some of the safety checks (which is what caused the issue this was meant to fix.)

The tricky thing is that I don't know how we can satisfy all the following at the same time:
1 - Bypass the UI.
2 - Do NOT bypass the safety checks Squad embedded into the UI call
3 - Avoid using the dirty trickery of accessing private members via reflection.

As near as I can tell, we cannot do all 3 unless I missed something.  So for now, we punt it ahead.

Dunbaratu:
I'd like to consider removing this from all milestones entirely (not just "postpone it", but "mark as wontfix").

The problems I mention above (the 3 requirements that can't be simultaneously satisfied) seem to me to be impossible to fix and we'll just have to live with it.

Dunbaratu:
Milestone cleared out because this issue is a wont-fix.

PR #1994, opened 2017-04-21
Fixes #1987 Fixes #1978 : multinational compiler

### LINKED ISSUE #1978: Support underscores in numbers
When writing big numbers (which, because KSP counts distance in meters, happens quite often), putting in underscores to group digits would greatly increase readability. This would be a great alternative to scientific notation, as I find numbers with less than 10 digits harder to read with scientific notation compared to the full number.

The rules for a valid number would be:

- Whatever consists of a valid number today
- An underscore must be surrounded by digits
- There are no limits to the number of digits between underscores. This avoids cultural issues (The [Indian numbering system](https://en.wikipedia.org/wiki/Indian_numbering_system) groups digits differently that the western system)

This would not change the meaning of the number, only how a human or kerbal would read the number.

Both Swift and C# 7 have the same feature.

Some examples:

| Name                | Raw      | Scientific | With underscores (western) | With underscores (indian) |
| ------------------- | -------- | ---------- | -------------------------- | ------------------------- |
| Kerbin Atmos        |    70000 |        7e4 |                     70_000 |                    70_000 |
| Kerbin spacestation |   150000 |       15e4 |                    150_000 |                  1_50_000 |
| Mun orbit           | 12000000 |       12e6 |                 12_000_000 |               1_20_00_000 |

### LINKED ISSUE #1987: Alter TinyPG file's regex'es to allow multilingual identifiers.
The current TinyPG grammar file specifies what counts as a letter in an identifier by using ASCII ranges such as:

``[A-Z][a-z]``

But that doesn't allow for identifiers in non-English languages.  What if a user wants to start a variable name with a ``Ü`` or a ``é`` or a ``Я``?  Also it would be nice if we could make some aliases like being able to type ``π`` for ``constants:pi``  It might make code more readable with all the mathematics going on in kerboscript.

Now that we can support any Unicode font the user has installed on their machine, we should also modify the language grammar to allow any language's letters to count as legal identifier characters.

Dunbaratu:
You can see the example script provided to see what has changed.

### DIFF
--- a/doc/source/language/syntax.rst
+++ b/doc/source/language/syntax.rst
@@ -52,9 +52,24 @@ operator symbols:
 .. highlight:: none
 
 **Identifiers**: Identifiers consist of: a string of (letter, digit, or
-underscore). The first character must be a letter. The rest may be letters, digits or underscores. **Identifiers are case-insensitive**. The following are identical identifiers::
+underscore). The first character must be a letter or an underscore.
+The rest may be letters, digits or underscores.
+**Identifiers are case-insensitive**. The following are identical identifiers::
 
     My_Variable  my_varible  MY_VARAIBLE
+
+.. note::
+  .. versionadded:: 1.1.0
+    Kerboscript accepts Unicode source code, encoded using the UTF-8
+    encoding method.  Because of this, the definition of a "letter"
+    character for an identifier includes letters from many languages'
+    alphabets, including accented Latin alphabet characters, Cyrllic
+    characters, etc.  Not all languages have been tested but in
+    principle they should work as long as they have a Unicode standard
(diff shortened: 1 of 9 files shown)


Dunbaratu:
Added not-ready flag because I realized we also need to edit the way the terminal reads user typing input.  Right now all characters other than printable ASCII are being interpreted as special chars (i.e. control codes).  This means any Unicode value > 128 will get rejected when a user tries to type it.

I'm not sure if we also want to implement the ALT-codes technique (i.e. hold ALT while typing, say 313, makes Unicode 313 appear).  Right now that's not implemented.  While a typical Windows standard, it doesn't happen at a "low level" enough layer of the system to come right from the keyboard driver.  Applications have to implement it to get it.

PR #2000, opened 2017-05-06
Fixes #1970 and adds altitudesealevel.

### LINKED ISSUE #1970: body:atm:sealevelpressure docs don't match reality
The documentation claims that a ``body:atm:sealevelpressure`` is measured in ATM's.

I just did this test (in Galileo Planet Pack where Gael is supposed to be just like Kerbin) and got the answer that ``gael:atm:sealevelpressure`` is 101.

If it's 101 then clearly the units aren't ATM's.

I'm marking it as both ``bug`` and ``documentation`` for now since I don't know which is wrong - are the docs wrong and should change or is the code wrong and it should change.

Dunbaratu:
In addition to changing the sealevelpressure to ATM's, I also noticed the pressure at a given altitude was missing from the atmosphere structure so I added that too.

### DIFF
--- a/doc/source/structures/celestial_bodies/atmosphere.rst
+++ b/doc/source/structures/celestial_bodies/atmosphere.rst
@@ -68,8 +68,27 @@ A Structure closely tied to :struct:`Body` A variable of type :struct:`Atmospher
     :type: :ref:`scalar <scalar>` (atm)
     :access: Get only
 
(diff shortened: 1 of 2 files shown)


Dunbaratu:
``Breaking`` label added because If anyone was using ``sealevelpressure`` beforehand, it just got changed from KPa to ATMs in order to match the documentation and the rest of kOS.

PR #2001, opened 2017-05-06
Fixes #1947 adding geoposition suffix to body.

### LINKED ISSUE #1947: Can't make a LATLNG of a different body
Under the hood in the C# code, a LATLNG is a ``GeoCoordinates`` class, which allows you to have a position on any body.  But in Kerboscript you can only make a LATLNG of the current ``ship:body``.  I had a look and there's no suffix of ``BodyTarget`` that takes as input a latitude and longitude to give a ``GeoCoordinates`` either.  There's just no way to do it in kerboscript.

That should be fixed.

Dunbaratu:
In addition to changing the sealevelpressure to ATM's, I also noticed the pressure at a given altitude was missing from the atmosphere structure so I added that too.

### DIFF
--- a/doc/source/math/geocoordinates.rst
+++ b/doc/source/math/geocoordinates.rst
@@ -18,7 +18,17 @@ Creation
     :parameter lng: (deg) Longitude
     :return: :struct:`GeoCoordinates`
 
-    This function creates a :struct:`GeoCoordinates` object with the given latitude and longitude. Once created it can't be changed. The :attr:`GeoCoordinates:LAT` and :attr:`GeoCoordinates:LNG` suffixes are get-only and cannot be set. To switch to a new location, make a new call to :func:`LATLNG()`.
+    This function creates a :struct:`GeoCoordinates` object with the given
+    latitude and longitude, assuming the current SHIP's Body is the body
+    to make it for.
+    
+    Once created it can't be changed. The :attr:`GeoCoordinates:LAT` and
+    :attr:`GeoCoordinates:LNG` suffixes are get-only (they cannot be
+    set.) To switch to a new location, make a new call to :func:`LATLNG()`.
+
+    If you wish to create a :struct:`GeoCoordinates` object for a latitude
+    and longitude around a *different* body than the ship's current sphere
+    of influence body, see :method:`Body:Geoposition` for a means to do that.
 
     It is also possible to obtain a :struct:`GeoCoordinates` from some suffixes of some other structures. For example::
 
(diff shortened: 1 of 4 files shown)


src/kOS/Suffixed/BodyTarget.cs
```diff
@@ -122,6 +122,10 @@ private void BodyInitializeSuffixes()
                       new OneArgsSuffix<ScalarValue, Vector>(
                               AltitudeFromPosition,
                               "Interpret the vector given as a 3D position, and return its altitude above 'sea level' of this body."));
+            AddSuffix("GEOPOSITION",
```
hvacengi: This suffix is hiding the suffix inherited from `orbitable`, with a very different behavior.  I think we should rename this new suffix to better represent that it is a different behavior from other suffix by the same name.  Maybe `childgeoposiiton` or `newgeoposition` or something like that.

Dunbaratu:
@hvacengi, I added another commit to the fork that should address the issues you brought up.

doc/source/math/geocoordinates.rst
```diff
@@ -18,7 +18,17 @@ Creation
...
+    latitude and longitude, assuming the current SHIP's Body is the body
+    to make it for.
+    
+    Once created it can't be changed. The :attr:`GeoCoordinates:LAT` and
+    :attr:`GeoCoordinates:LNG` suffixes are get-only (they cannot be
+    set.) To switch to a new location, make a new call to :func:`LATLNG()`.
+
+    If you wish to create a :struct:`GeoCoordinates` object for a latitude
+    and longitude around a *different* body than the ship's current sphere
+    of influence body, see :method:`Body:Geoposition` for a means to do that.
```
hvacengi: Should be `:method:'Body:GEOPOSITIONLATLNG'` I'll fix it in merging

doc/source/structures/celestial_bodies/body.rst
```diff
@@ -73,7 +73,8 @@ All of the main celestial bodies in the game are reserved variable names. The fo
     :attr:`MU`                       :ref:`scalar <scalar>` (:math:`m^3 s^{−2}`)
     :attr:`ATM`                      :struct:`Atmosphere`
     :attr:`ANGULARVEL`               :struct:`Vector` in :ref:`SHIP-RAW <ship-raw>`
-    :attr:`GEOPOSITIONOF`            :struct:`GeoCoordinates` in :ref:`SHIP-RAW <ship-raw>`
+    :method:`GEOPOSITIONOF`          :struct:`GeoCoordinates` given :ref:`SHIP-RAW <ship-raw>` position vector
+    :method:`GEOPOSITION`            :struct:`GeoCoordinates` given latitude and longitude values
```
hvacengi: Should be `:method:'Body:GEOPOSITIONLATLNG'` I'll fix it in merging

PR #2008, opened 2017-05-07
Fixes #1956 and changes font search order.

### LINKED ISSUE #1956: Reminder to self: URL to Commodore 64 font when new terminal fonts released.
The Commodore 64 - like font that the old ``font_sml.png`` file was built from has licensing terms that make it free to download but disallow any redistribution (meaning we can't package the TTF font file in our ZIP.)

When the terminal font change gets released, there should be some kind of message people see on migrating to it that gives out the URL where they can get that font from and install it on their computer, so they can get the old look back if they like.

Dunbaratu:
Fixes #1956 

Note, because the only places I could find to download the C64 fonts from were sites with possibly invasive banner ads I didn't feel comfortable linking to in a URL, I decided to just mention the existence of the fonts out there, and let the user search for them, instead of including a direct URL.

---

When making the font dialog, I discovered
that the font search list didn't actually
work properly when the first font in the
list was a bogus font name like "_not_picked_".

Instead of trying the next font in the list, it was
just rendering everything in Arial.

There are some code changes to fix that problem too
in here.

### DIFF
--- a/src/kOS/Screen/TermWindow.cs
+++ b/src/kOS/Screen/TermWindow.cs
@@ -38,9 +38,6 @@ namespace kOS.Screen
         
         private bool collapseFastBeepsToOneBeep = false; // This is a setting we might want to fiddle with depending on opinion.
 
-        private KeyBinding rememberThrottleCutoffKey;
-        private KeyBinding rememberThrottleFullKey;
-
         private bool allTexturesFound = true;
         private float cursorBlinkTime;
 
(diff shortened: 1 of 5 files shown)


src/kOS/Screen/TermWindow.cs
```diff
@@ -38,23 +38,23 @@ public class TermWindow : KOSManagedWindow , ITermWindow
         
         private bool collapseFastBeepsToOneBeep = false; // This is a setting we might want to fiddle with depending on opinion.
 
-        private KeyBinding rememberThrottleCutoffKey;
-        private KeyBinding rememberThrottleFullKey;
-
```
Dunbaratu: These two lines were unused fields causing a warning message.  (They stopped being used once we stopped needing to have an ugly ad-hoc fix for locking out the 'Z' and 'X' keys.)

Deleting them was not related to this PR's purpose, but I removed them since I was in this file anyway and I noticed the warnings.

PR #2010, opened 2017-05-10
Version 1.1.0 closeout fixes

### LINKED ISSUE #1783: Part Highlight not visible after part tree alterations
[ksp v1.1.3] [kos v1.0]
Tested with KerbalX (with KAL9000)

sorry for the long test script but because of the visual nature I went for user feedback at each step.

```
clearscreen.
set ag10 to false.
set foo to ship:rootpart.
set bar to highlight( foo, blue ).
print "The pod is highlighted blue. " + bar:enabled. 
print "press 0 for next step".

wait until ag10.
set ag10 to false.
sas on.
lock throttle to 1.
stage.
print "Highlighting still enabled. " + bar:enabled.
print "press 0 for next step.".

wait until ag10.
set ag10 to false.
                              //This is where it starts to go wrong
stage.
print "Highlight:enabled is now reporting " +bar:enabled.

//if this reported false i would assume it was a design choice to prevent having hard to remove
//highlights on detached parts.
//The highlight however has disappeared for me (will test different graphics settings in a bit)

print "Press 0 for next step.".
wait until ag10.
set ag10 to false.

set bar:enabled to true.
print "Highlight is now back. " +bar:enabled.
print "press 0 to end.".

wait until ag10.
set ag10 to false.
```

For me the highlight is disappearing as soon as the launch clamps are detached. It also happens with any other staging that fires decouplers. I will install bdArmory later and see if shooting parts off the vessel causes the same issue.
Will fiddle found with the graphics setting a bit too to see if that changes anything.

### LINKED ISSUE #1911: Ship cannot communicate with itself with CommnetConnectivityManager
kOS 1.0.3, KSP 1.2.2, ship has a cpu core and non-relay antenna, on the runway at KSC:

```
print ship:connection:isConnected.
False
```

### LINKED ISSUE #1977: SharedObjects.DestroyObjects throws exception that interrupts cleanup work.
```
ArgumentException: Set Method not found for 'DispatchManager'
  at System.Reflection.MonoProperty.SetValue (System.Object obj, System.Object value, BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] index, System.Globalization.CultureInfo culture) [0x00000] in <filename unknown>:0 
  at System.Reflection.PropertyInfo.SetValue (System.Object obj, System.Object value, System.Object[] index) [0x00000] in <filename unknown>:0 
  at kOS.SharedObjects.DestroyObjects () [0x00000] in <filename unknown>:0 
  at kOS.Module.kOSProcessor.OnDestroy () [0x00000] in <filename unknown>:0 
```

This seems to be happening because `DestroyObjects` wants to iterate over every Property inside itself (using Reflection) and set it to null.  But some of those properties are not settable in this context because they have `private set;` and the iterator that's going over them only gets the public members of them.  At least I think that's why it happens.

I get this message whenever a kOS part blows up in the game.

hvacengi:
UpdateHandler.cs
* Add a new method ClearAllObservers which will clear all observers and
Dispose of any observer that implements IDisposable

SharedObjects.cs
* Call the new ClearAllObservers method in SharedObjects.DestroyObjects
* Check to see if there is a Set method before setting a value to null

HighlightStructure.cs
* Add event subscription to monitor ship status
* Unsubscribe from events inside of Dispose

CommNetConnectivityManager.cs
* Update HasConnection to return true no matter what if checking
connection to the same cpu vessel
* Also fix a comment typo

Fixes #1977
Fixes #1911
Fixes #1783

### DIFF
--- a/src/kOS/Communication/CommNetConnectivityManager.cs
+++ b/src/kOS/Communication/CommNetConnectivityManager.cs
@@ -105,7 +105,7 @@ namespace kOS.Communication
                 return true;
 
             // We next need to query the network to find a connection between the two vessels.
-            // I found no exposed method for accessing chached paths directly, other than the
+            // I found no exposed method for accessing cached paths directly, other than the
             // control path.
 
             // WARNING: In stock this will only work for vessels with a relay antenna installed.
@@ -113,7 +113,7 @@ namespace kOS.Communication
             // there isn't a very good way around it.
             var net = CommNetNetwork.Instance.CommNet;
             tempPath = new CommPath();
-            return net.FindPath(vessel1.Connection.Comm, tempPath, vessel2.Connection.Comm) || net.FindPath(vessel2.Connection.Comm, tempPath, vessel1.Connection.Comm);
+            return vessel1 == vessel2 || net.FindPath(vessel1.Connection.Comm, tempPath, vessel2.Connection.Comm) || net.FindPath(vessel2.Connection.Comm, tempPath, vessel1.Connection.Comm);
         }
 
         public void AddAutopilotHook(Vessel vessel, FlightInputCallback hook)
--- a/src/kOS/Communication/CommNetConnectivityManager.cs
+++ b/src/kOS/Communication/CommNetConnectivityManager.cs
@@ -113,7 +113,7 @@ namespace kOS.Communication
             // there isn't a very good way around it.
             var net = CommNetNetwork.Instance.CommNet;
             tempPath = new CommPath();
-            return vessel1 == vessel2 || net.FindPath(vessel1.Connection.Comm, tempPath, vessel2.Connection.Comm) || net.FindPath(vessel2.Connection.Comm, tempPath, vessel1.Connection.Comm);
+            return vessel1.id == vessel2.id || net.FindPath(vessel1.Connection.Comm, tempPath, vessel2.Connection.Comm) || net.FindPath(vessel2.Connection.Comm, tempPath, vessel1.Connection.Comm);
         }
 
         public void AddAutopilotHook(Vessel vessel, FlightInputCallback hook)
--- a/src/kOS.Safe/UpdateHandler.cs
+++ b/src/kOS.Safe/UpdateHandler.cs
@@ -61,7 +61,31 @@ namespace kOS.Safe
                 observer.KOSFixedUpdate(deltaTime);
             }
         }
-        
+
+        public void ClearAllObservers()
+        {
+            var snapshot = new HashSet<IUpdateObserver>(observers);
+            foreach (var observer in snapshot)
+            {
+                IDisposable disp = observer as IDisposable;
+                if (disp != null)
+                {
+                    disp.Dispose();
+                }
+            }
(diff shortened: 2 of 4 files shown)


src/kOS/Communication/CommNetConnectivityManager.cs
```diff
@@ -105,15 +105,15 @@ public bool HasConnection(Vessel vessel1, Vessel vessel2)
...
+            // I found no exposed method for accessing cached paths directly, other than the
             // control path.
 
             // WARNING: In stock this will only work for vessels with a relay antenna installed.
             // This is a limitation put in place to improve performance in the stock game, and
             // there isn't a very good way around it.
             var net = CommNetNetwork.Instance.CommNet;
             tempPath = new CommPath();
-            return net.FindPath(vessel1.Connection.Comm, tempPath, vessel2.Connection.Comm) || net.FindPath(vessel2.Connection.Comm, tempPath, vessel1.Connection.Comm);
+            return vessel1 == vessel2 || net.FindPath(vessel1.Connection.Comm, tempPath, vessel2.Connection.Comm) || net.FindPath(vessel2.Connection.Comm, tempPath, vessel1.Connection.Comm);
```
Dunbaratu: Should we be using  a raw reference-equals check here?  We seem to be using ``vessel.id`` for this sort of check in a lot of places elsewhere in the code.

Granted, it would be really weird for there to be two instances of ``Vessel`` that are really the same vessel, so it probably doesn't matter, but I wouldn't put it past KSP to have something that weird happening.

hvacengi: That's a fair point.  The reason we use the vessel id comparison is that the default equality comparison walks the fields on the object to check equality, which compares ID, and Name, and all the underlying unity fields.  Which means that comparing a single GUID should be faster than comparing all of those fields.

Dunbaratu: I thought that ``obj1 == obj2`` just gives a reference equals, regardless of the equality overrides of the object type, and you have to use ``obj1.Equals(obj2)`` to get it to actually execute the equality comparison method.  (Otherwise how would you tell the compiler that you *are* trying to do a reference comparison, if ``==`` didn't always do it.)

Dunbaratu: Ah okay, another place where C#'s features are ones I do NOT consider a good idea.  The caller has to dig down deep into the implementation of a class to discover if  ``==`` is a reference equals or not.  It is, like 98% of the time, but to be sure about that other 2%, you have to explicitly call a method (``ReferenceEquals``) to do the more low-level thing of reference comparison.  That's massively un-intuitive (the thing that looks low level (``==``) is higher level than the thing that looks more high level (calling a method named something).

Dunbaratu:
To be honest, I didn't spend as long as I usually do on testing this one.  All the code changes were simple and straightforward.  I know I won't be around 'til late tonight and I didn't want to hold up a potential pre-release waiting on me.  I just did a very quick one-liner check on each of the issues the PR mentioned and verified they didn't happen anymore.  I didn't really try to think of any strange edge-cases or anything like that.

PR #2014, opened 2017-05-11
changelog for pre-release of 1.1.0, called 1.0.90

Dunbaratu:
Contains the edit to changelog in prep for pre-release.  The edit to changes.rst is not there yet because I will wait for full release to do that one and get all the links and URLs right.


Dunbaratu:
(Note on confusing branch name.  The branch is called v1.1.0_changelog because I named it before I decided to reduce the version number to 1.0.90.)

Issue #2030, opened 2017-05-19
Something makes Variables and ProgramContexts fail to orphan (memory leak)


Dunbaratu:
We haven't narrowed down the cause of it yet, but we detected that something is preventing variables and program contexts from entirely going away when a program ends.  When you break the execution either from normal termination or from ctrl-C or from a crash with error, it should clean up all the old program data and variables in memory.

According to the memory footprint, and a debug instance counter check in Variable and ProgramContext, it is not doing this.  We are not sure yet the cause of it, but this needs to be discovered and fixed.  It's a big problem if people are using scripts that use a lot of temporary data.  That data doesn't go away.  You don't notice if your scripts are small because with something the size of KSP, leaking about 30 or so Kilobytes every time you re-run a program won't be important.  But when your programs do use a lot of memory, then you start to notice the problem.

Dunbaratu:
Not easily fixable quickly, and it isn't something newly caused, so I'm moving it out of v1.1.0.

PR #2031, opened 2017-05-19
trigger closures and delegate memory leaks

### LINKED ISSUE #2028: UserDelegate callback hooks prevent orphaning program contexts (memory leak)
Let's say you have a ``UserDelegate`` that you use as a callback hook, any callback hook.

(i.e. You do something like this: ``set my_vecdraw:vecupdater to my_vec_update_func@.``.)

When the program that contained ``my_vec_update_func`` quits normally or crashes prematurely, the existence of that ``UserDelegate`` reference inside the ``VectorRenderer`` doesn't go away...
...and the ``UserDelegate`` has a reference to the ``ProgramContext`` that created it...
... which means the entire ``ProgramContext`` of that program that ended fails to orphan and get GC'ed.

We do have a protection against ever calling that ``UserDelegate`` ever again when its program is gone, because ``CPU.AddTrigger()`` will detect if the program context of the delegate differs from the current one, and refuse to schedule an invocation of the callback then.  But, this doesn't get rid of the problem that the reference is still there, still holding the ``ProgramContext`` around (and thus the whole program itself never goes away from memory entirely.)

Suggested solution:  Make ``UserDelegate.ProgramContext`` a weak reference.  It's only there to be able to test whether or not the delegate matches the current context.  It shouldn't prevent the removal of that program context.

### LINKED ISSUE #2029: Callback Delegates' closures don't seem to work
Here is an example of a callback (``vecdraw:vecupdater``) that is meant to call a script function.  It fails, generating the error message ``Undefined Variable Name 'vdVec'.``

```
function do_test
{
  local vdVec is v(1,1,1).
  set vd to vecdraw(v(0,0,0),v(1,1,1),white,"test",1,true).
  set vd:vecupdater to {
    // note use of local var vdVec on the next line, which is
    // within scope because it's local to function do_test:
    return vdVec * sin(time:seconds*180/3.14)*4.
  }.
}
do_test().

// Must have this wait to show the problem.
// The program has to be running long enough for
// a physics update to occur or else vd:vecupdater won't
// get called.
wait 1.
```
(Note the same problem happens whether an anonymous function is used (as shown here) or a named function that is nested inside of function do_test).

Dunbaratu:
Fixes #2029 
Fixes #2028

See the individual commits' logs within the branch for more messages describing what was done.

Note, the system invented to fix issue #2028 (See Cpu.AddPopContextNotifyees()) might be good for other parts of the system to use to solve #2030 (possibly).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,11 @@ and things are subject to change.)
 * **Choose the IP address** of the telnet server, from the ones your computer has available, instead
   of kOS picking one arbitrarily.
   [pull request](https://github.com/KSP-KOS/KOS/pull/1976)
+* **Allow local variables in triggers** In order to support the kOS callback
+  system used by the GUI, we also finally had to add support for proper local
+  variable scoping to triggers like WHEN and ON.  A trigger's condition
+  variables are no longer limited to having to be global.
+  [pull request](TODO URL HERE BEFORE RELEASE)
 * **Pressure at a given altitude** is now something you can query from an atmosphere.
   [pull request](https://github.com/KSP-KOS/KOS/pull/2000), documentation: search for "ALTITUDEPRESSURE" (TODO: replace with URL).
 * **Get a LATLNG for some other body than the current one.**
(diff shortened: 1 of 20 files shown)


Dunbaratu:
I won't mark this as "had to skip peer review" because I was basically merging in @hvacengi's edits that he made during his peer review of my PR, so I figure that counts.

I only backed out the final commit that added steering changes and wasn't ready yet.

PR #2043, opened 2017-06-04
Fixes #2035 gui suffix descriptions

### LINKED ISSUE #2035: GUI docs need detail sections for suffixes in tables
(This has been mentioned in the Slack channel, but it needs to be a GitHub issue so we remember to do it.)

RIght now the GUI docs page consists of just a list of suffixes and one-liner descriptions in tables.
We're lacking these sections for the suffixes:

```
.. attribute:: StructName:SuffixName

    :type: :struct:`blarg`

    A few sentences of description.
```

(Some of the info that should go into these sections is explained in prose text, but it's not formatted right, and it makes it impossible to link to it from elsewhere.)

As part of this edit we should also decide if we want to split the GUI doc into separate pages (one per structure type, i.e. a page for Slider, a page for TextField, a page for Button, etc), or leave it all in one.

There's good arguments for either approach.  Keeping it in one page makes it a nice top-down read-through, and such a large amount of it is classes inherited from other classes on the same page, that it does conceptually all go together well.  Spreading it to multiple pages makes it take less time to load the page when you click on a link to just one thing.

Dunbaratu:
Fixes #2035

I also had to edit ksdomain.ks in order to get it to understand ``.. method::`` tags where the structure isn't mentioned in the name but is presumed by which ``.. structure::`` section it is inside of.  This was supported by ``.. attribute::`` but not by ``.. method::`` before.


doc/ksdomain.py
```diff
@@ -154,12 +154,19 @@ class KOSAttribute(KOSObject):
...
         if m.group('prefix') is None:
             if current_struct is not None:
                 struct = current_struct
                 fullname = current_struct + ':' + name
+            else:
+                raise Exception("Attribute name lacks a prefix and isn't " +
+                    "indented inside a structure section.  Problem " +
+                    "occurred at " + self.env.docname + ", line " +
+                     str(self.lineno) + 
+                     ".")
```
Dunbaratu: There is probably a better way to throw this message, but I didn't have the time to research how you're meant to do it, and this gave me the necessary info to find where the problems were in my ReST files.

doc/ksdomain.py
```diff
@@ -190,11 +197,19 @@ class KOSMethod(KOSObject):
     def handle_signature(self, sig, signode):
         m = ks_sig_re.match(sig)
         name = m.group('object')
+        struct = None  #might override further down
 
         current_struct = self.env.temp_data.get('ks:structure')
         if m.group('prefix') is None:
             if current_struct is not None:
+                struct = current_struct
```
Dunbaratu: This is the line that was missing when this got copied from KOSAttribute above.  It's relevant because without it when you make a tag like ``.. method:: methname``, where the ``structname:`` prefix is implied rather than explicitly stated, the presumed default prefix wasn't being added due to this missing line.

doc/source/general/gui.rst
```diff
@@ -1,4 +1,4 @@
-.. _gui:
+.. _gui_overview:
```
Dunbaratu: It took a lot of trial and error to figure this one out:  It gets very confused if you include a label tag of ``_foo`` inside your ``foo.rst`` file when you also have a folder alongside it called ``foo/`` with other files in it.

When you tell Sphinx to automatically build a toctree from a folder, it also makes that folder name into a label tag.  Having both the two tags caused it to double everything in the folder walk, and mention it twice in the table of contents on the left, and mention it at the wrong nesting level.  It took me a while to work out that I had to rename this tag to fix that.  (Sphinx never issued any complaint about it.  It was happy to have the same file in two different places of the tree, so I had a hard time figuring out the cause.)

Dunbaratu:
I marked this as "not ready" because when I render it in Sphinx, the ``.. method`` sections render wrongly, and I THINK what I have in the ReST files is right, but something in our domain setup is wrong.

What I wrote seems to match what sphinx examples say to do, but we aren't rendering it like that does:

What I have is this:
```
    .. method:: suffixname(arg1)

    :parameter arg1: desc of arg1
    :type arg1: type of arg1
```
Which is meant to render like so:

*suffixname(arg1)*

- Parameters:
    - arg1 (type of arg1) desc of arg1

But instead it's rendering like so:

*suffixname(arg1)*

- Parameters:
    - arg1 desc of arg1

- type arg1: type of arg1


(the type isn't being "moved up" into the parameter, and I don't know why).

Dunbaratu:
Instead of fixing the problem, which I don't know how to do right now, I just punted and went back to formatting it the old way so I can move forward.

Dunbaratu:
Making a new label "Had to skip peer review" to use on merges where I didn't consult with someone else first because there were no other devs with the free time to do it.  These are marked as such to flag them as perhaps less reliable.

PR #2046, opened 2017-06-12
While it may be true that stability assist mode

Dunbaratu:
works in some cases, in practice it's a pain to figure out
what those cases are, and it's better to have the docs
just say not to use it at all then for them to be misleading
like this.  While this may have worked once upon a time,
it doesn't seem to anymore.

### DIFF
--- a/doc/source/commands/flight.rst
+++ b/doc/source/commands/flight.rst
@@ -24,4 +24,11 @@ Unless otherwise stated, all controls that a **kOS** CPU attempts will be done o
 
 .. warning:: **SAS OVERRIDES kOS**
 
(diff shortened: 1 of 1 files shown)


doc/source/commands/flight.rst
```diff
@@ -24,4 +24,11 @@ Unless otherwise stated, all controls that a **kOS** CPU attempts will be done o
...
+    However, SAS will tend to fight and/or override **kOS**'s attempts to
+    steer.  In order for **kOS** to be able to turn the ship, you need to
+    set ``SAS OFF``. You should take care in your scripts to manage the
+    use of ``SAS`` appropriately. It is common for people writing
+    **kOS** scripts to explicitly start them with a use of the
+    ``SAS OFF` command just in case you forgot to turn it off before
```
hvacengi: It looks like you already caught that this was missing a back tick to close the literal.  I'm noting it to coincide with my removal of the "had to skip peer review" label.

PR #2052, opened 2017-06-15
Irrelevant whitespace change to force travis build.

Dunbaratu:
This commit of pure whitespace-only changes is just so I can push a "code change" and see if travis's build will work with the new KSP 1.3 stub DLLs.


pipakin:
It IS possible I generated the 1.3.0 dlls wrong. I used the tool that existed to do it, and it seemed to build correctly locally.

PR #2062, opened 2017-07-06
Fixes #2061 unicode mangle

### LINKED ISSUE #2061: Multinational unicode getting mangled on input and local drive.
I got this report from user Sebra on the KSP forum, who was trying to use KSP 1.3 + kOS 1.1.2 with Russian text, and I've narrowed it down to this so far:

These are two different problems with kOS's new unicode character support where it doesn't work:

(1) While you can print unicode characters on *output* to the terminal, you can't type them as *input* to the terminal from your keyboard.  For example, ``print char(1081).`` will print Cyrillic character "й" correctly, but if you have a Russian keyboard layout and try to *type* "й" (which is located where the 'q' key is on a US layout keyboard), you will end up with the plain ascii character "9" instead.  This same problem happens with all the other Cyrillic characters as well.  (I told Windows to install a Russian language driver to test this out, which lets you toggle a US keyboard between US and Russian mapping with a hotkey.)

Note, in this mapping from "й" to "9", it's useful to note that "й" is Unicode ``0x0439`` while "9" is Unicode ``0x0039``, which is *exactly* what you'd get if you cleared out the high byte and preserved only the low byte.  The same thing happens with all the other cases I tested.  Somewhere along the way the data must be passing through a step that presumes characters are only 1 byte.

(2) If you create a ``.ks`` file that contains Cyrillic characters, it gets saved just fine in the archive volume, but *not* in the local volume.  In the local volume you get all the non-ascii characters mangled and turned into question marks ("?").  This has something to do with how it's saved in ``persistent.sfs``, because it gets mangled as soon as you switch scenes or do anything that will reload the local volume from the persistent file.  Given when and where this happens, I suspect this has something to do the steps we take when we pass it through Base64 and compression.

Dunbaratu:
Fixes #2061 

Two places where unicode support for characters with codes > 127 didn't quite work right.  See in-line comments in the PR for details.

### DIFF
--- a/src/kOS/Persistence/PersistenceExtensions.cs
+++ b/src/kOS/Persistence/PersistenceExtensions.cs
@@ -128,7 +128,7 @@ namespace kOS.Persistence
             {
                 if (SafeHouse.Config.UseCompressedPersistence)
                 {
-                    node.AddValue("binary", EncodeBase64(content.String));
+                    node.AddValue("binary", PersistenceUtilities.EncodeBase64(content.Bytes));
                 }
                 else
                 {
@@ -151,9 +151,5 @@ namespace kOS.Persistence
             }
         }
 
-        private static string EncodeBase64(string input)
-        {
-            return PersistenceUtilities.EncodeBase64(Encoding.ASCII.GetBytes(input));
-        }
     }
 }
--- a/src/kOS/Screen/KOSToolbarWindow.cs
+++ b/src/kOS/Screen/KOSToolbarWindow.cs
@@ -112,6 +112,7 @@ namespace kOS.Screen
             GameEvents.onGUIApplicationLauncherUnreadifying.Add(RemoveButton);
             GameEvents.onHideUI.Add(OnHideUI);
             GameEvents.onShowUI.Add(OnShowUI);
+            GameEvents.onGameStateSave.Add(OnGameStateSave);
             GameObject.DontDestroyOnLoad(this);
 
             fontPicker = null;
@@ -254,6 +255,7 @@ namespace kOS.Screen
 
             GameEvents.onHideUI.Remove(OnHideUI);
             GameEvents.onShowUI.Remove(OnShowUI);
+            GameEvents.onGameStateSave.Remove(OnGameStateSave);
 
             GoAway();
             SafeHouse.Logger.SuperVerbose("[kOSToolBarWindow] OnDestroy successful");
@@ -404,6 +406,18 @@ namespace kOS.Screen
         void OnShowUI()
         {
             uiGloballyHidden = false;
+        }
+
+        // As long as the user can get this toolbar menu dialog to show up, that means
+        // they can change a setting in it.  Those changes should be persisted even when
+        // there's no KOSProcessor modules loaded in the scene.  (Thus why the config
+        // save is done here as well as in KOSProcessor.)
+        void OnGameStateSave(ConfigNode node) // ConfigNode ignored.
+        {
+            if (SafeHouse.Config != null)
+            {
+                SafeHouse.Config.SaveConfig();
+            }
         }
 
         public void OnGUI()
--- a/src/kOS/Screen/TermWindow.cs
+++ b/src/kOS/Screen/TermWindow.cs
@@ -444,10 +444,14 @@ namespace kOS.Screen
             Event e = Event.current;
             if (e.type == EventType.KeyDown)
             {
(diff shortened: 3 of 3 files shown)


src/kOS/Persistence/PersistenceExtensions.cs
```diff
@@ -128,7 +128,7 @@ public static ConfigNode ToConfigNode(this HarddiskFile file, string nodeName)
             {
                 if (SafeHouse.Config.UseCompressedPersistence)
                 {
-                    node.AddValue("binary", EncodeBase64(content.String));
+                    node.AddValue("binary", PersistenceUtilities.EncodeBase64(content.Bytes));
```
Dunbaratu: This class's local EncodeBase64() method tried to encode the string you give it, by mapping it to a byte array using a vanilla ASCII interpretation (see deleted code below in this PR diff output), which will mangle anything that needs to use the higher 8 bits of the 16 bit Unicode char.

This was unnecessary as well, since ``content``, being of type ``FileContent``, already has its own logic to do its own UTF-8 encoding, built-in.  ``content.String`` is the unicode string in chars, while ``content.Bytes`` is the UTF-8 encoded byte array version of that string, and ``FileContent`` does the work to ensure the bytes are UTF-8 rather than just a vanilla dump of the unicode 2-byte patterns.  

Once I decided to use ``content.Bytes`` instead of ``content.String``, that also made the entire reason for ``EncodeBase64()`` go away (it was *just* to encode the string for you, which ``FileContent`` already does, better.)

src/kOS/Persistence/PersistenceExtensions.cs
```diff
@@ -151,9 +151,5 @@ private static FileContent Decode(string input)
             }
         }
 
-        private static string EncodeBase64(string input)
-        {
-            return PersistenceUtilities.EncodeBase64(Encoding.ASCII.GetBytes(input));
-        }
```
Dunbaratu: This no longer needed to exist since the one place it was being called got deleted, and I don't want anyone ever using this, since it encodes with vanilla ASCII instead of with UTF-8.

src/kOS/Screen/TermWindow.cs
```diff
@@ -444,10 +444,14 @@ void ProcessKeyEvents()
...
-                var c = (char)(e.character & 0x007f);
-
+                // This ugly hack is here to solve a bug with Unity mapping
+                // Keycodes to Unicode chars incorrectly on its Linux version:
+                char c;
+                if ((e.character & 0xff00) == 0xff00) // Only trigger on Unicode values 0xff00 through 0xffff, to avoid issue #2061
+                    c = (char)(e.character & 0x007f); // When doing this to solve issue #206
+                else
+                    c = e.character;
+
```
Dunbaratu: This silly workaround we had to put in a few years ago to work around a Unity on Linux bug, caused multi-byte Unicode to break now.

### Reason for the original workaround code:

When mapping a vanilla ASCII character into Unicode, you're supposed to pad the upper bits with zero.  So for example, 0x41 ("A") should become 0x0041.  But on their Linux port, for some reason, while Unity did this correctly for printable ASCII chars, they did not do so for control ASCII chars (i.e. (0x01 (control-A), 0x08 (control-H, or 'tab'), etc).  For them, Unity padded them with 1's instead of with 0's.  So 0x0d (control-M, or "return key"), instead of giving a unicode value of 0x000d like it should, was giving a value of 0xff0d.  This made the return key fail to work in kOS on Linux.  So at the time I added this ugly hack to 'fix' it by simply masking off all the high bits.

But this broke reading keys that send non-ascii Unicode chars, because the high byte was being *always* clobbered no matter what, whether it was the problem case in the Linux port (0x000d becoming 0xff0d) or not.

### The new version

Now it only clobbers the high byte when it's *exactly* all 1's, which should never occur for any normal printable character, as the Unicode from 0xff00 through 0xffff is considered "private use" that the Unicode consortium will never assign to any standard meaning.

src/kOS/Screen/KOSToolbarWindow.cs
```diff
@@ -112,6 +112,7 @@ public void Awake()
             GameEvents.onGUIApplicationLauncherUnreadifying.Add(RemoveButton);
```
Dunbaratu: The changes in this file are only here because a different PR hasn't been merged into develop yet, and I forgot it was still in my local copy.  When that PR is merged, this diff should go away.

PR #2063, opened 2017-07-07
Modify control hack even unhooking

### LINKED ISSUE #2059: Log spam with Timing0 threw during FixedUpdate
Log gets spammed with
```
[ERR 19:02:02.811] Timing0 threw during FixedUpdate: System.NullReferenceException: Object reference not set to an instance of an object
  at kOS.Module.kOSVesselModule.CacheControllable () [0x00000] in <filename unknown>:0 
  at (wrapper delegate-invoke) Callback:invoke_void__this__ ()
  at (wrapper delegate-invoke) Callback:invoke_void__this__ ()
  at (wrapper delegate-invoke) Callback:invoke_void__this__ ()
  at Timing0.FixedUpdate () [0x00000] in <filename unknown>:0 
```
Version: 1.1.0 according to kOS.version
Not sure what triggers it as it doesn't always affect the same vehicle.
[KSP.log.zip](https://github.com/KSP-KOS/KOS/files/1101835/KSP.log.zip)

hvacengi:
Fixes #2059

kOSVesselModule.cs
* Update unhooking of events to check a "enabled" variable instead of
checking the vessel type information again (KSP appears to change the
type occasionally).
* This does not fix the potential issue where a ship may initially be a
type that does not support control but gets's switched via the UI.  For
this to be an issue, the user would have to fly "debris" or an asteroid
then change it to another type, **and** have the affected remote
connection control setting enabled.  I don't think that justifies
checking the type every single tick.
* Also added a null check and error print if `parentVessel` is null,
that way we still get a message but don't have to go through the
computational expence of exception catching.

### DIFF
--- a/src/kOS/Module/kOSVesselModule.cs
+++ b/src/kOS/Module/kOSVesselModule.cs
@@ -245,6 +245,7 @@ namespace kOS.Module
             {
                 TimingManager.FixedUpdateAdd(TimingManager.TimingStage.ObscenelyEarly, CacheControllable);
                 TimingManager.FixedUpdateAdd(TimingManager.TimingStage.BetterLateThanNever, resetControllable);
+                workAroundEventsEnabled = true;
             }
         }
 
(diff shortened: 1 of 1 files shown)


Dunbaratu:
In the end, I decided my objections aren't this PR's fault.  In other words, if the problems I thought might exist are real, they were already there.  Before this PR.  They shouldn't stop this merge.

PR #2076, opened 2017-08-02
Fixes #1860 - progress bar terminal resize

### LINKED ISSUE #1860: Signal delay progress bar ignores terminal resize - draws in the wrong spot.
This has always annoyed me a little and I think I saw the cause of it.

In RemoteTechInterpreter.cs (which is about to become renamed to ConnectivityInterpreter.cs as part of PR #1850 , which will probably be merged in before this issue is resolved, so look for it there), we have these lines:

```
        private void CreateProgressBarSubBuffer()
        {
            progressBarSubBuffer = new SubBuffer();
            progressBarSubBuffer.SetSize(3, ColumnCount);
            progressBarSubBuffer.Fixed = true;
            progressBarSubBuffer.PositionRow = RowCount - progressBarSubBuffer.RowCount;
            AddSubBuffer(progressBarSubBuffer);

            var separator = new string('-', progressBarSubBuffer.ColumnCount);
            progressBarSubBuffer.Buffer[0].ArrayCopyFrom(separator.ToCharArray(), 0, 0);
        }
```
These measurements like RowCount and ColumnCount are only queried when the Interpreter was first constructed, and then they get stuck that way for the life of the interpreter.  If you stretch or shrink the window, then all the progress bars being drawn for signal delay get messed up because it never reassigns these values for `progressBarSubBuffer` to match the new size.

Dunbaratu:
Fixes #1860 

To test:

  - Be sure RemoteTech is installed.
  - Start new sandbox game.  Choose RemoteTechConnectivityManager.
  - Use Alt-F12 'set orbit' cheat to put a comsat into Kerbin orbit
  that can act as a relay to Duna.
  - Use Alt-F12 'set orbit' cheat to put a kOS-capable probe around Duna that
  relays through the satellite you made above.
  - Keep trying ``print "hello".`` in the probe's kOS Terminal,
  after resizing the terminal to different amounts.  Notice it will
  react to changes to the terminal made during the progres bar's
  wait time too, re-drawing the bar to fit the new size at the same
  ratio of completion it had before.

### DIFF
--- a/src/kOS/Screen/ConnectivityInterpreter.cs
+++ b/src/kOS/Screen/ConnectivityInterpreter.cs
@@ -27,11 +27,6 @@ namespace kOS.Screen
             AddResizeNotifier(CreateProgressBarSubBuffer);
         }
 
-        ~ConnectivityInterpreter()
-        {
-            RemoveResizeNotifier(CreateProgressBarSubBuffer);
-        }
-
         /// <summary>Whenever the terminal resizes, resize the progress bar,</summary>
         /// <param name="sb">The method operates on self (this), and the parameter sb would
         /// be unnecessary if it wasn't required by AddResizeNotifyer().</parm>
--- a/src/kOS/Screen/ConnectivityInterpreter.cs
+++ b/src/kOS/Screen/ConnectivityInterpreter.cs
@@ -27,6 +27,18 @@ namespace kOS.Screen
             AddResizeNotifier(CreateProgressBarSubBuffer);
         }
 
+        ~ConnectivityInterpreter()
+        {
+            // Normally this design pattern would fail because the notifier hooks
+            // would be references that prevent orphaning and thus we can't remove
+            // them in the destructor.
+            // 
+            // But in this case it probably can work, because this reference is
+            // circularly from the object to *itself* (these hooks are inside this
+            // instance itself), and thus probably don't prevent orphaning:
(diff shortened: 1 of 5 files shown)


src/kOS/Screen/ConnectivityInterpreter.cs
```diff
@@ -23,19 +23,32 @@ public class ConnectivityInterpreter : Interpreter, IUpdateObserver
...
+            CreateProgressBarSubBuffer(this);
+            AddResizeNotifier(CreateProgressBarSubBuffer);
         }
 
-        private void CreateProgressBarSubBuffer()
+        ~ConnectivityInterpreter()
         {
-            progressBarSubBuffer = new SubBuffer();
+            RemoveResizeNotifier(CreateProgressBarSubBuffer);
+        }
```
Dunbaratu: I just realized this won't work.  The destructor never gets called until after you orphan this instance.  This instance can't orphan until after you call ``RemoveResizeNotifier()``, so this can never actually get called.  This has to be moved elsewhere in the logic, not in the destructor.

Dunbaratu: Addendum: Okay, actually it probably did work all along.  I put it back, with the explanatory comment about why it probably was fine all along.  But even so, I still added the hook removal to SharedObjects's Destroy work just to be explicit about it.  Now there's two spots that can both catch this and do the job.

PR #2081, opened 2017-08-04
Fixes #2066 - font changes recalc better.

### LINKED ISSUE #2066: TERMINAL:CHARHEIGHT changes don't affect CHARWIDTH until a physics tick later.
Let's say your font aspect ratio is such:

```
When charheight is:    charwidth is:
          20                12
          16                10
```
If you do this:
```
set terminal:charheight to 20.
wait 0.
print "height x width is: " + terminal:charheight + " x " + terminal:charwidth.

print "okay, now changing it:".

set terminal:charheight to 16.
wait 0. // wait here.
print "height x width is: " + terminal:charheight + " x " + terminal:charwidth.
// commented-out: wait 0.
print "again!".
print "height x width is: " + terminal:charheight + " x " + terminal:charwidth.
```
You get the expected result:
```
height x width is: 20 x 12
okay now changing it:
height x width is: 16 x 10
again!
height x width is: 16 x 10
```
But if you move the second wait 0 further down, like so:
```
set terminal:charheight to 20.
wait 0.
print "height x width is: " + terminal:charheight + " x " + terminal:charwidth.

print "okay, now changing it:".

set terminal:charheight to 16.
// commented-out: wait 0.
print "height x width is: " + terminal:charheight + " x " + terminal:charwidth.
wait 0.
print "again!".
print "height x width is: " + terminal:charheight + " x " + terminal:charwidth.
```
Then you get this instead:
```
height x width is: 20 x 12
okay now changing it:
height x width is: 16 x 12
again!
height x width is: 16 x 10
```
Note, the 16 x 12 output line this time, where before it was 16 x 10.

Something is causing char width calculations to not be right until the next physics tick *after* the height changed, even though the height value changes right away.

This may seem minor, but it does cause boot scripts that try to resize the terminal get it wrong.

If they do this:
```
set terminal:charheight to 16.
set terminal:height to 20.
set terminal:width to 30.
```
Without a ``wait 0.`` after the change to charheight, then what happens is the terminal resizes itself wrongly because it is using the stale calculation of the char width when deciding the pixels size of the window.  (i.e. it resizes the window on *this* tick, using the old width, and it comes out all wrong).

Dunbaratu:
Fixes #2066 
The fix involved moving the order in which some things happen, so as to make it not use old stale font information when drawing the window (or at least waiting for the next pass so it won't change its mind halfway through about how to draw things).

### DIFF
--- a/src/kOS/Screen/TermWindow.cs
+++ b/src/kOS/Screen/TermWindow.cs
@@ -49,6 +49,7 @@ namespace kOS.Screen
         private bool resizeMouseDown;
         private int formerCharPixelHeight;
         private int formerCharPixelWidth;
+        private int postponedCharPixelHeight;
         
         private bool consumeEvent;
         private bool fontGotResized;
@@ -240,11 +241,12 @@ namespace kOS.Screen
             base.Open();
             BringToFront();
             guiTerminalBeepsPending = 0; // Closing and opening the window will wipe pending beeps from the beep queue.
-
-            GetFontIfChanged();
         }
-
-        private void GetFontIfChanged()
+            
+        /// <param name="recalcWidth">If true, recalc the font width from height if there's any font change.
+        /// Setting this to true should only be done from inside an OnGUI call.
+        /// If you call this from outside an OnGUI context, set this to false.</param>
+        private void GetFontIfChanged(bool recalcWidth)
         {
             int newSize = shared.Screen.CharacterPixelHeight;
             string newName =  SafeHouse.Config.TerminalFontName;
@@ -256,6 +258,15 @@ namespace kOS.Screen
 
                 terminalLetterSkin.label.font = font;
                 terminalLetterSkin.label.fontSize = fontSize;
+                if (recalcWidth)
+                {
+                    CharacterInfo chInfo;
+                    terminalLetterSkin.label.font.RequestCharactersInTexture("X"); // Make sure the char in the font is lazy-loaded by Unity.
+                    terminalLetterSkin.label.font.GetCharacterInfo('X', out chInfo);
+                    shared.Screen.CharacterPixelWidth = chInfo.advance;
+
+                    NotifyOfScreenResize(shared.Screen);
+                }
             }
         }
 
@@ -306,8 +317,8 @@ namespace kOS.Screen
         {
             if (!IsOpen) return;
 
-            GetFontIfChanged();
-            
+            GetFontIfChanged(true);
+
             ProcessUnconsumedInput();
 
             if (isLocked) ProcessKeyEvents();
@@ -333,7 +344,6 @@ namespace kOS.Screen
                 consumeEvent = false;
                 Event.current.Use();
             }
-
         }
         
         void Update()
@@ -768,7 +778,6 @@ namespace kOS.Screen
 
         void TerminalGui(int windowId)
         {
-
             if (!allTexturesFound)
             {
                 GUI.Label(new Rect(15, 15, 450, 300), "Error: Some or all kOS textures were not found. Please " +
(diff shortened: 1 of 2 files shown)


src/kOS/Screen/TermWindow.cs
```diff
@@ -256,6 +258,15 @@ private void GetFontIfChanged()
...
                 terminalLetterSkin.label.fontSize = fontSize;
+                if (recalcWidth)
+                {
+                    CharacterInfo chInfo;
+                    terminalLetterSkin.label.font.RequestCharactersInTexture("X"); // Make sure the char in the font is lazy-loaded by Unity.
+                    terminalLetterSkin.label.font.GetCharacterInfo('X', out chInfo);
+                    shared.Screen.CharacterPixelWidth = chInfo.advance;
+
+                    NotifyOfScreenResize(shared.Screen);
+                }
```
Dunbaratu: The work above is work that used to be done inside TerminalGUI(), partway through drawing things.  Now it's happening  here, which occurs at the very start of OnGUI() rather than halfway through it.

src/kOS/Screen/TermWindow.cs
```diff
@@ -803,13 +812,10 @@ void TerminalGui(int windowId)
...
-            Rect fontWidthLessButtonRect = new Rect(65, WindowRect.height-28, 10, 10);
-            Rect fontWidthMoreButtonRect = new Rect(90, WindowRect.height-28, 10, 10);
-            Rect fontHeightButtonRect = new Rect(140, WindowRect.height-32, fontHeightButtonImage.width, fontHeightButtonImage.height);
-            Rect fontHeightLabelRect = new Rect(160, WindowRect.height-28, 20, 10);
-            Rect fontHeightLessButtonRect = new Rect(185, WindowRect.height-28, 10, 10);
-            Rect fontHeightMoreButtonRect = new Rect(210, WindowRect.height-28, 10, 10);
+            Rect fontHeightButtonRect = new Rect(30, WindowRect.height-33, fontHeightButtonImage.width, fontHeightButtonImage.height);
+            Rect fontHeightLabelRect = new Rect(45, WindowRect.height-33, 20, 13);
+            Rect fontHeightLessButtonRect = new Rect(75, WindowRect.height-37, 12, 13);
+            Rect fontHeightMoreButtonRect = new Rect(100, WindowRect.height-37, 12, 13);
```
Dunbaratu: While I was working on this, I noticed that the various ``Rect``s for the font width button were still here, even though the code that used them had been deleted when we removed the ability to set the font width.

I moved the font height buttons over to where the font width ones were, which gives more space to handle drawing the terminal properly when the terminal size is tiny (at small sizes, the font height buttosns had been getting drawn on top of some of the other widgets before.)

src/kOS/Screen/TermWindow.cs
```diff
@@ -836,24 +842,21 @@ void TerminalGui(int windowId)
...
             screen.Brightness = (double) GUI.VerticalSlider(brightnessRect, (float)screen.Brightness, 1f, 0f);
             GUI.DrawTexture(brightnessButtonRect, brightnessButtonImage);
-            
-            int charHeight = screen.CharacterPixelHeight;
 
-            CharacterInfo chInfo;
-            terminalLetterSkin.label.font.RequestCharactersInTexture("X"); // Make sure the char in the font is lazy-loaded by Unity.
-            terminalLetterSkin.label.font.GetCharacterInfo('X', out chInfo);
-
-            int charWidth = chInfo.advance;
```
Dunbaratu: Above are the steps that got moved to GetFontIfChanged().

src/kOS/Screen/TermWindow.cs
```diff
@@ -836,24 +842,21 @@ void TerminalGui(int windowId)
...
+            int charHeight = screen.CharacterPixelHeight;
+            int charWidth = screen.CharacterPixelWidth;
 
+            // Note, pressing these buttons causes a change *next* OnGUI, not on this pass.
+            // Changing it in the midst of this pass confuses the terminal to change
+            // it's mind about how wide the window should be halfway through painting the
+            // components that make it up:
```
Dunbaratu: A side effect of moving the font size calculations was that when you clicked the font resize buttons, the new values were getting ignored and overwritten so the buttons no longer did anything.  To make them work I had to postpone their effect until this redraw pass is finished, and make them take effect next time.

src/kOS/Screen/TermWindow.cs
Dunbaratu: Moving the logic here also had another desirable effect:  It made the terminal now retain its previous width/height (in chars, not pixels) when the font changes.  Previously I had bent over backward to try to keep the pixel size the same and choose a new cell count (widthXheight) to hold the font, but that was stupid and never worked that well,  Now the terminal just draws bigger or smaller when you change the font, to keep the same number of rows and columns.

(I had previously worried about pushing the bottom of the window offscreen when you pick a bigger font, so you can't reach the resize corner, which was why I did that convoluted logic.  But in retrospect, you can drag the window from anywhere within the window, so this is not a problem - you can just drag the window up until you reach the resize corner.)

PR #2100, opened 2017-09-03
Migrate flight parameters

### LINKED ISSUE #1955: Steeringmanager settings don't get reset to defaults on CPU boot.
Let's say you do something like this to disable the YAW portion of the steeringmanager:
```
set steeringmanager:yawpid:Kp to 0.
set steeringmanager:yawpid:Ki to 0.
set steeringmanager:yawpid:Kd to 0.
```
(My terrain leveler for rover driving does this.)

Well, it turns out that settings like this one (and possibly lots of others) don't actually reset to the defaults when you reboot the CPU.  If you leave the scene and reload it, then the steeringmanager goes back to its default state, but if you just reboot the CPU (without reloading the scene) these settings still persist.

If it's hard to auto-reset them on CPU boot (because maybe multiple CPU's on the same vessel borrow the same steeringmanager and one CPU might reboot but the other doesn't?)  then maybe providing a quick "back to factory defaults" method on steeringmanager might be a good idea.

### LINKED ISSUE #2051: Script loses control after reloading saved game
My scenario is following:
0. I have vessel with 2 processor in it. It is saved just before separation.
1. Game is loaded
2. Vessel is separated
3. Each subvessel can control itself
4. The same game is loaded again
5. Vessel is separated
6. Some (both or none) of subvessels lose control

It seems to be random if subvessel looses control or not.
My (enhanced) logs are following:
[LOG 11:47:45.670] kOS: FlightControlManager.AddTo ce754e80-f6dc-4221-8144-c49ddd9c4270
[LOG 11:47:45.777] kOS: FlightControlManager.AddTo ce754e80-f6dc-4221-8144-c49ddd9c4270
[LOG 11:47:49.687] kOS: FlightControl Binding
[LOG 11:47:49.687] kOS: FlightControl Bound
[LOG 11:47:57.226] kOS: kOS.Binding.FlightControlManager:Update(currentVessel = null)!
[LOG 11:47:57.226] kOS: kOS.Binding.FlightControlManager:Update(currentVessel = Shared.Vessel)!
[LOG 11:47:57.254] kOS: FlightControl Binding
[LOG 11:47:57.254] kOS: FlightControl Bound
[LOG 11:48:06.275] kOS: kOS.Binding.FlightControlManager:Dispose()!
[LOG 11:48:06.275] kOS: kOS.Binding.FlightControlManager:Dispose: currentVessel: null
[LOG 11:48:06.279] kOS: kOS.Binding.FlightControlManager:Dispose()!
[LOG 11:48:06.279] kOS: kOS.Binding.FlightControlManager:Dispose: currentVessel: null
[LOG 11:48:08.712] kOS: FlightControlManager.AddTo ce754e80-f6dc-4221-8144-c49ddd9c4270
[LOG 11:48:08.717] kOS: FlightControlManager.AddTo ce754e80-f6dc-4221-8144-c49ddd9c4270
[LOG 11:48:19.905] kOS: kOS.Binding.FlightControlManager:Update(currentVessel = null)!
[LOG 11:48:19.905] kOS: kOS.Binding.FlightControlManager:Update(currentVessel = Shared.Vessel)!

Here I see that:
1. Two FlightControlManagers were created and set currentVessel
2. One of them detected that root part is changed and reset currentVessel
3. But when Dispose was executed both currentVessel are null
4. Thus in new load FlightControls think they are binded and not trying to bind again

Why currentVessel is null is complete mistery for me. I tried to replace currentVessel with property and log any value sets, but didn't catch the nulls. It seems to be some dark C# magic.

`set ship:control:neutralize to true.` helps

### LINKED ISSUE #2087: Object reference not set to an instance of an object
I have been working on a script and during my tests have done several "Load Save..." to reset the state of my experimentation.

After several loads, it looks like a `runpath("0:/boot/myscript.ks").` gives me the following error at random points in my program. Here is one example of such error:

```
At (kOS built-in Update)
<<System Built-In Flight Control Updater>>
Called from 0:/boot/myscript.ks, line 102
    local t is pidRCS:update(time:seconds, relVelocity:normalized:z).
                                                                  ^
```

Questions:
* Is there a leak or something that happens when I re-load a save too many times?
* Doing a power cycle on the kOS part does not help and still results in random failures.
* Is there a workaround other than exiting to the main menu?

### LINKED ISSUE #2092: Controls not really neutral?
I have a control program that ends with the following:

```
set ship:control:translation to v(0,0,0).
set ship:control:rotation to v(0,0,0).
unlock all.
set ship:control:neutralize to true.
rcs off.

print "Waiting for neutral".
until ship:control:neutral { }
print "Ship is neutral".
  
wait until false.
```

Except at the end of the program, after a while in `wait until false`, it looks like kOS is inducing a yaw to the ship and it starts spinning. If I hit ctrl-c it goes back to neutral.

If the control are neutral, why would kOS be affecting my yaw?

hvacengi:
Fixes #1955
Fixes #2051
Fixes #2087
Fixes #2092

This moves flight parameters other than steering into the new vessel module based system.  At the same time, it improves the stability of how the vessel module and associated parameters are handled between staging, docking, saving, loading, and reverting events.

PENDING DOCUMENTATION (mostly for the new `resettodefault` suffix on the steering manager, but also nothing to clarify the change with steering conflict errors)

Noteably this PR does not address moving raw controls to after the normal autopilot events, so SAS can still override raw control.

- [x] Documentation
- ~Re-implement UnbindAll logic~
- ~Add setting to change what happens when a program ends~


Dunbaratu:
I just ran into this problem using the new code (backported to KSP 1.2.2, but I doubt that matters for this):

Command attempted was "lock wheelthrottle to 0.", which should be perfectly legal.
Result was this:

```
(Filename: C:/buildslave/unity/build/artifacts/generated/common/runtime/UnityEngineDebugBindings.gen.cpp Line: 42)

kOS.Safe.Exceptions.KOSWrongControlValueTypeException: Cannot use a Scalar as the value for the STEERING.  Should use Direction, Vector, Maneuver Node, or special string "KILL" instead.
  at kOS.Control.SteeringManager.GetDirectionFromValue (System.Object value) [0x00000] in <filename unknown>:0 
  at kOS.Control.SteeringManager.kOS.Control.IFlightControlParameter.UpdateValue (System.Object value, kOS.SharedObjects shared) [0x00000] in <filename unknown>:0 
  at kOS.Binding.FlightControlManager.SetSteeringValue (System.Object val) [0x00000] in <filename unknown>:0 
  at kOS.Safe.Binding.BoundVariable.set_Value (System.Object value) [0x00000] in <filename unknown>:0 
  at kOS.Safe.Execution.CPU.SetGlobal (System.String identifier, System.Object value) [0x00000] in <filename unknown>:0 
  at kOS.Safe.Compilation.OpcodeStoreGlobal.Execute (ICpu cpu) [0x00000] in <filename unknown>:0 
  at kOS.Safe.Execution.CPU.ExecuteInstruction (IProgramContext context, Boolean doProfiling) [0x00000] in <filename unknown>:0 
```

src/kOS/Control/WheelSteeringManager.cs
```diff
@@ -0,0 +1,160 @@
...
+                        if (bearing < -180)
+                            bearing += 360; // i.e. 359 degrees to the left is really 1 degree to the right.
+                        else if (bearing > 180)
+                            bearing -= 360; // i.e. 359 degrees to the right is really 1 degree to the left
+                    }
+                }
+                catch
+                {
+                    throw new KOSWrongControlValueTypeException(
+                        "WHEELSTEER",
```
Dunbaratu: The name used in kerboscript is wheelsteerING - this message may confuse users if it refers to it as just WHEELSTEER.

src/kOS/Binding/FlightControlManager.cs
```diff
@@ -45,176 +41,80 @@ public override void AddTo(SharedObjects shared)
...
-            ConnectivityManager.AddAutopilotHook(currentVessel, OnFlyByWire);
 
-            AddNewFlightParam("throttle", Shared);
-            AddNewFlightParam("steering", Shared);
-            AddNewFlightParam("wheelthrottle", Shared);
-            AddNewFlightParam("wheelsteering", Shared);
+            shared.BindingMgr.AddBoundVariable("THROTTLE", GetThrottleValue, SetThrottleValue);
+            shared.BindingMgr.AddBoundVariable("STEERING", GetSteeringValue, SetSteeringValue);
+            shared.BindingMgr.AddBoundVariable("WHEELSTEERING", GetWheelSteeringValue, SetWheelSteeringValue);
+            shared.BindingMgr.AddBoundVariable("WHEELTHROTTLE", GetSteeringValue, SetSteeringValue);
```
Dunbaratu: I got an error when I tried using ``lock wheelthrottle``, where it claimed I can't use a Scalar for it (because it thought I was setting the steering instead).  This looks like a cut and paste error since wheelthrottle appears to be setting the same thing as steering does.

Dunbaratu:
The old problem where the throttle keeps getting mysteriously locked out of user control for no reason came back to me today but I can't identify why.  I don't think it's a new problem, though.

src/kOS/Control/WheelSteeringManager.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        void IFlightControlParameter.UpdateAutopilot(FlightCtrlState c)
+        {
+            if (!Enabled) return;
+
+            if (!(controlShared.Vessel.horizontalSrfSpeed > 0.1f)) return;
+
+            if (Mathf.Abs(VesselUtils.AngleDelta(VesselUtils.GetHeading(controlShared.Vessel), VesselUtils.GetVelocityHeading(controlShared.Vessel))) <= 90)
+            {
+                c.wheelSteer = Mathf.Clamp(Value / -10, -1, 1);
```
Dunbaratu: I never noticed this before in our code (and it was true before your PR, I checked.  You didn't change this.)  We divide the value by 10 (well, -10 technically, so we not only divide it, we invert the sign).  so the wheelsteer in the control is 1/10th as big as the one the script gives.  To get a real wheelsteer of 1, does this mean the script needs to set it to 10, rather than setting it to just 1.0?  Was there some reason we did this that I don't follow?  Wheelsteering appears to work just fine in practice, and doesn't feel like it's nerfed to only 1/10th deflection.

Is there some strange arcane rule in the KSP API that wheelSteer doesn't range from -1.0 to 1.0 like all the the other controls in FlightControlState do?

src/kOS/Binding/FlightControlManager.cs
hvacengi: @Dunbaratu This would be why you got a steering error for `wheelthrottle`, copy/paste error on my part.  I can fix it on my end or you you can fix it on your end, whichever is easier.

Dunbaratu: I just mentioned in in a review comment (just published a minute ago before I read this).  But I didn't change it.

src/kOS/Binding/FlightControlManager.cs
```diff
@@ -45,176 +41,80 @@ public override void AddTo(SharedObjects shared)
...
         {
-            foreach (var param in flightParameters.Values)
-            {
-                if (param.Enabled)
-                {
-                    param.OnFlyByWire(ref c);
-                }
-            }
+            var throttleManager = kOSVesselModule.GetInstance(Shared.Vessel).GetFlightControlParameter("throttle"); // will throw its own exception if the parameter doesn't exist
+            return throttleManager.GetValue();
```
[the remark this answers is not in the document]
hvacengi: This was an intentional choice to make it so that the binding did not need to be aware of when a vessel stages, docks, or disassembles.

src/kOS/Binding/FlightControlManager.cs
```diff
@@ -45,176 +41,80 @@ public override void AddTo(SharedObjects shared)
...
-            if (!flightParameters.ContainsKey(paramName.ToLower())) { Debug.LogError("no such flybywire parameter " + paramName); return; }
-
-            flightParameters[paramName.ToLower()].Enabled = enabled;
-
-            if (!enabled)
-            {
-                flightParameters[paramName.ToLower()].ClearValue();
-            }
+            var throttleManager = kOSVesselModule.GetInstance(Shared.Vessel).GetFlightControlParameter("throttle");
+            throttleManager.UpdateValue(value, Shared);
```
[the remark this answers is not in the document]
hvacengi: Good catch.  Odd that I would have done the switch to `val` in other sections of code but not here.  I agree, for the sake of clarity (even if just for the code highlights on github) this should consistently be `val`.

src/kOS/Control/SteeringManager.cs
```diff
@@ -307,24 +330,9 @@ public void InitializeSuffixes()
...
-                    throw new Safe.Exceptions.KOSObsoletionException("v0.20.3", "STEERINGMANAGER:SHOWTHRUSTVECTORS", "None, function removed", "");
-                }));
-            AddSuffix("SHOWRCSVECTORS", new SetSuffix<BooleanValue>(() =>
-                {
-                    throw new Safe.Exceptions.KOSObsoletionException("v0.20.3", "STEERINGMANAGER:SHOWRCSVECTORS", "None, function removed", "");
-                },
-                value =>
-                {
-                    throw new Safe.Exceptions.KOSObsoletionException("v0.20.3", "STEERINGMANAGER:SHOWRCSVECTORS", "None, function removed", "");
-                }));
```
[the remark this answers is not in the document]
hvacengi: Yeah, I kind of decided that between the age, obscurity, and the inaccuracy (it was kOS version v0.20.1) there wouldn't be any harm in removing the errors.  We really should make a policy.  While we want to remain helpful for users upgrading, I do like cleaning up no longer used methods.  We can talk about that balance.

Dunbaratu:
I'm definitely getting the problem where after a kOS program has been in control of the throttle, and then stops being in control of it, my manual human piloting control of the throttle remains locked out from then on.  I can only move the throttle with ``lock throttle`` commands from then on.

If I switch scenes and switch back to the craft, then the manual throttle starts working again.

I'm going to want to test that in KSP 1.3 (non RO) to see if it happens there too.

src/kOS/Control/ThrottleManager.cs
```diff
@@ -0,0 +1,118 @@
...
+        void IFlightControlParameter.DisableControl()
+        {
+            controlShared = null;
+            controlPartId = 0;
+            Enabled = false;
+        }
+
+        void IFlightControlParameter.DisableControl(SharedObjects shared)
+        {
+            if (shared.KSPPart.flightID == controlPartId)
```
hvacengi: @Dunbaratu found out why you were seeing the control act funny.  The code is returning if the currently controlling part **is** the part calling to disable, instead of returning if the part calling for disable **is not** currently controlling the manager.  The same issue is in wheel steer and wheel throttle.  I will modify accordingly.

src/kOS/Binding/BindingManager.cs
```diff
@@ -137,18 +137,14 @@ public void ToggleFlyByWire(string paramName, bool enabled)
...
 
         public void Dispose()
         {
-            if (flightControl != null)
-            {
-                flightControl.Dispose();
-            }
+            // TODO: Remove this method
+            // Do nothing, left over from old steering logic
         }
```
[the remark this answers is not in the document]
hvacengi: Probably not.  I wanted to make sure I didn't loose the reference to all the locations that called these methods.  I'm not convinced that I actually want to remove this logic, and it may be more tied to your requested suffix to control how kOS releases the controls on program end.  Based on a couple of things you reported as a bug with steering/throttle releasing after program termination, I think I want to re-add the unbind logic.

src/kOS/Control/ThrottleManager.cs
Dunbaratu: I was trying to read through the code to find the problem (because, hey, it's a good way to force me to understand the code for code review reasons), and I noticed this but didn't feel confident enough in my understanding of what you were doing here to raise a flag until I dug deeper.  It *felt* like it didn't make sense and would prevent the unlock from working most of the time, but I didn't know what you were *trying* to do here so I didn't want to say outright that it was wrong.  It didn't occur to me to think, "Would it have made sense if it was inverted?  Maybe that was what he meant.".

src/kOS/Module/kOSVesselModule.cs
```diff
@@ -228,20 +227,31 @@ private void ClearParts()
...
         {
-            AddFlightControlParameter("steering", new SteeringManager(parentVessel));
+            if (flightControlParameters != null)
+            {
+                flightControlParameters.Clear();
+            }
+            flightControlParameters = new Dictionary<string, IFlightControlParameter>();
+            AddFlightControlParameter("steering", new SteeringManager(Vessel));
+            AddFlightControlParameter("throttle", new ThrottleManager(Vessel));
+            AddFlightControlParameter("wheelsteer", new WheelSteeringManager(Vessel));
```
Dunbaratu: Should say "wheelsteering", as that's what this flight control parameter is called elsewhere.  This was making ``lock wheelsteering`` fail with a misleading error message that there is no such control variable.

src/kOS/Control/SteeringManager.cs
```diff
@@ -1319,7 +1319,7 @@ object IFlightControlParameter.GetValue()
         {
             return Value;
         }
-            
+
```
Dunbaratu: For whatever reason, Monodevelop inserts space into the file for indenting purposes *regardless* of whether you actually type anything on that line or not.  It can be frustrating to have to constantly have to manually re-edit it back out.

hvacengi: Yeah, I thought about not making the edit.  It just had stood out to me in the diff because it was not surrounded by changes.  I suspect they lack the logic that is used in visual studio, where it will indent your cursor but only add the spaces if you actually type at that indent.

Issue #1960, opened 2017-02-22
Potential memory leak with Lexicons


aidygus:
I'm working on a script in one of my  repos ([astar rover](https://github.com/aidygus/astar-rover)) and making heavy usage of lexicons and lists.  I've noticed that when executing the script KSP is chewing through huge chunks of memory which is not being released.

Normally my Dev KSP runs around 2.2 gig but after 7 or 8 executes of this script it's sitting up around 13 Gig.  Most I can run this script is 10 times before I have to kill the process to free memory again.

![Normal memory usage on starting](https://cloud.githubusercontent.com/assets/767909/23231210/e829585c-f93e-11e6-9173-8df7c6ddc920.png)
![After running the script for the 5th time](https://cloud.githubusercontent.com/assets/767909/23231215/edbbb3fa-f93e-11e6-9bcb-3878f5a47e13.png)

Most of these lexicons don't get more than a couple of hundred elements in size and I've added a quit feature to clear them out before exiting in case that was causing an issue.

Dunbaratu:
This issue's description says lexicons are the problem but the code provided in the repository you linked to contains no lexicons as far as I can tell.  It contains a lot of lists, but no lexicons.

Is the description incorrect (should it really be about lists not lexicons) or is the example code incorrect (wrong link maybe)?

aidygus:
Try the feature branch.  You'll be looking at the master

hvacengi:
I haven't done anything to try and run it, but could you provide some additional information about how you are running this script?  Are you running rover.ks 7 or 8 times, or just astar2.ks?  How fast are you running each instance, and how long do you wait to see if the memory is freed?  When you say you created a quit feature, what do you mean?

Have you tried isolating the problem to just issues with lexicons, eliminating the rest of your script?  (I know that the point is that it runs in a complicated script, but the more isolated/simplified the reproduction steps are the easier it will be for us to check the behavior)

aidygus:
I start the process with **runpath("/asrover/rover.ks").** and launch the astar algorithm through that.  It can be ran independently as well by specifying LAT/LAN values or a Waypoint name.

Within the Rover script are options to call astar, either using the cursor keys to move the indicated vector to where I want to go and hit return to execute the search or if in a contract situation press **w** and corresponding number for the waypoint.

[...]

PR #2109, opened 2017-09-13
Fix shutdown/reboot error

### LINKED ISSUE #2108: Error moving opcode with shutdown function
Apparently after I added the fix for breaking execution on power state toggle as part of commit fcb4ef073d01bbcb6a660d3c1565294e8e84fef1 it resulted in breaking the `shutdown` function.  It gave me a "Cannot jump to instruction 33.  No opcode there:..." error.  I'm not sure what needs to happen to fix it, but I'm looking into it.

hvacengi:
Fixes #2108
Misc.cs
* Move setting the opcode's AbortProgram property to before changing the
processor state, since BreakExecution is now called when setting the
mode.

### DIFF
--- a/src/kOS/Function/Misc.cs
+++ b/src/kOS/Function/Misc.cs
@@ -392,9 +392,9 @@ namespace kOS.Function
             if (shared.Processor != null)
             {
                 AssertArgBottomAndConsume(shared); // not sure if this matters when rebooting anwyway.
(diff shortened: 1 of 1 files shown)


Dunbaratu:
Do you think this might have been what was behind that old issue #1960 ?  It seemed that creating the problem required the user to run again and again.  And that meant using ctrl-C possibly.  So maybe the memory leak was caused by the program contexts staying around?

PR #2111, opened 2017-09-13
Added more places to warn about SAS vs lock steering.

### LINKED ISSUE #2110: Better clarification of how SAS fights lock steering
Given that SAS still fights lock steering, and this doesn't seem to be something we can really fix reliably (I just tried after the steering manager bug fixes and the SAS problem still exists), I think we need to be much much more obvious in the docs about how SAS fights steering.  New users still get messed up by that all the time.

I intend to add statements to the docs that make it clear in more than one place to make sure the user sees it.

Dunbaratu:
Fixes #2110 with adding the warning to a few more places where SAS and lock steering are mentioned.

### DIFF
--- a/doc/source/commands/flight/cooked.rst
+++ b/doc/source/commands/flight/cooked.rst
@@ -36,15 +36,18 @@ The special LOCK variables for cooked steering
 
 .. warning::
 
+    Do not use both ``SAS`` and ``lock steering`` at the same time.
+    See the explanation below.
+
+.. warning::
+
     It's a very bad idea to``WAIT`` during the execution of the expression in a
     LOCK THROTTLE.  See the note in the next section below.
 
 .. _LOCK STEERING:
 .. object:: LOCK STEERING TO expression.
 
-   This sets the direction **kOS** should point the ship where *expression* is a :struct:`Vector` or a :ref:`Direction <direction>` created from a :ref:`Rotation <rotation>` or :ref:`Heading <heading>`:
-
-    :ref:`Rotation <rotation>`
+    This sets the direction **kOS** should point the ship where *expression* is a :struct:`Vector` or a :ref:`Direction <direction>` created from a :ref:`Rotation <rotation>` or :ref:`Heading <heading>`:
 
         A Rotation expressed as ``R(pitch,yaw,roll)``. Note that pitch, yaw and roll are not based on the horizon, but based on an internal coordinate system used by **KSP** that is hard to use. Thankfully, you can force the rotation into a sensible frame of reference by adding a rotation to a known direction first.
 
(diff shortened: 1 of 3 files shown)


doc/source/commands/flight/systems.rst
```diff
@@ -25,7 +25,9 @@ RCS and SAS
 
     Turns the RCS **on** or **off**, like using ``R`` at the keyboard::
 
-        RCS ON.
+        RCS ON. // same as SET RCS TO TRUE.
+        RCS OFF. // same as SET RCS TO FALSE.
+        PRINT RCS.  // prints either "True" or "False".
```
Dunbaratu: Technically not part of the remit of the issue, but I didn't like how terse the example was originally, and wanted to make clear the point that RCS is really just a boolean variable and can be used like one.

doc/source/commands/flight/systems.rst
```diff
@@ -34,7 +36,16 @@ RCS and SAS
 
     Turns the SAS **on** or **off**, like using ``T`` at the keybaord::
 
-        SAS ON.
+        SAS ON. // same as SET SAS TO TRUE.
+        SAS OFF. // same as SET SAS TO FALSE.
+        PRINT SAS.  // prints either "True" or "False".
```
Dunbaratu: Technically not part of the remit of the issue, but I didn't like how terse the example was originally, and wanted to make clear the point that SAS is really just a boolean variable and can be used like one.

doc/source/commands/flight/cooked.rst
```diff
@@ -84,13 +87,25 @@ The special LOCK variables for cooked steering
...
+    notice them happening.  The problem is that SAS is unaware that what
+    kOS is triyng to do is deliberate and it sees it as something that
+    has to be countered.  (i.e. kOS causes the ship to start rotating to the
+    left to seek a new heading, then SAS sees this rotation and not knowing 
+    that it's deliberate, it tries to fight it by steering harder to the
+    right.) The effect of this in-fighting is that if you ``lock steering``
+    while ``sas`` is on, you will end up seeing the craft spin around
+    wildly in a seemingly nonsensical way.  It is a good idea to 
+    get in the habit of doing ``sas off.`` whenever you ``lock steering``
+    to something.
```
hvacengi: This isn't quite accurate as the root cause.  In the old version of the steering code this was true.  But the problem currently is that we tell SAS to point to our selected rotation, but it must not be the right rotation anymore.  I know for a fact that SAS integration worked for a time, and then it broke again.  If I had more time before release I'd try to find the actual fix, but I want to hold up publishing.  I'm not sure that the text here actually needs to change, because the explanation doesn't really help the user.  If we change it, I would simply leave it at "it doesn't work, don't do it".

PR #2124, opened 2017-09-25
Update kOS.netkan

politas:
"x_netkan_epoch" gives us a numbering restart allowing us to jump over the 1.3.3.1 release. The epoch is hidden by default in the CKAN client.
"x_netkan_override" forces the min ksp version for the 1.1.3.1 release, which is still "1.0.0" in the download

I'll duplicate the 1.1.3.0 .ckan to 1:1.1.3.0 in the CKAN repo and change any others with the 1.0.0 ksp_min version.

### DIFF
--- a/kOS.netkan
+++ b/kOS.netkan
@@ -3,6 +3,7 @@
     "identifier"     : "kOS",
     "$kref"          : "#/ckan/github/KSP-KOS/KOS",
     "$vref"          : "#/ckan/ksp-avc",
+    "x_netkan_epoch" : 1,
     "name"           : "kOS: Scriptable Autopilot System",
     "abstract"       : "kOS is a scriptable autopilot Mod for Kerbal Space Program. It allows you write small programs that automate specific tasks.",
     "author"         : "many KSP-KOS contributers, currently dunbaratu and hvacengi",
@@ -33,6 +34,14 @@
         {
             "file"   : "GameData/kOS",
             "install_to" : "GameData"
+        }
+    ],
+    "x_netkan_override": [
+        {
+            "version": "1:1.1.3.1",
+            "override": {
+                "ksp_version_min" : "1.2.0"
+            }
         }
     ]
 }


Dunbaratu:
To let KSP 1.2.2 users see the new update, do I need to incorporate these changes in and then cut a new release (with a new tag), or is this a change to just keep things from breaking a second time in the future (in which case adding them to the develop branch and waiting a few months for the next release is good enough)?  I'm not sure how to make git retroactively alter the contents of an already released tag.

politas:
You won't have to do anything else. KSP-CKAN/CKAN-meta#1293 will give users a clean upgrade path to the correct releases on your github

politas:
Ok, hang on, what's going on with that 1.1.0 deleted release? That is _weird_

politas:
@Dunbaratu, please merge this to stop the netkan crawler replacing the newly fixed kOS-1.1.3.1.ckan

Dunbaratu:
1.1.0 was deleted because Github's web interface lied to me.  I clicked "edit" on the release.  One of the fields you can edit on the "edit" release page is the tag name.  I edited the v.1.1.0 tag name to v1.1.0 as it should have been.  I clicked on save.  This is the user interface telling me I am *editing* the release in-place.  But this caused it to make a *new* tag called v1.1.0, which meant it had cut a *new* copy of the mod in it's *new* current state.  This is not "editing" a release.  This is "making a new" release.  It's just a new release that happens to overwrite the old one (i.e. in my reckoning, that's a delete and an insert operation, not an edit operation as the interface implied).  Thus instead of just renaming v.1.1.0 to v1.1.0 as the web interface claimed it was doing, what actually happened is that the old release v.1.1.0 was removed and a new release (with new timestamp, and with updated content because it cut a new tag from the repo) of v1.1.0 was made.)  This caused v1.1.0 to become the topmost, "latest release" in the list even though it's version number was below the recent v1.1.3.1 and v1.1.3.0.  I did not want that bogus stuff getting snarfed up by CKAN so I decided deleting it was far better than allowing the falsified bogus "new" release to  be seen.  People being unable to revert back several versions to v1.1.0 (because I deleted the release) is a less bad problem than people being told to "upgrade" from v1.1.3.0 to v1.1.0 as would have happened if I'd left it in place.

In the past I'd been able to "edit" an old release without it rising to the top of the list as the "latest" one.  But if you edit the tagname field, then it isn't really just an "edit", and it can rise to the top as "latest".  IMO the tagname field shouldn't be editable from the edit screen if editing it causes an operation that isn't actually an "edit" in-place.  It's really misleading.

Dunbaratu:
@politas - I am very confused about how CKAN gets its data.  The netkan file is part of the source code files that are in the repo, and it's part of the files that get a snapshot taken of them with the tag for a release.  Yet somehow you're saying that editing the netkan file on the develop branch, and not yet releasing that develop branch or making a new tag of the repo in any way, still makes some kind of difference to CKAN's database.

Is CKAN reading the netkan file directly from the repo, *prior* to those edits ever appearing in any release?

politas:
The Netkan file is read directly from the repo, yes. The [kOS.netkan file in CKAN's NetKAN repo](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/kOS.netkan) directs the robot to look at the GitHub repo via the URL https://raw.githubusercontent.com/KSP-KOS/KOS/master/kOS.netkan. The robot has no other information to point it to the kOS release list until it reads the kOS repo's kOS.netkan file (this file). When the robot reads this netkan file, it then knows where to look for new releases, and it looks only at the most recent release according to the GitHub Release presentation. The release number presented by GitHub is used as the release number.

It then downloads the zip file from the release, finds the kOS.version file inside that, and uses it to set the KSP compatibility for the release.

Does that clear it up for you?

hvacengi:
Ok, so we will need to merge into master not just develop. I'll make sure we have it in the right branch.

Dunbaratu:
> Does that clear it up for you?

@politas  - Yes.  My confusion was because I thought the order was this:
1 - CKAN daily checks the repo to see if it there's been a new release.
2 - If yes, *then* (and not before) it looks for the copy of the netkan file *that was in that release* (i.e. asking git to retrieve the file *in its historical form it was in when that tag got made*).

If the order is this way around:
1 - CKAN daily reads the current state of the netkan file in the repo whether that file is in a state intended for public release or not.
2 - CKAN uses the netkan file to look to find out if there's been a release.

Then that means we need to be much more careful about how that netkcan file is edited.  It means CKAN doesn't wait for us to officially release before it sees changes to the file.  Changes there can end up affecting CKAN immediately, before we intended to release them.

In practice it's *almost* the same thing because we try not updating the master branch until ready to make an official release, preferring to do all edits prior to release on the 'develop' branch.  But it does mean there's a small window of time when CKAN might pick up the change prior to us choosing to actually release.

Dunbaratu:
> Ok, so we will need to merge into master not just develop. I'll make sure we have it in the right branch.

@hvacengi have you done this?  I didn't see a merge message for master branch.

hvacengi:
@dunbaratu no I didn't get it merged. For some reason it looks like the master branch and the develop branch are out of sync and I was working on identifying those differences and looking up documentation for git cherrypick when I went to bed. If you want to handle it, that's fine by me. Otherwise I'll make it happen at lunch today.

hvacengi:
This should now be merged into the master branch.  @politas let us know if we need to make further changes.  But with any luck that should be good to go.

PR #2128, opened 2017-10-04
Issue 2127 RemoteTech local control

### LINKED ISSUE #2127: No access to KOS Terminal on a manned craft without connection to KSC using RemoteTech
When using KOS with RemoteTech connectivity Manager the terminal is inaccessible when there is no connection to KSC or another control station even on manned vessels. In this case the Message "Signal lost.  Waiting to re-acquire signal." is displayed at the bottom of the KOS terminal sreen and no further input is possible.

When using the CommNet connectivity manager, access to the KOS terminal on manned craft is possible, even though no connection to KSC / control station is availlable.

I think the  reason for the different behaviour between  RemoteTech connectivity manager and CommNet connectivity manager lies in the implementation of the "HasConnectionToControl" Method of the connectivity managers. This method is used by the "UpdateDeployment" Method of the ConnectivityInterpreter class to determine wether it should be allowed to type into the KOS Terminal

```
private void UpdateDeployment(double deltaTime)
{
            if (!ConnectivityManager.HasConnectionToControl(Shared.Vessel))
            {
                if (!signalLossWarning)
                {
                    DrawStatus("Signal lost.  Waiting to re-acquire signal.");
                    progressBarSubBuffer.Enabled = true;
                    signalLossWarning = true;
                }
            }
```

Within the "HasConnectionToControl" Method of the CommNet connectivity manager (in file: src\kOS\Screen\CommNetConnectivityManager.cs) having control is determined by either having a connection to KSC / ground station or by the presence of crew (vessel.CurrentControlLevel >= Vessel.ControlLevel.PARTIAL_MANNED):

```
public bool HasConnectionToControl(Vessel vessel)
{
            if (!IsEnabled)
                return true;

            // IsConnected is only set to true on the active vessel, so we have to manually
            // call the FindHome method to evaluate the home connection.

            // WARNING: In stock this will only work for vessels with a relay antenna installed.
            // This is a limitation put in place to improve performance in the stock game, and
            // there isn't a very good way around it.
            if (!vessel.isActiveVessel)
            {
                var net = CommNetNetwork.Instance.CommNet;
                tempPath = new CommPath();
                net.FindClosestControlSource(vessel.Connection.Comm, tempPath);
                return tempPath.signalStrength > 0 || vessel.CurrentControlLevel >= Vessel.ControlLevel.PARTIAL_MANNED;
            }
            return vessel.Connection.IsConnected || vessel.CurrentControlLevel >= Vessel.ControlLevel.PARTIAL_MANNED;
}
```

Within the RemoteTech connectivity manager (in file: src\kOS\AddOns\RemoteTech\RemoteTechConnectivityManager.cs) you need a vessel with a RemoteTech SPU and a working connection for this method to return true. There is no additional check for crew:

```
public bool HasConnectionToControl(Vessel vessel)
{
            if (!RemoteTechHook.IsAvailable(vessel.id))
                return false; // default to no connection if the vessel isn't configured for RT.
            return RemoteTechHook.Instance.HasAnyConnection(vessel.id);
}
```

Changing the last line to 

```
return RemoteTechHook.Instance.HasAnyConnection(vessel.id)  || vessel.CurrentControlLevel >= Vessel.ControlLevel.PARTIAL_MANNED;
```

fixed my problem. I still require a RemoteTech SPU on the vessel (due to:  if (!RemoteTechHook.IsAvailable(vessel.id)) return false;) but I think this is the design intend of KOS/ RemoteTech connectivity manager.

Sofar I did not run into additional problems that might be caused by this change. Since the check for crew is also included in the CommNet connectivity manager I think this addition might be a working solution.

Remark:
-The files quoted are from the current master branch, downloaded yesterday (october 3rd).
-I found a file called RemoteTechInterpreter.cs in the src\kOS\AddOns\RemoteTech2 folder. This file also includes a UpdateDeployment Method. In this Method the check for crew was included. However, this file is no longer part of the solution.

hvacengi:
Fixes #2127

Fixes the RemoteTechConnectivityManager to check HasLocalControl when evaluating HasConnectionToControl.  Also removes a few old files and a utility class.


Dunbaratu:
reminder to self: also check on ksp 1.2.2 with older RT.

PR #2137, opened 2017-10-19
Remove unnecessary identifier ToLower

tsholmes:
https://github.com/KSP-KOS/KOS/blob/develop/src/kOS.Safe/Compilation/KS/Compiler.cs#L210
The compiler already lowercases all identifiers, no need to re lowercase them on every variable access. Removing these lowercases reduces the runtime of my test script (sorting 200 random numbers in a `list`) from 18 seconds to 16 seconds.


hvacengi:
I ran into an error when attempting to access bound variables.  From discussion in slack it appears to be because bound variables call `AddVariable` directly without changing case.  My quick attempt at a fix was to modify `VariableScope.cs` to change line 48 from 
```
Variables = new Dictionary<string, Variable>();
```
to
```
Variables = new Dictionary<string, Variable>(StringComparer.OrdinalIgnoreCase);
```
But there would be merit in comparing the performance of `OrdinalIgnoreCase` vs calling `ToLower`.  My expectation is that they are essentially a wash, but that `OrdinalIgnoreCase` would be slightly less memory intensive due to differing allocation strategies.

Dunbaratu:
@hvacengi  
There was some discussion in the Slack about whether or not this really provides any speedup.
What was the final decision on that?  It was unclear.  Do we merge this and change everything to ToLower, or leave it using OrginalIgnoreCase like it is now?

tsholmes:
Running a sample program with `ToLower` took between 6.3 and 6.6 seconds (was pretty inconsistent), where running it with `OrdinalIgnoreCase` consistently took 6.2 seconds, so I'm gonna call that the winner. I'll switch this to put an OrdinalIgnoreCase into the VariableScope

PR #2138, opened 2017-10-20
Rewrite Stack to be double stack

tsholmes:
Stack operations had to shift everything in the above stack up and down on every push and pop. The above stack is usually small since it is just the context chain, but the extra shuffling adds up. This switches it to be two separate stacks (implemented as fixed-size arrays to avoid re-allocating), changing the `MoveStackPointer` into `PushAbove` and `PopAbove`, as it was never used for transferring between the stacks, just for accessing the above stack.

This reduces the runtime of my sort script from ~16 seconds to ~14.5 seconds.

### DIFF
--- a/src/kOS.Safe/Execution/Stack.cs
+++ b/src/kOS.Safe/Execution/Stack.cs
@@ -11,15 +11,15 @@ namespace kOS.Safe.Execution
         private const int MAX_STACK_SIZE = 3000;
         private readonly object[] stack = new object[MAX_STACK_SIZE];
         private int count = 0;
-        private readonly object[] aboveStack = new object[MAX_STACK_SIZE];
-        private int aboveCount = 0;
+        private readonly object[] scopeStack = new object[MAX_STACK_SIZE];
+        private int scopeCount = 0;
 
         private int triggerContextCount = 0;
 
         public void Push(object item)
         {
             ThrowIfInvalid(item);
-            if (count + aboveCount >= MAX_STACK_SIZE) {
+            if (count + scopeCount >= MAX_STACK_SIZE) {
                 // TODO: make an IKOSException for this:
                 throw new Exception("Stack overflow!!");
             }
@@ -43,28 +43,28 @@ namespace kOS.Safe.Execution
             return item;
         }
 
-        public void PushAbove(object item)
+        public void PushScope(object item)
         {
             ThrowIfInvalid(item);
-            if (count + aboveCount >= MAX_STACK_SIZE)
+            if (count + scopeCount >= MAX_STACK_SIZE)
             {
                 // TODO: make an IKOSException for this:
                 throw new Exception("Stack overflow!!");
             }
-            aboveStack[aboveCount++] = ProcessItem(item);
+            scopeStack[scopeCount++] = ProcessItem(item);
 
             checkTrigger(item, +1); 
         }
 
-        public object PopAbove()
+        public object PopScope()
         {
-            if (aboveCount == 0)
+            if (scopeCount == 0)
             {
                 return null;
             }
 
-            object item = aboveStack[--aboveCount];
-            aboveStack[aboveCount] = 0;
+            object item = scopeStack[--scopeCount];
+            scopeStack[scopeCount] = 0;
 
             checkTrigger(item, -1);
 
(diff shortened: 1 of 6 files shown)


src/kOS.Safe/Execution/Stack.cs
```diff
@@ -9,25 +9,73 @@ namespace kOS.Safe.Execution
     public class Stack : IStack
     {
         private const int MAX_STACK_SIZE = 3000;
-        private readonly List<object> stack = new List<object>();
-        private int stackPointer = -1;
+        private readonly object[] stack = new object[MAX_STACK_SIZE];
+        private int count = 0;
+        private readonly object[] aboveStack = new object[MAX_STACK_SIZE];
```
Dunbaratu: **We should stop calling this "above".**
If we're changing this into two different stacks, I think it we should also change the terminology.
This was called the "above" stack purely because of where it was located - above the other stack.
This terminology change should also affect all the other places the word "above" is used to refer to this stack, like PopAbove, PushAbove, aboveCount, etc.  There may also be mentions of this in comments trying to explain the "above stack", which should change too.

Ideas for new term: scopeStack, or callStack.

src/kOS.Safe/Execution/Stack.cs
```diff
@@ -168,31 +215,19 @@ public string Dump()
...
                 var builder = new StringBuilder();
-                builder.AppendLine("Stack dump: stackPointer = " + stackPointer);
+                builder.AppendLine("Stack dump: count = " + count + ", aboveCount = " + aboveCount);
 
-                // Print in reverse order so the top of the stack is on top of the printout:
-                // (actually given the double nature of the stack, one of the two sub-stacks
-                // inside it will always be backwardly printed):
-                for (int index = stack.Count - 1; index >= 0; --index)
+                // To simulate a single stack with the SP in the middle, we go through the aboveStack bottom up, then
+                // the normal stack top down
```
Dunbaratu: If we're changing the underlying implementation, I don't mind exposing that fact to users in the `Dump()` output.  This is meant to be a dump of what is really going on under the hood, so we should just print out the two stacks as two different stacks, labeled accordingly.

src/kOS.Safe/Execution/Stack.cs
```diff
@@ -203,6 +238,28 @@ public string Dump()
             }
         }
 
+        private void dumpItem(int index, bool isSP, object item, StringBuilder builder)
```
Dunbaratu: I like moving this out - it was getting nested a bit too deep.

src/kOS.Safe/Execution/Stack.cs
Dunbaratu: **Note to self**: this will be a source of merge conflict for me:  On the subject of creating the stack so it stays permanently at MAX_STACK_SIZE all the time - I had a change in my clock cycle branch designed to do this already but it wasn't ready to post it - I was keeping it as a stack (not an array) but telling .net to pre-allocate the stack to the full capacity.  I will need to take care when updating that branch to develop because this will conflict for sure.

tsholmes: ah yeah good point. i was just trying to keep things as similar as possible, but some clearer naming here would definitely make this more approachable in the future.

Dunbaratu:
A checklist of things I'm going to edit in my review - just to make sure I get all of them:
[ ] We should have two different MAX_STACK_SIZE values now, one for each stack, just in case we feel like setting their limits differently from each other later.
[ ] Some comments still refer to the way the "above stack" worked.  They are easy to miss because they're comments so don't affect compilation.  Edit comments to explain things the new way.
[ ] Instead of peeking a negative number to get to the scope stack, explicitly provide separate scope stack peeking API that uses positive numbers, and throw a "you should never see this" exception if we still pass in a negative number somewhere.  This change also has to propagate upward through the CPU and ICPU wrappers around the stack methods providing separate methods there too.
[ ] Instead of "scope stack" and "stack", rename the "stack" to "argument stack" so both stacks have descriptive names.  This includes:
-  [ ] Rename them inside Stack.cs and IStack.cs, and renaming all the API methods accordingly.
-  [ ] Rename their wrapper methods inside CPU.cs and ICPU.cs to give the same effect.

PR #2140, opened 2017-10-22
Skip interface cast to allow inlining InstructionPointer access

tsholmes:
This function is the vast majority of accesses to the InstructionPointer, but since it was being passed an interface type (where the only call site passes in a concrete type), the compiler couldn't inline the property access.

This is accessed so frequently that allowing it to inline here gives it a 5% speedup.

### DIFF
--- a/src/kOS.Safe/Execution/CPU.cs
+++ b/src/kOS.Safe/Execution/CPU.cs
@@ -1498,8 +1498,8 @@ namespace kOS.Safe.Execution
                 SafeHouse.Logger.Log(executeLog.ToString());
         }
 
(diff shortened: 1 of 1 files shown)


Dunbaratu:
I sometimes think we go overboard on the indirection through interface definitions.  It makes otherwise small edits a lot more tedious as you end up having to edit the same thing in 3 places, and sometimes it inserts a lot of extra complications to be able to read a piece of information from a location where the class has been abstracted into an interface with two implementations in different places.

But that was largely why it is that it's possible to decouple parts of kOS from the game and run a console command line version.

PR #2142, opened 2017-10-22
Avoid string StartsWith and EndsWith

tsholmes:
String functions in mono are extremely slow. This avoids calling `StartsWith` and `EndsWith` for cases where we could just check the characters directly.

This had a varying effect from almost no speedup to an 8% speedup in my test cases depending on how call heavy the script was (the `EndsWith("()")` check is the most expensive).

### DIFF
--- a/src/kOS.Safe/Utilities/StringUtil.cs
+++ b/src/kOS.Safe/Utilities/StringUtil.cs
@@ -1,0 +1,51 @@
+﻿using System;
+namespace kOS.Safe
+{
+    /// <summary>
+    /// String utility functions that avoid the culture-aware overhead of the builtin string functions.
+    /// </summary>
+    public static class StringUtil
+    {
+        public static bool EndsWith(string str, string suffix)
+        {
+            int strLen = str.Length;
+            int suffixLen = suffix.Length;
+
+            if (strLen < suffixLen)
+            {
+                return false;
+            }
+
+            for (int i = 0; i < suffixLen; i++)
+            {
+                if (str[strLen - i - 1] != suffix[suffixLen - i - 1])
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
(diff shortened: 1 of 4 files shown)


src/kOS.Safe/Utilities/StringUtil.cs
```diff
@@ -0,0 +1,23 @@
...
+            int len = str.Length;
+            return len > 0 && str[len-1] == '*';
+        }
+
+        public static bool HasVarPrefix(string str)
+        {
+            return str.Length > 0 && str[0] == '$';
+        }
+    }
+}
```
Dunbaratu: Should we just have a generic "endswith" and "startswith" call in our utility library that operates on generic strings but does it char by char?  Would it be an improvement over the generic mono one that way?  i.e:
```
public static bool EndsWith(string biggerString, string substring)
{
    int len = subString.Length;
    int i = biggerString.Length - len;
    if (i<0) return false;
    while (i < len)
    {
        if (subSrring[i] != biggerString[j]) // maybe allow for culture case-insensitive comparison here?  Nah, probably not.
            return false;
        ++i;
        ++j;
    }
    return true;
}
```
It seems weird to be writing one hardcoded primitive method per endswith comparison we want to do, but I can conceed it may be worth it if it's faster that the generic solution.

src/kOS.Safe/Utilities/StringUtil.cs
```diff
@@ -0,0 +1,51 @@
...
+            int suffixLen = suffix.Length;
+
+            if (strLen < suffixLen)
+            {
+                return false;
+            }
+
+            for (int i = 0; i < suffixLen; i++)
+            {
+                if (str[strLen - i - 1] != suffix[suffixLen - i - 1])
```
Dunbaratu: I'm going to switch to counting this forward to save some math.  (set index to strLen - i - 1 at first, then just increment it by +1 each iteration).

PR #2144, opened 2017-10-23
Cache type lookups on Suffix delegates

tsholmes:
Includes changes from #2139 

The type lookups (the largest of which by a huge margin was the check for varargs) were taking a decent amount of time. During profiling this only looked like it was going to result in a speedup of a few percent, but on running tests with profiling off, it resulted in a 10-25% speedup (depending on how suffix heavy the workload was).

### DIFF
--- a/src/kOS.Safe/Encapsulation/Suffixes/DelegateSuffixResult.cs
+++ b/src/kOS.Safe/Encapsulation/Suffixes/DelegateSuffixResult.cs
@@ -10,9 +10,11 @@ namespace kOS.Safe.Encapsulation.Suffixes
     public class DelegateSuffixResult : ISuffixResult
     {
         /// <summary>
-        /// The C# delegate that implements this built-in suffix.
+        /// Information about the delegate obtained via reflection.
+        /// This is retrieved once per suffix and cached thereafter to
+        /// avoid the expensive reflection lookups on every call.
         /// </summary>
-        private readonly Delegate del;
+        private readonly DelegateInfo delInfo;
         /// <summary>
         /// Although the delegate `del` knows via reflection how many
         /// arguments it should expect, using reflection to invoke it is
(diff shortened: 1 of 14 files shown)


src/kOS.Safe/Encapsulation/Suffixes/DelegateSuffixResult.cs
```diff
@@ -44,27 +41,24 @@ public void Invoke(ICpu cpu)
...
             {
+                DelegateParameter paramInfo = delInfo.Parameters[i];
+
                 object arg = cpu.PopValue();
                 Type argType = arg.GetType();
-                ParameterInfo paramInfo = paramArray[i];
-
-                // If this is the lastmost parameter then it might be a 'param' array which expects all the rest of
-                // the arguments to be collected together into one single array parameter when invoking the method:
-                isParamArrayArg = (i == paramArray.Length - 1 && Attribute.IsDefined(paramInfo, typeof(ParamArrayAttribute)));
```
hvacengi: I ran into an issue where kOS threw an `KOSArgumentMismatchException` when executing a delegate suffix.

It doesn't look like there is any point where you are checking to set `isParamArrayArg` to true after removing this line.  I suspect you either intended to check `delInfoParameters` for this information below, or we should add this line back without checking the attribute.  In my quick test, it looks like adding the following line fixes execution:
```
isParamArrayArg = i == delInfo.Parameters.Length - 1 && delInfo.Parameters[i].IsParams;
```

tsholmes: ah yeah i accidentally deleted that line 🤐  fixed it when i rebased

Dunbaratu:
This PR in github is still assigned to @hvacengi as the reviewer.  Does that mean you'll be the one merging it, hvacengi?  If you've reviewed it to your satisfaction but won't be the one merging it, let me know so I know to pick it up and do it.

PR #2148, opened 2017-10-26
Move as many functions as possible to kOS.Safe

### LINKED ISSUE #2147: Most builtin functions should be in Safe
Most of the builtin functions do not reference any KSP or Unity types and can be in Safe. Not having them there prevents us from being able to unit test them, and prevents us from setting up integration tests that actually run scripts and check their output.

tsholmes:
Fixes #2147 

This is the first step on the path to having a full test suite that runs scripts and validates their results.


hvacengi:
Just a summary of some of the conversation we had in Slack: @Dunbaratu and I are interested in looking into moving some of the structure constructor functions into nested classes of the structure they construct.  In other words the kerboscript `list()` function (C# class `FunctionList`) would be moved to being nested inside of `ListValue`.  This would be for the sake of helping simplify finding some of these functions in the source, instead of breaking them all into separate files.

Having done a little bit of testing, it appears that the above mentioned nesting idea will work just fine.  I'm working on preparing a PR to @tsholmes with those revisions.  After he's had a chance to review the changes and make sure the code looks good we should be able to merge into develop.

Issue #2152, opened 2017-10-28
Identifiers should not be on the stack


tsholmes:
Right now we push identifiers with various prefixes and suffixes onto the stack to signal that they are identifiers. These should be stored in the opcodes in a way that is clear they are identifiers.

```
push $identifier
```
changes to
```
pusheval identifier // pushes the value referred to by identifier onto the stack
```
---
```
push $identifier
push 1
storelocal
```
changes to
```
push 1
storelocal identifier
```
---
```
push $identifier
push suffix
getmember
```
changes to
```
pusheval identifier
getmember suffix
```

and similarly for the rest of the opcodes that reference identifiers.
This will allow us to know the types of the values on the stack at all times, and eventually restrict them to only be `Structure` types that the script can directly interact with

PR #2156, opened 2017-10-28
Include identifier in store opcode

tsholmes:
Includes the identifier name in the store, storeexist, storelocal and storeglobal opcodes.

This is the first step to fixing #2152

### DIFF
--- a/kerboscript_tests/integration/lock.ks
+++ b/kerboscript_tests/integration/lock.ks
@@ -1,0 +1,10 @@
+
+global a to 1.
+global b to 2.
+
+lock c to a + b.
+
+print(c).
+
+set b to 3.
+print(c).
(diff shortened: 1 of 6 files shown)


Dunbaratu:
@tsholmes Do you want to put the rest of fixing issue #2152 into this same PR or make separate PR's for them later?
The reason I ask is that *normally* when a PR is associated to an issue, if that PR gets merged then github automatically closes that issue, claiming the issue is done.  In this case you say this PR only goes partyway toward resolving the issue. That automated behaviour from github would be the wrong thing to do, so we need to be mindful of that if we merge this PR and you planned to have more PR's coming later on for the issue.  We'd need to re-open the automatically closed issue after merging.

tsholmes:
I'm going to make another PR for it later. I made sure to not use any of the trigger phrases (`fixing #2152` won't trigger the auto-close).

PR #2157, opened 2017-10-29
Fixes #2154

### LINKED ISSUE #2154: change program's outer scope to make locals really be local.
(BREAKING: This will mean local is really local and not global, at program level.)

Right now (master branch) if you do this:

**MyProgram.ks**
```
parameter x, y, z.
local a is 0, b is 0, c is 0.
global i is 0, j is 0 , k is 0.
```
Then a,b,c, and i,j,k are all really global, or even sillier, they end up in scope -999, which is the one used in the loader boilerplate that shouldn't be accessible by the program once it runs.
But we had a bug where two wrongs made a right, and it accessed the scope -999 when it shouldn't have, so we didn't notice.

But after the recent merge of PR #2141 , the new develop branch has a bug because of this previous bug: if you add additional nesting inside the program, it can't see the outer scope variables:

**called_program.ks**
```
local a is 0.
print "can print A here: " + a.
{
  print "cannot print A here- it will crash with 'a' not known:" + a.
  // NOTE THIS BUG IS ONLY TRIGGERED IF you run this program
  // FROM another program.  If you run it from the interpreter it
  // doesn't have the bug.  This is because the bug requires this
  // program to be run via the boilerplate program loader function
  // inserted at the top of all programs.
}
```

**calling_program.ks**
```
run called_program.ks.
```

(In a nutshell, we had two wrongs making a right, and PR #2141 deleted one of the two wrongs making us notice the other wrong.)

The fix to this is to make sure all programs have their own pushscope/popscope at the outermost level, which should fix the bug we've been having on ``develop`` but also introduce the breaking change that ``local`` really does mean ``local`` now.

Dunbaratu:
Fixes #2154 

### TODO list before this becomes ready:

- [x] - user docs need to talk about function's scoping changes and the local file scope changes.
- [x] - breaking warning drafted about local scope for the file now existing.
- [x] - breaking warning drafted about parameters having been global before but not now.

This change wraps a scope around a file (similar to putting an opening brace at the top of the file and a closing one at the bottom) so the file's local variables are really local, and so the functions declared in the file have proper scope captured in their closures.

### Doc changes needed:

Note, these will need to be fleshed out with good examples - these sentences below are just bullet-point reminders:

- Files now have a scope.  (**breaking**: Previously saying `local foo is 1.` or `global foo is 1.` meant the same thing at the file level.  Now they differ as one would expect.  lazyglobal will still make globals as it did before when you set an unknown variable name.  This only affects what happens when you explicitly say `local` - now it works - before it just became global anyway because the file has no scope.)
- Even though it's slightly wrong, to preserve backward compatibility functions declared at the file scope level will be call-able from other files because their function pointer variable will have global scope by default.  You can override this by explicitly saying the function is `local` when you declare it.
- However, UNLIKE before, even though the function's name can be seen and called from other files, its closure context inside its body will now be local as it should have been, so it can access local variables declared in the same file.
- With the explicit ``global`` keyword, you now should be able to make a function's name globally call-able, even when it is declared deeply nested in other things (this is consistent with using 'global' for variable identifiers).  It will still have local scope closure inside its body, however.  (This used to not be allowed - the ``global`` keyword was forbidden for function declarations.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,83 @@
 kOS Mod Changelog
 =================
+
+# VTODO_RELEASE_NUM_HERE_WHEN_WE_RELEASE TODO_RELEASE_CUTE_NAME_HERE
+
+This release was primarily focused on speedups and smoothness
+of execution.  We welcomed a new developer (github username tsholmes)
+who contributed a lot of bottleneck analysis and code speedups.  The
+goal was to reduce the burden kOS causes to the physics rate of the
+game, and consequently also allow tech tree scaled performance by era
+for the kOS computer parts themselves (slow at first, faster later).
+
+### BREAKING CHANGES:
+
+- Files now have an implied local scope, causing the following change:
+  - **Previously:** If you declared a variable as ``local`` at the
+    outermost scope of a program file (outside any curly braces),
+    then it had the same effect as ``global``, creating a varible
+    that you could see from anywhere outside that program file.
+  - **New behaviour:** Now that there is an outermost scope for a file,
+    ``local`` actually means something in that scope.  To get the
+    old behaviour you would need to explicitly call the variable
+    ``global``.i
+  (The variables magically created via the lazyglobal system will still
+  be global just like they were before.)
+- Parameters to programs now have local scope to that program file.
+  (Previously they were sort of global and visible everywhere, which
+  they shouldn't have been.  If you relied on this behaviour your
+  script might break.)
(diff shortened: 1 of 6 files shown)


Dunbaratu:
These doc edits are *enough* to make it merge-able, but you could argue we need to make a better example to really nail it down.  I don't want to make that example right now because it will take a bit of time and this is holding up everyone else working on kOS.

PR #2181, opened 2017-11-29
Add suffix names into opcodes

tsholmes:
Part 2 of 3 in fixing #2152

This puts suffix names in the opcodes that use them, putting us one step closer to only having encapsulated values on the stack.

### DIFF
--- a/src/kOS.Safe/Compilation/KS/Compiler.cs
+++ b/src/kOS.Safe/Compilation/KS/Compiler.cs
@@ -1754,7 +1754,13 @@ namespace kOS.Safe.Compilation.KS
 
                 // Push this term on the stack unless it's the name of the user function or built-in function or a suffix:
                 bool isDirect = true;
-                if (nodeIndex != 0 && !usingSetMember)
+
+                if (usingSetMember && suffixTerm.Nodes.Count == 1)
+                {
+                    // If this is the name of a suffix that we are setting, don't do anything with it.
+                    // ProcessSetOperation will handle putting the suffix name into the opcode.
+                }
+                else if (nodeIndex != 0 && !usingSetMember)
                 {
                     string suffixName = GetIdentifierText(suffixTerm.Nodes[0]);
                     AddOpcode(startsWithFunc ? new OpcodeGetMethod(suffixName) : new OpcodeGetMember(suffixName));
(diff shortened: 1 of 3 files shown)


src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -2024,6 +2030,49 @@ private bool VarIdentifierEndsWithSuffix(ParseNode node)
                  prevChild.Token.Type == TokenType.suffixterm);
         }
```
Dunbaratu: I sometimes hate the way git's diff shows things.  This ends up looking like a much more complex change until I realized it's mostly moving the clauses in the if/else ladder to a new order, which git diff doesn't quite "get" so it shows it as a lot of individual edits instead of just moved code.

src/kOS.Safe/Compilation/KS/Compiler.cs
```diff
@@ -2024,6 +2030,49 @@ private bool VarIdentifierEndsWithSuffix(ParseNode node)
...
+            }
+            if (descendedThroughAColon && 
+                (prevChild.Token.Type == TokenType.suffix_trailer ||
+                 prevChild.Token.Type == TokenType.suffixterm))
+            {
+                return GetIdentifierText(prevChild);
+            }
+
+            throw new KOSYouShouldNeverSeeThisException("VarIdentifier didn't end in a suffix");
+        }
```
Dunbaratu: It occurs to me that ``GetVarIdentifierEndSuffix()`` and ``GetVarIdentifierEndsWithSuffix()`` are almost identical cut and paste code.  Can we avoid repeated code here by having them share a commonality?  Both perform the same walk.  They only differ in what they do when they get to the bottom.  One returns the string and throws exception if there's no such string to return.  The other sees if there would be such a string and returns boolean whether or not there was.

These are almost exactly the same thing.

Could we have ``GetVarIdentifierEndSuffix()`` just return null (or ``""``) when there's no string (as opposed to throwing a never-see-this exception), and then completely gut the insides of ``GetVarIdentifierEndsWithSuffix()`` to have it just call ``GetVarIdentifierEndSuffix()`` and return true or false depending on whether a string was returned?

PR #2183, opened 2017-11-29
Properly unlock identifiers when setting them

### LINKED ISSUE #2182: Setting a locked variable does not unlock it
This was likely broken as part of #2156 

```
lock a to 1.
set a to 2.
print(a).
```

will print `1`.

tsholmes:
Fixes #2182 

The issue was that the changes in #2156 made it no longer visit the identifier node, which is where the auto-unlock code was.

Since this auto-unlock can only happen during a set, I moved the unlock code into `ProcessSetOperation` so it doesn't have to infer wether it needs to unlock based on the global state variables.

### DIFF
--- a/kerboscript_tests/integration/lock.ks
+++ b/kerboscript_tests/integration/lock.ks
@@ -8,3 +8,6 @@ print(c).
 
 set b to 3.
 print(c).
+
+set c to 5.
+print(c).
--- a/kerboscript_tests/integration/suffixes.ks
+++ b/kerboscript_tests/integration/suffixes.ks
@@ -1,0 +1,22 @@
+
+set a to lexicon().
+
+print(a:length).
+
+set a["a"] to 2.
+
+print(a:length).
+
+print(a["a"]).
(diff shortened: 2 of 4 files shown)


Dunbaratu:
This PR breaks ``SetSuffix``.

Example:
```
set ship:name to "hello".
```
throws a nullref error.

The opcodes it generates are now just this:
```
interpreter             6:18  0070 @0050   push hello 
interpreter             6:18  0071 @0051   setmember <<--INSTRUCTION POINTER--
```
when they used to be like this:
```
interpreter             1:5   0022 @0002   push $ship 
interpreter             1:10  0023 @0003   push name 
interpreter             1:18  0024 @0004   push hello 
interpreter             1:18  0025 @0005   setmember 
```
The "push $ship" and "push name" steps are getting left out.

I realize this entire thing may become moot after you redesign setmember, so it may be smartest to just leave this PR out and just check that this issue is resolved with the new setmember code.

PR #2185, opened 2017-12-01
Fixes #2178 allowing aborting queued triggers.

### LINKED ISSUE #2178: Unlocking steering doesn't work from gui button onclick callback.
As identified by Warrigal on discord, unlocking steering from a gui button callback appears to unlock for a brief instant (the debug console announces `kOS: FlightControlManager: ToggleFlyByWire: steering False` appropriately and the pitch/roll/yaw inputs appear to stutter towards 0), then it goes back to locked (without an announce on the debug console). From an `on` trigger, inline in the code, etc. `unlock steering.` works as expected. Tested against kOS 1.1.3.2 on kSP 1.3.1 (just kOS and module manager).

Any craft with a CPU appears to be sufficient to reproduce it.

```
clearscreen.
print "Demonstrating issues unlocking steering from a gui button:onclick.".
print "".
print "Hotkeys:".
print "AG8:    Steer On.".
print "AG9:    Steer Off.".
print "LIGHTS: Exit.".

clearguis().
lights off.

// Deligates
set f1 to {lock steering to heading(90,45).}.
set f2 to {unlock steering.}.

// Action groups.
on AG8 {f1(). preserve.}
on AG9 {f2(). preserve.} // Trigger. Works.

set w to GUI(200,200).

w:addlabel("Steering:").

set ba to w:addbutton("Steer On").
set ba:onclick to f1. // Works as expected.

set bb to w:addbutton("onclick={f2().} - broken").
set bb:onclick to {f2().}. // Indirect deligate call from button. Broken.

set bc to w:addbutton("onclick=f2 - broken").
set bc:onclick to f2. // Direct deligate call from button. Broken.

set bd to w:addbutton("onclick={unlock steering.} - broken").
set bd:onclick to {unlock steering.}. // Direct callback. Broken.

set be to w:addbutton("onclick={toggle AG9.} - works").
set be:onclick to {toggle AG9.}. // Indirectly trigger the action group. Works.

set bf to w:addbutton("Exit (lights on).").
set bf:onclick to {lights on.}.

w:show().

wait until lights.
print "Finished. Unlocking steering.".
unlock steering. // Inline. Works.
lights off.
w:dispose().
```

Dunbaratu:
Fixes #2178 allowing aborting queued triggers.

### In a nutshell, the problem is this:

**This is how the interrupts like triggers, callbacks, and the
locked steering are implemented in the system:**

- At the start of a physics tick pass, for each pending trigger
or callback, the CPU pushes a record onto the stack to simulate
as if that routine had just been called from the current spot in
the mainline code.  If there's 3 such triggers, A, B, and C, then
the CPU arranges the stack as if your mainline code had just called
A, which as its first instruction had just called B, which as its first
instruction had just called C.  Thus when it goes to execute the
instructions, it will call the nested routines and pop back to the main
code once they're all done.

^^ The logic to perform the above stack trickery is in ``Cpu.ProcessTriggers()``.

- That system encounters a problem when you want to have one of those
interrupt routines (the GUI callback in this case) cancel one of
the others (the steering trigger) - the stack has already been arranged
to call both routines, then one routine says to cancel the other.
The call to steering is already baked in to the stack by the time the
callback says to cancel it.  Even if the callback removes the trigger,
the steering routine still gets one last call to happen from this one last
scheduled stack call that's already baked in to the stack.  And in the
case of steering, that causes it to re-enable the steeringmanager as
it assigns something in to ``$steering`` again in the routine.

**The fix:**

The fix implemented here is to make it so that when a trigger is removed
by code explicitly trying to remove it (as happens in unlock steering),
it will also search for all pending calls to that trigger that are already
baked into the stack and flag them as cancelled.  Then any trigger that
wishes to allow itself to be cancel-able in this way must execute code
at the start of itself that will skip its body if it sees it's already been
thusly cancelled.  The new OpcodeTestCancelled was added to make this possible.

Right now that's just the steering/throttle/wheelthrottle/wheelsteering locks
that do this.  But in the future we could also do it for our own when/then or
on triggers if we want to make them cancel-able (I marked in the code some
comments to show where this could be done in the future in another PR if we
like.)

### DIFF
--- a/src/kOS.Safe.Test/Opcode/FakeCpu.cs
+++ b/src/kOS.Safe.Test/Opcode/FakeCpu.cs
@@ -55,6 +55,11 @@ namespace kOS.Safe.Test.Opcode
             throw new NotImplementedException();
         }
 
+        public SubroutineContext GetCurrentSubroutineContext()
+        {
+            throw new NotImplementedException();
+        }
+
         public IUserDelegate MakeUserDelegate(int entryPoint, bool withClosure)
         {
             throw new NotImplementedException();
@@ -201,6 +206,16 @@ namespace kOS.Safe.Test.Opcode
         }
 
         public void RemoveTrigger(TriggerInfo trigger)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void CancelCalledTriggers(int triggerFunctionPointer)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void CancelCalledTriggers(TriggerInfo trigger)
         {
             throw new NotImplementedException();
         }
--- a/src/kOS.Safe/Compilation/KS/Compiler.cs
+++ b/src/kOS.Safe/Compilation/KS/Compiler.cs
@@ -454,6 +454,13 @@ namespace kOS.Safe.Compilation.KS
             string triggerIdentifier = "on-" + expressionHash.ToString();
             Trigger triggerObject = context.Triggers.GetTrigger(triggerIdentifier);
 
+            // - - - - - - - - - - - - 
+            // TODO: If we ever implement triggers that can be named and cancelled by name later,
+            // then right here we'd need to add some opcodes that implement the same logic as
+            // the cooked steering triggers use (check OpcodeTestCancelled and premature return
+            // if this trigger has been cancelled elsewhere before we begin running it.)
+            // - - - - - - - - - - - - 
+
             currentCodeSection = triggerObject.Code;
             // Put the old value on top of the stack for equals comparison later:
             AddOpcode(new OpcodePush(triggerObject.OldValueIdentifier));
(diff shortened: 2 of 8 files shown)


Dunbaratu:
Here is one example test script that demonstrates the problem, and shows it fixed (try running it both before and after merging this fix to see the difference).

```
local gui is gui(0).

local steer_button is gui:addbutton("Steer over ocean").
set steer_button:onclick to {
   kuniverse:debuglog("-- steer_button:onclick starting.").
   sas off.
   lock steering to heading(90, 45).
   print "steer_button:onclick activated.".
   kuniverse:debuglog("-- steer_button:onclick ending.").
}.

local unsteer_button is gui:addbutton("unsteer").
set unsteer_button:onclick to {
   kuniverse:debuglog("-- unsteer_button:onclick starting.").
   unlock steering.
   print "steeringmanager:enabled:".
   print steeringmanager:enabled.
   kuniverse:debuglog("-- unsteer_button:onclick ending.").
}.

local throt_button is gui:addbutton("throttle to 75%").
set throt_button:onclick to {
   kuniverse:debuglog("-- throt_button:onclick starting.").
   sas off.
   lock throttle to 0.75.
   print "throt_button:onclick activated.".
   kuniverse:debuglog("-- throt_button:onclick ending.").
}.


local unthrot_button is gui:addbutton("unthrottle").
set unthrot_button:onclick to {
   kuniverse:debuglog("-- unthrot_button:onclick starting.").
   unlock throttle.
   kuniverse:debuglog("-- unthrot_button:onclick ending.").
}.

gui:show().

until false {
   wait 5.
   print steeringmanager:enabled.
}
```

You can also use the example on the associated issue, #2178 .

tsholmes:
Merged in ed613e0536916e0bb6af70684358db8a151f3a7e

PR #2189, opened 2017-12-08
Issue 2118 agx load

### LINKED ISSUE #2118: AGX no longer recognized
Caveat: @SirDiazo hasn't officially updated AGExt to KSP 1.3 yet (although there are reports of it not working on KSP 1.2.2) But I've tried with AGExt recompiled to work with 1.3 (which it does), but kOS won't recognize it at all. not via ADDONS:AGX:AVAILABLE, not via AG11-AG250. I did some cursory poking at the code, but I'm not familiar enough with KSP/Unity internals to suss out the problem. It looks ok(?) to me compared to addons that are working, so I'm just not sure what's up.

KSP: 1.3
kOS: 1.1.2
AGX: 2.2

hvacengi:
Fixes #2118

ActionGroupsExtendedAPI.cs
* Update type lookup to use the same logic as the IR and KAC wrappers
* The old method was returning null despite the presence of both the
assembly and the class

### DIFF
--- a/src/kOS/AddOns/ActionGroupsExtended/ActionGroupsExtendedAPI.cs
+++ b/src/kOS/AddOns/ActionGroupsExtended/ActionGroupsExtendedAPI.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Reflection;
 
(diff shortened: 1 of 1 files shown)


Dunbaratu:
This is really weird.  From reading the code I don't understand what the case is in which the new code works that the old code shouldn't have worked also.

Issue #2214, opened 2018-01-08
ADDONS:TR:PLANNEDVEC is broken


Efratror:
When using this code:

`IF ADDONS:TR:AVALABLE { 
     IF ADDONS:TR:HASINPACT { 
         PRINT ADDONS:TR:PLANNEDVEC.}
     }
}`

you get an exeption: 
> Exception has been thrown by the target of an invocation.

and this log entry:

> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.NullReferenceException: Object reference not set to an instance of an object
  at Trajectories.DescentProfile.GetAngleOfAttack (.CelestialBody body, Vector3d position, Vector3d velocity) [0x00000] in <filename unknown>:0 
  at Trajectories.NavBallOverlay.GetPlannedDirection () [0x00000] in <filename unknown>:0 
  at Trajectories.API.plannedDirection () [0x00000] in <filename unknown>:0 
  at (wrapper managed-to-native) System.Reflection.MonoMethod:InternalInvoke (object,object[],System.Exception&)
  at System.Reflection.MonoMethod.Invoke (System.Object obj, BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00000] in <filename unknown>:0 
  --- End of inner exception stack trace ---
  at System.Reflection.MonoMethod.Invoke (System.Object obj, BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00000] in <filename unknown>:0 
  at System.Reflection.MethodBase.Invoke (System.Object obj, System.Object[] parameters) [0x00000] in <filename unknown>:0 
  at kOS.AddOns.TrajectoriesAddon.TRWrapper.PlannedDirection () [0x00000] in <filename unknown>:0 
  at kOS.AddOns.TrajectoriesAddon.Addon.PlannedVector () [0x00000] in <filename unknown>:0 
  at kOS.Safe.Encapsulation.Suffixes.Suffix`1[kOS.Suffixed.Vector].Get () [0x00000] in <filename unknown>:0 
  at kOS.Safe.Encapsulation.Structure.GetSuffix (System.String suffixName) [0x00000] in <filename unknown>:0 
  at kOS.Safe.Compilation.OpcodeGetMember.Execute (ICpu cpu) [0x00000] in <filename unknown>:0 
[...]

PR #2220, opened 2018-01-26
Changes to support the Trajectories v2.0.0 to v2.2.0 API

PiezPiedPy:
Hi, I'm a Collaborator for the Trajectories mod and we have made changes to the API which will be included in the soon to be released v2.2.0

https://github.com/neuoy/KSPTrajectories/blob/master/Plugin/API.cs

I thought it would be wise to provide you with a pull-request, with changes needed to help stop any problems arising in future releases of kOS, when used with Trajectories.

Changes for v2.0.0 are:
* Changes to Trajectories addon classes to support the v2.0.0 release of Trajectories.
* New API function added  `HasTarget`.

Changes for v2.2.0 are:
* Changes to Trajectories addon classes to support the v2.2.0 release of Trajectories.
* New API functions added `TimeTillImpact`.
* For the Descent Profile `Prograde` and `Retrograde`.
* And some for version checking `GetVersion` `IsVerTwo` `IsVerTwoTwo`.

 There are also some changes that you need to make regarding Trajectories that are not correct in the kOS Documentation:
* `SetTarget` can be used regardless of a predicted trajectory calculated or not and thus a check for `HasImpact` is not required but an active vessel check is still required.
* `PlannedVector` and `CorrectedVector` require a valid target to be set before calling otherwise they will throw an exception, the Trajectories:API methods `PlannedDirection` and `CorrectedDirection` will return a null in such cases.


PiezPiedPy:
Issue #2214 should also be fixed with the Trajectories v2.0.0 release

Dunbaratu:
Will these proposed changes *only* work with Trajectories v2.0.0, or will they also work with the current Trajectories release?  (i.e. if we merge this PR, does that mean we should wait for Trajectories v2.0.0 before we let the public see this PR?)

PiezPiedPy:
@Dunbaratu 
Sorry I have not got back to you sooner, GitHub did not notify me of your post. 😠 

The changes in this PR will only work with Trajectories v2 which I'm going to release later Today.

Also regrettably Trajectories v2 will not work with the current release of kOS due to the API method name changes that where made 😞

parkerbossier:
I can confirm that Trajectories is working. I've got this PR locally merged into master [06ef52a](https://github.com/KSP-KOS/KOS/commit/06ef52ac6fefafa94c9981e4ee4e5246e39b4c35), and I correctly see `ADDONS:TR:AVAILABLE` and `ADDONS:TR:HASIMPACT`.

src/kOS/AddOns/Trajectories/TRWrapper.cs
```diff
@@ -38,35 +39,42 @@ private static void init()
                 wrapped = false;
                 return;
             }
-            trGetImpactPosition = trajectoriesAPIType.GetMethod("getImpactPosition");
+            trGetImpactPosition = trajectoriesAPIType.GetMethod("GetImpactPosition");
             if (trGetImpactPosition == null)
             {
```
Dunbaratu: I'd like to attempt to make this work regardless of which version of Trajectories is installed.
Would it be a bad thing if I changed this to look for *both* spellings and use whichever one it finds?
That will make life easier for me later if I choose to make a back-port for RO compatibility.

PiezPiedPy: Sounds like a good idea, only bad thing although not a big issue is anyone using an older Trajectories version will not gain the error checking in the newer API.
With the new method `HasTarget` you could just return a message stating something like, 'This function only available in Trajectories v2.x.x'
On a side note I've also created a branch of Trajectoriesv2.x.x for KSP1.3 (not built a release yet but tests have been positive so a release is imminent).

Thinking about it I am going to add a new method to the Trajectories API which returns its version number.

I can make the changes later today if you wish and add them to this PR.

Dunbaratu: I don't think older users will get the error checking, but at least it would still work as well as it did before before the error checking was present, i.e. using new kOS with old Trajectories wouldn't *break*.  By the way I was looking at this myself overnight and was going to send a PR back to you on top of your PR.

PiezPiedPy: I'll wait till you send your PR. 😉 
you could use the new API method to check the version if you wish, I've called it `GetVersion` returns a string i.e. `2.1.0`
Also now going to add `GetVersionMajor` , `GetVersionMinor` and `GetVersionPatch` that return an integer value.
Obviously a check for a version earlier than v2.x.x will be just to find if the Trajectories addon is available and if `GetVersion` exists.
Commit will be later today on Trajectories/master I'll let you know when I commit so you can have a look at the code.

Dunbaratu:
I had a few questions about the API but it's probably faster to PR back to you what I have and let you handle it than keep asking questions and doing it myself.  Let me make the PR.  It's not big.

src/kOS/AddOns/Trajectories/TRWrapper.cs
Dunbaratu: I had a few questions about the API but it's probably faster to PR back to you what I have and let you handle it than keep asking questions and doing it myself. Let me make the PR. It's not big.

Dunbaratu: So HasTarget didn't exist in older Trajectories versions?  That answers one of the questions I had (did you just happen to add the suffix to kOS it because it was forgotten before, or because it truly was new).

PiezPiedPy: `HasTarget` is new to v2.x.x and does not exist in the older versions.

Dunbaratu: Is your branch "Trajectories-v2.0.0-API-update" based on an older version of our ``develop``?  When I try to PR back to you I see a very big list of diffs, far more than the few I made.

Dunbaratu: The PR to your branch has been made.

PiezPiedPy: Thanks, will have a look soon 😉

PiezPiedPy:
@Dunbaratu Hi 
Can you tell me what return type I need to use for a Double with suffixes, I've used DoubleValue but it seems it is not defined, I tried to look but got lost lol. 
`AddSuffix("TIMETILLIMPACT", new Suffix<DoubleValue>(TimeTillImpact, "Remaining time until Impact in seconds."));`

hvacengi:
@PiezPiedPy `ScalarValue.Create(val)` to create any instance of a scalar.  Internally we treat doubles and scalars differently, but they identical to users, therefore the constructor is protected.

Dunbaratu:
@PiezPiedPy, basically in all these sorts of ``AddSuffix`` lines in the code, suffixes that return doubles or ints should be implemented as ``new Suffix<ScalarValue>(...)`` rather than ``<ScalarDoubleValue>`` or ``<ScalarIntValue>``.  Even when you know for sure that it will definitely be a double or definitely be an int, still package it as the generic type "ScalarValue" for the sake of the script reading it.

PiezPiedPy:
Is there any way to simulate a C# property { get; set; } with a single Suffix rather than having to define two Suffixes.

This is what I have done for the new API property `ProgradeEntry  { get; set; }`
**_get_**
` AddSuffix("ISPROGRADEENTRY", new Suffix<BooleanValue>(IsProgradeEntry, "Check whether the descent profile is set to Prograde."));`

**_set_**
` AddSuffix("PROGRADEENTRY", new OneArgsSuffix<BooleanValue>(ProgradeEntry, "Sets the descent profile to Prograde."));`

Also done the same for `RetrogradeEntry`

PiezPiedPy:
I'm all done here, also added what was needed for the v2.2.0 API additions.
* New API functions added `TimeTillImpact`.
* For the Descent Profile `IsPrograde` `IsRetrograde` `SetPrograde` `SetRetrograde`.
* And some for version checking `GetVersion` `IsVerTwo` `IsVerTwoTwo`.

I also checked that all versions of Trajectories will work with this merge. 😉 
There will be a release of v2.2.0 for KSP 1.4.2 and 1.3.1 very soon.

https://github.com/neuoy/KSPTrajectories/blob/master/Plugin/API.cs if you need/want to have a nose about.

Dunbaratu:
@PiezPiedPy  - Look for places in our code where the phrase ``new SetSuffix()`` appears.  That should show you what you want, a suffix that can both get and set at the same time.   (We don't have set-only suffixes, so the type ``SetSuffix`` presumes the suffix is *also* a getter even though the name of the type doesn't phrase it that way.)

The constructor typically looks like this:

SetSuffix<Type>( *getter delegate*, *setter delegate*, "text description")

(There's also a ``ClampSetSuffix`` where it will clamp the user provided value to a range before the setter sees it.)

Would you like to use this to combine ``IsPrograde`` and ``SetPrograde`` together into one suffix?  It sounds like this was what you wanted but couldn't find the way to do it.  (Same for ``IsRetrograde`` and ``SetRetrograde``.)

On another note, this does make me feel a lack in kerboscript is the lack of an enum type.  Because obviously you don't want *both* retrograde *and* prograde to be true at the same time, so having two independent Booleans for them seems wrong, but I admit there isn't really a good way around it without adding enums to kerboscript.

PiezPiedPy:
@Dunbaratu  Thanks for the help, I will have a play and see what happens, hopefully with the info I can successfully combine `IsPrograde` and `SetPrograde` into one suffix. (Same with `IsRetrograde` and `SetRetrograde`.)

PiezPiedPy:
Finally done, you can merge whenever you feel the urge 😉

Dunbaratu:
As I was writing the docs for this I realized there seems to be a lack of an important feature in the API:

Once a script calls ``Addons:Trajectories:SETTARGET()`` I can't figure out how it could un-set it again.

I was just going to add a suffix myself to this PR that does it.  But when I looked at the Trajectories code on Github, I can't find anything in ``Trajectories.API`` that would null out the ``Trajectories.Target.Body`` or ``Trajectories.Target.WorldPosition``, which is I think how you unset the target from what I can see.

Of course, I *could* bypass ``Trajectoreis.API`` and just call ``Trajectories.Target.Set()`` directly, but that feels like it would defeat the point of having the API class.

Dunbaratu:
There's a new PR to your branch you can look at.  Sorry, for constantly throwing this back - I swear I'm ready to merge it now, if you approve of the edits I did to the user docs and verify that I didn't say anything wrong in them.

PiezPiedPy:
@Dunbaratu Checked your PR and the docs look fine so I've merged your PR.

Re: The clearing of the Target, well spotted I missed that one 🙄. I will add it onto my To-do list and include it when I code the methods for setting the Descent Profile via the API, I wont be including them in this PR though but I will create a new one when the time comes.

Dunbaratu:
@PiezPiedPy - There's a few things to add to maybe the next update to the API, but they shouldn't prevent the merging of this PR, so I made a separate issue for them,  #2263.
If you are planning to add them to your TODO list, you can track them at that issue.

Dunbaratu:
Also fixes #2254 which is a duplicate issue.

PR #2238, opened 2018-03-05
triggers from delegate with bound args - fixes #2236

### LINKED ISSUE #2236: (GUI) triggers do not work with delegate with bound args
**EDIT: was able to make smaller example, after I looked in to the code
and I think that `CPU.AddTrigger` should merge in `PreBoundArgs`**
```
local wnd is gui(0).
local btn is wnd:addButton("TEST").
local function click {
	parameter text.
	print text.
}
set btn:onClick to click@:bind("TEST").
wnd:show().
wait until btn:takePress.
wnd:dispose().
```

**ORIGINAL:**
```
local wnd is gui(200).
local function init {
	parameter item.
	parameter width is 0.
	parameter stretch to true.
	parameter align is "center".
	set item:style:align to align.
	set item:style:width to width.
	if width = 0 set item:style:hstretch to stretch.
	return item.
}
local function button {
	parameter wnd.
	parameter text.
	parameter width is 0.
	parameter stretch to true.
	parameter align is "center".
	return init(wnd:addButton(text), width, stretch, align).
}
local function radio {
	parameter lst.
	parameter btn.
	lst:add(btn).
	set btn:toggle to true.
	return btn.
}
local dir is list().
local row is wnd:addHLayout().
local dirPro is radio(dir,button(row,"Prograde",100)).
local dirRet is radio(dir,button(row,"Retrograde",100)).
local row is wnd:addHLayout().
local dirUnl is radio(dir,button(row,"UNLOCK",100)).
local done is button(row,"CLOSE",100).

local function uncheck {
	parameter lst.
	parameter except.
	for btn in lst if btn <> except set btn:pressed to false.
}
local function check {
	parameter lst.
	parameter btn.
	uncheck(list, btn).
	set btn:pressed to true.
}
local function onToggle {
	parameter lst.
	parameter def.
	parameter btn.
	parameter fn.
	parameter on.
	if on { uncheck(lst,btn). fn(). }
	else {
		for b in lst if b:pressed return.
		set def:pressed to true.
	}
}
local function bindToggle {
	parameter btn.
	parameter fn1.
	parameter fn2.
	set btn:onToggle to { parameter on. fn1(btn,fn2,on). }.
	return btn.
}
local function bindToggle2 {
	parameter btn.
	parameter fn1.
	parameter fn2.
	set btn:onToggle to fn1:bind(btn,fn2).
	return btn.
}
local dirToggle is onToggle@:bind(dir,dirUnl).
bindToggle(dirPro, dirToggle, { sas off. lock steering to srfPrograde. }).
bindToggle2(dirRet, dirToggle, { sas off. lock steering to srfRetrograde. }).
bindToggle(dirUnl, dirToggle, { unlock steering. sas on. }).
set dirUnl:pressed to true.

wnd:show().
wait until done:takePress.
wnd:dispose().
unlock all.
```

Try pressing `Prograde` and `UNLOCK` and that works, because these two are created by `bindToggle`, but then try `Retrograde` which is bound by `bindToggle2`, which should work the same... or not?

`local dirToggle is onToggle@:bind(dir,dirUnl).` should bind parameters `lst` and `def` leaving `btn,fn,on`. Then `bindToggle` should bind `btn` and `fn` leaving `on` for the call. `bindToggle` and `bindToggle2` should do the same thing... or not?

firda-cze:
I have chosen to fix `CPU.AddTrigger` because this seems to be the point where `UserDelegate` gets converted to `EntryPoint` and `PreBoundArgs` were lost. Now passes the test with simple button:

```
local wnd is gui(0).
local btn is wnd:addButton("TEST").
local function click {
	parameter text.
	print text.
}
set btn:onClick to click@:bind("TEST").
wnd:show().
```
fixes #2236

P.S.: I seem to have different formatting options. Would it be possible to add [`.editorconfig`](http://editorconfig.org/)?

### DIFF
--- a/src/kOS.Safe/Encapsulation/KOSDelegate.cs
+++ b/src/kOS.Safe/Encapsulation/KOSDelegate.cs
@@ -47,6 +47,21 @@ namespace kOS.Safe.Encapsulation
             PreBoundArgs.Add(arg);
         }
 
+		/// <summary>
+		/// Get list of pre-bound args and provided args.
+		/// Used by CPU.AddTrigger.
+		/// </summary>
+		/// <param name="args">Arguments to add to returned list</param>
+		/// <returns>PreBoundArgs + args (or args if PreBoundArgs.Count = 0)</returns>
+		public List<Structure> GetMergedArgs(List<Structure> args)
+		{
+			if (PreBoundArgs.Count == 0) return args;
+			var merged = new List<Structure>(PreBoundArgs.Count + args.Count);
+			merged.AddRange(PreBoundArgs);
+			merged.AddRange(args);
+			return merged;
+		}
+
         public Structure CallPassingArgs(params Structure[] args)
         {
             if (Cpu == null)
--- a/src/kOS.Safe/Execution/CPU.cs
+++ b/src/kOS.Safe/Execution/CPU.cs
@@ -1136,7 +1136,7 @@ namespace kOS.Safe.Execution
         {
             if (del.ProgContext != currentContext)
                 return null;
-            TriggerInfo callbackRef = new TriggerInfo(currentContext, del.EntryPoint, del.Closure, args);
+            TriggerInfo callbackRef = new TriggerInfo(currentContext, del.EntryPoint, del.Closure, del.GetMergedArgs(args));
             currentContext.AddPendingTrigger(callbackRef);
             return callbackRef;
         }


firda-cze:
Would it be possible to add .editorconfig maybe like this?
```
[*.cs]
indent_style = space
indent_size = 4
charset = utf-8
csharp_space_after_keywords_in_control_flow_statements = true
csharp_space_between_method_declaration_parameter_list_parentheses = false
csharp_space_between_method_call_parameter_list_parentheses = false
```
or whatever you prefer (I prefer tabs - ts=4, or two spaces. Four spaces are a bit odd, but use whatever you like, that .editorconfig should make it easier for anybody using VS or anything supporting it and I made it according to what the current style appears to be).

..and add `.vs/` to `.gitignore`
Thank you

Dunbaratu:
I'm about to test this PR and probably merge it, but as for the visual studio settings, that would be a better question for @hvacengi than me.  I don't use Visual Studio so I can't really verify what those settings do.

Dunbaratu:
This PR seems like a good solution.  Good work, @firda-cze.  Ideally, it would probably be better to fix ``TriggerInfo`` so it is capable of understanding all the information about a delegate (including its bound args) rather than only knowing the entry point and args. 

But given the limitation that ``TriggerInfo`` doesn't have that information, this is the best way to do it.

Dunbaratu:
Merged, but first I had to strip some hard tab chars and replace them with equivalent spaces to match our standard.  The assumption that you are viewing the ascii text in a system that uses 4-char tabs isn't something you can always do in every tool you might use with the source code (especially as 8 char tabs is the more common standard).  If someone else views the code in a viewer with different tab sizes than you used, it looks all wrong to them as the tabbed lines don't match up with the spaced lines.  That's why tabs have to be a sort of all-or-nothing affair in the source code from day 1.  Either you have to say it's wrong to use tabs, or say it's wrong to use spaces.  You can't really mix and match the two styles in the same project.

Dunbaratu:
@firda-cze I delayed merging this a while because I was investigating a problem with delegates that showed up when I ran ``kerboscript_tests/delegatetest6.ks`` after merging this - but on further inspection that problem pre-dates your PR and had nothing to do with it - it happens even when I back out your PR, so it shouldn't prevent your PR.   (There seems to be some memory that's not orphaning properly in closures once the program ends and the closure is thrown away.  But given how indirect managed memory systems' relationship with the actual OS memory footprint of the process is, it's a royal PITA to tell if the problem is "real" or not.)

firda-cze:
@Dunbaratu I also use vim, but you can add `?ts=4` to any github url and you will see four spaces where tabs are.
I have noticed you use spaces and that was the reason I suggested using `.editorconfig`. I will not change my global VS settings, because that would break a lot of my projects, but that `.editorconfig` (especially the first three lines) will force spaces for KOS (or any solution/project having it).

And debate about what is best can be endless and pointless, so, I will only state my own opinion: four tabs are best for coding, it is just faster for navigation, easier and faster editing. But I can understand that some editors have problems with it, so, I accept two spaces as reasonable alternative (those two are mainstream, ts=4 or sw=2). Four spaces are cumbersome for me, but I can use it if you add the `.editorconfig` (and I added it for myself for future edits not to break your tabs/spaces).

firda-cze:
@Dunbaratu The *problem* with `kerboscript_tests/delegatetest6.ks` is probably not real. The memory drops when I force `GC.Collect()` in my tester, stacks are empty, it only appears to be the lack of full collection.

Issue #1119, opened 2015-08-03
Can't query maxthrust of engines that are off


burgled:
Maxthrust, availablethrust, their variants, and fuelflow return 0 if the engine is off. 
ISP also returns 0 but sealevelISP, vacuumISP and ISPat seem to work fine. Still, leaves no way to calculate thrust.

This is particularly a problem with solid rocket motors as you can't do the workaround of locking the throttle to 0, activating the engine to peek at thrust, and shutting it down when you're done.

hvacengi:
At the time I updated it for 1.x, we threw around the idea of making max thrust be completely independent of throttle and activated state.  Fuel flow and available thrust should still be dependent on firing state, but think it might make sense to modify it for max thrust.  @Dunbaratu @erendrake your thoughts?  I don't know if this would truly qualify as a breaking change, but it would change established practice.

Basic0:
I would take activation into account (or, perhaps, add a "max stage thrust").

What's the rationale behind typing it to throttle?

Dunbaratu:
I'd be in favor of making an additional suffix, to be used when you want to know the hypothetical thrust you'd get if the engine *was* turned on, rather than breaking all the scripts that use `maxthrust` to indicate that it's time to stage.

Patrykz94:
I think adding something like `MAXTHEORETICALTHRUST` and `MAXTHEORETICALTHRUSTAT(pressure)` would be a good option. I'm sure that `MAXFUELFLOW` would find it's uses as well.

And another suggestion from me would be to have a way to check what resources the engine consumes and in what ratios. This would be particularly useful in Realism Overhaul as there's lots of fuel types that are being used.

zegkljan:
I would also like to see some suffixes like @Patrykz94 mentioned. It makes so much sense - these numbers are available in the VAB/SPH anyway, so why not make them available in kOS as well?

My story: I wanted to make some non-kOS tool that calculates some launch-related data from a json that the kOS script running on the vessel dumps, and then tells the vessel the calculated stuff again through a json. At about 3 am I had everything ready and wanted to just test it. You can imagine the (almost) infinite disappointment finding out that the maximum thrust data were 0 and there was no way to get the real data unless I activate the engine(s).

roxtonro:
I'd like to respectfully request some traction on this issue.
[...]

Issue #551, opened 2015-02-08
Expose 'dtsg' stage number as a Part suffix


geoffromer:
The Part:STAGE suffix appears to expose what .craft files call 'istg'. However, there's another stage numbering scheme, stored in the 'dstg' field of a .craft file, which would be very useful to expose as well. According to [this thread](http://forum.kerbalspaceprogram.com/threads/56689-Any-doco-on-the-craft-file-format), dstg represents the decoupling structure of the ship, independent of how the decouplers are organized into stages in the UI: if a decoupler has dstg=N, the part 'above' it has dstg=N-1, and the part 'below' it has dstg=N+1. Confusingly, Part:STAGE already exposes this information for all parts _except_ the ones in the staging list, because KSP defines istg to be equal to dstg for parts without staging behavior.

This would be really valuable because it would enable kOS scripts to compute the properties of future stages, e.g. their mass and their delta-v (at least in the absence of crossfeed shenanigans).

firda-cze:
What would that be good for? I found this while trying to write staging script, that can find out which engines can be staged right now and which not (e.g. recoverable boosters that flameout before main engine). That at first lookled like needing this, but no, I could use e:parent:stage = stage:number-1 for this and think that the same can be used anywhere you may be tempted to look at dstg. Am I wrong?

firda-cze:
**`Part.separationIndex` seems to be what should be exposed by kOS** under different name, because `:STAGE` is activation number (or same as parent's value).


Well, although `dstg` can probably be correctly identified by searching for decoupler in parent list, it all got quite complicated when I wrote autostaging script and tested a rocket, where I added fuel tank (with cross-feed) on top of booster which was attached to decoupler. Normally, the tank would have same `:STAGE` as the decoupler, but not when attached to booster, which is engine, which has `:STAGE` reflecting the stage it is activated in.

Issue #1908, opened 2016-12-27
stage:liquidfuel returns wrong value


mrantmet:
Hello!

I've ecountered an issue with staging which I think was introduced either in KOS 1.0.2 or 1.0.3. After staging stage:liquidfuel and probably other resources are counted wrong way if (probably) not all the engines of the next stage are ignited. I've not yet figured out all the conditions for the issue to appear but I've managed to localize the probelm to some simple test craft, see attachement.

Steps to reproduce:
1. Launch the craft
2. in kos console type:
until false { print stage:liquidfuel. wait 0.5. clearscreen. }
3. observe sero value - that is fine
4. stage and observe non sero value - also ok
5. stage again and observe 0 value - not ok since the upper stage is full of liquidfuel and the engine is burning it quite successively.

Such ship configuration with sepratrons is used (by me) when i'd like to deorbit the upper stage or simply throw it away from the payload so the bug is quite annoying. Please consider this issue or advice if I'm doing something wrong.

[test.zip](https://github.com/KSP-KOS/KOS/files/673933/test.zip)

Tanks in advance,
mrantmet

hvacengi:
Have you compared this to the stock resources window, with "Stage View" enabled?  Our logic is nearly identical to the logic that Squad uses for their own UI.  That's actually on purpose, because we generally want to return the same value that a user would see in the user interface.

What's unusual in the small amount of testing I've done is that it isn't very consistent.  I have multiple vessels that use sepratrons to separate stages or begin deorbit burns, and haven't noticed this issue myself (I have a delta-v monitoring script that checks these values before every node burn).  I tried the testing script on those vessels and didn't replicate the issue.

Another odd thing is that the internal logic checks for any "active" engine, and I did confirm that your test craft's engine is active when it reads `0`.  I'll play with it a little bit in the debugger, but we may want to wait on publishing the fix until Squad also has a chance to push the update.

unixunion:
Im having this same issue. Is this case dead?

mrantmet:
looks dead, but I guess the problem still persists... was not looking into this since then. probably worth checking if bug is reported to ksp... have you discovered any workarounds btw?

firda-cze:
Isn't that the same problem as #2175 ?
Not sure about this one but that one is caused by stage without decoupler (e.g. last chute-only stage makes stage 1 return zeros).

Issue #11, opened 2014-03-15
discuss: NULL and NaN check ability?


Dunbaratu:
I think the ability to do the following would be useful:

```
if isNULL( expr ) { 
  // stuff
}
if isNaN( expr ) {
  // stuff
}
```

Only if _every_ place in the kOS C# code that could potentially return NaN or NULL to the script manages to check for _every_ case where that could happen and trap it and turn it into a different result would such functions be useless.  (And even then I would argue that doing that trap is the wrong solution as it hides relevant data from the script programmer.  There is a big difference between a valid number versus a NaN and a big difference between a valid object with zeroed-out data versus a NULL).

For example, you try to look at what your next encounter is.  But there is no next encounter.  NULL is the sensible way to report that, but currently returning a NULL to the script leaves the script programmer helpless to do anything to detect it and just means their code will crash when they try using the result.

There is also a difference between "your fuel tanks are empty" versus "you have no fuel tanks", so returning zero for cases where a resource is invalid and missing might not actually be the correct answer (see #5).

jenden0:
Assuming this has not been done already and I just haven't found it, would you guys be open to a pull request that implements isDef and isNull?  I needed isDef for a script I was writing so I implemented it in my local copy of the source, but I'd be happy to clean it up and add the related isNull so that other people can use them.

Dunbaratu:
Be aware that we are in the middle of a massive rewrite to handle local variables, and that affects what isDef would have to do (i.e. search the local vars, then the nextmost scope, and so on up the chain to global.)  How long ago did you get your develop copy ?  If it's more than a few weeks old, chances are it will be operating from the old global-only code and not quite work with the new design.

That being said, I'd be happy to look it over and see what you have.

jenden0:
Yea, its the last stable release so its well before those changes.  I knew that there was a lot of work going on related to functions and didn't think about the impact that would have on scoping.
That being said, if the scoping code at head is close enough to done that its safe to work off of I'd be happy to rewrite my change.
What I've got right now for isDef is pretty trivial though not exactly elegant.  I just created a global function that tries to access a variable then returns true if there's no exception.  In the case of an exception due to undefined variable it returns false.  Presumably as long as there is still some mechanism in the code for retrieving the value of a variable this would probably still work, though I'd kind of like to rewrite it without intentionally causing exceptions.

jenden0:
[...]

PR #2256, opened 2018-03-27
fix #2175 - stage resources when next stage does not contain decoupler

### LINKED ISSUE #2175: stage:resources / stage:liquidfuel display the wrong values
When on a two-stage rocket, activating the engine and running `stage:resources` (or `stage:liquidfuel`) returns a value of 0, even though the current stage has resources. Staging again fixes the issue, but also means the next stage has to fire prematurely.

When a second engine is added to the rocket, the `stage:resources` function works fine. I don't currently know if the issue fixes itself because of the second engine, or the fact that there is now a third stage.

*My KSP Version*: 1.3.1 (bug confirmed on 1.3.0 as well)
*kOS Version*: 1.1.3.2

---

Some images of the bug:
screen shot 2017-11-18 at 7 30 48 pm
screen shot 2017-11-18 at 7 33 01 pm
screen shot 2017-11-18 at 7 51 57 pm
screen shot 2017-11-18 at 7 53 27 pm
screen shot 2017-11-18 at 7 54 54 pm

firda-cze:
Maybe even for #1908

Tested with simple startup rocket - pod + flea + shute, works.

Added another flea and separator - works, resources properly separated between the stages (I use this for stage delta-v calculation which needs to know fuel mass, that is how I noticed it - in almost final stage in something I wanted to return in one piece including the engines, I had chutes in last stage, that caused the problems).

Deserves independent testing and more eyes,
this could potentially break things (although I could not think of any
... except not **identifying all decouplers/separators in the logic**).


firda-cze:
Damn those tabs again, because you still do not have that .editorconfig and I branched from your version.
Will fix that and pull that .editorconfig from my master.

hvacengi:
Well now I'm mildly torn.  I love the idea of fixing this, but I've been rationalizing the error for a while as "we match the UI".  So... is it more important to match the UI, or to be right.  I should also look at whether or not KSP has updated the method they use to check for this in 1.4.x, maybe the resource panel was fixed too.

hvacengi:
When testing, I'm going to take special care to look at the new multi-mount engine plates.  I know that they have been complicating the readouts from KER, and I want to make sure that they get treated as a decoupler the stage before the engine rather than after.

firda-cze:
There is potentially another problem, that exists in both current state and after my fix, see #551
`part.inverseStage` is not what you expect it to be. Try adding tank to booster or fairings and check its `inverseStage` - it is either the stage number of activation of that part (if it can be staged/activated - engine, decoupler and fairings) or a copy from parent, which means, that the tanks will get a copy not from nearest decoupler but from the booster or fairings. It does not pose a risk to the code because you will have the booster already activated, but it can be a problem with fairings if you attach anything to its base and have it staged much later.

In any case, I would like this be useful and not return zeros when it clearly should not. No matter what KSP's panel shows. This is a bit different scenairo - we are giving resources that belong to current stage, not list of resources per stage. I have seen MechJeb or Basic DeltaV sometimes showing correct deltaV for some later stage while current reading zero (which is obviously wrong if you have your engines fired and have thrust, then you have available DeltaV).

BTW: MechJeb is just too powerful, so I don't want to use it, because it's no longer the fun ;) That is why I am currently using DMagic's Basic DeltaV and want to remove that too in the future and rely solely on kOS and my own scripts. (Mainframe or at least editor-bound CPU would be nice.)

firda-cze:
**Warning:** The Small Fuel Tank has stage number from Hammer, not from the decoupler!
***Surprise:*** Long Fuel Tank and Battery did not get it from fairing but from the tiny decoupler inside. Looks like fairings are safe.
**Conclusion:** The code is safe, because you would have to add something to a booster that you do not plan to ignite... which is ridiculous. Ignited booster will be in current stage and thus anything attached to it will be counted.

obrazek

firda-cze:
Tested the engine plates and didn't find a problem.
There was small surprise when I decoupled the leftover plate by another decoupler and the stage was not ready (blinking), had to stage again... but the resource counting was correct in every test I did.

src/kOS/Suffixed/StageValues.cs
```diff
@@ -106,18 +106,34 @@ public void CreatePartSet()
...
+                    nextDecoupler = inverseStage;
+                }
+            }
+            foreach (var part in shared.Vessel.Parts)
+            {
+                if (part.State != PartStates.ACTIVE)
+                    continue;
+                foreach (var crossPart in part.crossfeedPartSet.GetParts())
+                {
+                    if (crossPart.inverseStage >= nextDecoupler)
```
firda-cze: This should probably be `>` and not `>=`

firda-cze:
Something is not quite right. It worked when I had separators in their own stage, but it counts with tanks in one more stage when I put the engines and separators in same stage.

I also don't understand why the (original) code checks the stage on the `crossPart` and not on the `part` itself.

This needs more testing.

src/kOS/Suffixed/StageValues.cs
```diff
@@ -106,18 +106,34 @@ public void CreatePartSet()
...
             {
-                part = parts[i];
-                if (part.State == PartStates.ACTIVE)
+                var inverseStage = part.inverseStage;
+                // Ignore leftover decouplers and/or docks with staging enabled
+                // that are in current stage (thus already activated)
+                if (inverseStage >= currentStage)
+                    continue;
+                // Another part in same stage we already found decoupler in, ignore that
+                if (inverseStage >= nextDecoupler)
```
firda-cze: Damn, highest inverse stage number, not lowest!
`<=` needed here

src/kOS/Suffixed/StageValues.cs
firda-cze: Why cross-feed? Why?
Parts upto next decoupler should do (those to be separated by next decoupler).

firda-cze:
The reason I did not find the bug during testing is probably in the order of parts and that I tested radial decouplers only. Added docks with staging enabled and tested with all kinds of decouplers.

I had problems with EP-18, but that plate is bogus, it didn't separate properly and caused some later stage to flash (not ready)... unreliable stuff, Squad should fix it first.

There is one more thing that crossed my mind while thinking about how it works (and worked before). There will probably be problem with ships built from bottom to top, e.g. probes you add on top of another rocket (with different purpose). It will work if you build it upside down (core first, engine last - on top), but not bottom-up. That problem applies to current state as well, this PR simply does nothing about it. (It may require using `separationIndex` instead of `inverseStage`, not sure.)

Another thing is that it won't count EC and other resources if not consumed by active engine. Alternative would be to add parts that are not dead and have inversStage >= nextDecoupler. Would work the same for fuel, but would produce more parts - including e.g. batteries you may wish to throw off, just anything that is to be decoupled next, anything, not just tanks.

firda-cze:
I am stuck, this makes no sense at all. The problem applies to current code as well.
`separationIndex` seems to work better than `inverseStage` (added `AddSuffix("SEPARATION", new Suffix<ScalarValue>(() => Part.separationIndex));` to `PartValue`), but the Cheetah attached through dock with staging enabled is scrued, `stage` works but `separation` does not. Contradiction, **does not compute**.

obrazek

```
for p in ship:parts
	if p:tag = "" print p:stage+"/"+p:separation+":"+p:title.
	else print p:stage+"/"+p:separation+":"+p:tag.
print " ".
for r in stage:resources if r:name:contains("fuel") print r.
```

firda-cze:
This got bigger than originally anticipated, but it was needed.
- `VesselTarget` now maintains all `PartValue`'s and `StageValues` and uses hooks to invalidate the list. It has part-value cache and `this[global::Part]` operator that must be used to get the `PartValue`s (constructors should never be called directly, except from `VesselTarget`).
- The list is built from root to leafs, each part gets link to `Parent PartValue`, `Children` and `Decoupler`. The last is the decoupler/separator (or dock with staging enabled) that will separate this part when activated. `DecoupledIn` is the stage number when that happens (or `-1` if no decoupler). fixes #551 
- Every `ListValue` is typed and gets `IsReadOnly = true` when the discovery process is finished. Cannot be modified and e.g. `ship:parts:add(stage:nextDecoupler)` will throw new `KOSModifyReadOnly` exception. I have enhanced the interface when working on this `IsReadOnly` feature. You will see `Count()` -> `Count` on many places as side-effect of the change (`ListValue` now implements `IList` interface).
- Two new types were introduced: `DecouplerValue` and `LaunchClampValue`. The first is base class of the second and also base class for `DockingPortValue`. New `ship:decouplers` (and alias `ship:separators`) were added. `stage:nextDecoupler` reads `None` or next decoupler including `LaunchClamp` which comes in handy for staging logic e.g. for **Kerbal X**. All the changes (especially in `StageValues`) are motivated to make staging logic easier and work with asparagus and similar designs. You can now e.g. check what fuel-resources are non-zero at stage activation and stage again when these become zero (works for *boosters* that are normal tanks and engines in Kerbal X - so called asparagus).
- Engines redone as well, `POSSIBLETHRUST` and `POSSIBLETHRUSTAT` added - fixes #1119 
- **Multi-Mode Engines need more tests!**
- Added `StringValue:Format`, can be reverted. Needed that during testing for some printouts.
- Used `RegisterInitializer` on many places. That should speed things up a bit. I plan to rewrite suffixes completely (make the dictionaries static, initialized once, not in every instance).
- `StringValue.None` used where `null` would be used. Plant to address that too (`NothingValue` or maybe reuse `PseudoNull`). Described here: https://github.com/KSP-KOS/KOS/issues/11#issuecomment-374750371

src/kOS/Binding/MissionSettings.cs
```diff
@@ -10,15 +10,15 @@ namespace kOS.Binding
...
-        private VesselTarget ship;
         private SharedObjects sharedObj;
+        private VesselTarget ship => VesselTarget.CreateOrGetExisting(sharedObj);
 
         public override void AddTo(SharedObjects shared)
         {
             sharedObj = shared;
 
             shared.BindingMgr.AddGetter("CORE", () => new Core((kOSProcessor)shared.Processor, shared));
-            shared.BindingMgr.AddGetter("SHIP", () => ship ?? (ship = VesselTarget.CreateOrGetExisting(shared)));
```
firda-cze: This was never counted (no ship.LinkCount++ after `??`) but probably never even executed, thanks to `Update` (ship assigned there).

src/kOS/Binding/MissionSettings.cs
```diff
@@ -84,32 +84,10 @@ public override void AddTo(SharedObjects shared)
...
-                ship = VesselTarget.CreateOrGetExisting(sharedObj);
-                ship.LinkCount++;
-            }
-            else if (ship.Vessel == null)
-            {
-                ship.LinkCount--;
-                ship = VesselTarget.CreateOrGetExisting(sharedObj);
-                ship.LinkCount++;
-            }
-            else if (!ship.Vessel.id.Equals(sharedObj.Vessel.id))
```
firda-cze: This probably handled docking (when the vessel becomes part of station).
Keeping the link for few updates after query by `:SHIP` could still be viable (to prevent re-creation of part-values),
but I would rather move such logic to shared objects (some `KeepVesselTargetAlive()` that would reset some counter which would be changed and checked every tick to remove the reference if it reaches 0 / max).

src/kOS/Suffixed/VesselTarget.Hooks.cs
```diff
@@ -0,0 +1,244 @@
...
+                ProcessorId = processor.KOSCoreId;
+                VesselId = vessel.id;
+            }
+
+            public bool Equals(InstanceKey other) =>
+                ProcessorId == other.ProcessorId && VesselId == other.VesselId;
+            public override bool Equals(object obj) =>
+                obj is InstanceKey key && Equals(key);
+            public override int GetHashCode() =>
+                VesselId.GetHashCode() ^ (3001 * ProcessorId); //3001 is prime number
```
firda-cze: Although default implementation of `Equals` may be good enough for structs (memberwise compare),
the default `GetHashCode` is often not (it can choose just one field, e.g. `VesselId` and completely ignore the reset, thus producing collisions in `Dictionary` degrading performance). XOR over fields multiplied by some bigger prime most often produces good hash.

src/kOS/Suffixed/Part/PartValueFactory.cs
```diff
@@ -6,45 +6,13 @@ namespace kOS.Suffixed.Part
...
 
-        public static ListValue<PartValue> ConstructGeneric(IEnumerable<global::Part> parts, SharedObjects shared)
-        {
-            var partList = parts.Select(part => Construct(part, shared)).ToList();
-            return ListValue<PartValue>.CreateList(partList);
-        }
+        public static ListValue<PartValue> ConstructGeneric(IEnumerable<global::Part> parts, SharedObjects shared) =>
+            ListValue<PartValue>.CreateList(parts.Select(part => Construct(part, shared)));
 
-        public static PartValue Construct(global::Part part, SharedObjects shared)
```
firda-cze: This logic (`PartValueFactory.Construct`) was moved to `VesselTarget.Parts:ConstructPart`.
The whole file is now just small set of tiny helpers that could be moved to `VesselTarget.Parts` as well.

src/kOS/Suffixed/VesselTarget.Parts.cs
```diff
@@ -0,0 +1,392 @@
...
+                self = new PartValue(Shared, part, parent, decoupler);
+            if (rootPart == null)
+                rootPart = self;
+            partCache[part] = self;
+            allParts.Add(self);
+            foreach (var child in part.children)
+                ConstructPart(child, self, decoupler);
+            self.Children.IsReadOnly = true;
+        }
+
```
firda-cze: The rest of this file is unchanged cut from original.

src/kOS/Suffixed/Part/PartValueFactory.cs
Dunbaratu: Was there a reason for using `partial class` like this and defining one class in multiple files?  Did it solve a problem or was it just an aesthetic preference?  (Full disclosure, my aesthetic preference is against using ``partial class``.  When asking "where do I look to see how this class works?" I like getting one answer, not three.)

firda-cze: Aesthetic/navigation. The file was just too big for effective navigation. I found myself scrolling a lot, splitting the window to see parts of the file at once... that is usually a sign to consider splitting it somehow, `partial class` is one possibility, creating helper classes is another (but often harder to execute).

I can merge it back to single file, if you don't like it. I separated it because it felt like three different parts that just happen to live inside single class, but has almost nothing to do with each other.

Dunbaratu: I'd feel different about it if the syntax of ``partial class`` required some kind of header statement that mentioned what the rest of the class is (i.e. "this class is defined as the union of A, B, and C.  Now you can go look up A, B, and C in other files to get the full definition of this class".)  But partial class instead says, "The definition of this class is:  just grep for the places it's mentioned and paste them together in your head.  There is no actual place it's listed.".

That's what makes ``partial class`` a very ugly construct, in my opinion.

Dunbaratu: I'll see what @hvacengi thinks about it.  Part of my resistance is that we use ``partial class`` nowhere else in the entire project.  So it's introducing a new feature to the source files.  But he may have a different opinion than me about it.

firda-cze:
I'd at least leave `Hooks` separated, at least the private class if not the instance cache logic as well (maybe rename the file to `VesselTarget.Instances` or `.InstanceCache` instead of `.Hooks`). And surround the parts with `#region` if not separated to its own file as I did. The rest is mainly orbit-logic, that makes the three parts distinct.

Let me know if you want me to do anything with this PR.

Reminder: I'd focus on

1. `MissionSettings` - the removed reference - maybe add it back somehow. I am a bit worried about references to parts. Maybe add reference to `TargetVessel` into `PartValue` to keep it alive as log as you have reference to any part. It should work as it is, but could probably be improved to prevent collection of VesselTarget of the ship you control by the script and duplication of PartValues if you hold them in a variable. I even wanted to create one instance, but found out that that would break some parts of the code that rely on SharedObjects-VesselTarget link especially for checks and exceptions. I'd like to make `PartValue` unique but that could cause similar problems, this reverted to core-vessel keys in the instance cache.

2. Engines, especially Multi-Mode. Never used them, removed the adapter because it contained lot of unnecessarily duplicate code. `ModleEnginesFX` is sublclass of `ModuleEngines`, thus the adapter is completely unnecessary, replaced it by extension methods (for `ModuleEngines` with work for `ModuleEnginesFX` as well). Implemented those `:POSSIBLETHRUST[AT]` suffixes on the way.

src/kOS/Suffixed/Part/PartValueFactory.cs
firda-cze: Adding &lt;DependentUpon&gt; in .csproj can help, can do that, it joins all the file under one node in VS - like `MainForm.Designer.cs` under `MainForm.cs` - and the naming scheme is obvious: `Class.Part.cs`, no need to search for references, you have all the parts nicely together, just look at it as if it was just one file merged, kind of `#include` with additional header in the parts - using, namespace and class name.

Parts can be separated to its own class (e.g. `VesselParts`) or completely joined with `StageValue` (moving most of the stuff there, possibly even removing `PartFactory`). `VesselCache` can be created as `static class` to hold the instance cache and `private class Hooks` (`VesselTarget` only needs `IDisposable` reference to it, does not really need to know the real class, `InvalidateParts` can be made public, no harm in it, Hooks only need this one method) . That would reduce the amount of code in VesselTarget and separate it without introducing `partial class`.

P.S.: I miss `friend` from C++.

firda-cze: Actually, you can have multiple source files (.cpp) for single class as well (some methods here, some there), the difference is that you need header file in C++, you don't in C#.

But general rule is: consider separate classes if you wish to split sources.

hvacengi:
Just wanted to give you an update: I'm working on testing this PR.  I have run into an error, but I haven't yet been able to identify the cause.  It appears to be that there is an issue with caching vessel part information.  After a sub-vessel undocks from the primary (primary remains the "active vessel") a call to `ship:partstagged` throws a "key not found" exception.  It looks to me like either the internal `VesselTarget.CreateOrGetExisting`  is correctly updating but the `ship` variable is not, or vice versa.

I'm trying to focus on functionality first before I look at code structure.  But I may end up having to do both because the changes list is so big that I'm having a hard time identifying everything that really changed.  It may end up as a little bit of time for me to fully review.

firda-cze:
@hvacengi `ship` is no longer a _variable_, it always calls `VesselTarget.CreateOrGetExisting` whenever accessed (MissionSettings.cs:14: `private VesselTarget ship => VesselTarget.CreateOrGetExisting(sharedObj);`). I have changed that as part of removing the link/reference counting. That should not be the cause.

Maybe I have missed some event, can you try adding some logging to `Hooks.OnVesselPartCountChanged` and such? Will try that myself, when I have the time.

Actually, when looking at the code, I spotted this inconsistency that could potantially cause problems:
```
            private void OnVesselPartCountChanged(Vessel v)
            {
                var target = VesselTarget;
                if (target != null && target.Vessel.Equals(v))
                    target.InvalidateParts();
            }
```
vs.
`InstanceKey.Equals => ProcessorId == other.ProcessorId && VesselId == other.VesselId;`

What if `Vessel.Equals` returns false, but `VesselId` does not change?! I have taken this pattern from your code without thinking about it. It just looked allright.

Your description just smells like `InvalidateParts()` was not called when it should.

hvacengi:
Having dug in a little further, I think you may be right.  The code to get tagged parts appears to use the raw vessel rather than the cached parts lists:
```
        private IEnumerable<global::Part> GetRawPartsTitledPattern(Regex partTitlePattern)
        {
            // Get the list of all the parts where the part's GUI title matches the pattern:
            return Vessel.parts.FindAll(
                part => partTitlePattern.IsMatch(part.partInfo.title));
        }

        private ListValue GetPartsTagged(StringValue tagName)
        {
            return PartValueFactory.Construct(GetRawPartsTagged(tagName), Shared);
        }
```
So if the cache isn't invalidated on decoupling, the raw vessel parts list is different from the cached list.  But in theory, if that's the issue, the opposite should be true.  The cached list should in fact have all of the parts from prior to undocking, unless for some reason the part object itself is re-constructed (which I wouldn't put past KSP and/or Unity).

I might as well post the actual error text (with line numbers) to help see what is happening:
```
[LOG 08:35:37.505] System.Collections.Generic.KeyNotFoundException: The given key was not present in the dictionary.
  at System.Collections.Generic.Dictionary`2[Part,kOS.Suffixed.Part.PartValue].get_Item (.Part key) [0x00000] in <filename unknown>:0 
  at kOS.Suffixed.VesselTarget.get_Item (.Part part) [0x00015] in D:\Shared\Users\Brad\Dropbox\Coding\Github\KOS\src\kOS\Suffixed\VesselTarget.Parts.cs:89 
  at kOS.Suffixed.Part.PartValueFactory.Construct (.Part part, kOS.SharedObjects shared) [0x00000] in D:\Shared\Users\Brad\Dropbox\Coding\Github\KOS\src\kOS\Suffixed\Part\PartValueFactory.cs:16 
  at kOS.Suffixed.Part.PartValueFactory+<>c__DisplayClass0_0.<Construct>b__0 (.Part part) [0x00000] in D:\Shared\Users\Brad\Dropbox\Coding\Github\KOS\src\kOS\Suffixed\Part\PartValueFactory.cs:10 
  at System.Linq.Enumerable+<CreateSelectIterator>c__Iterator10`2[Part,kOS.Suffixed.Part.PartValue].MoveNext () [0x00000] in <filename unknown>:0 
  at System.Linq.Enumerable+<CreateSelectIterator>c__Iterator10`2[kOS.Suffixed.Part.PartValue,kOS.Safe.Encapsulation.Structure].MoveNext () [0x00000] in <filename unknown>:0 
  at System.Collections.Generic.List`1[kOS.Safe.Encapsulation.Structure].AddEnumerable (IEnumerable`1 enumerable) [0x00000] in <filename unknown>:0 
  at System.Collections.Generic.List`1[kOS.Safe.Encapsulation.Structure]..ctor (IEnumerable`1 collection) [0x00000] in <filename unknown>:0 
  at kOS.Safe.Encapsulation.ListValue`1[T]..ctor (IEnumerable`1 listValue) [0x00000] in D:\Shared\Users\Brad\Dropbox\Coding\Github\KOS\src\kOS.Safe\Encapsulation\ListValue.cs:21 
  at kOS.Safe.Encapsulation.ListValue..ctor (IEnumerable`1 toCopy) [0x00000] in D:\Shared\Users\Brad\Dropbox\Coding\Github\KOS\src\kOS.Safe\Encapsulation\ListValue.cs:162 
  at kOS.Safe.Encapsulation.ListValue.CreateList[PartValue] (IEnumerable`1 toCopy) [0x00000] in D:\Shared\Users\Brad\Dropbox\Coding\Github\KOS\src\kOS.Safe\Encapsulation\ListValue.cs:169 
  at kOS.Suffixed.Part.PartValueFactory.Construct (IEnumerable`1 parts, kOS.SharedObjects shared) [0x0000d] in D:\Shared\Users\Brad\Dropbox\Coding\Github\KOS\src\kOS\Suffixed\Part\PartValueFactory.cs:10 
  at kOS.Suffixed.VesselTarget.GetPartsTagged (kOS.Safe.Encapsulation.StringValue tagName) [0x00001] in D:\Shared\Users\Brad\Dropbox\Coding\Github\KOS\src\kOS\Suffixed\VesselTarget.Parts.cs:248 
  at kOS.Safe.Encapsulation.Suffixes.OneArgsSuffix`2[kOS.Safe.Encapsulation.ListValue,kOS.Safe.Encapsulation.StringValue].Call (System.Object[] args) [0x00001] in D:\Shared\Users\Brad\Dropbox\Coding\Github\KOS\src\kOS.Safe\Encapsulation\Suffixes\OneArgsSuffix.cs:18 
  at kOS.Safe.Encapsulation.Suffixes.DelegateSuffixResult.Invoke (ICpu cpu) [0x00288] in D:\Shared\Users\Brad\Dropbox\Coding\Github\KOS\src\kOS.Safe\Encapsulation\Suffixes\DelegateSuffixResult.cs:142 
  at kOS.Safe.Compilation.OpcodeCall.StaticExecute (ICpu cpu, Boolean direct, System.Object destination, Boolean calledFromKOSDelegateCall) [0x003a4] in D:\Shared\Users\Brad\Dropbox\Coding\Github\KOS\src\kOS.Safe\Compilation\Opcode.cs:1793 
  at kOS.Safe.Compilation.OpcodeCall.Execute (ICpu cpu) [0x00001] in D:\Shared\Users\Brad\Dropbox\Coding\Github\KOS\src\kOS.Safe\Compilation\Opcode.cs:1639 
  at kOS.Safe.Execution.CPU.ExecuteInstruction (kOS.Safe.Execution.ProgramContext context, Boolean doProfiling) [0x0009b] in D:\Shared\Users\Brad\Dropbox\Coding\Github\KOS\src\kOS.Safe\Execution\CPU.cs:1466 
```

firda-cze:
I will try it more times, independent ships (separate launch), boot scripts that will touch the parts in a loop...
Was not able to reproduce it so far:
obrazek
Launched together (left terminal still reads some printouts from my launch script), separated, docked, undocked, docked again, undocked again... no exception.

src/kOS/Suffixed/Part/PartValueFactory.cs
```diff
@@ -6,45 +6,13 @@ namespace kOS.Suffixed.Part
...
-                }
-            }
-            if (ret != null)
-                return ret;
-
-            // Fallback if none of the above: then just a normal part:
-            return new PartValue(part, shared);
-        }
+        public static PartValue Construct(global::Part part, SharedObjects shared) =>
+            VesselTarget.CreateOrGetExisting(shared)[part];
```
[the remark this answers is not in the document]
firda-cze: `=> shared.VesselTarget[part]` then? I think I have created `SharedObjects.VesselTarget` shortcut later that can be used like this. So, shared is per-processor and per-ship? I thought that it is per-processor only.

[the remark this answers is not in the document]
firda-cze: `VesselTarget.CreateOrGetExisting(part.vessel, shared)[part]`! Right?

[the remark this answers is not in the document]
hvacengi: Yes, my change was to use `VesselTarget.CreateOrGetExisting(part.vessel, shared)[part]`

hvacengi:
@firda-cze I'm continuing to test the pull request.  After getting part lists on vessels other than the 
CPU vessel to work, I've started to look into some of the functionality a little more.  I think you're right.  We probably do want `SharedObjects.VesselTarget` to retain the object across GC collects.  We will want to make sure that `VesselTarget` is updated inside `kOSProcessor.UpdateVessel` too so that the reference doesn't get stale. Which probably means we have to think more about the key used to store `VesselTarget` objects, and make sure that the `VesselId` doesn't change when docking.

Another issue I found was with fuel transfer.  Right now we test the source and target parameter to see if it's a element, a part, or a list.  But because `ship:parts` is now a strongly typed generic `ListValue<PartValue>` instead of the separately defined `ListValue`, the test `is ListValue` returns false.  For the time being I've simply added another test for `is ListValue<PartValue>`, however we will need to make sure there aren't similar assumptions elsewhere in the project.  This isn't really an issue that you created as much as an issue that we had generics defined but never really used them.  I'm not sure that there is a good way to test if an object is a generic type based on `ListValue<>` (I know that you can find out if a type is generic itself, it's the actual tree to determine the generic parent that I am unsure of).

I am continuing to run through my colonization test suite (which tests much of the kOS system) and report/fix errors as I find them.

firda-cze:
https://stackoverflow.com/questions/5461295/using-isassignablefrom-with-open-generic-types
The crutial check is `givenType.IsGenericType && givenType.GetGenericTypeDefinition() == genericType`,
not sure if `IsAssignableFrom` can be used here (on those open-generics), but you can always loop/recurse as in the link.

But I would use marking `interface IListValue: IList, ISuffixed,... {}`.

firda-cze:
Core.cs:50 should maybe become:
```
        private ElementValue GetEelement()
        {
            var vessel = shared.KSPPart.vessel;
            var elList = vessel.PartList("elements", shared);
            var part = VesselTarget.CreateOrGetExisting(vessel, shared)[shared.KSPPart];
            return elList.Cast<ElementValue>().FirstOrDefault(el => el.Parts.Contains(part));
        }
```
Not sure, but this looks safer.

firda-cze:
I would like to create documentation for this, but I have a problem - it does not build:
```
Running Sphinx v1.7.4

Extension error:
Could not import extension ksdomain (exception: cannot import name 'Directive')
```
looking into `ksdomain.py` it probably complains about `from sphinx.util.compat import Directive`
because that was deprecated and removed in sphinx 1.7.

Do I need older version? Which?
Do you plan to upgrade it?

firda-cze:
Nevermind, deleting the line made it build the docs. The output has slightly different font that what is on the web, that that is not a problem. I will try to create the docs with this.

src/kOS/Suffixed/VesselTarget.Parts.cs
```diff
@@ -0,0 +1,392 @@
...
+                    break;
+                }
+                if (module is IStageSeparator)
+                {
+                    if (module is ModuleDockingNode dock)
+                    {
+                        var port = new DockingPortValue(Shared, part, parent, decoupler, dock);
+                        self = port;
+                        dockingPorts.Add(port);
+                        if (dock.stagingEnabled)
```
firda-cze: This would better be rewritten as
```
                        if (!dock.StagingEnabled())
                            break;
                        decoupler = port;
                        decouplers.Add(decoupler);
```
Mostly cosmetic, not changing the logic:
1. Rather `StagingEnabled()` instead of `stagingEnabled`
2. No need to check unchanged `decoupler` later, we can break here.

src/kOS/Suffixed/VesselTarget.Parts.cs
```diff
@@ -0,0 +1,392 @@
...
+                        {
+                            decoupler = port;
+                            decouplers.Add(decoupler);
+                        }
+                    }
+                    else
+                    {
```
firda-cze: I would like to add
```
                        if (!module.StagingEnabled())
                            break;
```
here, because:
1. It now matches heat shields, which is causing unexpected results - `stage:nextDecoupler` can return that heat shield instead of `None`.
2. `:DECOUPLER`/`:NEXTDECOUPLER` should not match any decoupler with staging disabled, not even launch clamps.

The question is:
Should everything with staging disabled be ignored? (that is the proposed change)
Or should the part still get the type but not be couned as decoupler (seen in `:DECOUPLER`/`:NEXTDECOUPLER`)?
The second option would be quite problematic. Is heat shield a decoupler or not?
Is decoupler with disabled staging still a decoupler?
Do you want to see launch clamp with disabled staging as launch clamp?

firda-cze:
Just found out that sepratrons (those tiny boosters) are confusing the staging logic, because their solid fuel is seen as accessible by engines (not sure if you need active boosters, but I had them). The following should fix it - StageValues.CreatePartSet:

```
            // Find all relevant parts that are to be separated by next decoupler/separator/dock
            foreach (var part in shared.Vessel.Parts)
            {
                // Engines only
                if (part.State != PartStates.ACTIVE)
                    continue;
                // All tanks accessible to this engine ...
                foreach (var crossPart in part.crossfeedPartSet.GetParts())
                {
                    var xpart = vesselTarget[crossPart];
                    // ... that are to be separated by next decoupler
                    if (xpart.DecoupledIn >= nextDecoupler
                        // ... but are not boosters
                        && !(xpart is EngineValue))
                    {
                        partHash.Add(crossPart);
                    }
                }
            }
```
Added
```
                        // ... but are not boosters
                        && !(xpart is EngineValue))
```

I'll be testing it for a while.

PR #2258, opened 2018-03-28
Target of inactive vessel

### LINKED ISSUE #2257: Target of inactive ship lost (after switching)
If I `lock steering to target:position` (or `target:position-ship:position` or even `target:position-kUniverse:activeVessel:position`) on one vessel and then switch to the other then the first one, now inactive, starts to steer to different direction. What is the correct way to do this, that works in all the cases (when active and also when inactive)?

firda-cze:
fix #2257
Works fine when switching vessels, I can even control the inactive from a GUI I have created,
I can make it point at active vessel or away from it on a press of a button, nice!

I am still not sure if it will work when I switch to KSC/Tracking and back, but that would require some clever boot script, but I seem to have problems with booting on loading. All ships boot correctly when I approach them, but not when I switch to them from Tracking Station.

### DIFF
--- a/src/kOS/Binding/MissionSettings.cs
+++ b/src/kOS/Binding/MissionSettings.cs
@@ -60,36 +60,24 @@ namespace kOS.Binding
 
             shared.BindingMgr.AddGetter("TARGET", () =>
             {
-                if (shared.Vessel != FlightGlobals.ActiveVessel)
-                {
-                    throw new kOS.Safe.Exceptions.KOSSituationallyInvalidException("TARGET can only be returned for the Active Vessel");
-                }
-                var currentTarget = FlightGlobals.fetch.VesselTarget;
+                var target = shared.Vessel == FlightGlobals.ActiveVessel ?
+                    FlightGlobals.fetch.VesselTarget : shared.Vessel.targetObject;
 
-                var vessel = currentTarget as Vessel;
-                if (vessel != null)
-                {
+                if (target is Vessel vessel)
                     return VesselTarget.CreateOrGetExisting(vessel, shared);
-                }
-                var body = currentTarget as CelestialBody;
-                if (body != null)
-                {
+                if (target is CelestialBody body)
                     return BodyTarget.CreateOrGetExisting(body, shared);
-                }
-                var dockingNode = currentTarget as ModuleDockingNode;
-                if (dockingNode != null)
-                {
+                if (target is ModuleDockingNode dockingNode)
                     return new DockingPortValue(dockingNode, shared);
-                }
 
                 throw new kOS.Safe.Exceptions.KOSSituationallyInvalidException("No TARGET is selected");
             });
 
             shared.BindingMgr.AddGetter("HASTARGET", () =>
             {
-                if (shared.Vessel != FlightGlobals.ActiveVessel) return false;
                 // the ship has a target if the object does not equal null.
-                return FlightGlobals.fetch.VesselTarget != null;
+                return (shared.Vessel == FlightGlobals.ActiveVessel ?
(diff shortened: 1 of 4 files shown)


src/kOS/Binding/MissionSettings.cs
```diff
@@ -60,36 +60,24 @@ public override void AddTo(SharedObjects shared)
 
             shared.BindingMgr.AddGetter("TARGET", () =>
             {
-                if (shared.Vessel != FlightGlobals.ActiveVessel)
-                {
-                    throw new kOS.Safe.Exceptions.KOSSituationallyInvalidException("TARGET can only be returned for the Active Vessel");
-                }
-                var currentTarget = FlightGlobals.fetch.VesselTarget;
+                var target = shared.Vessel == FlightGlobals.ActiveVessel ?
+                    FlightGlobals.fetch.VesselTarget : shared.Vessel.targetObject;
```
firda-cze: It would probably work with just `shared.Vessel.targetObject` but I wanted to be safe and preserve the original without any change.

src/kOS/Binding/MissionSettings.cs
```diff
@@ -60,36 +60,24 @@ public override void AddTo(SharedObjects shared)
...
-                var body = currentTarget as CelestialBody;
-                if (body != null)
-                {
+                if (target is CelestialBody body)
                     return BodyTarget.CreateOrGetExisting(body, shared);
-                }
-                var dockingNode = currentTarget as ModuleDockingNode;
-                if (dockingNode != null)
-                {
+                if (target is ModuleDockingNode dockingNode)
```
hvacengi: How is it that I've been writing C# code for 12 years and I've never seen this syntax?  I admit that entire time has been as a hobyist, but it sure seems like a very useful feature.  I'm a little curious how it compares on performance in terms of allocations, but if it is essentially the same or better I'll gladly stash this away in my toolbox

hvacengi:
I had actually looked into using `targetObject` once before, but it seemed to be a little unreliable.  I want to say the issue was with setting the target, so it's possible that getting the target is perfectly acceptable.

If you can reproduce that tracking station bug please fill out an issue for it.  I don't often switch using the tracking station, but I don't recall ever having issues.

src/kOS/Binding/MissionSettings.cs
Dunbaratu: I've never seen this either.  C# is full of weird obfuscations like this that are IMO just there to be different, not because they're truly useful.  Never in a million years would I look at that line of code and guess what it does if not for seeing the old lines that it replaces in this diff output.  There's nothing in this syntax that visually hints at what it does.  You'd never guess that this is an assignment statement that puts a value into ``dockingNode`` with syntax like this.  You'd definitely be able to guess that there's a cast or type check happening, but not that there's any assignment occurring.

firda-cze: It is one of the newer features added since open-sourcing Roslyn, C#6 or 7, not sure.
They have created many shortcuts I really like (e.g. expression body: `prop => value`),
less typing. Visual Studio is even suggesting such syntax.

firda-cze:
I don't think that changing (assigning to) it would be safe, that is a limitation that shall persist,
but this is still very helpful and mitigates the surprise of not being able to use target after switching.

Documentation should be changed too.

I will thest the booting a bit more, not sure what is causing it, but I am **starting on archive**, that can be relevant. Never boots when I use Tracking Station or KAC, terminal looks like there is no boot at all, but it boots perfectly on rendezvous.

firda-cze:
separated, will do the same with the rest

PR #2269, opened 2018-04-07
Fixes #1942 - Remove bad links from Vector doc page

### LINKED ISSUE #1942: [documentation] Busted links on the vector doc page
```
The vector cross product of two vectors in the order (v1,v2) returning a new Vector:
```
has a link to `http://ksp-kos.github.io/KOS_DOC/math/veccross` which doesn't exist.

and
```
This is the dot product of two vectors returning a scalar number. This is the same as v1 * v2:
```
has a link to `http://ksp-kos.github.io/KOS_DOC/math/vecdot` which does't exist

dhawton:
Reference issue #1942

### DIFF
--- a/doc/source/math/vector.rst
+++ b/doc/source/math/vector.rst
@@ -33,12 +33,12 @@ Structure
     .. list-table:: **Members**
         :widths: 4 2 1 1
         :header-rows: 1
-        
+
         * - Suffix
           - Type
           - Get
           - Set
-          
+
         * - :attr:`X`
           - :ref:`scalar <scalar>`
           - yes
@@ -110,7 +110,7 @@ Structure
     :access: Get only
 
     This creates a unit vector pointing in the same direction as this vector. This is the same effect as multiplying the vector by the scalar ``1 / vec:MAG``.
-    
+
 .. attribute:: Vector:SQRMAGNITUDE
 
     :type: :ref:`scalar <scalar>`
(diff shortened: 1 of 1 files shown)


Dunbaratu:
These were meant to be links that point to wikipedia generic pages about what dot products and cross products and so on are.  We were supposed to have alias terms defined to those pages, and that's what the terms <vecdot>, <vecmult> and so on were meant to be.  Rather than delete the links, I think I'd rather fix them so they work.

Dunbaratu:
Ah forget it - let's just delete the links, as this PR does.

dhawton:
I'll see what I can come up with unless you want to take it over. I have no problems doing it.

Dunbaratu:
I was looking into it myself just now and I'm having a heck of a time trying to find good external links that explain it in ways that are nice for the user new to vectors.  Wikipedia is terrible at teaching anything mathematical because it cares more about being a precise reference than about *teaching*. It gives explanations that only make sense to people who already know the material and all the lingo to begin with and don't need any help.

I tried Wulfram MathWorld and it has a good page for dot product and for vector addition, but it's cross product page is just like the WIkipedia probelm - it only would make sense to someone who already knows the topic anyway.

dhawton:
@Dunbaratu I've found tutorial sites for the concepts and linked to them. Not Wikipedia, but teaching pages.

Dunbaratu:
I decided I liked the explanations on mathinsight.org to the ones on betterexplained.com, so I used them exclusively.  I also added text anchors with "#" so the URLs will jump directly to the right spot on the page instead of to the generic top of the vector_indroduction page.

PR #2291, opened 2018-06-06
Fixes #2228 triggers in anonymous delegates

### LINKED ISSUE #2228: WHEN statement not working in ANONYMOUS functions
First of all, thank you for such a wunderful mod! I've had quite a lot of fun with it recently!

But while extending my library today, I noticed the following code would not print "hi":
```
set f to {
  when true then {
    print "hi".
  }
}.
f().
wait until false.
```
On the other hand, this one does:
```
function f {
  when true then {
    print "hi".
  }
}.
f().
wait until false.
```
And even this does work:
```
function f {
  when true then {
    print "hi".
  }
}.
set x to f@.
x().
wait until false.
```

Is there a reason for this, that I'm just unable to comprehend, or is there something broken?
Thanks a lot for your time and effort!

lucaelin:
Make the compiler visit any node in the parse tree when looking for ON and WHEN.
This fixes #2228

### DIFF
--- a/src/kOS.Safe/Compilation/KS/Compiler.cs
+++ b/src/kOS.Safe/Compilation/KS/Compiler.cs
@@ -422,7 +422,7 @@ namespace kOS.Safe.Compilation.KS
                     PreProcessChildNodes(node);
                     PreProcessWhenStatement(node);
                     break;
-		default:
+                default:
                     PreProcessChildNodes(node);
                     break;
             }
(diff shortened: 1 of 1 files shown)


Dunbaratu:
Although this slightly slows down the compiler because now it has to visit every part of the parse tree to the leaf nodes, the difference is negligible and I cannot find any other way to do it that would be guaranteed to work.  Given that an anonymous delegate is an expression, and an expression can appear inside just about *any* other syntax construct, the only way to be sure to check through all of them is to just check all the way down to the leaf nodes.

PR #2292, opened 2018-06-06
Fixes #2283 Mission Builder boot system

### LINKED ISSUE #2283: DLC's Mission Builder breaks the boot system
Right now, the kOS boot system contains this important step:

(From kOSProcessor.ks, line 463):
```
                // populate it with the boot file, but only if using a new disk and in PRELAUNCH situation:
                if (vessel.situation == Vessel.Situations.PRELAUNCH && path != null && !SafeHouse.Config.StartOnArchive)
                {
```

This is the section of code that copies the boot filename to the local disk when you spawn the vessel on the launchpad or runway.  This is the only place the boot filename gets copied to the vessel.

The problem is that Mission Builder provided a new means to go from the VAB editor to having a launched vessel, that bypasses the launchpad.  Thus this vessel never is in PRELAUNCH state and the boot filename never copies over.

That new way is this:

  - Build a mission.  Make the mission spawn a new hardcoded vessel (rather than one the player is meant to build when playing the mission).
  - The Mission Builder will ask you to go to the VAB to design the craft it will spawn.
  - You use the VAB to pick a boot file for this craft, then save it and go back to the Mission Builder.

Now the mission will spawn that vessel already in orbit, having bypassed the launch step.  This bypasses the loading of the boot file onto the local drive and the vessel can't boot properly.

Given that this section is inside a check like so:
```
            if (HardDisk == null)
            {
```
I am not convinced that PRELAUNCH condition is even needed.  If the vessel was coming from a persistent.sfs file that was already launched and then later went "on rails", it would have a HardDisk already and thus not need to go through this part anyway when that vessel gets re-loaded.

lucaelin:
Don't require a vessel to be in `PRELAUNCH` situation for a bootfile to be copied to the vessels HardDisk.
This fixes #2283

### DIFF
--- a/src/kOS/Module/kOSProcessor.cs
+++ b/src/kOS/Module/kOSProcessor.cs
@@ -460,8 +460,8 @@ namespace kOS.Module
                 }
 
                 var path = BootFilePath;
-                // populate it with the boot file, but only if using a new disk and in PRELAUNCH situation:
-                if (vessel.situation == Vessel.Situations.PRELAUNCH && path != null && !SafeHouse.Config.StartOnArchive)
+                // populate it with the boot file, but only if using a new disk:
+                if (path != null && !SafeHouse.Config.StartOnArchive)
                 {
                     var bootVolumeFile = Archive.Open(BootFilePath) as VolumeFile;
                     if (bootVolumeFile != null)


Dunbaratu:
I tested this and it didn't seem to fix the base problem so I guess it requires more than I thought.

This is the test it fails:

- In Mission Builder, make a tiny mission that only has the one start node and nothing else.  Make that start node spawn a vessel on the launchpad, that you create in the mission builder.  When making the vessel in the VAB, assign a simple boot script to it that just does ``print "IF YOU SEE THIS, I AM RUNNING THE BOOT SCRIPT.".``.
- Test the mission.  It appears to work and the boot script gets copied to the vessel and runs.
- Quit the test and go back to the Mission Builder.
- In the MIssion Builder, change the vessel spawn so it now spawns in orbit instead of on the launchpad.
- Test the mission again.
- Now the boot script does *NOT* copy to the vessel, and does *NOT* run.

So I must have been wrong about assuming it would be fixed just by removing the PRELAUNCH check.  There's got to be something more to it.

Dunbaratu:
I couldn't re-create the problem mentioned above when I tried again later.  I don't know why. 
Whatever the case is, this PR doesn't make the problem any *worse*, so I feel okay merging it.

PR #2293, opened 2018-06-07
Fixes #2290 Optional GUI parameters

### LINKED ISSUE #2290: Make parameters to gui elements optional
Currently methods like `:addlabel` or `:addcheckbox` require one or more arguments to be passed. 
I suggest to have default values for those like `:addlabel("")` and `:addcheckbox("", true)` because their values can usually be dynamically adjusted using `label:text` and `button:pressed`. Whats missing here is a way to get and set a buttons `text` attribute.

The reason I ask for this is because i created a small tool (https://lucaelin.github.io/ksml/) that generates kos gui code from xml-files. It currently requires a special `param:` attribute for elements that have required parameters. I'd like to remove those.

lucaelin:
So this is my first taste of C# so I guess theres room to improve :) Is this a "valid" thing to do? I greatly appreciate any feedback!

### DIFF
--- a/src/kOS/Suffixed/Widget/Box.cs
+++ b/src/kOS/Suffixed/Widget/Box.cs
@@ -34,21 +34,21 @@ namespace kOS.Suffixed.Widget
 
         private void InitializeSuffixes()
         {
-            AddSuffix("ADDLABEL", new OneArgsSuffix<Label, StringValue>(AddLabel));
-            AddSuffix("ADDTEXTFIELD", new OneArgsSuffix<TextField, StringValue>(AddTextField));
-            AddSuffix("ADDBUTTON", new OneArgsSuffix<Button, StringValue>(AddButton));
-            AddSuffix("ADDRADIOBUTTON", new TwoArgsSuffix<Button, StringValue, BooleanValue>(AddRadioButton));
-            AddSuffix("ADDCHECKBOX", new TwoArgsSuffix<Button, StringValue, BooleanValue>(AddCheckbox));
+            AddSuffix("ADDLABEL", new OptionalArgsSuffix<Label>(AddLabel, new Structure[] { new StringValue ("") }));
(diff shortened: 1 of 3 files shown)


src/kOS/Suffixed/Widget/Box.cs
```diff
@@ -34,21 +34,21 @@ public Box(Box parent, LayoutMode mode, WidgetStyle style) : base(parent, style)
...
         {
-            AddSuffix("ADDLABEL", new OneArgsSuffix<Label, StringValue>(AddLabel));
-            AddSuffix("ADDTEXTFIELD", new OneArgsSuffix<TextField, StringValue>(AddTextField));
-            AddSuffix("ADDBUTTON", new OneArgsSuffix<Button, StringValue>(AddButton));
-            AddSuffix("ADDRADIOBUTTON", new TwoArgsSuffix<Button, StringValue, BooleanValue>(AddRadioButton));
-            AddSuffix("ADDCHECKBOX", new TwoArgsSuffix<Button, StringValue, BooleanValue>(AddCheckbox));
+            AddSuffix("ADDLABEL", new OptionalArgsSuffix<Label>(AddLabel, new Structure[] { new StringValue ("") }));
+            AddSuffix("ADDTEXTFIELD", new OptionalArgsSuffix<TextField>(AddTextField, new Structure [] { new StringValue ("") }));
+            AddSuffix("ADDBUTTON", new OptionalArgsSuffix<Button>(AddButton, new Structure [] { new StringValue ("") }));
+            AddSuffix("ADDRADIOBUTTON", new OptionalArgsSuffix<Button>(AddRadioButton, new Structure [] { new ScalarDoubleValue(0), new ScalarDoubleValue (0), new ScalarDoubleValue (1) }));
```
Dunbaratu: This looks like an error (ScalarDoubles?) I'll change it as I work the review.

PR #2296, opened 2018-06-26
Interrupt priorities

### LINKED ISSUE #2239: Delayed onClick/onToggle after set button:pressed

### LINKED ISSUE #2272: Triggers behave unexpectedly (was: "when using locals") because two trigger instances from the same source line can't coexist
I'm trying to create triggers for each science module with a loop, and running into some trouble. The following should print the numbers 1 through 10 (not necessarily in order, I'm not sure if triggers are always checked in order of creation?), but it doesn't. Now, if the use of locals was totally broken in triggers, you might expect that it would print "10" ten times, or maybe do nothing at all. 
Instead, none of those things happen. It just prints "1", once.
```
{
  local trig to false.
  local a to 0.
  until a > 10 {
    local b to a.
    on trig {
      print b.
    }
  }
  set trig to true.
}
```
Here's another case, this time with a `for` loop.
```
set offsets to list(1,2,3,4,5).
for t in offsets {
	local offset to t.
  local utime to time:seconds + offset.
  when time:seconds > utime then {
    print "t: " + t + "  offset: " + offset.
  }
}
wait until 0.
```
This should print something every second for five seconds. At worst, you might expect it to wait five seconds, then print five lines. Instead, it simply prints `t: 5  offset: 1`.

### LINKED ISSUE #2278: Change respawning triggers system (WHEN/ON/LockSteering) into UserDelegate callbacks
Right now triggers use several different systems that are very similar but use slightly different infrastructure, which could be combined into one common system, thus reducing code complexity.  These are the multiple ways that triggers which run every update are handled.

1. Originally the only triggers that run every update were the cooked controls (lock steering, lock throttle, etc).
These triggers used a system for the CPU to re-insert them every update unless they were still running from a previous update.

2. Then we had triggers like WHEN and ON, which also run every update. and were piggybacked onto the same system, but slightly differently.  They use OpcodeReturn to re-insert themselves whenever they exit.

3. Then we added user-definable Delegate callbacks, some of which can be called by other places in our C# code other than the CPU.  These use a different system that allows us to schedule the callback to occur later, then lets the CPU execute it, then detects that it is done on a future Unity update and picks up the resulting value and uses it, then reschedules another call to the Delegate and does it again.

It is possible to actually collapse all these systems together to all use technique (3).  We could make, for example, a RecurringTriggerManager object that schedules executions of the When's and On's, waits to see if they're all done, then schedules calls to them again if they are.  We could make the SteeringManager do the same thing for the lock steering and lock throttle and so on triggers.

What this would buy us is a simplification of some of the weird logic in the CPU and in OpcodeReturn that's there to allow *some* triggers to respawn themselves and other triggers not to.  This weird logic also handles trying to make sure that recurring triggers don't recur fast enough to prevent non-trigger code from ever executing.

With this change, *all* triggers would be one-shot triggers and it would be up to the SteeringManager and/or RecurringTriggerManager to decide to keep scheduling new calls to the one-shot triggers when it sees the old ones finished, rather than that being the job of the CPU to do in its ContinueExecution() logic, or the OpcodeReturn checking if the trigger returns true.

However, as a refactor it has the potential to accidentally invoke places in the code that still expect things to work the old way.  It would have to be carefully checked to ensure everything had switched over.  It's a case where *after* the switch the system would be cleaner, but making that switch could be hairy.

I'm not sure if I want to do this yet, but in case I do, this issue is here as a placeholder for it.  At any rate it's the sort of refactor that shouldn't be attached to anything else (no other bug fixes, no other enhancements) so it can be done as cleanly as possible.

Dunbaratu:
Fixes #2239
Fixes #2272
Fixes #2278

(Not sure if it fixes 2280, so I won't tag it here.)

I believe this is a more complete solution to the problems of these issues:  (Note this includes some of the code that started a branch that was mean to solve those issues).  If this is accepted, it should supplant PR #2277.

To get the idea of the change, I recommend starting by reading the user docs changes in doc/source/general/cpu_hardware.rst before looking at the code.  It should (hopefully) make the code make more sense.

### DIFF
--- a/src/kOS/Suffixed/Widget/Button.cs
+++ b/src/kOS/Suffixed/Widget/Button.cs
@@ -82,7 +82,6 @@ namespace kOS.Suffixed.Widget
             if (UserOnToggle != null)
             {
                 UserOnToggle.TriggerOnNextOpcode(new BooleanValue(pressed));
-
             }
 
             if (parent != null && parent.UserOnRadioChange != null)
--- a/src/kOS/Suffixed/Widget/Button.cs
+++ b/src/kOS/Suffixed/Widget/Button.cs
@@ -81,7 +81,7 @@ namespace kOS.Suffixed.Widget
 
             if (UserOnToggle != null)
             {
-                UserOnToggle.TriggerNextUpdate(new BooleanValue(pressed));
+                UserOnToggle.TriggerOnNextOpcode(new BooleanValue(pressed));
 
             }
 
@@ -100,7 +100,7 @@ namespace kOS.Suffixed.Widget
             // not the button-goes-out state that should auto-activate when it's read:
             if (UserOnClick != null && (IsToggle || pressed))
             {
-                UserOnClick.TriggerNextUpdate();
+                UserOnClick.TriggerOnNextOpcode();
                 if (!IsToggle)
                     causeRelease = true;
             }
--- a/src/kOS/Suffixed/Widget/Button.cs
+++ b/src/kOS/Suffixed/Widget/Button.cs
@@ -81,7 +81,10 @@ namespace kOS.Suffixed.Widget
 
             if (UserOnToggle != null)
             {
-                UserOnToggle.TriggerOnNextOpcode(new BooleanValue(pressed));
+                if (guiCaused)
+                    UserOnToggle.TriggerOnFutureUpdate(InterruptPriority.CallbackOnce, new BooleanValue(pressed));
+                else
+                    UserOnToggle.TriggerOnNextOpcode(InterruptPriority.NoChange, new BooleanValue(pressed));
             }
 
             if (parent != null && parent.UserOnRadioChange != null)
@@ -99,7 +102,11 @@ namespace kOS.Suffixed.Widget
             // not the button-goes-out state that should auto-activate when it's read:
             if (UserOnClick != null && (IsToggle || pressed))
             {
-                UserOnClick.TriggerOnNextOpcode();
+                if (guiCaused)
+                    UserOnClick.TriggerOnNextOpcode(InterruptPriority.CallbackOnce);
+                else
+                    UserOnClick.TriggerOnNextOpcode(InterruptPriority.NoChange);
+
                 if (!IsToggle)
                     causeRelease = true;
             }
--- a/doc/source/general/cpu_hardware.rst
+++ b/doc/source/general/cpu_hardware.rst
@@ -129,51 +129,113 @@ Triggers
     moot, which caused a re-write of most of this section of the
     documentation.
 
-Many of the warnings and cautions mentioned below can really be boiled
-down to this one phrase, which is a good idea to memorize:
+There are multiple things within kerboscript that run "in the background"
+always updating, while the main script continues on. The way these work is
+a bit like a real computer's interrupt handling system, but not *quite*.
 
-*Main-line code gets interrupted by triggers, but triggers don't get
-interrupted by main-line code (or other triggers).*
+Collectively all of these things are called "triggers".
 
-There are multiple things within kerboscript that run "in the background" always updating, while the main script continues on. The way these work is a bit like a real computer's multithreading, but not *quite*. Collectively all of these things are called "triggers".
+Triggers come in these varieties:
 
-Triggers are all of the following:
+.. _recurring_trigger:
 
--  LOCKS which are attached to flight controls (THROTTLE, STEERING,
-   etc), but not other LOCKS.
--  ON condition { some commands }.
--  WHEN condition THEN { some commands }.
+- **Recurring triggers:** Triggers that once they are started keep getting
+  called again and again on a regular basis, until they are made to stop.
+  - LOCKS which are attached to flight controls (THROTTLE, STEERING,
+    etc), but not other LOCKS.
+  - User Delegates assigned to recurrently updating suffixes such as
+    :attr:`VecDraw:VECUPDATER`.
+  - WHEN and ON triggers:
+    - ``WHEN condition THEN { some commands }``
+    - ``ON condition { some commands }``
 
-The way these work is that once per **physics tick**, all these
-trigger routines get run, including those locks that are always
-re-evaluated by the cooked steering, and the ``ON`` and ``WHEN``
-triggers.  (This isn't *quite* true.  The real answer is more
-complex than that - see :ref:`CPU Update Loop <cpu_update_loop>`
-elsewhere on this page).
+.. _callback_once_trigger:
 
-Each of the steering locks behaves like a function that returns
-a value, and is re-called to get the new value for this **physics
-tick**.  Each of the ``ON`` and ``WHEN`` triggers also behave
+- **CallbackOnce triggers:** Triggers that only happen once per event.  To
+  make the trigger happen again, the event has to happen again:
+  - Callback delegates you tell the system to call when the user
+    performs GUI events (for example a button's ONCLICK).
+
+These two types of trigger don't have the same priority level.
+It is possible for a recurring trigger to interrupt a callback-once
+trigger, but not the other way around.  Further information about
+this is described in the :ref:`interrupt priority <interrupt_priority>`
+documentation below.
+
+All triggers work essentially like this:
+
+The kOS CPU decides it's time to cause a call to the trigger.  (How it
+does this is explained below in
+:ref:`interrupt priority <interrupt_priority>`.)  Once it decides its
+time to call the trigger, it does so by inserting a subroutine call
+at the current moment that interrupts the normal program flow and
+jumps to the trigger's subroutine *as if* the program itself had chosen
+to call the subroutine.  It manipulates the call-stack in such a way
+that the normal work of the ``Return`` instruction at the end of the
+trigger routine will pop back to the current location of the program
+flow.  This system works because all variables in kOS are on the
+stack without any registers, and so popping back to where the
+interruption happened puts everything back in the state it was in
+before the interruption so the program can continue as if nothing
+had happened.
+
+.. _trigger_steering:
+
+Triggers for Cooked Steering
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+*This is a kind of :ref:`recurring trigger <recurring_trigger>`.*
+
+The ``lock`` expressions associated with
+:ref:`Cooked Control <commands/flight/cooked.html>`__,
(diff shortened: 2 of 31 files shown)


Dunbaratu:
> Note to B: Can probably be fixed 

What do you mean "fixed"?  I thought you said this was exactly the behaviour *you* claimed you preferred - that callbacks caused by scripts altering a value rather than by user GUI activity happen synchronously, and at the same priority level.  The ability to do this is precisely WHY I made ``InterruptPriority.NoChange`` be a thing.

> Last question: Can you find some example where trigger could try to interrupt itself and being blocked by the second rule (identity) and not by the first rule (lower priority)? It has same priority as self, thus cannot interrupt self.

There's two trigger queues.  The active one and the pending one.  First things are moved into the pending one, then they only move from there into the active one at boundaries between physics ticks.  The identity rule prevents it from even entering the pending queue one in the first place.  The priority rule only really affects the active queue.

identity rule -> Should I put this trigger in the pending queue, meaning it will move to the active queue *on the next physics tick*?

priority rule -> Given that a trigger is in the active queue, does the priority allow it to happen right now?

If it's the same identity, then it should never even reach the priority rule test.

firda-cze:
And I thought *you* don't like *run levels* (interrupt priorities) and specificaly wrote that changing `pressed` from main loop would make the event handler interruptible if you do not rise its run level, which is even **bigger problem** than not lowering it to GUI-level if executed from top priority trigger. I think that CallbackOnce should always be executed on that run level / interrupt priority, no matter what.

The identity rule seems to prevent triggers/callbacks from flooding the queue (you can click the button like mad, but the callback won't enter the pending queue if it is already there or executing). Thanks, that is understandable.

Dunbaratu:
> I think that CallbackOnce should always be executed on that run level

Be clear on what "that" refers to here.

src/kOS/Suffixed/Widget/Button.cs
```diff
@@ -81,8 +81,10 @@ protected virtual void ScheduleCallbacks()
 
             if (UserOnToggle != null)
             {
-                UserOnToggle.TriggerNextUpdate(new BooleanValue(pressed));
-
+                if (guiCaused)
+                    UserOnToggle.TriggerOnFutureUpdate(InterruptPriority.CallbackOnce, new BooleanValue(pressed));
+                else
+                    UserOnToggle.TriggerOnNextOpcode(InterruptPriority.NoChange, new BooleanValue(pressed));
```
firda-cze: I would change these and all similar lines to this:
```
UserOnToggle.Trigger(InterruptPriority.CallbackOnce, !guiCaused, new BooleanValue(pressed));
```
Implement `UserDelegate.Trigger` like this:
```
public TriggerInfo Trigger(InterruptPriority priority, bool immediate, params Structure[] args) =>
    CheckForDead(false) ? null : Cpu.AddTrigger(this, priority, Cpu.NextTriggerInstanceId, immediate, args);
```
`Cpu.AddTrigger` like this:
```
        public TriggerInfo AddTrigger(int triggerFunctionPointer, InterruptPriority priority, int instanceId, bool immediate, List<VariableScope> closure)
        {
            TriggerInfo triggerRef = new TriggerInfo(currentContext, triggerFunctionPointer, priority, instanceId, closure) {
                IsImmediateTrigger = immediate
            };
            currentContext.AddTrigger(triggerRef);
            return triggerRef;
        }
```
And finally `ProgramContext.AddTrigger` like this:
```
        public void AddTrigger(TriggerInfo trigger)
        {
            bool immediate = trigger.IsImmediateTrigger && trigger.Priority >= CurrentPriority;
            if (immediate)
                Triggers.Add(trigger)
            else if (! ContainsTrigger(trigger))
                TriggersToInsert.Add(trigger);
        }
```

The only thing I am not quite sure about is whether the check `ContainsTrigger` should be there for both or, as written above, for pending only. `WinForms` (and similar GUI toolkits) do not check such a thing and you have to do it manually in code (either use some flag or temporarily unsubscribe the callback), but that practice is so common, that I am tempted to make it a rule and always check, thus making it:
```
        public void AddTrigger(TriggerInfo trigger)
        {
            if (ContainsTrigger(trigger)) return;
            bool immediate = trigger.IsImmediateTrigger && trigger.Priority >= CurrentPriority;
            if (immediate)
                Triggers.Add(trigger)
            else 
                TriggersToInsert.Add(trigger);
        }
```

Reasoning for it all:
1. `set button:pressed to true.` **from main** should execute the callback immediatelly, but with elevated priority, that no other (non-immediate) callback can interrupt it (like click on other button).
2. `set button:pressed to true.` **from recurring trigger** should delay the execution after all top-priority triggers are finished.
3. `set button:pressed to true` or more common `set editBox:text to "something"` from inside the very callback that is handling the change, should, probably, do nothing. This is about the `ContainsTrigger` check.

P.S.: `InterruptPriority.NoChange` won't be a thing after this.

firda-cze: Myabe the check should rather be written like this:
```
        public void AddTrigger(TriggerInfo trigger)
        {
            bool immediate = trigger.IsImmediateTrigger && trigger.Priority >= CurrentPriority;
            if (immediate) {
                if (! Triggers.Contains(trigger)) Triggers.Add(trigger)
            } else if (! TriggersToInsert.Contains(trigger))
                TriggersToInsert.Add(trigger);
        }
```
Because
1. `guiCaused` callbacks happen asynchronously together with setting the property. You can execute `set button:pressed to false`, which would execute the callback, get interrupted by IPU, now you can click on the button but original won't schedule the callback now, althought you will now read the state as true and `onTrigger` never notified you about it. Ignoring current callback when scheduling pending should fix the situation.
2. Executing the callback immediately from inside itself is probably not desirable. Not a big deal, but rather prevention. It is quite common mistake WinForms starters forget about (that changing the text of edit box from inside OnTextChanged will call the handler recursively). I am still not 100% sure about it, as it could have some consequences I cannot see right now, but still think it is the better way to do it (prevent that recursion, even indirect if you change property of some other control that will in turn backfire at you changing the original property/attribute/suffix).

firda-cze: https://github.com/Dunbaratu/KOS-1/blob/3edfc4294cceb47876c11193e5e1db92760b0ede/src/kOS.Safe/Execution/CPU.cs#L1434

This looks like running triggers are removed from `ProgramContext.Triggers` an I assumed they are still there, until they exit/return. It would require third list for those triggers (that were removed from Triggers and didn't finish yet).

BTW: The `if (trigger is NoDelegate)` can never be true, the classes are not compatible.
https://github.com/Dunbaratu/KOS-1/blob/3edfc4294cceb47876c11193e5e1db92760b0ede/src/kOS.Safe/Execution/CPU.cs#L1390
I have already changed that in one of my PRs to `if (trigger.EntryPoint < 0)`

Dunbaratu: In OpcodeReturn, a *new* trigger is created when the old one returns.  Triggers are removed entirely once they start executing, and the "return" from the code causes a new instance of the trigger to get constructed (a new structure pointing to the same entry point, same program context, etc) and put on the stack.  This is done so that the infrastructure for handling recurring triggers was easily laid atop the already existing infrastructure for handling one-time triggers.  It also ensures recurring triggers can't possibly re-interrupt when a trigger's code is already in the callstack, because its; the `return` at the end of the currently executing trigger subroutine that *causes* a new instance of the trigger to get put into the list, if the trigger is a recurring one.

firda-cze: I will leave that to your decision. The first version:
```
        public void AddTrigger(TriggerInfo trigger)
        {
            bool immediate = trigger.IsImmediateTrigger && trigger.Priority >= CurrentPriority;
            if (immediate)
                Triggers.Add(trigger)
            else if (! ContainsTrigger(trigger))
                TriggersToInsert.Add(trigger);
        }
```
fixes the first two points (callback from main or recurring trigger), but does nothing with the third (allowing recursion of callbacks).

I could have said that I am OK (not preferring but OK) with callbacks being executed inside recurring triggers without changing the priority, but I have definitely warned about immediate callbacks executed from main. Why not solve both at once, especially when I now see the code and that it is not so difficult to actually delay the callback when inside recurring trigger.

I am still not sure about the third point (prevent recursion) and it would obviously need more book-keeping (third list for running triggers - added here when removed from Triggers and removed from it in OpcodeReturn, probably). Then the check could be made like this:

```
        public void AddTrigger(TriggerInfo trigger)
        {
            bool immediate = trigger.IsImmediateTrigger && trigger.Priority >= CurrentPriority;
            if (immediate) {
                if (! RunningTriggers.Contains(trigger)) Triggers.Add(trigger)
            } else if (! ContainsTrigger(trigger))
                TriggersToInsert.Add(trigger);
        }
```
Assuming `ContainsTrigger` is unchanged (checking `Triggers` and `TriggersToInsert` as active triggers are still somehow pending, but passed the priority test, if I understand it correctly).
Maybe it should check for both `RunningTriggers` and `Triggers` when handling immediate. You know the infrastructure much better than me, you know I was not able to implement it.

Dunbaratu:
There has been a minor update to the branch (a new commit).  Nothing to do with the comments you made - just some problems I found in one of my scripts highlighted an issue and I wanted to deal with that first before looking at what you wrote.

Dunbaratu:
Also Fixes #2278

PR #2315, opened 2018-08-08
Fixes #2314 - detect and detach parameters on other vessels.

### LINKED ISSUE #2314: After decoupling, the upper stage is being controlled by the lower stage's CPU.
Here's an exact set of circumstances to create the problem, shown in screenshot form.

The summary is this:  If a lower-stage kOS CPU had a ``lock steering`` in effect when you decoupled it away, it seems to be trying to steer the upper stage vessel even though it's no longer attached to it.

bug before
bug invoke
bug after

This workaround seems to work, but it's ugly:
When you're about to stage, do it this way:
1 - First unlock steering, then ``wait 0.`` to ensure it really took effect.
2 - Then decouple, then ``wait 0.`` to ensure it really took effect.
3 - Then Re-lock steering.

The steering has to be locked at the moment of decoupling to cause the issue, and it has to be the lower stage (the "branch") that did the lock steering command, not the upper stage (the "root").

Dunbaratu:
Fixes #2314 

The problem is that while there was logic to attach flight parameters to the new kOSVesselModule when there's a vessel split, there wasn't any logic in the old kOSVesselModule to *detach* from flight parameters when there's a vessel split.

Now whenever it re-builds the part count because of a split or join, it also performs a check to see if any of the currently existing flight parameters are now referring to a different vessel.  If so, remove them and re-default them.

Before this PR: the flight parameters themselves are assigned a new Shared like they should be, and thus the new Vessel (so ``param.GetShared().Vessel`` is the new vessel), but the old ``kOSVesselModule`` still refers to these now changed flight parameters in its list of flight parameters even though they are now for a different vessel.  This causes the autopilot hooks to move the controls of the wrong vessel when the kOS core tries to steer the new vessel.

So the fix is to detect when the flight parameters have had their Shared Vessel ID re-assigned so they don't match anymore.  When that happens, this kOSVEsselModule needs to detach from them too.


src/kOS/Module/kOSVesselModule.cs
```diff
@@ -79,7 +79,7 @@ protected override void OnStart()
         public override void OnLoadVessel()
         {
             base.OnLoadVessel();
-            SafeHouse.Logger.SuperVerbose(string.Format("kOSVesselModule OnLoadVessel()!  On {0} ({1})", Vessel.vesselName, ID));
+            SafeHouse.Logger.SuperVerbose(string.Format("kOSVesselModule OnLoadVessel()!  On {0}({1}), {2}({3})", Vessel.vesselName, ID, vessel.vesselName, vessel.id));
```
Dunbaratu: This was just to verify that Vessel (capitalized) and vessel (lowercase) are the same thing.  They're both inherited from KSP's VesselModule class and I wasn't sure if they might differ.  I figured it didn't hurt to leave this information in there so I didn't revert this edit.

src/kOS/Module/kOSVesselModule.cs
```diff
@@ -234,7 +235,48 @@ private void AddDefaultParameters()
...
+                else if (key.Equals("wheelsteering"))
+                    p = new WheelSteeringManager(Vessel);
+                else if (key.Equals("wheelthrottle"))
+                    p = new WheelThrottleManager(Vessel);
+                else if (key.Equals("flightcontrol"))
+                    p = new FlightControl(Vessel);
+
+                if (p != null)
+                    AddFlightControlParameter(key, p);
+            }
```
Dunbaratu: The reason I have two loops - the first to find the keys to remove and the second to remove them - instead of just one loop is that I couldn't remember if .Net's rules that ban deleting from a collection while iterating over it counted for iterating the keys of a Dictionary.  It doesn't make any sense for them to since the keys are lookups rather than integer index that would change position, but a lot of what .Net bans when it comes to this don't make sense to me anyway so it seemed like it might ban it.

src/kOS/Module/kOSVesselModule.cs
```diff
@@ -350,13 +392,26 @@ private void CheckRehookAutopilot()
...
-            if (childParts.Count > 0)
+            if (vessel != null)
             {
-                foreach (var parameter in flightControlParameters.Values)
+                if (childParts.Count > 0)
                 {
-                    if (parameter.Enabled && parameter.IsAutopilot)
+                    foreach (var parameter in flightControlParameters.Values)
                     {
-                        parameter.UpdateAutopilot(c);
```
Dunbaratu: The above is the typical thing that happens in git diff output where a diff looks a lot more complex than it really is when you wrap existing code in a new indented section (in this case all I really did is wrap everything in the ``if (vessel != null)`` check).

src/kOS/Module/kOSVesselModule.cs
```diff
@@ -166,6 +166,7 @@ public void FixedUpdate()
                     ClearParts();
                     HarvestParts();
                     partCount = Vessel.Parts.Count;
+                    ResetPhysicallyDetachedParameters();
```
Dunbaratu: This seemed like the right place to put it, as this is the stuff where it checks for when the vessel adds or removes parts due to staging, docking, etc.

Dunbaratu:
Added NOT READY because I just detected a fatal problem.

The new code breaks RAW control.
For example just type this on the terminal:
```
set ship:control:yaw to -1.
```
This *should* deflect the yaw slider all the way left.  It has no effect on the yaw slider at all and the controls don't move.  

I just discovered this using my airplane autopilot, which runs on raw controls.  I think the new code is disabling the raw controls because raw control doesn't actually have a vessel ID properly assigned (I remember this from when I tested this before, but didn't think of the implications).

I'll come back to this later.

Dunbaratu:
NOTE: TO fix the travis build error - replace the places where I use ``vessel` (lowercase v) in kOSVesselModule.cs and instead use ``Vessel`` (uppercase V).  ``Vessel`` is a Property, and ``vessel`` is the backing field behind it.  In older versions of the KSP DLL's, only the property was public, and the field was private.  Now the field became protected in KSP's VesselModule class so it can be seen by derived classes like kOSVesselModule.  But we didn't really want to use it - that was just a typo mistake and it prevents backporting to KSP 1.3.1.

Dunbaratu:
Updated to latest 'develop', and also made the edits talked about above to see if that will make it pass Travis on KSP 1.4.1 (which will be important for the backport later).

PR #2324, opened 2018-08-27
added basic tutorial in the documentation

Toukie:
Added basictutorial.rst in /doc/source/tutorials/
edited tutorials.rst in /doc/source/

### DIFF
--- a/doc/source/tutorials/basictutorial.rst
+++ b/doc/source/tutorials/basictutorial.rst
@@ -46,7 +46,7 @@ To run the file, type:
 Making scrips with an editor
 ____________________________
 
-To use a text editor instead go to your KSP folder and go to ``/Ships/Script/`` and create an empty file and name it whatever you want
+To use a text editor instead go to your KSP folder and go to ``Ships/Script/`` and create an empty file and name it whatever you want
 (be cautious using spaces in filenames because this might mess stuff up, also, avoid using capital letters in .ks files if you are running Linux). Not sure where your KSP folder is? If you used steam to install KSP, go to your steam library and right click Kerbal Space Program in the list of games. Next click on ``properties`` and go to the ``local files`` tab. Finally, press on the ``browse local files`` button. (KSP doesn't require you to use steam, you can copy the KSP folder from the Steam folder to your desktop so you can play without using Steam).
 
 All KOS script files should end with ``.ks``.
--- a/doc/source/tutorials/basictutorial.rst
+++ b/doc/source/tutorials/basictutorial.rst
@@ -1,0 +1,654 @@
+********************************
+KOS and programming introduction
+********************************
+
+This tutorial is for people who are new to programming and want to start programming using KOS.
+
+.. contents:: Contents
+    :local:
+    :depth: 2
+
+==========================
+Accessing the KOS terminal
+==========================
+
+To use KOS you need to have a KOS processor on your vessel, they can be found under the 'control' tab in the editor and are under
+the same tab as the RCS thrusters and the reaction wheels.
+Right clicking the KOS processor in the editor after adding it to your vessel shows a screen where you can tweak settings of the
+KOS processor.
+
+There are two ways you can create KOS scripts: either you make them in game and edit them in game or you use a text editor.
+
+Making scripts in-game
+______________________
+
+First we'll look at making and editing scripts in-game.
+Put your vessel with KOS processor on the launchpad and right click the processor.
+Press the button that says 'open terminal' and click on the window that just popped up.
+
+typing the following will create a file called 'hello', to enter the command press ``ENTER`` (don't forget the period at the end):
+
+::
+
+	edit hello.
+
+
+you can now edit the file and make your script for the processor to execute, the other tutorials will show how you can make a
+working script. Keep in mind that files made in game dissapear after the vessel is gone. This happens because you locally made a file on the ship's processor.
+
+To run the file, type:
+
+::
+
+	run hello.
+
+
+Making scrips with an editor
+____________________________
+
+To use a text editor instead go to your KSP folder and go to ``/Ships/Script/`` and create an empty file and name it whatever you want
+(be cautious using spaces in filenames because this might mess stuff up). All KOS script files should end with ``.ks``.
+An example of a valid KOS script file name would be:
+
+``hello.ks``
+
+Running files you made using text editors is a bit more complex than if you'd make a file in game.
+This is because when you make a script in game the script is locally stored on the ship.
+Files you made with a text editor are saved in the archive, can be accessed from different saved games and won't be lost after
+the vessel is gone. You could compare it with ground control, which has infinite data storage and a spaceship which has limited
+space for data.
+
+To access the archive type:
+
+::
+
+	switch to 0.
+
+To run ``hello.ks`` type the same as above:
+
+::
+
+	run hello.
+
+=============
+Print and set
+=============
+
+Before we can send ships to space we first need to know the basic features of KOS.
+Let's begin with the ``print`` command. Type the following in your terminal and press ``ENTER`` to enter the piece of code. Keep in mind that **ALL** lines of code require a period
+at the end of the line, there are some exceptions but more about that later. ::
+
+	print "hello world".  // shows: hello world
+
+See those two forward slashes? That means that a comment has been made, anything
+after the double slashes won't be read by the script so you don't have to worry
+about what it says. Here's an example of what happens if you make something a comment: ::
+
+	// print "hello world 1".
+	print "hello world 2".
+
+If you'd run these two lines of code you'd **ONLY** see:
+``hello world 2``
+
+You'd might be wondering why you'd want to use the forward slashes if the script doesn't read them.
+The computer ignores them, but they're there to help leave explanations for the humans to see who might have to read your program.
+
+To clear the screen type: ::
+
+	clearscreen.
+
+Now say that we really like to use our hello world command but don't want to type
+the entire sentence every time, we could use the set command. ::
+
+	set x to "hello world".
+
+The set command 'sets' a certain value to the given variable. Everytime we refer
+to ``x`` we will actually refer to ``"hello world"``. ::
+
+	set x to "hello world".
+	print x. // shows: hello world
+
+Of course we can choose other things to print other than ``hello world``.
+Keep in mind that 'normal' text requires ``""`` and variables you made using the
+set command, numbers and booleans (true or false) don't need ``""``. ::
+
+	set x to "hello world".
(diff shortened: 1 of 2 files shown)


doc/source/tutorials/basictutorial.rst
```diff
@@ -0,0 +1,654 @@
...
+
+::
+
+	run hello.
+
+
+Making scrips with an editor
+____________________________
+
+To use a text editor instead go to your KSP folder and go to ``/Ships/Script/`` and create an empty file and name it whatever you want
```
lucaelin: Maybe explain where the KSP folder usually is, because people continue to use the wrong folders especially when using CKAN for mod installation..

doc/source/tutorials/basictutorial.rst
```diff
@@ -0,0 +1,654 @@
...
+(be cautious using spaces in filenames because this might mess stuff up). All KOS script files should end with ``.ks``.
```
lucaelin: Windows has the option to hide file-extensions on by default. This causes many people to name their files ``something.ks.txt`` without even knowing it. Maybe some additional instructions on how to properly create this file would be useful?

doc/source/tutorials/basictutorial.rst
```diff
@@ -0,0 +1,656 @@
...
+
+  set SomeAnimal to "Dog".
+
+  if SomeAnimal = "Cat" {
+    print "this is a cat".
+  } else if SomeAnimal = "Dog" {
+    print "this is a dog".
+  } else {
+    print "this is neither a cat nor a dog".
+
```
Dunbaratu: missing closing '}'.

doc/source/tutorials/basictutorial.rst
```diff
@@ -0,0 +1,656 @@
...
+Equals to or smaller than
+
+1 <= 1
+
+1 <= 2
+
+Is not equal to:
+
+1 <> 2
```
Dunbaratu: REST markdown likes to do silly things with '<' and '>' sometimes.  It's probably best to protect the above examples with double-backticks.

doc/source/tutorials/basictutorial.rst
```diff
@@ -0,0 +1,656 @@
...
+
+  set x to 1.
+
+  if x = 1 {
+    print "x is one".
+  }
+
+This will show: ``x is one``.
+
+You've probably noticed the curly brackets ``{ }`` after an ``if`` statement. You don't need a period at the end of an ``if`` statement but you need these brackets.
```
Dunbaratu: The claim "you need these brackets" isn't really true in the example you're showing, since the example you're showing only uses one statement in the body.  It would be more accurate to say "you need these brackets *IF* you have more than one statement in the body, otherwise they're optional."  

This is valid:
```
if x = 1
  print "x is one".
```

doc/source/tutorials/basictutorial.rst
```diff
@@ -0,0 +1,656 @@
...
+  print time:seconds.
+
+
+will print the current time in seconds. Let's say the time is 1 minute.
+It would print ``60``. You can also set the current time as a variable: ::
+
+  set CurrentTime to time:seconds.
+
+The variable ``CurrentTime`` will stay 60 seconds. Using the ``set`` command will look at
+a value and pick that value to stay the  same, even if the value it was set to changes.
```
Dunbaratu: What?

doc/source/tutorials/basictutorial.rst
```diff
@@ -0,0 +1,656 @@
...
+For instance, printing ``CurrentTime`` would give 60. Not only at ``time:seconds`` = 60,
+also at any other time like ``time:seconds`` = 4000. Eventhough ``time:seconds`` is 4000,
+``CurrentTime`` is still 60.
```
Dunbaratu: This is really really unclear.  I *know* the language and had to look at this several times to figure out what you were trying to say.  You are contrasting ``set`` to ``lock``, taking time to emphasize the point that ``set`` only assigns the value once and then it stays the same after that, but you haven't introduced ``lock`` yet, so you are drawing a comparison that doesn't mean anything *yet* to the reader
A better example would be something a bit more realistic - like this:

"Consider what would happen if you ran this (type it in and try it):"
```
print time:seconds.
wait 1.
print time:seconds.
wait 1.
print time:seconds.
```
"Notice it prints 3 different numbers, because you waited a second between each print statement, time:seconds was different each time."

"Now compare that with what happens if you set it into a variable first, then print that variable:"
```
set CurrentTime to time:seconds.
print CurrentTime.
wait 1.
print CurrentTime.
wait 1.
print CurrentTime.
```
"Notice that CurrentTime prints the same number each time, even though several seconds have passed.  This is because the ``set`` command stores the result *at the time it was run* into the variable, and the value stays the same until you set it again with another ``set`` command."

doc/source/tutorials/basictutorial.rst
```diff
@@ -0,0 +1,656 @@
...
+
+  until time:seconds > TimePlusFive {
+    print Multiplier.
+    set Adder to Adder + 1.
+    wait 1.
+  }
+
+So an easy way to read the until loop is to cover what ever is inside of the curly brackets
+and say: until ``time:seconds`` is bigger than our current time plus 5 seconds, print
+``Multiplier``, increase the value of ``Adder`` and wait 1 second.
```
Dunbaratu: This use of "cover up what ever is inside of the curly brackets" doesn't make sense.  You tell the reader to cover up the body of the loop, then tell them they can look at it that way and say: printing ``Multiplier``. and increasing the value of ``Adder``, and "wait 1 second".  All of these are in the section you just got done explicitly telling the reader to cover up and ignore at the moment.  There needs to be a clean break where you say, "and now take your hand away and look at what is inside the curly brackets and .... (etc)" before you get to the stuff about Multiplier, Adder, and waiting 1 second.

doc/source/tutorials/basictutorial.rst
```diff
@@ -0,0 +1,656 @@
...
+
+  set CurrentTime to time:seconds.
+
+The variable ``CurrentTime`` will stay 60 seconds. Using the ``set`` command will look at
+a value and pick that value to stay the  same, even if the value it was set to changes.
+For instance, printing ``CurrentTime`` would give 60. Not only at ``time:seconds`` = 60,
+also at any other time like ``time:seconds`` = 4000. Eventhough ``time:seconds`` is 4000,
+``CurrentTime`` is still 60.
+
+The ``lock`` command updates constantly the variable, for example: ::
```
Dunbaratu: This sounds very German.  English grammar would be more like "updates the variable constantly".

doc/source/tutorials/basictutorial.rst
```diff
@@ -0,0 +1,656 @@
...
+ print "done waiting".
+
+
+It will take 10 seconds before ``done waiting`` shows up.
+Using ``wait 0`` will let the script wait for one physics tick, this can be handy for
+when you're doing stuff with maneuvers. Maneuvers don't show up instantly but show up
+after one physics tick. More about maneuvers in part ???.
+
+The ``until`` command will keep looping a piece of code until the given value has been met.
+Before we can talk about until loops let's first talk about ``time:seconds`` and the ``lock`` command. ::
```
Dunbaratu: I noticed that this little "side note" about time:seconds, and ``lock`` takes more space than the actual information about ``until`` does.  Also, while you do use ``time:seconds`` in your ``until`` example loop below, you don't really use ``lock`` at all in the example, so did you have to introduce it now to get to the stuff about the ``until`` loop?

My concern is that you start off talking about ``until`` here, then spend a long time on things that are NOT the ``until`` loop, before you get to the ``until`` loop.  Can this be changed so that the first ``until`` example people see is just a lot simpler?  A simple counting loop rather than one based on ``time:seconds`` would help because then you don't have to explain ``time:seconds`` yet, and can just get to the information about the loop right away.  THEN *after* that you can talk about the interactions between ``set``, ``lock``, and give the more complex loop example.

doc/source/tutorials/basictutorial.rst
```diff
@@ -0,0 +1,656 @@
...
+  4
+  6
+  8
+  10
+
+Wait until
+__________
+
+You can also use the ``wait until`` command, this blocks all other code until the
+conditions have been met. ::
```
Dunbaratu: The phrase "blocks all other code until" isn't 100% correct because of triggers, which haven't been explained yet.  Perhaps change it to just say "pauses the program here until"...

doc/source/tutorials/basictutorial.rst
```diff
@@ -0,0 +1,656 @@
...
+
+  set TimePlusFive to time:seconds + 5.
+  wait until time:seconds > TimePlusFive.
+  print "done waiting".
+
+It will take 5 seconds for ``done waiting`` to show up.
+
```
Dunbaratu: I'm not sure if this is the right place to mention this or not, but the fact that ``wait until`` only checks the condition "once per physics tick" is relevant, from a cpu usage point of view.  It means this isn't a "busy loop" that drives up the cpu usage, and more importantly for the beginner that doesn't care about that, it means it's physically impossible to wait a fraction of a physics tick, and the wait time always "rounds up" to the nearest start of a new tick.

doc/source/tutorials/basictutorial.rst
```diff
@@ -0,0 +1,656 @@
...
+(For each item in the list called ``ValueList``, which we call ``Whatever``, do whatever is inside of the brackets).
+
+In this case it prints: ::
+
+  0
+  5
+  10
+  15
+  20
+
```
Dunbaratu: Also give an example where the index is itself a variable, maybe?
show that:
```
set x to 3.
print ValueLilst[x].
```
does the same thing as:
```
print ValueList[3].
```

doc/source/tutorials/basictutorial.rst
```diff
@@ -0,0 +1,656 @@
...
+the function it is in. ::
+
+  Function VelocityCalculator {
+    Parameter OrbitHeight.
+
+    set KerbinRadius to 600000.
+    set TotalRadius to OrbitHeight + KerbinRadius.
+    set OrbitalPeriod to ship:orbit:period.
+    return (2 * 3.1416 * TotalRadius) / OrbitalPeriod.
+    print "this will be skipped".
```
Dunbaratu: Maybe explain why it will be skipped.

```
    return (2 * 3.1416 * TotalRadius) / OrbitalPeriod.
    // A return statement ends the function, meaning
    // nothing that comes after it in the function will
    // actually happen.  For example, you'll never see
    // the following line print out, because the return
    // statement ended the function early:
    print "You Will Never See This Print.".
```

PR #2325, opened 2018-08-29
Fixes #1669 - changing clamping of hyperbolic orbits

### LINKED ISSUE #1669: obt:meananomalyatepoch should not be modulo 360
It looks like Mean Anomaly At Epoch as reported by kOS is always clamped to be in the range of 0...360.

This is fine for actual orbits (eccentricity < 1) and all of the normal trigonometric functions treat 0=360=720 degrees anyways.  However, it matters when it comes to hyperbolic orbits and hyperbolic trigonometry functions -- where `cosh(360) != cosh(0)`

KSP's own internal state doesn't appear to be clamped.  For my current game state, KER reports a mean anomaly of 433.15 degrees and a recent KSP autosave reports `MNA = 7.55618641172975` (well over 2*pi).

pand5461:
The position on an elliptic orbit is expressed in terms of trig functions of mean anomaly, and on a hyperbolic trajectory in terms of hyperbolic trig functions. The latter are non-periodic, so when mean anomaly is clamped to [0; 360] range, it gives identical answers for essentially different orbital points.
The true anomalies for hyperbolic trajectories make more sense in [-180; 180] range: negative values mean points before periapsis, and positive ones after periapsis.
Simple eccentricity check is added to dispatch the values returned by `obt:trueanomaly` and `obt:meananomalyatepoch` suffixes.
For elliptic orbits, this keeps the [0; 360] clamping, as before.


Dunbaratu:
I added some paragraphs to the user documentation describing the new way it works, them merged it.

PR #2336, opened 2018-10-05
Fixes #2317 - Disable Wheels, if cpu shut down due to electric starving

### LINKED ISSUE #2317: Raw controls should let go when out of power
When you do something like this:
```
set wheelthrottle to 0.2.
```
and then the rover runs out of power, the ``kOSVesselModule`` still keeps pushing the control even though the CPU doing that should by all rights be "dead".  This causes it to never recharge properly because the instant there's a tiny bit of power it gets spent by the ``kOSVesselModule`` pushing on the control - and it's a control you cannot turn off because you don't have power anymore.

To Fix:  have kOSVesselModule detect a lack of power and shut down all of its ``IFlightControl``s - remember all the kOS CPUs have to do a reboot when the charge went to zero so you're not confusing any programs by doing this.

PascalOtto:
Rover can now load energy, without wasting it immediatly on the motor, what prevented the user from taking control.

### DIFF
--- a/src/kOS/Module/kOSProcessor.cs
+++ b/src/kOS/Module/kOSProcessor.cs
@@ -884,6 +884,8 @@ namespace kOS.Module
                     if (shared.Window != null) shared.Window.IsPowered = false;
                     if (shared.SoundMaker != null) shared.SoundMaker.StopAllVoices();
                     foreach (var w in shared.ManagedWindows) w.IsPowered = false;
+                    // Give rovers a chance to load energy
+                    kOSVesselModule.GetInstance(shared.Vessel).GetFlightControlParameter("wheelthrottle").UpdateValue(0.0d, shared);
                     break;
             }
 
--- a/src/kOS/Module/kOSProcessor.cs
+++ b/src/kOS/Module/kOSProcessor.cs
@@ -884,8 +884,8 @@ namespace kOS.Module
                     if (shared.Window != null) shared.Window.IsPowered = false;
                     if (shared.SoundMaker != null) shared.SoundMaker.StopAllVoices();
                     foreach (var w in shared.ManagedWindows) w.IsPowered = false;
-                    // Give rovers a chance to load energy
-                    kOSVesselModule.GetInstance(shared.Vessel).GetFlightControlParameter("wheelthrottle").UpdateValue(0.0d, shared);
+                    kOSVesselModule vesselModule = kOSVesselModule.GetInstance(shared.Vessel);
+                    if (!vesselModule.ProcessorAvailable()) vesselModule.OnAllProcessorsStarved();
                     break;
             }
 
--- a/src/kOS/Module/kOSVesselModule.cs
+++ b/src/kOS/Module/kOSVesselModule.cs
@@ -435,6 +435,35 @@ namespace kOS.Module
         }
 
         /// <summary>
+        /// Checks if there is any kOSProcessor active
+        /// </summary>
+        public bool ProcessorAvailable()
+        {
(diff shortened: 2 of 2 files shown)


Dunbaratu:
Hello @PascalOtto. Thanks for contributing a PR to the project.  I don't think I've seen your name before.

While I understand the problem, I don't know if this is the best way to solve it, for two reasons:

(1) - This fix is specific to *only* wheelthrottle.  Is this a problem that occurs on wheelthrottle and wheelthrottle ONLY, or is it a general problem with raw control movement of any kind being on during power-off?  (There's no 'issue' with the PR describing the problem it fixes, so I'm trying to guess).

(2) - If two kOS CPUs exist on the same vessel and only *one* of them is power-starved (i.e. it's electric power is deliberately locked out by the user for hibernation reasons), we shouldn't kill the ``wheelthrottle`` because the *other* kOS CPU might still be getting power just fine and be trying to use ``wheelthrottle``.  (This is the entire reason why the class ``KOSVesselModule`` exists - while there can be multiple ``KOSProcessor``s in the same vessel, a vessel can only have one notion of where its controls are set.  The vessel's throttle control cannot be simultaneously at 100% and 0%.  Therefore we force ``KOSProcessor``s to do all the vessel controls setting and getting via the ``KOSVesselModule``, of which there is only one per vessel.)

Because of (1), this might need to be done more generically, for all the controls.
Because of (2), this should probably get triggered based on when the entire vessel is out of power and be done by something inside ``KOSVesselModule``, rather than being based on just this one part being out of power.

PascalOtto:
Hello @Dunbaratu. Thats right, I am new to all of this. I study Computer Science in germany for a year now and so I decided it would be cool to get some practical experience. I will try to get some easy bugs fixed to get things going. 

So I thought about doing a check in KOSVersselModule, but I can't find a good way to check if the electricity is starved. Correct me if I'm wrong but wouldn't we have to do the check frequently, which consumes some process time. So what do you think about KOSProcessor calling KOSVesselModule, which does a check if there is any other active KOSProcessor and releases the Controls if not?

By the way, how is it possible to seperate electricity, how would a vessel look like, which could get one CPU starving, but another one working?

Dunbaratu:
The way you starve one part of power deliberately is by clicking the little power icon next to it in the context menu panel.  You can lock it out as being unable to use power.

(Also, even though the entire vessel shares electriccharge without pathing to it, that is not something guaranteed to always be true necessarily.)

How about this technique?

- Keep the check where it is (in ``KOSProcessor.ProcessorModeChanged()``) because this is a good spot that doesn't get called again and again all the time.
- But make a new method inside ``kOSVesselModule``, with a name like ``AllKOSProcessorsStarved()``.  Make it iterate over all the KOSPartModules inside the vessel, asking each of them if they are in STARVED state.  Only if *all* of them say yes, it returns true.  (You can iterate over all modules of a particular type in a vessel, but it's an expensive serial walk.  You can see an example of code that does this in ``VesselTarget.GetModulesNamed()``.
- Make a second method inside ``kOSVesselModule``, with a name like ``OnAProcessorStarved()`` that your code in ``kOSProcessor.ProcessorModeChanged()`` will call instead of you directly setting the WHEELTHROTTLE.  Have this method be the one that disables WHEELTHROTTLE, but also have it disable everything else that a kOSProcessor might be doing to the controls - wheelsteering, raw controls, etc.  (Thus why it's a good idea to move it here, so that longer list of commands isn't cluttering up the code in kOSProcessor).

I think that might work.

PascalOtto:
Whenever the last active Processor gets shut down or starved out, ALL controls will be released now.

PR #2354, opened 2018-11-07
Compatibility for IR Next

### LINKED ISSUE #2316: kOS didn't work with IR_v3.0.0_beta3_p7
I find the kOS_v.1.1.5.0_for_KSP_1.3.1 didn't work with InfernalRobotics_3.0.0_beta3_p7_for_KSP_1.3.1.
When I check ADDONS:IR:AVAILABLE, the result is FALSE.
And I can't control IR parts by ADDONS:IR structure.

kos didn t work with ir

Maybe the reason is IR has updated. It's module name and part-menu has changed.
Have you the plan to repair this problem in next kOS version?
Thnaks.

tony48:
Replace the old IRWrapper with the new one by meirumeiru.
Fix #2316

### DIFF
--- a/src/kOS/AddOns/InfernalRobotics/IRWrapper.cs
+++ b/src/kOS/AddOns/InfernalRobotics/IRWrapper.cs
@@ -1,756 +1,677 @@
-﻿using System;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 
+// TODO: Change this namespace to something specific to your plugin here.
 namespace kOS.AddOns.InfernalRobotics
 {
-    public class IRWrapper
-    {
-        private static bool isWrapped;
-        private static bool? hasAssembly = null;
-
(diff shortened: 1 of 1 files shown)


Horogg:
Please, is there any chance to re-add support for old IR?
- New IR Next looks good, but UI has many bugs and integrity of builded structures is very unstable after load game from save.
- I use old IR with KSP 1.6.1 without problems.
- Here is worked fork - IR 2.1.3 ... https://github.com/tony48/InfernalRobotics/releases

tony48:
kRPC is compatible with the two IRs, so it's definitely possible.

hvacengi:
It is certainly possible to support both.  Given that the initial PR was replacing support, and the appearance on the forums and on the git repositories was that IR Next was getting at least some support, I decided not to modify the PR to support both mods.  It should be relatively easy to simply include two versions of IRWrapper, for for each mod, and then test each one when looking for support.

I'll create a new issue to track this ability and for future conversation.

Issue #2319, opened 2018-08-13
Expose body suffixes for surface, water, and children.


Dunbaratu:
KSP's API has these bits of information about a ``Body`` that we aren't exposing to kOS:

``Body.ocean`` - bool - True if the body has a liquid sphere at sea level.  False if its dry and "sea level" is just a number.
``Body.hasSolidSurface`` - bool - True if the body has any terrain polygons.  False if it's all gas like Jool.
``Body.orbitingChildren`` - List<CelestialBody> - the children of this body (i..e moons of a planet or planets of a star).  Right now you can only walk the tree of bodies *upward* from child to parent via the :BODY suffix (i.e. ship:body:body is the body the body you orbit orbits).  There's no way to walk the tree downward from parent to child.  This suffix would provide that.

PR #2355, opened 2018-11-12
Added features described in issue #2319

evandisoft:
This is in reference to issue #2319.

I don't have much experience contributing to projects on Github like this. Hoping to start doing this sort of thing a lot.

Hope I did everything correctly. If not, let me know and I'll fix it.

Also, I modified the .gitignore so that it stopped committing my vim .swp files. Hope that is ok. If not, I can just not use Vim.

### DIFF
--- a/doc/source/structures/celestial_bodies/body.rst
+++ b/doc/source/structures/celestial_bodies/body.rst
@@ -67,6 +67,9 @@ All of the main celestial bodies in the game are reserved variable names. The fo
     :attr:`NAME`                     :ref:`string <string>`
     :attr:`DESCRIPTION`              :ref:`string <string>`
     :attr:`MASS`                     :ref:`scalar <scalar>` (kg)
+    :attr:`OCEAN`                    :ref:`boolean <boolean>`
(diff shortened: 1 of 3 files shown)


doc/source/structures/celestial_bodies/body.rst
```diff
@@ -67,6 +67,9 @@ All of the main celestial bodies in the game are reserved variable names. The fo
     :attr:`NAME`                     :ref:`string <string>`
     :attr:`DESCRIPTION`              :ref:`string <string>`
     :attr:`MASS`                     :ref:`scalar <scalar>` (kg)
+    :attr:`OCEAN`                    :ref:`boolean <boolean>`
```
Dunbaratu: I think this should probably be named with the word "HAS" in front of it, to make it consistent with many of the other similar boolean suffixes like this (i.e. call it "HASOCEAN").

Thanks for the contribution, by the way!

evandisoft: You're welcome. Fixed

PR #2357, opened 2018-11-17
Fixes #2356 and clears some 'eraseme' printouts.

### LINKED ISSUE #2356: Error after program quit if both interpreter and program have "lock steering".
(This is messy.)

Lock Steering from the interpreter interfers with Lock Steering from a program, if Lock Steering is left on while the program is invoked, and the program does its own Lock Steering.

Interestingly, the error is only thrown when the program ENDS, not while it's running.

To re-create it, use any craft at all.  Make a program that contains a lock steering, EVEN IF that lock steering is never actually invoked, as in the example below:
Make the following small example program:
**Filename "test_bug.ks"**:
```
// ______ filename test_bug.ks ______
print "---program starting---".
wait 5.
print "---program ending---".
if false {
  // Note this line NEVER actually runs because it's in an "if false",
  // but its presence still causes the bug, probably because of how
  // the compiler builds the boilerplate for it at the top of the opcodes:
  lock steering to north.
}
```
Then at the interpreter, run these things in this order:
```
lock steering to heading(90,0). wait 2. run test_bug.ks.
```
This causes you to see:
```
---program starting---
---program ending---
Program ended.
```
(as expected), but then as soon as the program ends, it does this:
```
lock steering to heading(90,0). wait 2. run test_bug.
                                         ^

 
(Filename: C:/buildslave/unity/build/artifacts/generated/common/runtime/DebugBindings.gen.cpp Line: 51)

kOS.Safe.Exceptions.KOSNotInvokableException: Attempted to make a function call on a non-invokable object:
   steering
  at kOS.Safe.Execution.CPU.GetVariable (System.String identifier, Boolean barewordOkay, Boolean failOkay) [0x00000] in <filename unknown>:0 
  at kOS.Safe.Execution.CPU.GetValue (System.Object testValue, Boolean barewordOkay) [0x00000] in <filename unknown>:0 
  at kOS.Safe.Compilation.OpcodeCall.StaticExecute (ICpu cpu, Boolean direct, System.Object destination, Boolean calledFromKOSDelegateCall) [0x00000] in <filename unknown>:0 
  at kOS.Safe.Compilation.OpcodeCall.Execute (ICpu cpu) [0x00000] in <filename unknown>:0 
  at kOS.Safe.Execution.CPU.ExecuteInstruction (kOS.Safe.Execution.ProgramContext context, Boolean doProfiling) [0x00000] in <filename unknown>:0 
```

There are things in place to prevent a delegate from a "stale" program context from getting called.  Those protections are being triggered  in some way as the program ends and returns control back to the interpreter.

Note, the interpreter's steering updater ``$steering*`` function is NOT being called while the program is running (this seems correct), but as soon as the program ends, it tries calling it again and it complains like this.

Dunbaratu:
(Getting rid of the "eraseme" printouts was not part of the
fix, but is something that needed cleaning up anyway.  Whenever
I put the word "eraseme" in a line of code it is meant to be
a reminder to go back and delete that line before merging it
in, because it only existed to log some debug info for a
specific problem I was looking at at the time.  I must have
forgotten to do so at some point in the past for those lines
to still be there now.)

The bug fix is to unconditionally restore the global pointers,
regardless of whether the program context also set them.  The
fact that the restoration had been inside an "else" meant it
was being skipped when the "if" happened, and that's incorrect.

As part of this, the logic of when to remove triggers from a
program context has been moved to the PopContext() so that all
old triggers from a context being popped will always go away.
The logic that previously tried doing it inside RestorePointers()
was wrong because by the time that is called, the currentContext
no longer points at the Program that's going away.  It points to
the new context being popped TO, not the one popped FROM.  So
the RemoveTrigger call was trying to remove it from the wrong context
anwyway.  It was meant to remove it from the stale context being
popped, which it was not doing given where it was located.  Luckily
it never caused a bug, because of a *second* bug that made it never
trigger, which is that the values it removes are no longer of type
(int) like they used to be and are now of type (UserDelegate) so the
check against their type was skipping over the offending code anyway.
(Two wrongs kind of made a right - sort of.)

### DIFF
--- a/src/kOS.Safe/Execution/CPU.cs
+++ b/src/kOS.Safe/Execution/CPU.cs
@@ -210,6 +210,7 @@ namespace kOS.Safe.Execution
                 // remove the last context
                 ProgramContext contextRemove = contexts.Last();
                 NotifyPopContextNotifyees(contextRemove);
+                contextRemove.ClearTriggers();
                 contexts.Remove(contextRemove);
                 shared.GameEventDispatchManager.RemoveDispatcherFor(currentContext);
                 contextRemove.DisableActiveFlyByWire(shared.BindingMgr);
@@ -435,19 +436,14 @@ namespace kOS.Safe.Execution
             {
                 if (globalVariables.Contains(item.Key))
                 {
-                    // if the pointer exists it means it was redefined from inside a program
-                    // and it's going to be invalid outside of it, so we remove it
+                    // If the pointer exists it means it was redefined from inside a program
+                    // and it's going to be invalid outside of it, so just to be sure, remove
+                    // it entirely in preparation for restoring the old one:
                     globalVariables.Remove(item.Key);
                     deletedPointers++;
-                    // also remove the corresponding trigger if exists
-                    if (item.Value.Value is int)
-                        RemoveTrigger((int)item.Value.Value, 0);
                 }
-                else
-                {
-                    globalVariables.Add(item.Key, item.Value);
-                    restoredPointers++;
-                }
+                globalVariables.Add(item.Key, item.Value);
+                restoredPointers++;
             }
 
             SafeHouse.Logger.Log(string.Format("Deleting {0} pointers and restoring {1} pointers", deletedPointers, restoredPointers));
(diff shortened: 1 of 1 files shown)


evandisoft:
I didn't find any problems with it. Worked with my default "load the right libraries and then get to orbit" program that has some complexity.

Dunbaratu:
Merging myself even though that's sort of against the rules I set up - just because it did get a review so that's probably okay.

PR #2362, opened 2018-11-25
Added KUNIVERSE:REALTIME and :REALWORLDTIME

lucaelin:


lucaelin:
Looks like monodevelop messed with the entire file just before committing.. how do I make it stop?

lucaelin:
This RP is closed in favor of PR 2363

PR #2371, opened 2018-12-06
Better Fixes #2345 solution.

### LINKED ISSUE #2345: kOS reading UTF-8 can't handle a file with a BOM in it.
(This is something that was raised by issue #2265.)

kOS can read UTF-8 files on the archive (volume 0), but not when they are preceded by a BOM (Byte Order Mark) marker.  kOS ignores the BOM and fails to realize it doesn't count as char data.  Thus it's offset by one byte when rest of the bytes are read, making it mangle unicode characters above 127.

BOM is a short byte sequence optionally inserted at the front of a UTF_8, UTF-16, or UTF-32 file that tells the reader whether the file is encoded as UTF-8, UTF-16, or UTF-32.  It also specifies in UTF-16 whether the bytes are encoded in big-endian or little-endian order, which isn't even relevant information in UTF-8.  In UTF-8 the only purpose of the BOM is to identify the file as definitely being UTF-8.  Most editors that save UTF-8 don't even use the BOM mark.

The fix of telling users "Find your text editor settings about the BOM and tell it not to include it" is what we do for now.  But it is hypothetically possible to make kOS recognize the BOM when present and just skip past it as long as it says the file is UTF-8.

What is the BOM? -> https://en.wikipedia.org/wiki/Byte_order_mark

Dunbaratu:
Fixes #2345 better.

This still leaves raw binary files unaffected when they
do not start with the magic first 3 bytes of the BOM,
thus it still handles KSM files correctly.

### DIFF
--- a/src/kOS.Safe/Persistence/FileContent.cs
+++ b/src/kOS.Safe/Persistence/FileContent.cs
@@ -38,7 +38,7 @@ namespace kOS.Safe.Persistence
 
         public FileContent(byte[] content) : this()
         {
-            Bytes = content;
+            Bytes = CopyBytesSkippingBOM(content);
         }
 
         public FileContent(List<CodePart> parts) : this()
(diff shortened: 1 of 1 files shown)


Dunbaratu:
Closing because it's been superceeded by PR ##2374

PR #2378, opened 2018-12-13
Fixes #2377 - Added cid suffix to PartValue

### LINKED ISSUE #2377: Part-ID that is unique to the craft
Trying to reliable get the same part between launches can be done using nametags, but assigning huge amounts of nametags with each nametag being unique to a part is a lot of work.
The KSP Part class seems to include a so called craftID that is stored inside the .craft file and equal for "the same part" across multiple launches. 
Exposing this ID to KOS would i.e. allow scripts to mimic the action-group system without having to assign nametags to all parts or relying on a not-changing order of the `ship:parts` list.

lucaelin:
Fixes #2377 
This PR is NOT READY yet as i have not tested the cid behavior in case of docking and such...
It's also missing documentation as it is super hard for me to correctly formulate things like this :see_no_evil:


lucaelin:
Exactly.

lucaelin:
I still havn't tested any docking scenarios... I'm gonna do that today. But in addition to testing the exact behavior, the PR still needs documentation.

lucaelin:
I just tested docking and it works just like i expected it to. Once two crafts created using the same craft file dock to eachother, there are two parts on the (joined) craft with the same cid. Once undocked, each ship has its own part back :) The cid doesn't change in this process.

PR #2408, opened 2019-01-13
Now both OrbitInfo constructors fill body name.

### LINKED ISSUE #2406: ship:patches documentation is either misleading or is bugged
I'm unsure if this is a bug with `ship:patches` or if the documentation is incorrect/misleading, hence the title.

**To reproduce:** get in an orbit with multiple patches, then `print ship:patches.`

This returns a list of all of the ship's patches like so:
```
LIST of n items:
[0] = "ORBIT of <unnamed>"
[1] = "ORBIT of <unnamed>"
[2] = "ORBIT of <unnamed>"
.
.
.
[n-1] = "ORBIT of <unnamed>"
```
[From the docs](https://ksp-kos.github.io/KOS/structures/vessels/vessel.html#attribute:VESSEL:PATCHES), I don't think this is the intended behavior:

"The list of orbit patches that describe this vessel’s current travel path based on momentum alone with no thrusting changes. If the current path has no transitions to other bodies, then this will be a list of only one orbit. **If the current path intersects other bodies, then this will be a list describing the transitions into and out of the intersecting body’s sphere of influence.**"

Emphasis mine. The list of patches returned contains no useful information beyond the number of patches. It is only through the use of patch suffixes that any information can be gleaned.

Dunbaratu:
Fixes #2406 by populating the name of the body on all OrbitInfos.

### DIFF
--- a/src/kOS/Suffixed/OrbitInfo.cs
+++ b/src/kOS/Suffixed/OrbitInfo.cs
@@ -1,4 +1,4 @@
-﻿using kOS.Safe.Encapsulation;
+using kOS.Safe.Encapsulation;
 using kOS.Safe.Encapsulation.Suffixes;
 using kOS.Safe.Exceptions;
 using kOS.Serialization;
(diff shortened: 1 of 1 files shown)


src/kOS/Suffixed/OrbitInfo.cs
```diff
@@ -1,4 +1,4 @@
-﻿using kOS.Safe.Encapsulation;
+using kOS.Safe.Encapsulation;
```
Dunbaratu: Go home, git diff, you're drunk.  There's literally no diff between these two lines.  Not even in whitespace.

PR #2415, opened 2019-01-28
added constant:g0, and edited the doc more about G vs G0.

### LINKED ISSUE #2413: Expose 'g' (lowercase) through constant:
We expose G (uppercase, Newton's Universal Gravatational Constant) through the ``constant:G`` technique.
We do this because the game's idea of G might be a bit off from a textbook's idea.  The official standard value according to NIST keeps changing every few years as better measurements get made.  And it's the game's idea that matters in kOS, not the value you read from a textbook or website.  I recently made a PR that calculates it from the game itself.

But we don't expose g (lowercase, acceleration due to gravity at sea level on Earth/Kerbin).

We probably should expose it, for the same  reason - because the value the game uses is what matters in things like ISP calculations, not the value that may be printed in this textbook or that.

One problem with implementing this is kOS's case-insensitive nature, and backward compatibility.  We already called Newtons' gravitational constant by the name ``constant:G`` so we can't call accel at sea level ``constant:g`` because that's the same "spelling".

Dunbaratu:
Fixes #2413.
g0 is simpler than G was - g0 actually *is* in the physics globals.

### DIFF
--- a/doc/source/math/basic.rst
+++ b/doc/source/math/basic.rst
@@ -22,7 +22,9 @@ constants about the universe that you may find handy in your math operations.  P
       - Description
 
     * - :global:`G`
-      - Newton's Gravitational Constant
+      - Newton's Gravitational Constant.
+    * - :global:`g0
+      - gravity acceleration (m/s^2) at sea level on Earth.
     * - :global:`E`
       - Base of the natural log (Euler's number)
     * - :global:`PI`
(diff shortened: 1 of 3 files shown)


Dunbaratu:
Travis build failed because it's using KSP 1.4.1 and it turns out PhysicsGlobals.GravitationalAcceleration was only just recently added for KSP 1.6 and up - I didn't know that.  (I suspect SQUAD only started caring about ensuring they were using a standard constant g0 across all their code when they wanted to do dV calculations in KSP 1.6 and those calculations need that g0 because of that weird way ISP measures fuel by Newtons instead of by mass.)

To support backports, I'm going to have to change this implementation into something involving reflection where it tests to see if PhysicsGlobals has a GravitationalAcceleration member first, and if it does not then it falls back to a hardcoded literal instead.

PR #2423, opened 2019-02-04
fix #2422 broken ship:maxthrust

### LINKED ISSUE #2422: ship:maxthrust is broken in latest release
It looks like SHIP:MAXTHRUST now returns thrust at zero air pressure, instead of thrust at current pressure as it's used to be and supposed to be in the docs.

```
print ship:maxthrust.
214.999984741211
print ship:maxthrustat(0).
214.999984741211
print ship:maxthrustat(body:atm:altitudepressure(altitude)).
168.421600341797
```

firda-cze:
fix #2422 

Changed all `double atmPressure = -1.0` to `double? atmPressure = null` and altered `GetIsp` to `... .Evaluate((float)(atmPressure ?? engine.part.staticPressureAtm));`

### DIFF
--- a/src/kOS/Suffixed/Part/EngineValue.cs
+++ b/src/kOS/Suffixed/Part/EngineValue.cs
@@ -256,11 +256,7 @@ namespace kOS.Suffixed.Part
 
     public static class ModuleEnginesExtensions
     {
-        public static float GetThrust(this ModuleEngines engine, bool useThrustLimit = false, float throttle = 1.0f, bool operational = true)
-        {
-            return GetThrust(engine, engine.part.staticPressureAtm, useThrustLimit, throttle, operational);
-        }
-        public static float GetThrust(this ModuleEngines engine, double atmPressure, bool useThrustLimit = false, float throttle = 1.0f, bool operational = true)
+        public static float GetThrust(this ModuleEngines engine, double? atmPressure = null, bool useThrustLimit = false, float throttle = 1.0f, bool operational = true)
         {
             if (engine == null || operational && !engine.isOperational)
                 return 0f;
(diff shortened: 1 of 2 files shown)


Dunbaratu:
Thanks for noticing the issue, recognizing what caused it, and reacting quickly.  But can you add a "///" comment to the method to document the meaning of the defaulted null?

Changing the default from -1 to null doesn't do anything to fix the primary problem that calling it without knowing the undocumented effect it had caused unexpected behaviour.  It's still undocumented what it actually means to pass a null there, still requiring that someone know the innards of the method to realize the effect it will have.

Dunbaratu:
As far as the meaning of the old -1 default, as far as I can tell that appears to be how SQUAD designed it not us.  The -1 was just passed directly into the KSP API call `engine.atmosphereCurve.Evaluate()` itself before.  I'm not sure how to verify that the new way matches the old way when I can't tell what the old way was or what that meant.

Dunbaratu:
> Current: Obvious null having special meaning (quite understandable what the choice for the pressure is if you do not specify).

No. Null for pressure does not automatically imply "read current pressure at the engine's position in space".  My first impression would be that null means pressure info doesn't exist, which is not what it means here.   I'm going to add the documentation.  It's the exact same problem as before.  You have to read the insides to know what the default null does, just like you had to read the insides to know what the default -1 does.

firda-cze:
I will repeat the cause: `VesselUtils` uses default `-1.0` for *special value* and instead of calling `Engine.Get...` **without any parameter**, in that case, **it passes that specific arbitrary `-1.0` into it**. I did not know it and created two separate methods, one having `atmPressure` argument, one not. What happened is, that `VesselUtils` called the wrong version with wrong value, because I had no idea that `-1.0` had any meaning OUTSIDE of `EngineValue`.

The difference is that I did not spot there is any special value accepted from outside. It has nothing to do with what the meaning of that special value is (nothing against documenting it, but **that was not the cause**).

I was about to write the documentation myself, but you did not answer my question: *Was kOS supposed to use current pressure for e.g. ship:maxThrustAt(-1)? Should the value be clamped instead? (treating negative values as zero, for sure ... or producing error)* ...what about some calculation `ship:maxThrustAt(someVar-someDelta)` producing negative value?

Dunbaratu:
The point is this change fixes the bug.  It also makes a coding style change you think is relevant to the original mistake and I don't.   But since my point is that this change is irrelevant, that also means I see no reason to bother fighting it if you're more comfortable with it being this way.

The user documentation doesn't seem to make any claims one way or the other about passing negative values to ``ship:maxThrustAt()``.  Therefore I'd say that if it did anything weird with negative values before, this was probably accidental behaviour, not deliberate.  So clamping it to never go negative seems like a sane thing to do.

Dunbaratu:
Okay the docs look good.  Thanks for the edit.

I'm compiling and running some script tests on it now.

firda-cze:
1. Documented `GetThrust`/`GetIsp` in `EngineValue` and the references I found (in `VesselUtils`)
2. Joined the two `GetMaxThrust` and `GetAvailableThrust` in `VesselUtils` into one (DRY, the difference was only in the parameter `useThrustLimit `).
3. Removed `GetVacuumSpecificImpluse` and `GetSeaLevelSpecificImpulse` from `EngineValue`, unused (should use `GetIsp(0)` or `GetIsp(1)`).

P.S.: I know that *negative is special* pattern from C/C++ well enough, but this is C# (and it won't be good in *Modern C++* either).

PR #2429, opened 2019-02-07
 Fixes #2425 - Replace kOS GUI PNG icons with DDS.

### LINKED ISSUE #2425: Replace PNG images with DDS in GUI icons too
There was a PR already merged to replace textures with DDS format files (PR #2389).  However it did not do it to ALL the files.  It only did it to the decals for the 3D meshes.  It did not do it for the GUI icon files - leaving them still as PNG's.  However, it DID resize those PNG files a bit, which seems to have have caused the problem mentioned below:

So one user reported that as of kOS-v1.1.6.1,  some of our icons in the toolbar widget are coming out solid black, with an error in the player log that says things like this:
```
d3d: failed to create 2D texture id=2151 w=256 h=198 mips=9 d3dfmt=894720068 [invalid call]
Load(Texture): kOS/GFX/monitor_minimal_frame
```
And this:
```
d3d: failed to create 2D texture id=2169 w=16 h=14 mips=5 d3dfmt=894720068 [invalid call]
Load(Texture): kOS/GFX/terminal-icon-open
```
And so on.

It seems to be depending on precise factors: it has to be using Direct3D (If Unity is using OpenGL it does not happen).  It only happens on some exact video drivers for specific hardware combinations.

There was some barking up the wrong tree, claiming that the images had to have dimensions that are powers of 2, but that's not true.  As long as we produce the DDS files ourselves and don't rely on Kerbal to do the conversion for us, it works on arbitrary sizes.  It's only when we let Kerbal do it at runtime (by providing PNGs it has to convert) that it breaks on some computers.

Therefore the fix will have to be to finish the job that PR #2389 started, and convert the REST of the images to DDS as well.  As long as they stay as PNG's, they won't load into Kerbal's GameDatabase on some people's computers if they are at some resolutions.  It has been nearly impossible to pin down what those resolution restrictions for shipping them as PNGs actually are (many google searches say they must be powers of 2, but experimentation shows that's not true as they sometimes work when not powers of 2 on the same computers where they fail at other resolutions (i.e. work at height 27, fail at height 28, neither of which is a power of 2).)

Dunbaratu:
#2425 

Also, moved some of the old kOS terminal icon loading code over
to use Squad's GameDatabase that loads the textures at game
start, instead of loading the textures at runtime once per
terminal window using the Unity texture methods directly like
it used to.


Dunbaratu:
Okay removed the NOT READY, because it's ready.

lucaelin:
lgtm

PR #2432, opened 2019-02-09
stage:resources(lex) vs. sepratrons

firda-cze:
Before:
obrazek

After:
obrazek

Difference:
When any booster (that Hammer) is activated then any next-stage-booster's (including those non-active sepratron's) fuel is considered available (broken KSP cross-feed API) to all activated boosters (thus `stage:resources(Lex)` includes the fuel of them all - 37.5+6*8=85.5, not only for the Hammer = 37.5).

I beleive this is SQUAD's fault... should I file some bug report? Where and how? `part.crossfeedPartSet` does not return what is trully accessible (boosters cannot drain fuel from other boosters).

P.S.: I think that kOS should have some kind of `OrderedDictionary` for `stage:resource` to act both like `Lexicon` and `List` so that `stage:resources["SomeFuel"]` acts just like `stage:resourcesLex["SomeFuel"]`. Do you agree? I can do that (and making it callable which means `stage:resources("SomeFuel")` act the same... I hope I can do that too if you like the idea).

### DIFF
--- a/src/kOS/Suffixed/StageValues.cs
+++ b/src/kOS/Suffixed/StageValues.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using KSP.UI.Screens;
 using kOS.Module;
 using kOS.Safe.Utilities;
+using kOS.Suffixed.Part;
 
 namespace kOS.Suffixed
 {
@@ -127,9 +128,14 @@ namespace kOS.Suffixed
                 // All tanks accessible to this engine ...
                 foreach (var crossPart in part.crossfeedPartSet.GetParts())
                 {
+                    var xpart = vesselTarget[crossPart];
                     // ... that are to be separated by next decoupler
-                    if (vesselTarget[crossPart].DecoupledIn >= nextDecoupler)
+                    if (xpart.DecoupledIn >= nextDecoupler
+                        // ... but are not other boosters
+                        && !(crossPart != part && xpart is EngineValue))
+                    {
                         partHash.Add(crossPart);
+                    }
                 }
             }
             partSet.RebuildInPlace();


src/kOS/Suffixed/StageValues.cs
```diff
@@ -127,9 +128,14 @@ public void CreatePartSet()
                 // All tanks accessible to this engine ...
                 foreach (var crossPart in part.crossfeedPartSet.GetParts())
                 {
+                    var xpart = vesselTarget[crossPart];
                     // ... that are to be separated by next decoupler
-                    if (vesselTarget[crossPart].DecoupledIn >= nextDecoupler)
+                    if (xpart.DecoupledIn >= nextDecoupler
+                        // ... but are not other boosters
+                        && !(crossPart != part && xpart is EngineValue))
```
Dunbaratu: So is this basically a test for "is the resource (solidfuel) contained inside a part that is also an engine?", and that check is suposed to be a check to indicate that this is a solid booster with fuel and engine in one part?

That might work, but I wonder if any mods also make "all in one" parts that are both fuel and engine, which don't stop the cross feeding?

There's got to be something in the KSP fuel feeding system that recognizes not to feed fuel from the sepratrons into anything else, and I wonder if it's purely the fact that it's solidfuel that does it?  It might not be "don't feed this fuel to other parts because it's in a sepratron", but rather it might be "don't feed this fuel to other parts because it's solidfuel, and solidfuel has some sort of 'non-transferrable' tag on it."  I never have been able to transfer solidfuel manually (which makes sense why that's not allowed), and maybe that's the sort of thing we should be keying this on.  If we can find the flag the game uses to decide whether a resource is cross-feedable, that might differentiate solidfuel from other fuels.

firda-cze: I belive you understood my logic and my worries very well, even exactly as far as I can tell reading what you wrote. My other worries are that you will dismiss this as not-perfect as that happend to me before. My logic is: if it improves the state, not breaking what works, then it should be incorporated. But the sad truth is that some refuse to even think about it any more when they find out that it can cause problems... the point is that it does not cause problems that where not there before. It solves some but not all. Are you gonna put your head into the sand as mathuin? I hope not, but we now interract in a bit more personal level and I am doing my best to show you my logic. I am logic, imperfect, but more logical than irrational. Proove me wrong where I am wrong. I am not AI, but the more I think about it, I am not so sure.

firda-cze: I have no idea how to find such flag. Fuel inside unlit TwinBear (or how is that booster called) won't be seen after this. The key word here is **unlit**, booster that is not activated. Unlikely in real design and sepratrons do produce wrong values now. Something should be done, but if you do not like my solution then close this PR. That is at least clear signal, better then leaving this dormant for years.

Dunbaratu:
Opening an issue so the problem isn't forgotten as I cut off communication.

PR #2433, opened 2019-02-10
Fixes #2259 Boot from archive - minimal

### LINKED ISSUE #2259: Booting from archive
Booting from archive fails when loading the ship (e.g. jumping from Tracking Station or Kerbal Alarm Clock).
The problem appears to be that:
1. The boot script is not copied from archive to local disk if starting from archive.
2. Home connection is not yet ready, thus booting from archive fails.

```
[LOG 20:55:23.877] [FLIGHT GLOBALS]: Switching To Vessel Relay ---------------------- 
...
[LOG 20:55:26.697] kOS: OnStart: Orbital READY
[WRN 20:55:26.700] kOS: First Update()
[LOG 20:55:26.717] kOS: Pushing context staring with: File                 Line:Col IP   label   opcode operand
[LOG 20:55:26.722] kOS: Saving and removing 0 pointers
[LOG 20:55:26.746] kOS: FlightControlManager.AddTo ddcc7e58-aec0-4084-9ebc-f7fc2f92fc05
[EXC 20:55:26.772] KOSVolumeOutOfRangeException: The Boot volume is out of range and not accessible.
	UnityEngine.Debug:LogException(Exception)
	kOS.KSPLogger:LogException(Exception)
	kOS.Module.kOSProcessor:CheckCanBoot()
	kOS.Safe.Execution.CPU:Boot()
	kOS.Module.kOSProcessor:FixedUpdate()
```

So, it either has to copy the boot script to local drive even if starting from archive is enabled,
or there needs to be some mechanism added to wait until the ship is fully loaded or maybe even delay the boot until it can trully boot from archive (when connection is lost - wait for connection, signal that in the terminal allowing to cancel the boot by ^C).

Better explanation in the terminal would also help, but that is #2121

ghost:
Will delay booting if needed using `YieldFinishedCompileBoot`.

### DIFF
--- a/src/kOS.Safe/Execution/CPU.cs
+++ b/src/kOS.Safe/Execution/CPU.cs
@@ -136,15 +136,13 @@ namespace kOS.Safe.Execution
                 shared.Screen.Print(bootMessage);
             }
 
-            if (!shared.Processor.CheckCanBoot()) return;
-
             VolumePath path = shared.Processor.BootFilePath;
             // Check to make sure the boot file name is valid, and then that the boot file exists.
             if (path == null)
             {
                 SafeHouse.Logger.Log("Boot file name is empty, skipping boot script");
             }
-            else
+            else if (shared.Processor.CheckCanBoot())
             {
                 // Boot is only called once right after turning the processor on,
                 // the volume cannot yet have been changed from that set based on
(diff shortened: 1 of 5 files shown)


Dunbaratu:
This will get merged because it's not filled with refactor nonsense that doesn't improve anything, and because you did *just* spend the effort separating it yesterday.

firdacz:
Five years later....

PR #2467, opened 2019-03-20
Fixes #2457 by removing not-working IsAvailable checks.

### LINKED ISSUE #2457: With RT, Inter-vessel comms failing if other vessel is beyond load distance.
Our docs about inter-vessel comms imply that delay over lightspeed is a thing with Remote Tech installed.  This implies sending to a distant vessel should be possible.  It claims that you can do ``Vessel("foo"):connection:delay`` to find out this number in seconds, which would be pointless if the other vessel is within loaded range just a few km away.

But it does not work now.  Now any call to ``vessel("foo"):connection:isconnected`` will *always* return false when the other vessel is more than a few km away so that it is unloaded, regardless of what sorts of antenna or probe cores or "ModuleSPU"'s it has on it..  **As far as I can tell, the root cause of this is that the function call ``HasFlightComputer(Guid)`` in the RemoteTechApi is now behaving differently than it used to**.  I don't normally use this feature and only just discovered this as I was trying to test another change I was working on.

It seems that now, RemoteTech's ``HasFlightComputer(Guid)`` is always returning False for any unloaded vessel, when it looks like before it used to work for distant vessels.  Because of that change, kOS is claiming no connection exists to the vessel, and the ability to send a message from the current vessel to a distant vessel got broken.

I know that used to work because I know kOS has a lot of code in it to handle the fact that the target vessel might not have its PartModules loaded when sending it a message, and nothing in our source code file that calls Remote Tech's API has changed since 2016.

Dunbaratu:
Fixes #2457 

The RT call HasFlightComputer(guid) will always return false
for any unloaded vessel.  Our protection check, IsAvailable(guid)
was there to stop us from calling API routines on vessels where
doing so would throw nullref, but it also used HasFlightComputer(),
which has this undesirable side-effect of failing on distant
unloaded vessels.

In the meantime since that code was written, Remote Tech has
made changes to their API routins that added protections against
these nullrefs so we can now unconditionally call them without
worry of a crash.  That removed the need for us to call
IsAvailable(guid) all over the place, but we still call
the other version of IsAvailable() (no guid), that is just
checking for whether RT exists period and not checking for
any specific vessel.

(Also, shifted some Infinity checks to the slightly more proper
Double.IsInfinity, just because IEEE suggests using that over
direct value equals).

### DIFF
--- a/src/kOS/AddOns/RemoteTech/Addon.cs
+++ b/src/kOS/AddOns/RemoteTech/Addon.cs
@@ -3,6 +3,7 @@ using kOS.Safe.Encapsulation.Suffixes;
 using kOS.Suffixed;
 using kOS.Suffixed.Part;
 using System.Linq;
+using System;
 
 namespace kOS.AddOns.RemoteTech
 {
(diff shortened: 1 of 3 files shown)


src/kOS/AddOns/RemoteTech/Addon.cs
```diff
@@ -30,31 +31,32 @@ private void InitializeSuffixes()
...
+            return Double.IsPositiveInfinity(waitTotal) ? -1 : waitTotal;
         }
 
         private static BooleanValue RTAntennaHasConnection(PartValue part)
         {
             bool result = false;
 
+            // IsAvailable(Id) is only able to return True on loaded vessels, but this
+            // is a test for a specific PART on a vessel, and individual parts on
+            // vessels don't exist when the vessel is unloaded anyway.
```
Dunbaratu: The above comment is my justification why here the IsAvailable(id) is still being used.

src/kOS/AddOns/RemoteTech/RemoteTechHook.cs
```diff
@@ -24,6 +24,15 @@ public static IRemoteTechAPIv1 Instance
...
 
+        /// <summary>
+        /// True if Not ONLY does the vessel exist and Remote Tech is enabled,
+        /// but ALSO, the vessel is *loaded* and has a ModuleSPU.  Note that if
+        /// the vessel does have a ModuleSPU but is *NOT* loaded (i.e. it's outside
+        /// the physics bubble), then this will return False *no matter what*,
+        /// because Remote Tech removes the Flight Computer from all distant vessels.
+        /// </summary>
+        /// <param name="vesselId"></param>
+        /// <returns></returns>
```
Dunbaratu: No change here - just felt it was important to mention this rather important fact for any future mantainers.

PR #2493, opened 2019-04-22
Fix connectivity manager appearance

madman2003:
I tried to explain how the dialog position was chosen, because this is really non-trivial.

Fixes the connectivity manager appearing off-screen for me.


Dunbaratu:
Testing this it all looks good.  I even fiddled with 200% UI scale on minimum resolution (it looks quite bad that way, but it does put the dialog box in the right place.)

Those dialog box parameters make no sense without being documented.  Thanks for doing all the trial and error to reverse-engineer it and provide the "missing documentation".

PR #2498, opened 2019-04-25
Support for parts with multiple engines which aren't multi-mode.

RCrockford:
This fixes an issue encountered with RO engines which have multiple engine modules in a single part (for main engine plus built-in verniers), but which aren't multi-mode engines.
Engine parts will support any number of engine modules, with aggregated values returned for the various suffixes.
Multi-mode engines continue to behave as before.

### DIFF
--- a/src/kOS/Suffixed/Part/EngineValue.cs
+++ b/src/kOS/Suffixed/Part/EngineValue.cs
@@ -146,13 +146,19 @@ namespace kOS.Suffixed.Part
         public void Activate()
         {
             ThrowIfNotCPUVessel();
-            FilteredEngineList.ForEach(e => e.Activate());
+            foreach (ModuleEngines e in FilteredEngineList)
+            {
+                e.Activate();
+            }
         }
 
         public void Shutdown()
         {
             ThrowIfNotCPUVessel();
-            FilteredEngineList.ForEach(e => e.Shutdown());
+            foreach (ModuleEngines e in FilteredEngineList)
+            {
+                e.Shutdown();
+            }
         }
 
         // The following functions calculate the correct Isp for multiple engines, regardless of individual performance.
--- a/src/kOS/Suffixed/Part/EngineValue.cs
+++ b/src/kOS/Suffixed/Part/EngineValue.cs
@@ -13,17 +13,22 @@ namespace kOS.Suffixed.Part
     [kOS.Safe.Utilities.KOSNomenclature("Engine")]
     public class EngineValue : PartValue
     {
-        public ModuleEngines Engine1 { get; private set; }
-        public ModuleEngines Engine2 { get; private set; }
+        public List<ModuleEngines> EngineList { get; private set; }
         public MultiModeEngine Multi { get; private set; }
         public GimbalFields Gimbal { get; private set; }
-
-        public ModuleEngines Engine {
+        public IEnumerable<ModuleEngines> Engines {
             get {
-                return Engine2 == null || Multi.runningPrimary ? Engine1 : Engine2;
+                if (EngineList.Count > 0 && Multi != null)
+                {
+                    return EngineList.Count < 2 || Multi.runningPrimary ? EngineList.GetRange(0, 1) : EngineList.GetRange(1, 1);
+                }
+                else
+                {
+                    return EngineList;
+                }
             }
         }
-        public bool MultiMode { get { return Engine2 != null; } }
+        public bool MultiMode { get { return Multi != null; } }
         public bool HasGimbal { get { return Gimbal != null; } }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Dunbaratu:
This looks good on first inspection (I haven't had a chance to try it yet because I was sick the last few days.)

One comment - I think "ignition" should be true if *any* of the sub-engines inside this engine are ignited, and not require that *all* are ignited.  At any rate, it would be a good idea to make sure the user documentation makes it clear to the user what value to expect these to return in the case of an engine with multiple sub-engines some of which are flamed out or ignited and some of which are not.

RCrockford:
Agreed, I was a bit unsure on the boolean functions so just went with "all" for now.
I think there are cases where you might want it either way so it's a bit of a tricky decision. Maybe go with any for ignition and flameout and document it, and add a new suffix for all?

Dunbaratu:
I think "any" will more accurately be right most of the time.  In the odd case where there are two engines, one with flameout but one engine still going, that returning "true" for flameout will break a lot of scripts that are looking for "flameouts" to decide to stage and decouple the engine away.

At any rate, I think I'll just make that edit (to "any"), test it, and accept it like that if it works.  I want to get something out today or tomorrow because we still don't officially claim to be KSP 1.7 compliant and I want to fix that with a new release since as far as I can tell everything is just fine with 1.7.

Dunbaratu:
I just realized you said this was for RO.  That means I don't have to get it into the KSP 1.7 release I want to make today, since a KSP 1.7 release doesn't help you anyway if you need this for Realism Overhaul.  I would like to work on a "backport" for Realism Overhaul soon, and it would need to be merged for *that*, but until then it doesn't matter because you can't use it in RO anyway if it's compiled for KSP 1.7.

Dunbaratu:
@RCrockford - to help with testing, can you mention a specific engine that needs this PR, and which mod supplies that part?  (Not just "RO" but the specific mod within RO that provides the engine in question.)

If the mod in question is available without the rest of RO installed, that will make this easier to test in isolation, and could also mean this PR has a utility beyond just RO, i.e. it could be useful in KSP 1.7 even though RO isn't released for KSP 1.7 yet.

RCrockford:
@Dunbaratu I have pushed a change with the any semantics rather than all (except for throttle lock where I felt it was more useful to tell if all of them are locked).

For testing, the engine is the RD-108 from RealEngines (https://forum.kerbalspaceprogram.com/index.php?/topic/171335-realenginespack-201-close-010418/). It is modified by RO, but I think it should work without it installed.

Also, for RO I have a 1.3.1 backport of develop in my fork, I can submit a pull request for this if you would like me to.

Dunbaratu:
Oh, by the way, I need to write an issue for this - as in what is the currently incorrect behaviour this is meant to fix?  I just did a small bit of testing with RD-108 on KSP 1.7 (RealEngines isn't officially supporting KSP 1.7 but installing it anyway seems to work), and what I see (without this pull request) is that both engines inside the one part *are* reported in "list engines", but it's showing it as two clones of the same center engine, instead of one being the center engine and one being the side gimbal engines as it really should be.

IMO it would be fine to be two different engines like that if they were properly reported as two different engines rather than the same engine twice.

RCrockford:
@Dunbaratu Looks correct to me. I have commented on the issue with multiple entries showing in list engines in the issue.

Dunbaratu:
I'm in the middle of regression testing and noticed there may be a second bug that predates your PR. 

Older versions of kOS would report a multi-mode engine as 1 engine in LIST ENGINES. Newer versions of kOS (even without your PR merged) report a  multi-mode engine as 3 engines in LIST ENGINES - they appear to be 3  references to the same engine and all 3 work identically though (they  are not the different modes of the engine, all 3 of them will report the  new ":mode" when you toggle just one of them, so they are the same  engine just referenced 3 times.  This is clearly wrong.)

I'm trying to trace down exactly when that happened.  It wasn't  noticed at the time but we may have broken something in the last few  months. Therefore "try to make it match the previous behavior" might not be  the right thing to do because the previous behavior seems to also have  been a bug.

Dunbaratu:
That duplication of engine modules that you noticed (and fixed) in PartsToList wasn't always that way.  It used to return just one item for a multi-mode engine then something changed in the stock game and we didn't notice it, it seems.  I suspect it's this:  Of the 3 engine modules, only one of them was an IEngineStatus and the other 2 were not so it didn't end up adding all 3 of them.  But SQUAD might have changed that to make all 3 of them be IEngineStatus's.

Okay I'm fine with that part of your PR then.

Dunbaratu:
@RCrockford - I did notice another problem in this PR, though, in regression testing that *does* need to be addressed.  I hope it's not too complicated to fix.

With this PR installed, try these steps:

Build a ship with a "Panther" jet engine - it's a stock multi-mode engine with a "wet" and a "dry" mode, so that means it has multiple engine modules to toggle between but they shouldn't both be active at the same time.

Try this (assuming that engs[0] is the panther engine):
```
list engines in engs.
engs[0]:shutdown.
engs[0]:activate.
```
If you look at the staging list with the fuel indicators, you see that it will now have two fuel bars, just like the RD-108 multiple engine would, because it activated both of the engine's PartModules just like it would have for the RD-108 engine.  But this is a multi-mode engine that is supposed to toggle modes.  It's not supposed to have both of its engine PartModules active at the same time.  I suspect both are drawing fuel at the same time and both are adding thrust at the same time, which really shouldn't be happening.

Presumably the same problem exists with the other multi-mode engines but I only tested with the Panther for now.  @hvacengi had a quick look at the PR code and warned me that he thought this problem might exist based on what he read, and when I tested it, lo he was right.  (Thanks for the sharp eye, hvacengi.)

Dunbaratu:
@RCrockford based on the fact that I am suspicious that KSP itself might have changed something important here between versions, I think it would be a good idea to also test if the above problem behaves differently in KSP 1.7 than it does in KSP 1.3.1 (I did not try your PR in KSP 1.3.1.  I tried it in KSP 1.7 by installing the RealEngines on KSP 1.7 to get the RD-108 engine.)

Dunbaratu:
Oh, by the way, I didn't explicitly say it yet, so let me say it now: Thank you for the contribution and the effort you're putting into this.  I don't want to come down all negative just because I'm finding a problem.    I'm just being a stickler for regression checking because I'm the one who has to field any complaints if people's previously working scripts break in the next update we put out.

RCrockford:
Hmm, interesting findings on the Multimode issue. I'll see if I can give that a test in 1.3.1.

Looking at the diff, I think instead of:
            `EngineList.ForEach(e => e.Activate());`
it should be
            `Engines.ForEach(e => e.Activate());`

Does that look more valid to you?

Dunbaratu:
> Does that look more valid to you?

From the look of it, I think that would fix it.  It may also be a good idea to do a file-wide rename on "EngineList" and "Engines" to make the names more indicative of what the relevant difference is between them for any future devs.  A person looking at the code wouldn't immediately see what the meaningful difference is between calling something "EngineList" and calling something "Engines".  They're just two ways to say "collection of more than 1 engine".

Maybe something like ``RawEngineList`` versus ``CulledEngineList`` or something along those lines would make it clearer.  At least it gives a developer a good hint that one is a subset of the other for some relevant reason, which will make them go look at their definitions to see why and which one they should use.

I can do that as a later refactor though.  You don't have to include it in this PR.

RCrockford:
I have pushed a commit with the regression issues fixed. Also I did a quick rename on the lists. ;)

Dunbaratu:
Okay looks good.

I'm ready to merge it.

PR #2510, opened 2019-05-06
Fixes #2499 disallow unset bindings, also document.

### LINKED ISSUE #2499: UNSET breaks bound variables like GEAR, TARGET, etc.
To recreate:

- Have a savegame in which there is at least *something* in kerbin orbit already.
- Put a new vessel on launchpad with a kOS unit on it.
- While controlling that new vessel on launchpad, go to map view and use mouse to select a ship in orbit as target.
- In the terminal window of the new vessel, type UNSET TARGET.

It throws a NullRef, and from then on, SET TARGET will not work anymore, and any attempt to get at the "real" target fails.

(I have no idea if this is related to #2495 or not.  It probably isn't but it could be.)

Dunbaratu:
Fixes #2499 by disallowing the UNSETing of things that were not created by the user.

In trying to document the changes to UNSET, I found out that UNSET was not even really documented at all prior to this, so I also added UNSET documentation.

### DIFF
--- a/doc/source/commands/flight/systems.rst
+++ b/doc/source/commands/flight/systems.rst
@@ -59,7 +59,7 @@ RCS and SAS
 
         SET SASMODE TO value.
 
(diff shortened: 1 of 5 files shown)


doc/source/commands/flight/systems.rst
```diff
@@ -59,7 +59,7 @@ RCS and SAS
...
+    It is the equivalent to clicking on the buttons next to the nav ball while manually piloting the craft, and will respect the current mode of the nav ball (orbital, surface, or target velocity - use NAVMODE to read or set it).  Valid strings for ``value`` are ``"PROGRADE"``, ``"RETROGRADE"``, ``"NORMAL"``, ``"ANTINORMAL"``, ``"RADIALOUT"``, ``"RADIALIN"``, ``"TARGET"``, ``"ANTITARGET"``, ``"MANEUVER"``, ``"STABILITYASSIST"``, and ``"STABILITY"``.  A null or empty string will default to stability assist mode, however any other invalid string will throw an exception.  This feature will respect career mode limitations, and will throw an exception if the current vessel is not able to use the mode passed to the command.  An exception is also thrown if ``"TARGET"`` or ``"ANTITARGET"`` are used when no target is set.
```
Dunbaratu: Github's diff'er didn't highlight the relevant part of the line like it usually does.  Hint to reviewer: The only difference is in the last sentence of this paragraph.

PR #2519, opened 2019-05-20
Fixes #2518 Fix for empty part module field names.

### LINKED ISSUE #2518: Part module fields sometimes appear with no name
In certain cases part module fields can appear with no name if BaseField.guiName has zero length.
In this case it would be better to fall back to BaseField.name; the PAW in KSP appears to do this.

Repro: Latest RO in KSP 1.6.1, engines show a field at the top of the PAW "pressureFed". In kOS this is shown as an unnamed field.

RCrockford:
Fixes #2518.

Adds a fall back to BaseField.name for part module field names in the event that BaseField.guiName is empty.

### DIFF
--- a/src/kOS/Suffixed/PartModuleField/PartModuleFields.cs
+++ b/src/kOS/Suffixed/PartModuleField/PartModuleFields.cs
@@ -1,4 +1,4 @@
-﻿using kOS.Safe.Encapsulation;
+using kOS.Safe.Encapsulation;
 using kOS.Safe.Encapsulation.Suffixes;
 using kOS.Safe.Exceptions;
 using kOS.Suffixed.Part;
@@ -196,7 +196,7 @@ namespace kOS.Suffixed.PartModuleField
             {
                 returnValue.Add(new StringValue(string.Format(formatter,
                                               IsEditable(field) ? "settable" : "get-only",
-                                              field.guiName.ToLower(),
+                                              field.guiName.Length > 0 ? field.guiName.ToLower() : field.name.ToLower(),
                                               Utilities.Utils.KOSType(field.FieldInfo.FieldType))));
             }
             return returnValue;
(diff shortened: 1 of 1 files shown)


src/kOS/Suffixed/PartModuleField/PartModuleFields.cs
```diff
@@ -196,7 +196,7 @@ protected virtual ListValue AllFields(string formatter)
             {
                 returnValue.Add(new StringValue(string.Format(formatter,
                                               IsEditable(field) ? "settable" : "get-only",
-                                              field.guiName.ToLower(),
+                                              field.guiName.Length > 0 ? field.guiName.ToLower() : field.name.ToLower(),
```
Dunbaratu: This seems to fix the problem only for the suffix ``:ALLFIELDS``.   But the name it returns still wouldn't be recognized in all the other places where you can use these names, like ``:ALLFIELDNAMES``, or ``:HASFIELD``, or ``:GETFIELD``, or ``:SETFIELD``, etc.

For those, they're still only looking at the ``guiName``.  So this would make the user think kOS will let them use the name when it won't.

To really fix the problem would require coming up with some kind of method that can be re-used in all the places the ``guiName`` is being used now, to perform this same "use guiName unless it's blank then use not-gui name." technique all over the place.

Dunbaratu: I also wonder if the same logic needs to be applied not just to Field names but also to Event names (buttons on the menu) and Action group actions.

RCrockford: Hmm, good point, I'll do a bit more testing on this. I've not seen it for actions and events, but that's not too say it couldn't happen.

RCrockford:
I've extended the field name lookup changes to cover all uses of fields and also events and actions.

Dunbaratu:
The update looks good by eye.  I plan to put in a little time to kOS tomorrow, and I'll have a look at it in more detail (as in, install it and test it out) then.

Dunbaratu:
I don't really have the time to install all of RO to test this right now.  (The only mod I know of that has this problem of not using the guiName is realfuels), so I'm going to have to trust you that you tested this and it worked.

PR #2543, opened 2019-06-12
Fixes #2531 this is ugly

### LINKED ISSUE #2531: Wrong conversion for PartModuleField values using Robotic Parts
Trying to set the PartModuleFields for the DLCs robotic parts (ie. Target Angle), causes values to be clamped between 0 and 100 and rounded to the nearest integer. This is wrong for most sliders in the new DLC.

Cause of the issue seems to be https://github.com/KSP-KOS/KOS/blob/d6258fad1c4a35e2fa3b885e3fc0879b9c140ca3/src/kOS/Suffixed/PartModuleField/PartModuleFields.cs#L174

`UI_FloatRange` claims that its range is always 0 to 100 with a step-size of 1.

Ignoring the clamp entirely seem to fix the issue and still cause values to be properly converted to the UIs possible values, maybe Squad moved the logic from the UI to the actual API, causing this step to be unnecessary on kOSs side?

Dunbaratu:
Fixes #2531 

I still believe proper support of the Breaking Ground DLC will require much more work than just this, using special case part abstractions similar to the engine part abstractions we have, but this at least fixes the problem with the slider range limits being reported wrong by the Breaking Ground DLC's UI sliders.

These new sliders have min/max values that can be different per part depending on what the limits were set to in the VAB.  That system seems to utterly bypass the usual field range values, rather than *setting and using* them as I'd prefer. 

EDIT: Also, this includes some new suffixes so a script can read the min/max values of a robot servo part module.

### DIFF
--- a/doc/source/structures/vessels/baseservo.rst
+++ b/doc/source/structures/vessels/baseservo.rst
@@ -1,0 +1,16 @@
+.. _baseservo:
+
+BaseServo
+=========
+
+This is a special type of :struct:`PartModule` used to handle the robotic
+servos that came with the Breaking Ground DLC.  If you examine a PartModule
+that controls such a part, it will be this type.
+
+This type has no extra suffixes (yet??) beyond those that come with every
+:struct:`PartModuile`.  The reason it has to be a separate type is that
+it needs to perform some strange work under the hood to mimic how normal
+PartModules work.  (The Breaking Ground DLC PartModules have some strange
+implementation changes about how their Part Action Window sliders work
+that required extra work for kOS to support.)
+
--- a/doc/source/structures/vessels/partmodule.rst
+++ b/doc/source/structures/vessels/partmodule.rst
@@ -47,6 +47,15 @@ Once you have a :struct:`PartModule`, you can use it to invoke the behaviors tha
         * - :meth:`GETFIELD(name)`
           -
           - Get value of a field by name
+        * - :meth:`GETFIELDMIN(name)`
+          -
+          - The field's minimum legal value, if scalar
+        * - :meth:`GETFIELDMAX(name)`
+          -
+          - The field's maximum legal value, if scalar
+        * - :meth:`GETFIELDSTEP(name)`
+          -
+          - The mandatory "rounding" increment, if scalar
         * - :meth:`SETFIELD(name,value)`
           -
           - Set value of a field by name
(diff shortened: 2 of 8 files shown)


Dunbaratu:
One problem just discovered:  using SETFIELD() to move a servo's target slider seems to ignore the part symmetry grouping, UNLIKE using the slider by mouse.  (If clicking the slider by mouse, and 4 parts are locked together in symmetry, all 4 parts move.  If moving the same slider with SETFIELD(), and 4 parts are locked together in symmetry, it still only moves the 1 part not all 4.  Again, there is some wacky magic happening where mouse clicks on the slilder don't act the same as the KSPField does.

Dunbaratu:
This is so frustrating.

I am coming to the awful conclusion that we just cannot let people use the Part Action Window directly for the Breaking Ground's robotic servos.  Even to the point of maybe having to utterly ban attempts to SETFIELD on them.  (i..e make ``BaseServoModuleFields`` *remove* the standard PartModuleFields suffixes and instead map them to error messages explaining the situation.)

Why?  Because I keep coming across cases where they only work right if the Part Action Window happens to be open.  If it's closed, then things get mysterious.  I suspect something tries to "save cycles" in the servo's code by truncating work when the PAW is closed, under the false assumption that it's impossible for the servo to be told to move if the window is closed.

Example: Pistons that jitter like mad after a few consecutive SETFIELD("target extension", [..] ) calls, in utter disregard of what value you're setting it to (it keeps resetting to zero extension for one physics frame, then popping back out to the value you picked for like the next 2 or 3 frames, then resetting to zero again (which you did not set it to), and repeating the pattern.) - the instant the PAW for the servo is opened in the game, the servo becomes sane (the kOS code is still doing the exact same thing - no difference - but the random resets to zero deflection stop happening).

tl;dr - it's not JUST the "current Extension" value that is messed up when the PAW is closed - the SETFIELD on the "target extension" *also* comes out all wrong if the PAW has been closed a while (but works briefly it was open recently).

I think we may need to just write a special case set of code and disallow the normal PartModuleFields suffixes for these, which hurts.

Dunbaratu:
Okay I found the specific exact things that need to happen to trigger the bug (and I AM calling it a bug now) in Breaking Ground's robotic servos:

https://forum.kerbalspaceprogram.com/index.php?/topic/130742-15x-to-122-krpc-control-the-game-using-c-c-java-lua-python-ruby-haskell-c-arduino-v048-28th-october-2018/&do=findComment&comment=3618637

Essentially you have to set BOTH the Motor boolean field to true (enabled) AND the Target Extension, to trigger the bug.  (Even if the Motor boolean field was already true anyway before you set it).  This appears to be confusing the heck out of the DLC.)

Dunbaratu:
I will likely be leaving this PR un-merged, as I got advice from TriggerAU that works more cleanly and doesn't require this ugly solution.  The better PR is #2554.

PR #2544, opened 2019-06-12
Principia addon support

RCrockford:
Adds a new addon for Principia, allowing direct (read only) access to the Principia flight planning system.
Will require a new build of Principia, but should be safely dormant until then.


Dunbaratu:
@RCrockford  - This looks like it took some work.  Thank you for the contribution.

I know next to nothing about Principia so I'll have to largely just trust you that this works if I want to merge it.  (I'm not really in a position to test it against the mod.)

Along those lines, I really don't know how to write the user-facing documentation for this (basically telling people about these new suffixes and what they mean).  I'll have to rely on you to fill that in too.

If you want, I can create stubs of the user docs with a lot of "TODO" paragraphs in them, and hand that back to you to ask you to fill in all the TODO's - that way I deal with the Sphinx formatting and Markdown and just leave it up to you to type in the actual information.

Dunbaratu:
@RCrockford The automatic build tool 'Travis" is claiming this won't build, and I think I see why:

Tt gives this error:

CSC: error CS2001: Source file '/home/travis/build/KSP-KOS/KOS/src/kOS/AddOns/Principia/PRManoeuvre.cs' could not be found.

It took me a while to notice why it was complaining - because "that file *IS there", I kept thinking.

I finally noticed the capitalization.  The csproj file in this PR is calling it "PRManouevre.cs" with a capital "R", but the actual filename is "PrManoeuvre.cs" with a lowercase "r".  Travis is running on a Linux server (case sensitive filenames).  This will probably have to be fixed because some contributing kOS devs compile kOS on Linux.

RCrockford:
Ah, I'd wondered why the Travis build was falling, I didn't even think it might be that. Too much time on windows. I'll get the capitalisation fixed.

I'll write some docs for the new addon suffixes over the next couple of days.

RCrockford:
I have committed a fix for the Travis build, and some documentation.
I've not tried to build the docs with Sphinx (I don't think I can install it here), but I think it should be fine.

Dunbaratu:
Okay I reviewed it an the code and docs look good.
Like I said, I can't test it right now, but I can verify that it won't cause any problems for a stock user so I'll be merging it later.  I hope you can field any issues that come up about it later, if any do, @RCrockford.

Dunbaratu:
@RCrockford I already merged this, as it is acceptable as-is, but as a followup - would be be a hassle to do either of these two things as a followup?

1: Hypothetically changing this:
```
 public class PRManoeuvre : Structure
```
to this:
```
 public class PRManoeuvre : Node
```
?

(i.e. make it so that PRManoeuvre's are a class derived from stock kOS Maneuver Nodes).

Thus ``obj:isType("Node")`` would return true for both.

2: Adding aliases so where a script has to say "Manoeuvre", it could also be spelled "Maneuver"  and work that way too.  (For an example of this, search the code for the word "Color" and the word "Colour" to see how we have some similar spelling aliases implemented already.)

RCrockford:
1. I like the idea of that, but PR Manoeuvre has a much more limited interface than Node. If I inherit from node it would gain all suffixes from node which then wouldn't behave as expected. That might be more confusing to the user than just having them distinct.

2. I can definitely go through and add aliases. I used the spelling there as it's consistent with Principia.

Dunbaratu:
@RCrockford I reverted this PR by reverting the merge commit I did.  My commit that reverted the commit is :  #cbdc1a7

I have no clue how to tell github to remove the "Merged" label on this PR, though.  Oh well.  If we want to put this back in place again, it should be possible to re-make the PR for it.

PR #2546, opened 2019-06-12
Fixes #2545. Adds control locking from rp-1. 

### LINKED ISSUE #2545: Locked avionics state in RP-1 is ignored.
In rp-1 if you have insufficient avionics for the mass of the craft the controls are locked. kOS currently ignores this and allows control anyway.
It would be nice to lock out controls, similar to how the connectivity manager blocks access.

RCrockford:
Fixes #2545.
Adds a read of the rp-1 control locking mechanism so that craft with insufficient avionics are uncontrollable. No effect for non rp-1 games (as the lock will not exist).

### DIFF
--- a/src/kOS/Module/kOSVesselModule.cs
+++ b/src/kOS/Module/kOSVesselModule.cs
@@ -394,6 +394,11 @@ namespace kOS.Module
         /// <param name="c"></param>
         private void UpdateAutopilot(FlightCtrlState c)
         {
+            // Lock out controls if insufficient avionics in RP-0.
+            ControlTypes RP0Lock = InputLockManager.GetControlLock("RP0ControlLocker");
+            if (RP0Lock != 0)
+                return;
+
             if (Vessel != null)
             {
                 if (childParts.Count > 0)


src/kOS/Module/kOSVesselModule.cs
```diff
@@ -394,6 +394,11 @@ private void CheckRehookAutopilot()
         /// <param name="c"></param>
         private void UpdateAutopilot(FlightCtrlState c)
         {
+            // Lock out controls if insufficient avionics in RP-0.
+            ControlTypes RP0Lock = InputLockManager.GetControlLock("RP0ControlLocker");
```
Dunbaratu: Question:  Is this RP0 lock still called RP0Lock in RP-1?  And if so, does that constitute a promise from the RP folks that this will be the public API call name of it for other mods to check for through all future versions?  I suspect it's only called ``RP0Lock`` purely because nobody noticed the name still said "RP0" when RP-1 came out.  That sounds like the sort of thing that might get fixed when someone notices it, and then that breaks mod compatibility for a mere name change.  Do you know if this is going to be the permanent name for this?

RCrockford: The name is taken directly from the current rp-1 source.
That said it's entirely possible someone could unknowingly change it in the future. I might restructure the way this works so that kOS provides a function for rp-1 to call to lock controls. That way it is obvious that there is an external interaction.

Dunbaratu: I got a reply in the RP-0 Github about this - they claim it will remain "RP0" forever and it's safe to use this name.

Dunbaratu:
@RCrockford This looks like it might have the effect of locking kOS's inputs to whatever they were *before* when the avionics support goes away.  Can you test what happens if you lose avionics control WHILE KSP was in the midst of moving the controls?  Do they get stuck in the moved position?  (i.e. imagine a decoupling makes a probe core go away and what's left is not sufficient to allow Avionics to work on the mass that remains.  If kOS was in the middle of, say, pushing YAW 30% left just when that happened, will the controls keep on pushing 30% left from then on?

RCrockford:
I have tested this (it's actually very easy as controls are unlocked on the pad then lock as you launch). Raw steering controls return to 0 when control is locked. As soon as control unlocks they return to their previously set values. In the case of cooked steering, controls return to zero and then cooked steering resumes on unlock.

PR #2549, opened 2019-06-15
Fixes #18

### LINKED ISSUE #18: Tertiary operator ("?")  (with short-circuit logic) to protect against infinity in a single expression.
The change for #14 broke the ability to store infinity in a variable.

Now if there is going to be a chance that there might be a divide by zero anywhere in an expression, that has to be checked for ahead of time, rather than allowing it to return infinity.  As soon as any math expression or sub-expression returns infinity, the script dies.

That means things like this:
`set var to x / y.`

If there's a chance Y might be zero, have to be turned into this instead:

```
set var to bigNum.
if y > 0 or y < 0 { 
  set var to x / y.
}.
```

(I'd use if/else instead for that if there was an 'else'.)

This makes writing lock expressions currently a bit of a pain because there's no mechanism currently to protect a lock expression with an "if" like you can with a 'set'.  The value of a lock must be written out as a single expression, which you must assume could be executed at any random time not directly under your control (because flybywire might execute it even if you don't).

So what I'm talking about is the ability to do this:

lock var to x /y, unless y happens to be zero at this instant, in which case lock it to bigNum instead.

With a 'set' command, you can control when the check happens as shown above.

But since a lock has to all be one expression, the technique used for the set isn't quite right.

If something like C's trinary operator "?" was implemented, then you could protect lock expressions like this:

```
lock var to ( (y = 0) ? bigNum : x /y)
```

Obviously the syntax there isn't quite right for kosscript's terms.  Perhaps something like:

```
lock var to CHECK y = 0 TRUEVAL bigNum FALSEVAL x/y .
```

might look a bit more 'kosscript'-like.

As it stands, the only way to protect the expression is to lie about the contents of 'y' and pretend it's something like 0.000001 whenever it's really 0.0, which if you're using it elsewhere besides this lock expression, could be messy.

That's what I'm doing for now in my script to deal with this new change.  When the air pressure is really 0, I'm pretending it's 0.000001 just to keep the script from barfing out.

Dunbaratu:
This has been sitting a long time, but was stuck on not
 liking any syntax choice - then it came up in Discord
 again and people seemed to like this, so I finally did
 it:

    CHOOSE expr1 IF bool_expr ELSE expr2.


kerboscript_tests/ternary.ks
```diff
@@ -0,0 +1,38 @@
...
+set x to true.
+set y to false.
+set a to 1.
+print "  Next 2 lines should be 'A':".
+print "  " + (choose "A" if (choose x if a=1 else y) else "B").
+print "  " + (choose "A" if choose x if a=1 else y else "B"). // same without parens should work.
+print "  Next 2 lines should be 'B':".
+print "  " + (choose "A" if (choose x if a=2 else y) else "B").
+print "  " + (choose "A" if choose x if a=2 else y else "B"). // same without parens should work.
+
```
teleksterling:
```suggestion
+
+print "Third nesting case - nesting choose in the first term.".
+print " Next two lines should be 'A'.".
+print "  " + (choose (choose "A" if true else "B") if true else "C").
+print "  " + (choose choose "A" if true else "B" if true else "C"). // same without parens should work.
+print " Next two lines should be 'B'.".
+print "  " + (choose (choose "A" if false else "B") if true else "C").
+print "  " + (choose choose "A" if false else "B" if true else "C"). // same without parens should work.
+print " Next four lines should be 'C'.".
+print "  " + (choose (choose "A" if true else "B") if false else "C").
+print "  " + (choose choose "A" if true else "B" if false else "C"). // same without parens should work.
+print "  " + (choose (choose "A" if false else "B") if false else "C"). // result of inner test should affect result
+print "  " + (choose choose "A" if false else "B" if false else "C"). // same without parens should work.
+
+
```

PR #2553, opened 2019-06-24
[Pending documentation] Lexicon suffixes (Fixes #2551)

### LINKED ISSUE #2288: Do not make list#x deprecated it works for OOP
Hello

Not sure this discussion is not appeared in the past.
I looked through issues but found nothing similar.
Sorry if is duplicate.

I noticed that list#x marked as deprecated here
http://ksp-kos.github.io/KOS_DOC/structures/collections/list.html?highlight=deprecated
Accidently I found that # works also for lexicons.
And voila - we get good OOP syntax out of box right now.

```
function kob_atm_ascent_new {
    declare local parameter myMaxQ.
    local obj is lexicon().

    // class attributes
    set obj#q_lim to myMaxQ.
    set obj#q_cur to 0.
    set obj#q_factory to 0.
    ... more attributes

    // class methods
    set obj#update to kob_atm_ascent_update@:bind(obj).
    set obj#test to kob_atm_ascent_test@:bind(obj).
}

function kob_atm_ascent_test{
    declare local parameter obj, mult.
    return obj#q_cur * mult.
}

// Update variables according to ship current state
// Param1: object
function kob_atm_ascent_update {
    declare local parameter obj.
    obj#q_factor to obj#q_cur * 12345.
}
```
class usage example somewhere in code
````
    local ascent is kob_atm_ascent_new(15).	
    until ship:altitude < 60 {
        ascent#update().
        print round(ascent#q_cur, 2) + at ( 0,  1).
        ...
    }
````
That looks exciting to me.
It is absolutely true OOP like in other scripting languages.
Just don't depreciate it.

Thanks for excellent mod.

### LINKED ISSUE #2551: Lexicons could use suffix syntax by overriding GetSuffix().
It has been asked for many times before to make it so that lexicons can use the syntax of suffixes, like so:

```
set my_var to lexicon(
  "key0", 0,  // key looks like an identifier
  "key1", 100,  // key looks like an identifier
  "key2", 200, // key looks like an identifier
  "key with spaces", 300, // key cannot be an identifier
  List("key that isn't even a string", 0), 400 // key cannot be an identifier
).
print my_var:key0. // prints 0.
print my_var:key1. // prints 100.
print my_var:key2. // prints 200.
print my_var:key with spaces. // doesn't parse - not do-able
print my_var:List("key that isn't even a string",0). // doesn't parse - not do-able
```
(I kept examples of keys that wouldn't work in the example to show what would and would not be supported by this proposed change.  This proposed change still wouldn't allow those last two examples to work, but would allow the first 3 to work.).

The inability to make identifiers become lexicon keys in this way was the main reason for this long thread from before asking us to keep deprecated syntax around:  #2288.

I had pushed back on it because I thought it required special parser and compiler support for just one type of class, and that felt wrong.  (Like the way the C# keyword "yield" and the .Net type IEnumerable are intertwined in a way which never sat well with me.)

But I hit on a possible way to do it that has nothing to do with the compiling and lives entirely inside the class Lexicon itself, and that might work better.

When the compiler sees the colon-syntax, that causes it to call ``Structure.GetSuffix()`` or ``Structure.SetSuffix()`` on the result.  Well we already have a few instances where classes derived from ``Structure`` override ``GetSuffix()`` - ``BodyTarget`` does, as does ``Config``, ``StageValues``, and ``VesselTarget``.  (They cook the value a little bit, then call ``base.GetSuffix()``.)

So we could put the logic entirely inside ``Lexicon`` instead of touching the compiler at all.  We could have ``Lexicon`` override ``GetSuffix()`` by having it first try ``base.GetSuffix()`` and if that fails then instead look inside itself for keys that match.

This way the logic lives entirely inside the class definitions and NOT in the compiler.

The compiler's rule would thus remain unchanged: "When I see ``AAA:BBB``, then I will eval AAA, and whatever object that ends up being, I will call that object's ``GetSuffix("BBB")`` and put its result on the stack.  That would be the same for lexicons or anything else.  It's the *lexicon* that would define that it does something unusual when asked to GetSufffix.

A similar logic would work for SetSuffix, I presume.

Dunbaratu:
Fixes #2551 Fixes #2288 

Making this work required a few things:

1 - Overriding GetSuffix and SetSuffix so that if a Lexicon
fails to Get or Set suffix on the normal suffixes, it falls
back to trying it on the keys.

2 - Making it so that GetSuffix and SetSuffix are sometimes
allowed to fail without throwing an exception, since expecting
them to sometimes fail is now part of Lexicon's normal execution
path it doesn't seem right to "communicate" that fact to
Lexicon with an exception it has to catch.

3 - Also overriding HASSUFFIX and SUFFIXNAMES so they reflect
the fact that some lex keys effectively become "suffixes" now.

### DIFF
--- a/kerboscript_tests/lex_suffix_test1.ks
+++ b/kerboscript_tests/lex_suffix_test1.ks
@@ -1,0 +1,37 @@
+local my_lex is lexicon(
+  "Key0" , "A",
+  "Key1" , "B",
+  "Key2" , "C",
+  "Key3" , "D",
+  "Key4" , "E",
+  "Key5" , "F").
+
+print "Expect: ABCDEF".
+print "Actual: " + concat_lex(my_lex).
+
+set my_lex:key1 to "_".
+set my_lex:key3 to "_".
+set my_lex:key5 to "_".
+
+print "Expect: A_C_E_".
+print "Actual: " + concat_lex(my_lex).
+
+set my_lex:key1 to { return "%". }.
+set my_lex:key3 to { local a is 1. local b is 4. return a+b. }.
+set my_lex:key5 to { return 3/2. }.
+
+print "Expect: A%C5E1.5".
+print "Actual: " + concat_lex(my_lex).
+
+function concat_lex {
+  parameter the_lex.
+
+  local str is "".
+  for key in the_lex:keys {
+    if the_lex[key]:istype("Delegate") 
+      set str to str + the_lex[key]().
+    else
+      set str to str + the_lex[key].
+  }
+  return str.
+}
--- a/kerboscript_tests/lex_suffix_test2.ks
+++ b/kerboscript_tests/lex_suffix_test2.ks
@@ -1,0 +1,39 @@
+print "Testing using Lex as a psuedo-class.".
+print "------------------------------------".
+
+
+print "Making 'fred', an instance of person'.".
+local fred is construct_person("Fred", 23).
+print "fred:greet() prints this:".
+print fred:greet().
+
+print "Making 'henri', an instance of frenchperson.".
+local henri is construct_frenchperson("Henri", 19).
+print henri:greet().
+
+function construct_person {
+  parameter name, age.
+
+  local myself is LEX().
+
+  set myself:name to name.
+  set myself:age to age.
+  set myself:greet to {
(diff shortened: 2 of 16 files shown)


Dunbaratu:
I Just noticed that this PR is missing Sphinx doc changes to describe the new feature.  It should not be merged without that.

PR #2554, opened 2019-06-25
Implemented TriggerAU's suggested fix to range limit

### LINKED ISSUE #2531: Wrong conversion for PartModuleField values using Robotic Parts

Dunbaratu:
Fixes #2531

I got some feedback from SQUAD team member TriggerAU about this problem and he pointed me to the API call that SQUAD uses to obtain the float ranges, and it encapsulates some of the work needed to find the cases where a float range is getting overridden by another float range.  This method call TryGetFieldUIControl seems to do everything right and obtain the true ranges that game is actually using.

### DIFF
--- a/src/kOS/Suffixed/PartModuleField/PartModuleFields.cs
+++ b/src/kOS/Suffixed/PartModuleField/PartModuleFields.cs
@@ -64,29 +64,6 @@ namespace kOS.Suffixed.PartModuleField
         }
 
         /// <summary>
-        /// Return true if the field in question is editable in the KSP rightclick menu
-        /// as an in-game tweakable right now.
-        /// </summary>
-        /// <param name="field">the BaseField from the KSP API</param>
-        /// <returns>true if this has a GUI edit widget on it, false if it doesn't.</returns>
-        private bool IsEditable(BaseField field)
-        {
-            return GetFieldControls(field).Count > 0;
-        }
-
-        /// <summary>
-        /// Get the UI_Controls on a KSPField which are user editable.
-        /// </summary>
-        /// <param name="field"></param>
-        /// <returns></returns>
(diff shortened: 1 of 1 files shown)


src/kOS/Suffixed/PartModuleField/PartModuleFields.cs
```diff
@@ -133,49 +122,40 @@ private bool IsLegalValue(BaseField field, ref Structure newVal, out KOSExceptio
...
+            var vector2 = control as UI_Vector2;
+            if (vector2 != null)
+            {
+                // I have no clue what this actually looks like in the UI?  What is a
+                // user editable 2-D vector widget?  I've never seen this before.
+                if (convertedVal != null)
                 {
-                    // I have no clue what this actually looks like in the UI?  What is a
-                    // user editable 2-D vector widget?  I've never seen this before.
-                    if (convertedVal != null)
```
Dunbaratu: The above looks like a large edit, but if you look carefully, all it is is just a change of indent level.

PR #2563, opened 2019-07-18
Bounding box

Dunbaratu:
Exposes bounding box information from the  part meshes to the kOS script.

Should allow you to detect altitude from the bottom of a landing gear part, for example, or see how big a ship is when docking so you can safely go around it.


Dunbaratu:
The commit #fb08141, attempting to reduce the load on the computer by only using 2 corners instead of all 8 to calculate the bounds was broken logic that squishes the box in some orientations.

I will be reverting that commit and re-pushing an update to this.

PR #2602, opened 2019-08-24
Minor Formatting update to CONTRIBUTING.md

jonnyboyC:
Not sure if this is just the github variant of markdown but the headers appears as ####Header instead of the larger format with a space


jonnyboyC:
Associated Issue #2603

PR #2629, opened 2019-10-14
Bugfixes and enhancements

### LINKED ISSUE #2530: Add the ability to construct arbitrary orbit objects
I've noticed that orbit objects are not attached to existing orbits or maneuver nodes. One can create a maneuver node, save `nextnode:orbit` into a variable and delete the node. The variable still holds the orbit, although it's nowhere in the flight plan.

So, it could be useful to create orbital patches with user-defined parameters without maneuver node trickery. One example is the PEG algorithm where an auxiliary orbit is used to estimate gravity losses during a burn.

### LINKED ISSUE #2556: floor() and ceiling() could use 10ths, 100ths, etc like round() has.
With ``round()`` you can pick the number of digits after the decimal to round to, i.e. ``round(x, 2)`` to round to the nearest hundredth.

But floor() and ceiling() only work for going to an integer, not tenths or hundredths, etc.

They could use a similar optional second parameter like Round() has, that does the same thing. 

i.e. while ``round(1.666, 2)`` would give 1.67, ``floor(1.666, 2)`` would give 1.66.

Without that, people have to resort to the workaround of multiplying by a power of ten to shift the decimal point, then calling floor() then shifting it back, like so:
to get the effect of ``floor(x, 2)`` the script has to do: ``floor(x*100) / 100``.

### LINKED ISSUE #2587: Means to detect if body exists
For people writing scripts meant to be played in multiple different modded solar systems (i.e. RSS versus Stock versus JNSQ versus GPP), it would be nice if a script could safely query (without crashing) whether or not a body exists.  Right now I don't see a means to do that.

i.e. if you do ``BODY("Mun")`` and there is no such body as Mun, the script crashes, but the script has no way to test with ``if BODYEXISTS("Mun")`` first, since no such test exists.

### LINKED ISSUE #2599: GUI: let kerboscripts turn off wordWrap so things format properly
Unity's IMGUI has a field on GUISkin called ``wordWrap` that decides whether or not it's allowed to break up labels on word boundaries to make them fit.  We have it on by default.  But that is the main reason the GUI displays come out all wonky by default.  Someone makes a gui with a label like "This is My Label" and it displays like this instead:
```
This
is My
Label
```
The only way to stop it doing that is to set a fixed width to the GUI box like so: GUI(400).  If the hardcoded size is long enough to make it not need to wrap, that's the only way to stop it wrapping.  As long as you're letting it stretch and flow the content and decide the box size for you, it will always seem to prefer the narrower window, and if it has to break up things in a silly way to do that, it will choose to do so, giving the effect shown above.

We should expose wordWrap so users can fix this annoying feature of Unity IMGUI.  I might even want to see it be the default but that might mess it up for other people.

### LINKED ISSUE #2606: Get rid of startup message about terminal fonts
There's something wrong with the startup dialog box that describes the terminal font, but it's hard to trace why it's happening because it might be an interaction with other mods.  The description is here:

https://forum.kerbalspaceprogram.com/index.php?/topic/165628-173-kos-v1190-kos-scriptable-autopilot-system/&do=findComment&comment=3661594

But in reality it doesn't matter how to fix it, because that dialog box doesn't need to exist anymore.  The proper solution should be to just get rid of it entirely.  The warning that says to go open the panel and pick a font should be replaced with just silently accepting the default and setting it as if it was the "user's choice".  If the player doesn't see that message, that's not the end of the world.  They just play with the default font kOS picked and don't know they could go change it on the toolbar dialog box, but not knowing that won't ruin any gameplay.

The original reason for the dialog was to inform people of the transition from the built-in bitmap letter images kOS shipped with to a system font (mostly so Unicode glyphs could be supported without me having to draw the bitmap images of hundreds of different multinational alphabets by hand).  But that change was years ago.  The box doesn't need to exist anymore.

### LINKED ISSUE #2609: Make HEADING() accept optional 3rd arg for ROLL
It would be handy if HEADING allowed you to say HEADING(compass,pitch,roll).
Note that with the default steeringmanager settings, roll won't do anything until you get close to the desired direction, so this would have to be documented with a note that you should also increase ``steeringmanager:ROLLCONTROLANGLERANGE`` to 180 make it have any effect.

### LINKED ISSUE #2613: [feature] DockingPort suffix(es) to find other DockingPort in docked pair or Element thereof
I'm writing a script to undock and deorbit a vessel after its crew / passengers have been transferred to the vessel it's docked with. In order to avoid colliding after undocking, I want to keep track of the position of the other vessel after undocking. In order to get a reference to that vessel, I'm pretty sure I first need a reference to a part or element on the "other side" of the docking port before undocking, then simply get its parent vessel after undocking. The most straightforward thing to do, I think, would be (assuming PORT holds a reference to the DockingPort in the same Element as the CPU which will be told to UNDOCK) to write something like LOCAL OTHER IS PORT:PARTNER, where :PARTNER would be a new suffix of DockingPort returning the other DockingPort that it's docked to (if applicable).

Also possibly helpful, though perhaps almost redundant if DockingPort:PARTNER is implemented, would be DockingPort:OTHERELEMENT, returning the element containing the other DockingPort — essentially equivalent to DockingPort:PARTNER:ELEMENT.

As it is, it seems I have to check whether DockingPort:PARENT is also a DockingPort, else look through DockingPort:CHILDREN to find the other DockingPort. I find it a bit messy.

ThunderousEcho:
Resolves #2613 
Resolves #2609 
Resolves #2606 
Resolves #2599 
Resolves #2587 
Resolves #2556 
Resolves #2530 

Added suffixes PARTNER and HASPARTNER to DockingPort, an optional third argument to the heading function, suffix WORDWRAP to Style, a global BODYEXISTS function, optional second arguments to FLOOR and CEILING, and an orbit constructor.
Got rid of startup message about terminal fonts.


ThunderousEcho:
Thanks for catching those. Is there any way to preview .rst files without
installing Python? (For some reason, nothing related to Python works on my
machine.)

Dunbaratu:
Sadly no - you really need python to run the sphinx scripts.  I can't seem to get it to work right on my Windows install at all either (not python, but the rest of the sphinx stuff).  I have a remote server with Linux on it where I run them the script generation there.  I often don't run it until about to release, which means often releasing includes one more PR where I fix up the minor errors from .RST stuff that didn't parse right, or misspelled links, etc from the last few months.  I usually accept RST edits in PRs without testing that they have proper syntax, because I feel confident I can fix them up later while still keeping the meaning right if the author typed all the text but just messed up the formatting syntax.

Issue #2075, opened 2017-07-27
Discussion: Features for Realism Overhaul's mods


Dunbaratu:
(Note: the "Not Ready" label, which is normally for Pull Requests, is on this Issue to represent that the issue *itself* isn't even done being typed up yet.  I'm adding more to it as I go over the next few weeks.)

While I'm not sure if this belongs in an Addons module, or if it's even do-able that way, I'm experimenting with Realism Overhaul to discover what things kOS doesn't expose for RO that make RO and kOS hard to use together.

### This should be split into more than one issue before starting to work on it.

I'm using this as a single checklist for discussion and brainstorming.  When actually implementing things from this list, it would be best to split them off into their own issues.

### Which mod does what?

One complication is that RO isn't one mod.  It's a mod PACK, which means each of these features really comes from some other mod, which could be installed independently without the rest of RO present.  Therefore the trigger conditions for whether or not to add these features would have to be per mod inside RO, rather than RO itself.  (i.e. instead of "If RO installed, then.....", we'd have to make it "If RealFuels installed, then...." and "If Deadly Re-Entry installed then...", etc for each mod that matters to the bullet points below.)  For each point, we'd have to identify the actual mod that causes the feature to exist, and test for THAT mod (rather than RO) to enable the feature.  On a first pass, I haven't even identified which mod does each of these things.  Having help with that would be tremendous.

**This is a checklist I plan to add more to later.**

For each item in the list, I will mention if I found a workaround that lets you get the info some other (but maybe more complex) way.  Things for which there is no workaround should have higher priority.

I categorize workarounds into these three types:

- Yes : There is a workaround that's only somewhat cumbersome.
- Yes But : There is a workaround but it's ugly and may not be practical.
- No : There is no workaround at all that I could think of.

(For this list, the name "**RCM**" will be used as an abbreviation for the "Right Click Menu" of a part).

### Engine part suffixes:
[ ] *Current Ullage state*:  Ullage is the requirement that makes engines fail if the fuel is floating around in freefall.  Ullage state ("Very stable", "unstable", "risky", etc....) is communicated to the user via a string on the RCM.
    - Workaround Exists? Yes - you can access this as a ``KSPField`` on the RCM.
[ ] *Ullage Required*: Some engines never worry about Ullage.  Others do.  This should be communicated to the user as a Boolean.  (A script's behaviour when dealing with an engine that can never have Ullage problems can differ from when dealing with an engine that can.)
[...]

PR #2662, opened 2020-01-13
Engine Value improvements

### LINKED ISSUE #2658: Engine suffix to report consumed fuels
A new suffix to return a list of fuels this engine consumes. Mainly useful for RO where there are a large number of fuels, but could be useful for stock fuelling calcs to determine if an engine is monoprop or liquid fuelled.

My current proposed solution for this just returns a list of strings. I noticed in #2075 that a lexicon with proportions was proposed; I haven't found a way to get this information yet.

RCrockford:
Fixes #2658, see also #2075.

Adds suffix to provide consumed fuel information to `ENGINE`.
Adds suffixes for `MASSFLOW`, `MAXFUELFLOW`, and `MAXMASSFLOW` to `ENGINE`.
Adds a number of RealFuels informational suffixes (these return sensible defaults for Stock configs).
Adds documentation for the above.

Documentation fix for `AggregateResource`.

Tested on Stock and RealFuels builds - should return consistent numbers for both configurations now.

### DIFF
--- a/doc/source/structures/vessels/engine.rst
+++ b/doc/source/structures/vessels/engine.rst
@@ -54,8 +54,17 @@ Some of the Parts returned by :ref:`LIST PARTS <list command>` will be of type E
           - :ref:`scalar <scalar>` (kN)
           - Possible thrust at the specified pressure (in standard Kerbin atmospheres), when the engine is enabled.
         * - :attr:`FUELFLOW`
-          - :ref:`scalar <scalar>` (l/s maybe)
-          - Rate of fuel burn
+          - :ref:`scalar <scalar>` (unit/s)
+          - Current volumetric flow rate of fuel.
+        * - :attr:`MAXFUELFLOW`
+          - :ref:`scalar <scalar>` (unit/s)
+          - Untweaked maximum volumetric flow rate of fuel at full throttle.
+        * - :attr:`MASSFLOW`
+          - :ref:`scalar <scalar>` (Mg/s)
+          - Current mass flow rate of fuel.
+        * - :attr:`MAXMASSFLOW`
+          - :ref:`scalar <scalar>` (Mg/s)
+          - Untweaked maximum mass flow rate of fuel at full throttle.
         * - :attr:`ISP`
           - :ref:`scalar <scalar>`
           - `Specific impulse <http://wiki.kerbalspaceprogram.com/wiki/Specific_impulse>`_
@@ -113,6 +122,24 @@ Some of the Parts returned by :ref:`LIST PARTS <list command>` will be of type E
         * - :attr:`GIMBAL`
           - :struct:`Gimbal`
           - Gimbal of this engine (only if available)
+        * - :attr:`ULLAGE`
+          - :ref:`Boolean <boolean>`
+          - Does this engine need ullage (for RealFuels)
+        * - :attr:`FUELSTABILITY`
+          - :ref:`scalar <scalar>`
+          - How stable is the fuel for this engine (for RealFuels)
+        * - :attr:`PRESSUREFED`
+          - :ref:`Boolean <boolean>`
+          - Is this engine pressure fed? (for RealFuels)
+        * - :attr:`IGNITIONS`
+          - :ref:`scalar <scalar>`
+          - Number of ignitions remaining for this engine (for RealFuels)
+        * - :attr:`MINTHROTTLE`
+          - :ref:`scalar <scalar>`
+          - The minimum throttle setting for this engine (for RealFuels)
+        * - :attr:`CONFIG`
+          - :struct:`String`
+          - Engine configuration name (for RealFuels)
 
 
 .. note::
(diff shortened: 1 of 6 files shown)


Dunbaratu:
I don't currently have a working install with RealFuels enabled.  I'm just going to have to trust you on this one that it works there.

Dunbaratu:
@RCrockford What does the following line of documentation in ``consumedresource.rst`` mean *exactly*?

> (only valid while engine is running)

Does it mean:

(A) "is only correct if thrust is happening" (engine is active and thrustlimiter and throttle are both nonzero).

or

(B) is also accurate if "engine is active" even if it's currently throttled all the way back to zero.

In RO engines often stop working when you throttle to zero, but even there some have more than one ignition, so the difference is still relevant.  I want to be careful to clarify the description in the docs.

Dunbaratu:
Additionally, I assume the meaning of "Untweaked" is this:

> Untweaked maximum volumetric flow rate of fuel at full throttle.

"untweaked" means "thrust limiter is at 100%".

Is that right?

RCrockford:
It's option A - you only get a useful value when the engine is consuming fuel. I assume this is because ksp doesn't update propellant flow when the engine is off.

That is what I mean by untweaked (it was copied from elsewhere in the documentation). It's analagous to `MAXTHRUST`.

doc/source/structures/vessels/engine.rst
```diff
@@ -357,5 +405,47 @@ Some of the Parts returned by :ref:`LIST PARTS <list command>` will be of type E
     :type: :struct:`Gimbal`
 
     Returns the :struct:`Gimbal` attached to this engine. Only accessible if the gimbal is present (Use :attr:`Engine:HASGIMBAL` to check if available).
+    
+.. attribute:: Engine:ULLAGE
+
+    :access: Get only
+    :type: :ref:`Boolean <boolean>`
+    
+    If RealFuels is installed, returns true if this engine needs ullage, otherwise returns false.
```
Dunbaratu: I could see a user finding the terminology ambiguous here.  What does it return when the engine is a type that requires ullage, but currently DOES have sufficient ullage to ignite properly therefore it doesn't "need" it (it already has it).  i.e. is this a (static) property of the engine type, or is this a current, dynamic property of the engine's current condition.

RCrockford: Hmm yes, I se what you mean.
Ullage is a static property if the engine, namely that it is a type that requires ullage.
The related suffix `FUELSTABILITY` is used to check the current condition of the engine.
I'm not at my computer right now, but I will revise the documentation tomorrow to clarify.

src/kOS/Suffixed/Part/EngineValue.cs
```diff
@@ -213,6 +246,215 @@ public ScalarValue GetPossibleThrustAtAtm(ScalarValue atmPressure)
...
+            bool ullage = false;
+
+            if (HasRealFuels)
+            {
+                foreach (ModuleEngines e in FilteredEngineList)
+                {
+                    var ullageField = e.GetType().GetField("ullage");
+                    if (ullageField != null)
+                    {
+                        ullage |= (bool)ullageField.GetValue(e);
```
Dunbaratu: Is there a reason this uses bitwise OR rather than Boolean OR?  I guess because there's no `||=` operator in C#, but if the goal is to short-circuit once `ullage` is true, wouldn't it be better to just completely abort the `foreach` loop once it's true?

RCrockford: Yes, the lack of a logical operator of that form.
It could short circuit as soon as a true result is found.

src/kOS/Suffixed/Part/EngineValue.cs
```diff
@@ -213,6 +246,215 @@ public ScalarValue GetPossibleThrustAtAtm(ScalarValue atmPressure)
...
+            bool pressureFed = false;
+
+            if (HasRealFuels)
+            {
+                foreach (ModuleEngines e in FilteredEngineList)
+                {
+                    var pressureFedField = e.GetType().GetField("pressureFed");
+                    if (pressureFedField != null)
+                    {
+                        pressureFed |= (bool)pressureFedField.GetValue(e);
```
Dunbaratu: (same comment as above, as this is the same pattern - using `|=` to affect a boolean technically works but feels so very wrong to me.  Also, why not break the foreach loop once its true, if that's the goal of this.

doc/source/structures/vessels/engine.rst
Dunbaratu: Thanks.  So far other than the few comments I think I'm ready to merge this, but I'll wait for tomorrow for those changes you mention.

src/kOS/Suffixed/Part/EngineValue.cs
Dunbaratu: Bitwise OR won't short-circuit like Boolean OR does.  I thought maybe there's a chance that Microsoft inserted some weird special case logic when both operands are bool.  But I looked it up and they didn't.  Logical OR is guaranteed to short circuit but bitwise OR runs both operands, even for bool.  That's one of the reasons I wish there *was* a ``||=`` operator, so it could short circuit properly, but alas there isn't.

RCrockford:
I have clarified the documentation and tidied up those boolean checks. I think that covers everything we talked about.

Dunbaratu:
Okay it all looks good.

Like I said, I'll just have to trust you on this one, as it would take too much messing about for me to test it on realfuels.  But visually scanning the code it looks good.

PR #2667, opened 2020-02-10
Fixes #2666 prefer picking visible over invisible field.

### LINKED ISSUE #2666: Stock ModuleControlSurface "authority limiter" getfield/setfield broke in KSP updates.
**Problem discovered with version/part:**

First discovered in KSP 1.8.1 - used to work a year ago.  I don't know exactly which update broke it.
This probably occurs on other control surface parts, but the test case was for the stock part "AV-R8 Winglet".

**Shallow look at the problem:**

```
// The following two commands give an error
// claiming the field is not available, even
// though you can see it, and manually
// manipulate it, in the PAW:
print ship:partstagged("fin")[0]:getmodule("ModuleControlSurface"):getfield("authority limiter").
print ship:partstagged("fin")[0]:getmodule("ModuleControlSurface"):setfield("authority limiter", 20).

// For reference, the following commands on a similar field
// work just fine, even though it looks like the same exact
// type of field (Single, a slider):
print ship:partstagged("fin")[0]:getmodule("ModuleControlSurface"):getfield("deploy angle").
print ship:partstagged("fin")[0]:getmodule("ModuleControlSurface"):setfield("deploy angle", 20).
```

Dunbaratu:
Fixes #2666 - see comments in the code to explain how.

### DIFF
--- a/src/kOS.Safe/Exceptions/KOSLookupFailException.cs
+++ b/src/kOS.Safe/Exceptions/KOSLookupFailException.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 
 namespace kOS.Safe.Exceptions
 {
@@ -30,7 +30,7 @@ namespace kOS.Safe.Exceptions
         /// <param name="exist">True if the reason for the message is that the thing exists at
         ///   other times, but is just not available at the moment.</param>
         public KOSLookupFailException(string category, string lookupName, object obj, bool exist = false) :
-            base(String.Format( (exist ? TERSE_NOT_EXIST_MSG_FMT : TERSE_NOT_EXIST_MSG_FMT), category, lookupName.ToUpper(), obj.ToString()) )
+            base(String.Format( (exist ? TERSE_DOES_EXIST_MSG_FMT : TERSE_NOT_EXIST_MSG_FMT), category, lookupName.ToUpper(), obj.ToString()) )
         {
             this.category = category;
             this.lookupName = lookupName;
--- a/src/kOS/Suffixed/PartModuleField/PartModuleFields.cs
+++ b/src/kOS/Suffixed/PartModuleField/PartModuleFields.cs
@@ -235,8 +235,23 @@ namespace kOS.Suffixed.PartModuleField
         /// <returns>a BaseField - a KSP type that can be used to get the value, or its GUI name or its reflection info.</returns>
         protected BaseField GetField(string cookedGuiName)
         {
-            return partModule.Fields.Cast<BaseField>().
-                FirstOrDefault(field => string.Equals(GetFieldName(field), cookedGuiName, StringComparison.CurrentCultureIgnoreCase));
+            // Conceptually this should be a FirstOrDefault(), because there should only be one Field
+            // with the given GUI name.  But Issue #2666 forced kOS to change it to a list of hits
+            // because KSP started naming two fields with the same gui name, only one of which is visible
+            // at a time:
+            IEnumerable<BaseField> allMatches = partModule.Fields.Cast<BaseField>().
+                Where(field => string.Equals(GetFieldName(field), cookedGuiName, StringComparison.CurrentCultureIgnoreCase));
+            // When KSP is *not* doing the weird thing of two fields with the same name, there's just one hit and it's simple:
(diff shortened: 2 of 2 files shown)


src/kOS.Safe/Exceptions/KOSLookupFailException.cs
```diff
@@ -30,7 +30,7 @@ public class KOSLookupFailException: KOSException
         /// <param name="exist">True if the reason for the message is that the thing exists at
         ///   other times, but is just not available at the moment.</param>
         public KOSLookupFailException(string category, string lookupName, object obj, bool exist = false) :
-            base(String.Format( (exist ? TERSE_NOT_EXIST_MSG_FMT : TERSE_NOT_EXIST_MSG_FMT), category, lookupName.ToUpper(), obj.ToString()) )
+            base(String.Format( (exist ? TERSE_DOES_EXIST_MSG_FMT : TERSE_NOT_EXIST_MSG_FMT), category, lookupName.ToUpper(), obj.ToString()) )
```
Dunbaratu: Not related to the issue, but I noticed it when debugging the issue - this exception message has been wrong for ages (always gave the "not exist" message even when the field does exist but is invisible.)

src/kOS/Suffixed/PartModuleField/PartModuleFields.cs
```diff
@@ -235,8 +235,23 @@ public virtual BooleanValue HasField(StringValue fieldName)
         /// <returns>a BaseField - a KSP type that can be used to get the value, or its GUI name or its reflection info.</returns>
         protected BaseField GetField(string cookedGuiName)
         {
-            return partModule.Fields.Cast<BaseField>().
-                FirstOrDefault(field => string.Equals(GetFieldName(field), cookedGuiName, StringComparison.CurrentCultureIgnoreCase));
+            // Conceptually this should be a FirstOrDefault(), because there should only be one Field
+            // with the given GUI name.  But Issue #2666 forced kOS to change it to a list of hits
+            // because KSP started naming two fields with the same gui name, only one of which is visible
+            // at a time:
+            IEnumerable<BaseField> allMatches = partModule.Fields.Cast<BaseField>().
```
jonnyboyC: You may want to call `.ToArray()` here to avoid a few unnecessary enumerations of the partModule fields. in the subsequent `Count()`'s

Dunbaratu: Can you extrapolate on this?  Are you saying that partModule.Fields is of a type that requires a walk of the whole list to return a count?

Dunbaratu: Ah I get it you were actually commenting on the line below this one.  I thought you were saying add a ToArray() to the line above, when actually you mean to the line below with the "where" clause.

Dunbaratu: I think I understood what you asked for and put up an update to the PR - can you look at it?

I don't know if it's needed, though.  partModule.Fields is of type BaseFields, which already does implement its own Count() getter.  Without peeking at SQUAD's code illegally with a decompiler, I can't tell whether that Count() is just returning an already known integer or if it's running a walk, of the list to generate it.  I assume when you ask C# to do a Where(), the subset returned is the same type of collection as the original bigger collection, so I had assumed it was returning a BaseFields.

jonnyboyC: It's honestly not really necessary just a small optimization. `.Count()` is a Linq extension method and will call the `IEnumerable<T>` methods `Current` and `MoveNext` to determine the count. [This SO post](https://stackoverflow.com/questions/4098186/lists-count-vs-count/4098586) explains the main gist. Linq typically has specialized versioned when the `IEnumerable<T>` is really a `ICollection<T>` or `IList<T>`. Since we call four additional methods on the `IEnumerable<T>` after the where save a little time.

Dunbaratu: yeah the actual collection under the hood *does* have a simple integer count - the only question is whether or not Linq can detect that and use it, which I don't know.  At any rate I did change it.

PR #2673, opened 2020-02-20
Add string escaping according to string regex format

### LINKED ISSUE #2672: Escaped quotes in string literal are not unescaped
The syntax allows using quotes in a string by prepending it by another quote: https://github.com/KSP-KOS/KOS/blob/develop/src/kOS.Safe/Compilation/KS/kRISC.tpg#L91

This works during parsing but when you try to print the following string: `"foo""bar"` it will show `foo""bar` instead of `foo"bar`.

Is this expected behavior?

thexa4:
Based on: https://github.com/KSP-KOS/KOS/blob/b56c1b5709df897308c579843e02b9e163973484/src/kOS.Safe/Compilation/KS/kRISC.tpg#L91

Fixes #2672

(Warning: has not been tested yet, unable to compile right now)

### DIFF
--- a/src/kOS.Safe/Compilation/KS/Compiler.cs
+++ b/src/kOS.Safe/Compilation/KS/Compiler.cs
@@ -2016,7 +2016,7 @@ namespace kOS.Safe.Compilation.KS
             bool shouldEscape = true;
             if (value[0] == '@')
             {
-                value = value.substring(1);
+                value = value.Substring(1);
                 shouldEscape = false;
             }
             
--- a/src/kOS.Safe/Compilation/KS/Compiler.cs
+++ b/src/kOS.Safe/Compilation/KS/Compiler.cs
@@ -2012,7 +2012,19 @@ namespace kOS.Safe.Compilation.KS
         private void VisitString(ParseNode node)
         {
             NodeStartHousekeeping(node);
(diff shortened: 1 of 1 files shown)


jonnyboyC:
@Dunbaratu Does this now allow certain string escapes? Could you give an example? Mostly just trying to keep https://github.com/jonnyboyC/kos-language-server as in sync with master here a possible

thexa4:
It makes this `print "hello""world".` output `hello"world`

jonnyboyC:
Thanks!

7ranceaddic7:
@thexa4  Is that really the desired result?

I would think that output should be `hello""world`,  no?

thexa4:
The advantage of treating a repeated quote as a single quote is that it allows you to represent all strings. If you use it as-is, you would not be able to write down a string literal that only contains `"`.

The way this is usually solved is by defining an escape character like `\`. `"\""` would then become a `"` string literal. The other option is escaping by repeating the character. An example of this is printf. It uses the `%` character to as a field placeholder. This makes it impossible to print a string that contains an actual `%` character. The solution they use interprets a `%%` as a single literal `%`.

The existing parser syntax seems a lot like the printf case which is why I used this behavior.

PR #2675, opened 2020-02-21
Fix broken documentation links

DasSkelett:
## Motivation
The `Downloads and Links` (https://ksp-kos.github.io/KOS_DOC/downloads_links.html) part of the documentation has some broken links to external sites.

## Changes
This PR fixes these broken links.
To be precise, the download link to SpaceDock (which wrongly had an `www.` subdomain prefixed) and CurseForge (I think the current link was valid once, but they changed their main domain + path structure some time ago). Both are now HTTPS links instead of HTTP, too.

Furthermore the forum thread link wasn't working, it needs a `-` and something else attached after the thread id, I chose `-*` which works quite well.


DasSkelett:
Pinging @Dunbaratu in case you missed this PR.

Issue #2681, opened 2020-03-08
The CreateOrbit function takes Mean Anomaly At Epoch As Radians


JCSkyKnight:
I think this is unintentional as I have checked the other angles and they all seem to be taken in degrees. Can test with a simple script.

set orb to createOrbit(20, 0, 270000, 20, 20, 290, 0, Mun).
print(orb:meanAnomalyAtEpoch).

set orb to createOrbit(20, 0, 270000, 20, 20, 290*constant:degToRad, 0, Mun).
print(orb:meanAnomalyAtEpoch).

Dunbaratu:
For some reason this didn't close when the PR for it got merged, so I'm closing it manually now.

Issue #2687, opened 2020-04-01
Make time to periapsis / apoapsis obtainable for future (or planned) orbits


vyznev:
Situation: I'm trying to write a script that would plan a circularization maneuver at the periapsis of a future encounter.  However, kOS seems to provide no way of determining the time at which the periapsis will occur until I've actually entered the SOI of the body around which I want to circularize (at which point I can use `ETA:PERIAPSIS`).

I believe this information is in fact available from KSP, as it's clearly displayed in the map UI.  It would be very useful if kOS could provide some way to expose it to scripts too.  One way to do that could be to add a new `:ETA` suffix to `Orbit` and make it return a structure similar to the current global `ETA` structure, but for that particular orbit segment.  (The existing `:NEXTPATCHETA` suffix could then be deprecated in favor of `:ETA:TRANSITION`.)

Note: This is basically a duplicate of issue #550 from 2015.  However, that issue was closed four years ago without a full fix ever having been implemented.

pand5461:
The ETA to periapsis can be derived from the `orbit` object, specifically, from the mean anomaly at epoch and semi-major axis. For predicted orbits after transition, the epoch is at the SOI transition, so that MNA almost directly translates into the time from transition to periapsis.

nuggreat:
the ETA to your periapsis will always be 1/2 way between entering the SOI and leaving it so something like this `LOCAL etaPE IS (preOrbit:NEXTPATCHETA + preOrbit:NEXTPATCH:NEXTPATCHETA) / 2.` or you can calculate it from the orbital elements

vyznev:
@nuggreat That'll work for simple fly-by orbits, but not if the orbit patch begins or ends with a satellite encounter. So it won't work e.g. if you're trying to plan a circularization maneuver around Duna and your fly-by orbit happens to pass near Ike.

For closed orbits, it's possible to determine the periapsis time from `:EPOCH`, `:MEANANOMALYATEPOCH` and `:PERIOD`, which is actually what I tried first. But that fails for open fly-by orbits, since those have an infinite period (and thus even trying to access `:PERIOD` will crash the kOS script).

[...]

PR #2689, opened 2020-04-05
Make createorbit() take mean anomaly at epoch in degrees

vyznev:
This fixes issue #2681. I ran into this issue myself while testing a fix for #2687, so I fixed it and then found this earlier report.

In particular, with this patch, the following function now successfully clones an orbit:

```
function cloneorbit {
	parameter o is orbit.
	return createorbit(
		o:inclination,
		o:eccentricity,
		o:semimajoraxis,
		o:lan,
		o:argumentofperiapsis,
		o:meananomalyatepoch,
		o:epoch,
		o:body
	).
}
```

### DIFF
--- a/src/kOS/Function/Suffixed.cs
+++ b/src/kOS/Function/Suffixed.cs
@@ -110,7 +110,7 @@ namespace kOS.Function
             } else 
             {
                 double t = GetDouble(PopValueAssert(shared));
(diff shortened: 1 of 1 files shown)


Dunbaratu:
"Breaking" label added because previously written scripts would have called it with radians.

PR #2690, opened 2020-04-06
Add :ETA suffix to OrbitInfo

vyznev:
This fixes issue #2687 by adding an `:ETA` suffix to all Orbit structures and making `ETA` an alias for `SHIP:ORBIT:ETA`.

To account for all corner cases, I ended up calculating the apoapsis and periapsis ETAs semi-manually from `orbit.epoch`, `orbit.ObTAtEpoch` and `orbit.period`.  While simply using `orbit.StartUT + orbit.timeToAp` and `orbit.StartUT + orbit.timeToPe` would work for most orbits, it fails e.g. for custom orbits constructed with `CREATEORBIT()`.

In particular, this implementation has been tested and appears to work with:

* active vessel orbits while at warp 0 (where epoch and StartUT both track current time),
* active vessel orbits while at warp 1+ (where the epoch stays fixed while StartUT tracks current time),
* future predicted orbit segments (where epoch and StartUT stay fixed),
* future planned orbit segments after a maneuver (and possibly ending in another maneuver),
* current orbits of other vessels,
* future predicted orbit segments of other vessels (e.g. after selecting the vessel as target),
* orbits of other vessels for which KSP has not calculated future patches,
* orbits of celestial bodies, and
* custom orbits created with `CREATEORBIT()`.

(The last three kinds or orbits may have StartUT set to 0.) Also, I've done my best to test all these cases with both elliptical and hyperbolic orbits, and with orbits ending in encounters, although there are probably a few cases I haven't tested the latest version with. If anyone feels like doing another exhaustive testing pass, please do!

In any case, at the very least, I'm confident that this code introduces no regressions. The one notable change to current behavior is that `ETA:TRANSITION` will now return `float.MaxValue` if the current vessel is in an orbit with no predicted SOI transitions; previously it would return an arbitrary meaningless value (typically close to the period of the orbit).

Ps. This is only a code patch. The documentation of course also needs to be updated to reflect these changes. I'll submit a separate patch for that later, also fixing a bunch of related documentation issues I found while writing this.


src/kOS/Suffixed/OrbitInfo.cs
```diff
@@ -109,33 +111,33 @@ public OrbitableVelocity GetVelocityAtUT( TimeSpan timeStamp )
...
+        /// <returns>an OrbitEta structure.</returns>
+        private OrbitEta GetETA ()
         {
-            if (GetHasNextPatch())
-            {
-                return new OrbitInfo(orbit.nextPatch, Shared);
+            if (eta == null) {
+                eta = new OrbitEta(orbit, Shared);
             }
-            throw new KOSSituationallyInvalidException("Cannot get next patch when no additional patches exist.  Try checking the HASNEXTPATCH suffix.");
```
Dunbaratu, line 121: Man, git's diff output can really make things confusing at times.  It looked to me like you had replaced GetNextPatch with GetETA, which made no sense at all.  It took a while to realize that GetNextPatch was just pushed down a few lines in the file by GetETA, rather than replaced by it as the diff makes it look.

vyznev: Yeah, that's really not the greatest diff I've ever seen.

BTW, thanks to your comment, I noticed that I'd left in some unnecessary commit changes from an earlier iteration. Let me undo those and see if the diff gets more readable too.

vyznev: Turns out moving the new `GetETA` method earlier in the file so that there's enough other code in between makes diff actually realize that it's a completely new method. :)

Dunbaratu:
Reminding myself - add documentation of this change before merging the PR

Issue #2698, opened 2020-05-10
New maxTemp for parts .cfg


robopitek:
So I don't know how it looks in old kerbals but I sure about one.
**Today you can use these  Scriptable Control Systems as heatshields** because their `maxTemp = 3400` (except KOSCherryLight it `maxTemp = 2400`) for comparision maximum temperature for random stock heatshield is 3300 and that brokes game balance!
So I think that maximum temperature for these computers should be 1200 because stock electronics (eg science parts, probe cores) `maxTemp` is also 1200. (except KOSCherryLight it `maxTemp` should be 2000 because stock illuminators `maxTemp` is also 2000)

I think you understand and sorry for my english.

Dunbaratu:
This should have been closed back when  #2699 was merged.

PR #2699, opened 2020-05-10
New maxTemp for parts .cfg

robopitek:
This is about this issue #2698 

I not 100% that "Resources/GameData/kOS/Parts/kOSMachine0mLegacy/part.cfg" should be changed but other `maxTemp` should works good.


Dunbaratu:
I disagree that they should be as burn-able as science instruments, as a kOS part could be vital to returning properly and should be treated more similar to probe cores.  I think the values should be copied from probe core values.

Dunbaratu:
Hmm.  Turns out probe cores are also at 1200, so it's the same.  I guess that makes it okay to use that value.

PR #2719, opened 2020-07-08
Vessel delta-v suffixes

### LINKED ISSUE #2655: DeltaV structure
KSP now can display the DeltaV. So I think it would be a great idea to implement a new structure (or a Suffix) for DeltaV.

### LINKED ISSUE #2674: support reading stock DeltaV values
since the stock game now does delta v calculations it would be great if we could have access to read these and use them in scripts instead of redoing the math in a kos script.

ThunderousEcho:
Added DELTAV, DELTAVATASL, DELTAVVACUUM, and BURNTIME suffixes to Vessel.
Resolves #2655
Resolves #2674

There are a lot of other useful things in Vessel.VesselDeltaV that we should consider adding bindings for.


Dunbaratu:
Thank you for writing this up.
I am significantly changing it though, to provide access to the stage-specific info too.

I will make a new class, ``DeltaVCalc`` that is a wrapper around either ``VesselDeltaV`` or ``DeltaVStageInfo`` depending on if it was for a single stage or the whole vessel.

PR #2725, opened 2020-07-30
Added gui fields to a gui group

RCrockford:
Places all kOS fields into a group in the PAW.
In Edtior:
image
In Flight:
image


Dunbaratu:
Oh good.  When I saw KSP now had these I was wondering how we go about using them.  Thanks for figuring out how.

PR #2729, opened 2020-08-07
Fixes issue 2697 clickthrublocker

Dunbaratu:
This edit makes kOS use the ClickThroughBlocker mod to handle its keyboard focus.  ClickThroughBlocker is a nice mod, but it does require all mods to get on board and use it or else it starts acting very hostile toward the ones that don't, stealing their focus away.  (Ironically it was *causing* clickthrough bugs when one of ClickThroughBlocker's windows was stacked underneath kOS's terminal window and stealing its focus.)

The only way to fix the fact that CTB windows were stealing focus from the kOS terminal even when they were stacked underneath it was to make the kOS terminal *become* a CTB window too so CTB knows it exists and can track its ``rect`` and stacking order relative to the other CTB windows.

To make this work, I spoke at length with @LinuxGuruGamer (maintainer of ClickThroughBlocker) on our Twitch channels and spoke of how moving kOS to using ClickThroughBlocker would change its behavior to focus-follows-mouse, which wasn't necessarily desired.  At the time focus-follows-mouse was the only option ClickThroughBlocker had.  He has since added the option to use click-to-focus depending on player preference.  It is for THAT reason that this mod requires ClickThroughBloocker to be at least version 0.10.  Old versions of ClickThroughBlocker will be missing the ``focusFollowsClick`` boolean field that our code in kOS explicitly reads to change its own behavior.  With that field missing, there would be errors.

### A NOTE ABOUT THIS PR and the kOS.netkan file:

This PR "wants" to edit the kOS.netkan file to reflect the new mod dependency on ClickThroughBlocker, but it can't.  That had to be put in a separate PR ( To be called #2730 ).  For reasons, see the edits to documentation files in PR #2730. )

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -213,7 +213,14 @@ Setting Up Your Environment
                * Click on the resulting link to go to that class's documentation page.
                * The Unity documentation for that class will start with some faint grey text at the top of the page that says, "Implemented in:", which tells you which DLL you need to reference to get code using that class to compile properly.
 
-	3. If you do not have a copy of KSP locally, you may
+        3. Make sure your installation of KSP has LinuxGuruGamer's ClivkThroughBlocker
+           mod installed.  kOS now needs it in order to compile.  After it is
+           installed, create a reference for it in your kOS project's References
+           like so:
+
+               * $KSP_INSTALL_DIR/GameData/00_ClickThroughBlocker/Plugins/ClickThroughBlocker.dll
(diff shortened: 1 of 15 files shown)


Dunbaratu:
**REVERTED**

NOTE that although this PR is merged, I have had to revert the changes in this merge because I need to put kOS's window handling back to what it was before ClickThroughBlocker until it adds an API that's pending.

PR #2730, opened 2020-08-07
Netkan for issue 2697

Dunbaratu:
This is associated with PR #2729.  When PR #2729 is merged, then wait until that merge makes its way into a public release ZIP, and immedaitely after that ZIP is made public, THEN merge this PR into develop and into master.

To see why, see the documentation files that this PR itself edits, which describe this new policy I'll try to adhere to from now on.

### DIFF
--- a/RELEASECHECKLIST.md
+++ b/RELEASECHECKLIST.md
@@ -39,6 +39,10 @@ note where it says "(*backport*)" below if you are doing a backport.
 - [ ] Copy Github release to [Spacedock](http://spacedock.info/mod/60/kOS:%20Scriptable%20Autopilot%20System)
   - [ ] If this is a *normal release* for the most recent KSP version supported, Make sure it is set as the default (using the "changelog" link on the Spacedock page).
   - [ ] If this is a *backport*, make sure it did NOT become the default, and set the default back to the most recent non-backport version (using the "changelog" link on Spacedock).
+- [ ] NOW that all the ZIP releases are out there on the various sites, NOW go back and merge any netkan_issue_NNNN PRs into develop and into master:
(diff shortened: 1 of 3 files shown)


Dunbaratu:
NOTE - This is not merged because I had to back-out the merge of #2729 and the clickthroughblocker dependency isn't there yet.  Unless the issue I had with it is addressed, I can't make kOS depend on it.

HebaruSan:
Hi, I just stumbled on this while investigating something else and read your `kOS.netkan.README_SUPER_IMPORTANT.md` documentation (which says `kOS.version` in a few places where I think you meant `kOS.netkan` towards the end), and I wanted to suggest a change that might help with maintainability here:

https://github.com/KSP-CKAN/CKAN/wiki/Adding-a-mod-to-the-CKAN#internal-ckan-files

1. Create a `kOS.ckan` file and include it inside the ZIP files for new releases
2. Move any properties that you consider version-specific from `kOS.netkan` to `kOS.ckan` (e.g., `conflicts`, `depends`, etc.)

A file with a `.ckan` extension inside of a release ZIP is treated specially by the bot: it will parse it as CKAN metadata and merge it the fields from whatever .netkan it is processing. After these changes, the NetKAN bot would only see the appropriate metadata for each release, and you could go back to updating metadata in the same pull request as relevant changes are made, as long as `kOS.ckan` is edited instead of `kOS.netkan`. Best of all, you could drop the "after the next release" pull requests from your work flow.

I have been using this approach for my own mods for a while now and I like it pretty well. Astrogator now has a very minimal .netkan file and a more verbose .ckan file:

- https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/Astrogator.netkan
- https://github.com/HebaruSan/Astrogator/blob/master/GameData/Astrogator/Astrogator.ckan

PR #2736, opened 2020-08-13
Fix #2669: additional asteroid functionality

### LINKED ISSUE #2669: [Feature Request] Expose IDiscoverable properties of vessels to detect asteroid classes
https://kerbalspaceprogram.com/api/interface_i_discoverable.html

In "twitch plays ksp" we have no way to know the class of an asteroid.  That's a pretty niche usage, but any kind of automated system that wants to go out and grab an asteroid will probably want to know its class first!

JonnyOThan:
add STOPTRACKING and SIZECLASS suffixes to Vessel

### DIFF
--- a/src/kOS/Suffixed/VesselTarget.cs
+++ b/src/kOS/Suffixed/VesselTarget.cs
@@ -274,6 +274,8 @@ namespace kOS.Suffixed
             AddSuffix("MESSAGES", new NoArgsSuffix<MessageQueueStructure>(() => GetMessages()));
 
             AddSuffix("STARTTRACKING", new NoArgsVoidSuffix(StartTracking));
+            AddSuffix("STOPTRACKING", new NoArgsVoidSuffix(StopTracking));
+            AddSuffix("SIZECLASS", new Suffix<StringValue>(GetSizeClass));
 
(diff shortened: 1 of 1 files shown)


Dunbaratu:
I see you did a few PR's recently.  Thank you.

I was about to merge this one but I want to ask a few clarifications because I'm trying to write user-facing documentation for it and want to say the right things:

1 - Not all ``Vessel`` objects are asteroids.  They might just be a normal ship.  What is the expected result if a user attempts to call ``:STOPTRACKING`` on a ``Vessel`` that isn't actually an asteroid?  Or on a vessel that is an asteroid, but hasn't had tracking started in the first place?  Will it throw an exception or will it fail silently?   (As a general rule, if it does throw an exception I prefer to detect the problem case ourselves and throw our own homemade exception rather than letting some built-in .Net exception "bleed" through to where the user sees it in the terminal and gets confused by what it means.)

2 - Same question, but for ``:SIZECLASS`` - what is the SIZECLASS of a non-asteroid vessel going to say?

3 - I haven't yet played with the new comets in KSP 1.10.x.  Do you know if comets are basically just asteroids in the KSP API? (Meaning should the documentation say these suffixes are for "Asteroids" or should it say "Asteroids and Comets").

Dunbaratu:
Well, I'm going to merge what I have, even if it's lacking some explanation in the docs, because I want to move on.

PR #2737, opened 2020-08-13
add kuniverse:launchcraftwithcrewfrom

JonnyOThan:


JonnyOThan:
Bleh, no, I'm a github noob.  How do I undo this?

luxgladius:
I think you can just git reset HEAD~ and then git add only the correct files before commiting again. I believe GitHub looks at the branch, not the commits at the time of submission.

luxgladius:
You might have to force a push though with git push origin +career

Issue #2263, opened 2018-04-06
A few missing features for Trajectories Addon


Dunbaratu:
Here's a few features I noticed were missing from the Trajectories Addon when I was testing PR #2220.
These shouldn't delay merging PR #2220 because that PR is desperately needed to make the Trajectories Addon work *at all*.  But they shouldn't be forgotten either, so I'm making this a separate issue to correct after PR #2220 is merged:

- Once ``Addons:TR:SETTARGET()`` has been called, there's no way to actually un-set the target from the script.  (Also no way for the script to un-set it after the user manually set it).
- A script cannot query *what* the Trajectories target is currently set to.  It can query that the target exists, but not what it is.  There should be a way to obtain the GEOPOSITION from it.
- The manual GUI interface lets a user pick a descent profile that is neither prograde nor retrograde, but something in between, with the sliders.  There is no way for a script to do this right now.  Perhaps making a script able to set/get the value of the 4 variables attached to those slider values would help.

PiezPiedPy:
@Dunbaratu Thanks for the heads up, I'll add a method to extract the Target position along with a Clear Target method.
Also going to add methods to set and retrieve the Descent Profile nodes.

rlangeoi:
Since PR #2220 has been merged, shouldn't Trajectories 2.0 compatibility be available in the latest v1.1.5.2 release for 1.4.1 (at least with basic support) ? It doesn't seem to be working for me and since TR < 2.0 doesn't exist for 1.4 I can't downgrade it

austinjames314:
Is there any way to get the existing functionality into a game now, in an as-is state? I don't care about anything other than getting the coordinates of the impact point, so I don't need any of the functions about the current Trajectories target.

PiezPiedPy:
Due to the amount of requests on the Trajectories forum thread I made a build of kOS and posted a link to this zip https://www.dropbox.com/s/ykvqsqlgc8714h3/kOS_develop.zip?dl=0

PiezPiedPy:
New API methods where added for the Descent Profile for Trajectories v2.3.2 `ResetDescentProfile DescentProfileAngles DescentProfileModes`.

The next release of Trajectories being v2.4.0 will have new API methods `GetTarget` and `ClearTarget`.

I plan on adding them to kOS in the near future 😉

PiezPiedPy:
@Dunbaratu or whoever can help 😉 
[...]

PR #2747, opened 2020-08-17
Addon update for Trajectories v2.4.0

PiezPiedPy:
An update to the addons for Trajectories, bringing in support for new API functionality in Trajectories v2.4.0.

See Issue #2263 - A few missing features for Trajectories Addon.

New Suffix's have been added as follows:

`IsVerTwoFour()` returns true if the installed version of Trajectories is >= 2.4.0
`GetVersionMajor, GetVersionMinor, GetVersionPatch` return version information.
`GetTarget()` returns the `GeoCoordinates` of the set target.
`ClearTarget()` clears the target.
`ResetDescentProfile(ScalarValue aoa)` resets all descent profile nodes to the passed angle (in degrees).
`DescentAngles` returns or sets all the angles of the descent profile nodes via a list of values (in degrees).
`DescentModes` returns or sets all the modes of the descent profile nodes via a list of booleans.
`DescentGrades` returns or sets all the grades of the descent profile nodes via a list of booleans.

Documentation has also been updated to include the new changes.

### DIFF
--- a/src/kOS/AddOns/Trajectories/Addon.cs
+++ b/src/kOS/AddOns/Trajectories/Addon.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using kOS.Safe.Encapsulation;
 using kOS.Safe.Encapsulation.Suffixes;
 using kOS.Safe.Exceptions;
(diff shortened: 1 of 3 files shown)


nuggreat:
To match with almost all other angles in kOS it would be better if it took/returned the angles in degrees and not radians.  As a result of kOS working with degrees is all most all cases if a degree to radian conversion is needed for a given API call it is much more covenant for the users if that is done automatically on the C# side of things as apposed to requiring the users to do the conversion.

PiezPiedPy:
I don't use kOS myself and was unaware that kOS used degrees for most things, I'll add conversion to the addon as Trajectories uses radians internally.

konstantinua00:
Is it possible to change `IsVerTwoFour()` to... something more extendable into the future?  
Maybe function that returns version as single number, or major-minor-bugfix trio of functions?

PiezPiedPy:
@konstantinua00 There is already a 'GetVersion()' suffix that returns the installed version number as a string

src/kOS/AddOns/Trajectories/Addon.cs
```diff
@@ -54,23 +83,30 @@ private BooleanValue IsVerTwoTwo()
...
                 CelestialBody body = shared.Vessel.orbit.referenceBody;
                 Vector3? impactVect = TRWrapper.ImpactVector();
                 if (impactVect != null)
                 {
-                    var worldImpactPos = (Vector3d)impactVect + body.position;
-                    var lat = body.GetLatitude(worldImpactPos);
-                    var lng = Utils.DegreeFix(body.GetLongitude(worldImpactPos), -180);
+                    Vector3d worldImpactPos = (Vector3d)impactVect + body.position;
+                    double lat = body.GetLatitude(worldImpactPos);
+                    double lng = Utils.DegreeFix(body.GetLongitude(worldImpactPos), -180);
```
Dunbaratu, line 106: BTW, thank you for this change.
The "var, go look up method elsewhere to see what type it really is" design pattern is one of the most annoying misfeatures of C#.  I call it "IDE-think" - people think "well, the IDE tells me the type so I don't have to state it", which is a PITA when reading it somewhere other than an IDE, like on github diff output.

Dunbaratu: The only place I consider it good is when calling a constructor, because there the constructor's name IS the type so it's already clear from that.

Dunbaratu:
@PiezPiedPy I am considering taking the risk of merging this without me testing it, and taking your word for it that it's good.  I don't normally use Trajectories and you are the expert here on it.  Nothing in the code *looks* bad or out of place and I had a read through it.  Are you okay with that?

src/kOS/AddOns/Trajectories/Addon.cs
PiezPiedPy: I have to agree, I rarely use var, usually only when throwing code together quick, once I'm happy with how the code works I will change the var's to the actual types.

PiezPiedPy:
@Dunbaratu I have no issue with a merge and have already tested so there should be no problems, but those pesky bugs always find a way 😜.

PR #2754, opened 2020-09-06
Force all input through buffer unless flushing buffer.

### LINKED ISSUE #2753: Pasting a long chunk text fast through a telnet client can get commands out of order.
@JonnyOThan found this in a telnet client he wrote for TwitchPlaysKSP.  His telnet client takes a chunk of text from twitch chat and shoves it into the kOS telnet very quickly as if a user typed it really really fast.

When the chunk of text was big enough, one of the lines got processed out of order.

This is not an exact example, but it is an idea of what the problem was like:
Chunk of Text shoved fast through telnet:
```
print "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 01".
print "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 02".
print "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 03".
print "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 04".
print "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 05".
print "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 06".
print "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 07".
print "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 08".
print "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 09".
print "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 10".
print "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 11".
print "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 12".
print "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 13".
print "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 14".
print "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 15".
print "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 16".
```
Resulting printout in terminal (note where line 11 ends up):
```
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 01
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 02
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 03
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 04
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 05
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 06
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 07
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 08
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 09
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 10
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 12
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 13
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 14
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 15.
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 16
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 11
```
I don't have the exact example, but it was like that.

There is a suspicion that the problem may be how kOS handles buffered char input that comes in while the interpreter is busy responding to a command.  When the interpreter gets a line of input, compiles it into opcodes, and runs it, until those opcodes finish the interpreter is blocked from taking more input, and instead any input that arrives while those opcodes aren't done yet gets queued and a regular OnGUI update dequeus them later.  It seems there may be a chance that new input that comes in might jump in front of that queued input, as the queued characters don't get processed until the interpreter is idle again, and new input could be getting in there just at the very moment the interpreter was idle but before the queue got flushed.

This may be difficult for me to test as I don't have the software that triggers it (Jonny's chat-to-telnet client).

Dunbaratu:
Fixes #2753 (I hope - I can't test it because I couldn't cause the bug in the first place - I need @JonnyOThan to test it for me.)

Now the only input that is allowed to immediately get
sent to the interpreter is that which came FROM the
buffer flusher routine ProcessUnconsmedInput().  It
uses flags that allow its input to come through
directly, but *everything else* that types input is
now required to queue it (so it will pass through
ProcessUnconsumedInput() a moment later).  Hypothetically
this should fix the ability for input to skip the buffer
and jump in line.

The one exception is the immediate characters related
to terminal closing or program interruption like CTRL-C.
They are allowed to jump in out of order and happen immediately.
And in the case of CTRL-C, it also flushes whatever was in the
buffer so if you typed anything ahead it will no longer
get executed after a CTRL-C like it used to.

### DIFF
--- a/src/kOS/Screen/TermWindow.cs
+++ b/src/kOS/Screen/TermWindow.cs
@@ -325,8 +325,6 @@ namespace kOS.Screen
 
             GetFontIfChanged(true);
 
-            ProcessUnconsumedInput();
-
             if (isLocked) ProcessKeyEvents();
             if (FlightResultsDialog.isDisplaying) return;
             if (uiGloballyHidden)
@@ -362,6 +360,7 @@ namespace kOS.Screen
             GetNewestBuffer();
             TelnetOutputUpdate();
             ProcessTelnetInput(); // want to do this even when the terminal isn't actually displaying.
+            ProcessUnconsumedInput(); // Moved here from OnGUI because it needs to run even when the GUI terminal is closed.
             if (telnetsGotRepainted)
             {
                 // Move the beeps from the screenbuffer "queue" to my own local terminal "queue".
--- a/src/kOS/Screen/TermWindow.cs
+++ b/src/kOS/Screen/TermWindow.cs
@@ -471,7 +471,7 @@ namespace kOS.Screen
                 // command sequences
                 if (e.keyCode == KeyCode.C && e.control) // Ctrl+C
                 {
-                    ProcessOneInputChar((char)UnicodeCommand.BREAK, null);
+                    ProcessOneInputChar((char)UnicodeCommand.BREAK, null, true, true);
                     consumeEvent = true;
                     return;
                 }
@@ -481,21 +481,21 @@ namespace kOS.Screen
                     (e.keyCode == KeyCode.D && e.control) // control-D to match the telnet experience
                    )
                 {
-                    ProcessOneInputChar((char)0x000d, null);
+                    ProcessOneInputChar((char)0x000d, null, true, true);
                     consumeEvent = true;
                     return;
                 }
                 
                 if (e.keyCode == KeyCode.A && e.control)
                 {
-                    ProcessOneInputChar((char)0x0001, null);
+                    ProcessOneInputChar((char)0x0001, null, true, true);
                     consumeEvent = true;
                     return;
                 }
                 
                 if (e.keyCode == KeyCode.E && e.control)
                 {
-                    ProcessOneInputChar((char)0x0005, null);
+                    ProcessOneInputChar((char)0x0005, null, true, true);
                     consumeEvent = true;
                     return;
                 }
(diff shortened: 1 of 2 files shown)


Dunbaratu:
I just discovered a bug this PR has - the telnet terminal is frozen when the in-game GUI terminal is not open.  (If it's not open, the buffer flush never happens and the keypresses that are getting queued up never get used until you open the terminal and then it all comes flooding in.)

Dunbaratu:
I think the recent change fixes the problem I found - and it looks good now.

PR #2765, opened 2020-10-13
Fixes #2764 - Prevents duplicate KOSNameTag.

### LINKED ISSUE #2764: kOS is adding extra copies of the kOS Name Tag partmodule in KSP 1.10.x
The "Name Tag" system in kOS is a tiny PartModule that is added to every single part by a ModuleManager rule.
As of KSP 1.10.x I started noticing that sometimes a part will get two Name Tag buttons in it, as if that rule is being triggered twice and adding the module twice.  This should be looked into, but it's hard to replicate and is probably a race condition that varies depending on what order partmodules get added.

Dunbaratu:
Fixes #2764 

Note, the problem wasn't that ModuleManager was wrong,
but that SQUAD does something really wacky with the
KerbalEVA parts to handle having part variants
for DLC (i.e. if you have Breaking Ground, kerbals
have an additional science module for ground deployment.)

To make this work, SQUAD set it up to *pretend* DLC Eva kerbal
is a totally different part from stock EVA kerbal - a part that
only contains the DLC info.  Then at a late stage (after
MM has done its work), SQUAD melts these two different
EVA kerbal parts together into one, which is where the
duplicates came from.

Since kOS can't really handle two name tags on one part anyway
(maybe that could be another PR another day to support that),
the fix was to make KOSNameTag refuse to duplicate itself into
a part by actively deleting other copies of itself when it's
being inserted into a part in OnAwake().

### DIFF
--- a/src/kOS/Module/KOSNameTag.cs
+++ b/src/kOS/Module/KOSNameTag.cs
@@ -1,5 +1,6 @@
 using kOS.Screen;
 using kOS.Suffixed;
+using kOS.Safe.Utilities;
 using UnityEngine;
 
 namespace kOS.Module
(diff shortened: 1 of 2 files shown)


src/kOS/Module/KOSNameTag.cs
```diff
@@ -37,10 +38,31 @@ public void PopupNameTagChanger()
                     return;
                 }
             }
+            // Make a new instance of typingWindow, replacing the existing one if there was one:
+            KOSNameTagWindow oldTypingWindow = gameObject.GetComponent<KOSNameTagWindow>();
+            if (oldTypingWindow != null)
+            Destroy(oldTypingWindow);
```
Dunbaratu, line 44: The above change wasn't part of the fix.  I just noticed the fact that i was orphaning windows for the GC to collect when I could have been explicitly telling Unity they were going away, and probably should have been since Unity doesn't really notice orphaning (it's C++, not C#).

PR #2784, opened 2020-10-23
Resolving several github issues about documentation

### LINKED ISSUE #2701: variable name incorrect in iterator docs
On this page: https://ksp-kos.github.io/KOS/structures/collections/iterator.html

This code example changes variable names partway through:
```
// Starting with a list that was built like this
SET MyList To LIST( "Hello", "Aloha", "Bonjour").

// It could be looped over like this
SET MyCurrent TO MyList:ITERATOR.
PRINT "before the first NEXT, position = " + MyCurrent:INDEX.
UNTIL NOT MyCurrent:NEXT {
    PRINT "Item at position " + MyIter:INDEX + " is [" + MyIter:VALUE + "].".
}
```
The variable is called ``MyCurrent`` when first declared, but then in the loop it calls it ``MyIter``, which is undeclared and therefore an error.

### LINKED ISSUE #2709: missing suffix in geocoordinates documentation
The suffix `:BODY` is missing from the list of suffixes available on [geocoordinates](http://ksp-kos.github.io/KOS_DOC/math/geocoordinates.html#structure:GEOCOORDINATES).

### LINKED ISSUE #2727: [Documentation] Updates per Second is incorrect in docs
[Relevant documentation page - The kOS CPU Hardware](https://ksp-kos.github.io/KOS/general/cpu_hardware.html#update-ticks-and-physics-ticks)

The kOS updates should run at a max of about 50 UPS, not the listed 25 UPS. It appears that users cannot adjust the updates per in-game second now like described, but the frequency may still vary slightly. It looks like the other documentation about kOS running on physics ticks and not graphics ticks remains accurate.

### LINKED ISSUE #2756: Explode documentation for operators
Operators are mentioned here: http://ksp-kos.github.io/KOS_DOC/language/syntax.html#general-rules

But they are only listed.  There is no description.
Some are obvious, like "=" or "<".  But it's still worth explaining, for example, that "=" is the comparator not the assignment (i.e. it's like "==" in C-like languages), or that not-equal is "<>", and so on.

### LINKED ISSUE #2757: Document that INDEXOF returns -1 if search fails.
The various collections in kOS inherit their behaviors from the .NET collections they're built on top of.  This means that everywhere there's an INDEXOF it returns -1 to indicate the search failed.  However, the documentation doesn't mention this.

Dunbaratu:
Fixes #2709 - Note it turns out the relevant suffix wasn't even
mentioned for List and was missing entirely.

Fixes #2701
Fixes #2727
Fixes #2756
Fixes #2757

### DIFF
--- a/doc/source/language/syntax.rst
+++ b/doc/source/language/syntax.rst
@@ -23,13 +23,38 @@ operator symbols:
 
 .. highlight:: none
 
-**Arithmetic Operators**::
+.. _operators:
 
-    +  -  *  /  ^  e
+**Arithmetic Operators**: These are listed in order of precedence from first to last:
 
-**Logic Operators**::
+    * ``(`` ..to.. ``)`` -> Grouping with brackets makes the expression evaluate first.
+    * ``e`` -> Scientific notation operator, for example: ``1.23e-4`` is ``0.000123``.
+    * ``^`` -> Exponent operator, for example: ``2^3`` is ``8``.
+    * ``*``, ``/`` -> Multiplication and Division, having equal precedence.  Note that
+      division never truncates remainders, even when given two integer operands.
+      (i.e. ``1 / 4`` will return ``0.25``, not ``0``).
+    * ``+``, ``-`` -> Addition and Subtraction, having equal precedence.
 
-    not  and  or  true  false  <>  >=  <=  =  >  <
+**Logic Operators**: These are listed in order of precedence from first to last:
+
+    * ``(`` ..to.. ``)`` -> Grouping with brackets makes the expression evaluate first.
+    * ``not`` -> ``not A`` Logical inversion of A.  Due to the high precedence,
+      if A is some complex expression you may need to write this with brackets
+      as ``not(A)`` in kerboscript to make it parse right.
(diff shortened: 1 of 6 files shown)


doc/source/language/syntax.rst
```diff
@@ -41,7 +66,28 @@ operator symbols:
...
+      a function being called, like ``min(a, b)``.
+    * ``{  }`` -> Used to group a list of statements together into a
+      single block of statements.  Used to define the beginning and
+      ending of a function body, loop body, or conditional body.
+    * ``[  ]`` -> Denotes a list index or lexicon index.
+    * ``#`` -> An archaic way to index lists that is mostly supplanted by
+      the ``[ ]`` square bracket operator, but exists for backward
+      compatibility.
+    * ``,`` -> Separates arguments when calling a function, or parameters
+      when defiing a function.
```
teleksterling, line 82: There's a typo here - 'defiing' missing an 'n'.

Dunbaratu: Thanks.  I fixed it in a PR-less naked commit (not the best practice but I couldn't bring myself to make a new PR for just this.)

PR #2794, opened 2020-10-26
Put max fontsize in one place, and made it 48

### LINKED ISSUE #2711: config.xml TerminalFontDefaultSize continually overwritten with smaller value
Inside config.xml, if you try to set TerminalFontDefaultSize to 24 (which, on my high DPI monitor, is about the equivalent of 12 on a typical monitor), it keeps changing it down to 20, which is just a bit short of being comfortable to read.

kOS in general is difficult to use on this monitor due to it ignoring KSP's UI scale setting, but that's another issue... I can script around it for my own UI screens, but I scale them based on Config:defaultFontSize, which is being clamped as noted above. :(  Fixing this would at least keep my own scripts' windows readable.

Dunbaratu:
Fixes #2711

While I was at it, I moved all the mentions of
font min and max size to one place to use the
same constant everywhere to fix the way the
mod disagreed with itself about the setting.

### DIFF
--- a/src/kOS/Suffixed/Config.cs
+++ b/src/kOS/Suffixed/Config.cs
@@ -63,7 +63,7 @@ namespace kOS.Suffixed
             AddSuffix("DEBUGEACHOPCODE", new SetSuffix<BooleanValue>(() => DebugEachOpcode, value => DebugEachOpcode = value));
             AddSuffix("BLIZZY", new SetSuffix<BooleanValue>(() => UseBlizzyToolbarOnly, value => UseBlizzyToolbarOnly = value));
             AddSuffix("BRIGHTNESS", new ClampSetSuffix<ScalarValue>(() => TerminalBrightness, value => TerminalBrightness = value, 0f, 1f, 0.01f));
-            AddSuffix("DEFAULTFONTSIZE", new ClampSetSuffix<ScalarValue>(() => TerminalFontDefaultSize, value => TerminalFontDefaultSize = value, 6f, 30f, 1f));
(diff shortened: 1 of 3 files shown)


src/kOS/Suffixed/Config.cs
```diff
@@ -63,23 +63,33 @@ private void InitializeSuffixes()
...
-            AddConfigKey(PropId.TerminalFontName, new ConfigKey("TerminalFontName", "FONTNAME", "Font Name for terminal window", "_not_chosen_yet_", "n/a", "n/a", typeof(string)));
-            AddConfigKey(PropId.TerminalBrightness, new ConfigKey("TerminalBrightness", "BRIGHTNESS", "Initial brightness setting for new terminals", 0.7d, 0d, 1d, typeof(double)));
-            AddConfigKey(PropId.TerminalDefaultWidth, new ConfigKey("TerminalDefaultWidth", "DEFAULTWIDTH", "Initial Terminal:WIDTH when a terminal is first opened", 50, 15, 255, typeof(int)));
-            AddConfigKey(PropId.TerminalDefaultHeight, new ConfigKey("TerminalDefaultHeight", "DEFAULTHEIGHT", "Initial Terminal:HEIGHT when a terminal is first opened", 36, 3, 160, typeof(int)));
-            AddConfigKey(PropId.SuppressAutopilot, new ConfigKey("SuppressAutopilot", "SUPPRESSAUTOPILOT", "Suppress all kOS autopiloting for emergency manual control", false, false, true, typeof(bool)));
+            AddConfigKey(PropId.EnableTelnet,
```
Dunbaratu, line 74: The git diff for these next few lines makes it look like I made a lot of changes, but mostly I just split up the very long lines into multiple lines, except for the change involving TerminalStruct.MAXCHARSIZE - those were real.

PR #2805, opened 2020-11-09
fixes #2767 but in a different way, with zeroMode

### LINKED ISSUE #2767: PR: a 2 argument flavor of the TIME function
I'm respectfully requesting a two timestamp scalar argument version of TIME function.  If two args are given to TIME it would return the real span of time between the two scalar time values without adding an extra day to ETAs, otherwise, if given only one argument, it keeps doing what it is doing for all the old code out there that is already coded to treat the TIMESPAN struct delivered by TIME in its peculiar non-timespan non-timestamp way when the scalar arg is an ETA

Dunbaratu:
Fixes #2767 but not in the way requested. 

I added a flag to change how the TimeSpan interprets the meaning of Year and Day, whether it counts them from zero or from one.  Under the hood nothiing has changed about the actual data in the TimeSpan since it never really *stored* year and day - it just stores the second since epoch and *calculates* the year and day from that.  this just changes how the calculation is done.

Note that while I was in there I also made the year, day, hour, minute, and second suffixes set-able, which they were not before.

The fact that it only really stores the seconds-since-epoch behind the scenes also means you can safely mix and match timespans that have zeroFlag false with ones that have zeroFlag true, doing adding and subtracting on them without issue since they do not care how the year and day are being extracted when the arithmetic itself happens just on the "unix time"-like data.

Synopsis of change:
```
set t1 to time(24*60*60*4 + 30). // 4 days and 30 seconds.
set t1:ZEROMODE to false. // this is the default anyway
print t1:calendar + ", " + t1:clock.
year 1, day 5, 00:00:30
set t1:ZEROMODE to true.
print t1:calendar + ", " + t1:clock.
year 0, day 4, 00:00:30 
```

### DIFF
--- a/doc/source/structures/misc/time.rst
+++ b/doc/source/structures/misc/time.rst
@@ -25,6 +25,26 @@ the following points illustrate:
 
 This allows you to use a :struct:`TimeSpan` such as is returned by the :global:`TIME` special variable to make correct physics calculations.
 
+TimeSpan has two modes: "one" mode and "zero" mode
+--------------------------------------------------
+
+By default, a :struct:`TimeSpan` will use a convention where the
+year starts at 1, and the day starts at 1, such there there is
+no such thing as Year 0, and Day 0.  This is how the Kerbal game
+itself shows times, and it matches how our Gregorian Calendar works
+(which was invented when Europe hadn't adapted the number zero yet).
+
+However, if you would prefer it to calculate times starting at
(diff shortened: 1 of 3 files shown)


Dunbaratu:
Marked as duplicate because another PR (#2837) will render this one moot if it gets merged.

Dunbaratu:
I am going to close this as duplicate because it was rendered moot by merging #2837 instead.  (They were two different approaches to addressing the issue).

PR #2818, opened 2020-11-26
Workaround to fix wrong position of packed vessels

marianoapp:
There's a bug in KSP where it returns the wrong position for packed vessels. Instead of returning the current position is returning the position where it will be in the next simulation frame.

Ideally we shouldn't be working around KSP bugs but this one is severe enough that I think merits the effort, and if KSP ever fixes the underlying issue it's easy to rollback.

poserror

In the image there are two arrows, the red one points to the position reported by KSP and the green one to the position using the workaround. Getting closer to the target vessel caused it to unpack and then the red arrow started pointing to the correct position.

In the example above the workaround was done in a kOS script BEFORE applying this patch and that's why I had access to the "wrong" position reported by KSP.

### DIFF
--- a/src/kOS/Suffixed/Part/PartValue.cs
+++ b/src/kOS/Suffixed/Part/PartValue.cs
@@ -191,7 +191,8 @@ namespace kOS.Suffixed.Part
 
         public Vector GetPosition()
         {
-            return new Vector(Part.transform.position - Shared.Vessel.CoMD);
+            Vector3d positionError = VesselTarget.CreateOrGetExisting(Part.vessel, Shared).GetPositionError();
+            return new Vector(Part.transform.position - Shared.Vessel.CoMD + positionError);
         }
 
         private void ControlFrom()
--- a/src/kOS/Suffixed/VesselTarget.cs
+++ b/src/kOS/Suffixed/VesselTarget.cs
@@ -38,8 +38,12 @@ namespace kOS.Suffixed
 
         private Vector3d GetPositionInternal()
         {
-            Vector3d position = Vessel.CoMD - CurrentVessel.CoMD;
+            return Vessel.CoMD - CurrentVessel.CoMD + GetPositionError();
+        }
 
+        public Vector3d GetPositionError()
+        {
+            Vector3d positionError = Vector3d.zero;
             // Workaround to fix a KSP bug:
             // When the target vessel is packed KSP returns the position where it's going to be
             // in the next simulation frame instead of the position where it is now.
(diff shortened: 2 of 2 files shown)


nuggreat:
So I have some questions about this.

Does the issue still exist when on rails if so then your fix would cause the same problem just in a different circumstance?

It also looks like you are hard coding `0.02` as the time step what about differing time steps that occur at different levels of physics warp wouldn't that simply cause the same issue as you are only correcting for the 1x time step?

marianoapp:
I would need to test how it behaves when the target is on rails, but you're right about the time step, that was an oversight.

mgalyean:
Given that the patch is generating the correct location, and KSP is going to generate the incorrect one for now either way, it shouldn't be hard for the code to recognize if/when KSP fixes the underlying problem by comparing the fixed location with the KSP reported location.  In which case it could automatically bypass the fix code until the code could be properly revisited and re-optimized.  This would keep it from breaking noticeably and turn it into a minor maintenance task done at leisure rather than a fire to be stomped out.  Just a thought

nuggreat:
Another question does this only affect `SHIP:POSITION` or does it also affect the position of parts?

marianoapp:
Added a few changes to limit the scope of the workaround to make sure it's only active when the target vessel is loaded but not yet unpacked. Also added support for TimeWarp, both normal and physics.

@mgalyean I don't have a way to know if KSP is returning the correct position because I don't know what that should be, and the position I'm calculating is using the value reported by KSP as base. Basically what I'm doing is instead of returning "x" I'm returning "x - k".

@nuggreat It only modifies `SHIP:POSITION` and `SHIP:DISTANCE`, but part positions may also suffer from the bug, I haven't tested that.
But I get where you're going, if we only modify the ship position we could end up with parts being reported as floating in space far away from the vessel. On the other hand, I'm not sure if you can access parts positions when the vessel is still packed.

nuggreat:
Just checked and you can access the position of parts on packed craft and they do report a different position than the ship's position.

marianoapp:
Parts belonging to a packed vessel now return the correct position too. Since part positions are used to calculate a vessel bounds this change shifts the bounding box to the correct position.

Dunbaratu:
I was about to look into this last night but then I saw the chatter in the comments and realized you were about to change it some more, so I put it off.

Is it ready for merging now?  I may have some time to look at it tonight.

marianoapp:
Yes, it's ready to merge

Issue #2807, opened 2020-11-15
lock steering ignores wheel strength slider on torque wheels


Dunbaratu:
When calculating how to tune the PIDs, SteeringManager makes use of the stock call ``ITorqueProvider.GetPotentialTorque()`` to find out how capable the ship is of turning.

The problem is that GetPotentialTorque() always returns the max torque that a reactionwheel COULD provide, ignoring the slider that can reduce reaction wheel strength.

(To test, ``print steeringmanager:rawtorque.`` in a terminal, move the reaction wheel's slider, and print it again - there is no difference.)

I tested it with an RCS block and the RCS blocks don't seem to have the same problem - they do vary with the thrust limiter.  I also tested with a gimbal-able engine - it also seems to change with the thrust limiter since it pays attention to the actual thrust the engine is currently providing when calculating it (if the throttle is off, a gimbal engine does not contribute to GetPotentialTorque, as it should be.)

It seems that ONLY reaction wheels have this problem.

Dunbaratu:
Should have been closed when #2820 got merged, but I didn't include this in the description of #2820

PR #2820, opened 2020-12-09
Steering null retune 1

### LINKED ISSUE #2814: SteeringManager is confused about RCS because API call GetPotentialTorque() is wonky
The SteeringManager tries to tune things based on what it knows of the torque of the vessel, from information it gets from the API call ITorqueProvider.GetPotentialTorque().
(Search the code for "GetPotentialTorque" to see where it's using it).

The problem is that KSP returns inconsistent results for this call depending on what kind of torque providing part it actually is, and this is messing with the SteeringManager in cases where the majority of the torque comes from RCS rather than other sources.

Through experimental testing, this seems to be what it does:

- If the ITorqueProvider is a reaction wheel
    - Position of part from Center of Mass doesn't matter.  Torque provided is magically constant regardless of distance from center of mass.  This might actually be "correct" in the game universe if reaction wheels really do work this way in game.
    - Because position doesn't matter, returns a constant amount in all X, Y, Z axes.  i.e. the vector returned is V(10,10,10) for a reaction wheel that provides a torque of 10.
- If the ITorqueProvider is an engine with gimbal
    - Position of part from Center of Mass matters.
    - Vector returned is in the reference frame of the ship itself, with the usual swapping of Y and Z axes that controls seem to have.
    - Magnitude changes depending on engine thrust, which seems correct.
- If the ITorqueProvider is an areodynamic control surface
    - Position of part from Center of Mass matters.
    - Vector returned is in the reference frame of the ship itself, with the usual swapping of Y and Z axes that controls seem to have.
    - Magnitude changes depending on speed and air density, which seems correct.
- If the ITorqueProvider is an RCS part
    - Position of part from Center of Mass matters.
    - **Vector returned is NOT IN the reference frame of the ship itself**.  It appears to be in the reference frame of Unity's global **world coords**.  As the ship rotates in space around the world coords' x, y, z axes, it keeps providing a different answer.  Pointing the ship directly along the world X, world Y, and world Z axes returns values that appear to swap the 3 numbers around in the tuple exactly as one would expect if it was in the reference frame of world.

That last one is the problem.  The code in SteeringManager seems to presume that GetPotentialTorque() always returns values in the reference frame of the ship.  And that's true ... in all cases *except* for ModuleRCS for some weird reason.

Possible fixes.  Pick one:

1. Continue using the stock call, but check for the case where the ITorqueProvider happens to be a ModuleRCS, and if it is perform a reference frame transform to fix this problem and make it give the answer in ship relative coordinates.  Pro: least code change to kOS.  Con: It's designing it so two wrongs make a right, which isn't very future proof if KSP ever fixes this.

2. Throw away the use of ITorqueProvider and just make our own.  Pro: future-proof against KSP ever fixing the problem. Con: Duplicating KSP's code and thus not future proofed against a new type of ITorqueProvider coming out later that our homemade call doesn't know about.

Dunbaratu:
Fixes #2814 - had to make our own homemade replacement for GetPotentialTorque() in the case of RCS parts because the stock one is so very broken.

This starts from work other KSP modders did who also realized the stock ModuleRCS.GetPotentialTorque() is so wrong as to be impossible to just 'patch up', and wrote their own.

### DIFF
--- a/doc/source/commands/flight/cooked.rst
+++ b/doc/source/commands/flight/cooked.rst
@@ -296,10 +296,10 @@ settings that might help solve some problems, in the list below:
   This is specifically for when it wiggles the controls *during* its
   rotation to the destination orientation.)
 
-  - **solution**: Increase :attr:`STEERINGMANAGER:ROTATIONEPSILONMAX` to make
+  - **solution**: Increase :attr:`STEERINGMANAGER:TORQUEPSILONMAX` to make
     it "not care" about the exact rotation rate until it gets closer to
     the target orientation.
-    Increasing :attr:`STEERINGMANAGER:ROTATIONEPSILONMIN` can help also, but
+    Increasing :attr:`STEERINGMANAGER:TORQUEPSILONMIN` can help also, but
     making it too high could prevent the steering from holding the nose on
     target once it does reach the desired direction.
 
@@ -309,7 +309,7 @@ settings that might help solve some problems, in the list below:
   controls at all).
 
   - **solution**: You might have to either decrease
-    :attr:`STEERINGMANAGER:ROTATIONEPSILONMAX` or increase
+    :attr:`STEERINGMANAGER:TORQUEEPSILONMAX` or increase
     :attr:`STEERINGMANAGER:MAXSTOPPINGTIME`.
 
   - **explanation**: The problem may be that your vessel is so slow at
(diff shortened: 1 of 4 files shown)


Dunbaratu:
This should have also closed issue #2807 about torque wheel slider being ignored.  The homemade GetPotentialTorque() handles this as well.

PR #2839, opened 2021-01-14
Fixes #2838 with a little button in the toolbar window.

### LINKED ISSUE #2838: New boot files don't show up in editor within core PAWs without exiting/re-entering editor
I tried and failed to find a dupe of this but I'd be surprised if it isn't a dupe.  Given how messy exiting the editor can be when one has detached subassemblies lying about it would be a big win if when any time the PAW for a core is brought up in the editor and the boot file selector clicked that a check is done for new files in Script/boot and they are added to the available boot file options. I'm not trying to get this in as a last minute add to the upcoming release, timing is coincidental.  It is completely understandable that it wouldn't be in the upcoming release if ever

Dunbaratu:
Fixes #2838 

In the end I went with the quick and simple solution of putting the button in the toolbar window.

It still doesn't force the PAW to re-read the value if you have it open when you click the button, but if you close the PAW and re-open it, it has the new list.

### DIFF
--- a/src/kOS/Module/kOSProcessor.cs
+++ b/src/kOS/Module/kOSProcessor.cs
@@ -439,7 +439,14 @@ namespace kOS.Module
 
         public void ForcePAWRefresh()
         {
-            part?.PartActionWindow?.UpdateWindow();
+            // Thanks to https://github.com/blowfishpro for finding this API call for me:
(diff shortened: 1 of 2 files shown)


mgalyean:
That is far simpler than abandoning the editor and re-entering, very much appreciated, thank you

PR #2879, opened 2021-02-18
Add ReactionWheel partmodule handling

thexa4:
Added documentation but haven't been able to run the compile on it so I'm unsure whether I did the links right.

Adds support for reading out the torque of reaction wheels to enable players to tune their own steering managers.

Code has manually been tested in-game.

### DIFF
--- a/doc/source/structures/vessels/reactionwheel.rst
+++ b/doc/source/structures/vessels/reactionwheel.rst
@@ -1,0 +1,76 @@
+.. _reactionwheel:
+
+Reaction Wheel
+======
+
+Some of the Parts returned by :ref:`LIST PARTS <list command>` will be a Reaction Wheel. It is also possible to get just the Reaction Wheel parts by executing ``LIST ReactionWheels``, for example::
+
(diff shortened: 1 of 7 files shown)


Dunbaratu:
Possible problem: Some parts are more than one kind of part (this is a big problem in kOS in general).  A part may be both a reaction wheel and a command pod, for example.
Perhaps we could expose the internal method used in SteeringManager for parts: https://github.com/KSP-KOS/KOS/blob/6ace6c0dfd49d89d262e77dd8a0f8ac2dabf3e2b/src/kOS/Control/SteeringManager.cs#L685
so it isn't just reaction wheels it's also RCS thrusters and engine gimbals, etc.

nuggreat:
If we expose torque we should also expose the moments as right now you need to calculate them indirectly from angularVelocity and angularMomentum.  The main reluctance I have put forward in the past as to why we don't expose that info is due to airo surfaces and how the torque from those tends to be more of a guess than a true value.

thexa4:
Good points, I'll rethink the design

Issue #2889, opened 2021-03-09
Some type of support for crude multithreading with more than one kOS part.


Dunbaratu:
There's been enough people trying to use multiple kOS parts in conjunction as a sort of multi-threaded approach that I guess it's time to try to support at least *some* sort of crude model that will allow that.  The key things that cannot be done by user-level code alone and need kOS support are:

1. Some way to share memory between cores on the same vessel.  This would be a way for one core to set a variable that another core can read.)  One possible way is to use the current ship message queue, but it would be nice if they could just share variables more directly with, say, one universal variable dictionary that all kOS parts can see (a sort of "even more global than global" namespace.)  One possible way to do this with minimal effort would be to just provide a single lexicon everyone can see, like if one core does ``set ship:sharedlex["foo"] to 5.`` then another core can do ``print ship:sharedlex["foo"].`` (Intuitively that might seem like a slow way to access memory, but in kOS even the "direct" variables are still dictionary lookups in the C# under the hood so that doesn't really matter too much.  A program could just first do ``set sharedFoo to ship:sharedlex["foo"].`` up front, then refer to ``sharedFoo`` from then on, if the author is concerned about speed of accessing the variable.)
2. Some way for a script to have "atomic access" to a semaphore or mutex, so that when the script's algorithm requires it, it can declare atomic sections where the other cores requesting the same "baton" have to wait their turn.  Please note, this would *NOT* be an "atomic section" from the point of view of KSP itself.  This kind of "atomic section" could still run out of IPU and have to continue next physics update.  But what this would do is make it so that if a core that "has the baton" runs past its IPU and has to continue next physics update, the other kOS cores that requested access to the same "baton" won't wake up and run their opcodes until that "baton" is freed by this one.  (I envision something like this:  Currently each physics update, a core wakes up and runs IPU opcodes.  The change would be that a core wakes up, checks to see if it's requesting a baton that's in use, and only if that baton is not in use will it proceed to execute the next IPU worth of opcodes, otherwise it will to go back to sleep and try again next update.)  Obviously, there will need to be failsafes to make sure "the baton" gets freed when a program dies or a the core part explodes.

[...]

Issue #145, opened 2014-06-16
new keyword proposal for marking sections that should happen in one update.


Dunbaratu:
I was thinking that what might make a lot of sense is to let users mark sections of the code as atomic, meaning they must occur without being interrupted partway through.  This is for solving the problem where the number of instructions per update runs out in the midst of trying to take physical measurements, so that there's a large "jump" in time between adjacent measurements that the script programmer didn't expect, throwing off the calculations.

i.e. this:

```
UNTIL 0 {
  SAMETIME {  // < -- proposed new keyword
    set prevVel to ship:velocity:orbit.
    set prevPos to ship:position.
    set prevFuel to ship:liquidfuel.
    set prevTime to time:seconds.
  }
  WAIT 0.0001.
  SAMETIME {  // < -- proposed new keyword
    set newVel to ship:velocity:orbit.
    set newPos to ship:position.
    set newFuel to ship:liquidfuel.
    set newTime to time:seconds.
  }

  // Now the code does stuff comparing the new and prev values..

}.
```

To a certain extent you can get this right now with an explicit WAIT, because you can WAIT 0.0001 just before entering the "SAMETIME" sections instead, but that depends on knowing that your statements all fit within one CONFIG:IPU, and being aware of the compiler details to know how many instructions your script commands become seems wrong.

As to the intended behavior under the hood, I was thinking of this:
- At the start of SAMETIME, put the program to sleep until the next Update tick, so the first line of the SAMETIME section is the first line of code executed in a new update.  Also, set a runtime flag for "I am in a SAMETIME section right now" that will get cleared at the end of the braces section.
- When a SAMETIME is currently active, if you reach the end of CONFIG:IPU instructions before you reached the end of the SAMETIME section, then we do one of two things: Either we:
  - Issue a warning to the terminal stating that the SAMETIME section is too big to fit in one update, and continue behaving as normal - the program stops and waits for the next update tick.  This would make SAMETIME into merely a sort of warning system.  It doesn't force the code to exist in one tick - it just lets the script programmer know that it doesn't so they have to redesign their code.
  - Or, Ignore the normal rules and continue executing until the end of the SAMETIME section is encountered before finishing the update tick, even though that exceeds CONFIG:IPU.  Thus SAMETIME becomes an enforcement on kOS demanding it behave like the programmer wants.  This carries risk of stuttering slowness or even infinite hang, but we _already_ have exactly that risk in how triggers execute, which don't adhere to any limit of instructions per Update().
[...]

PR #2890, opened 2021-03-09
Add TICKSLEFT binding

thexa4:
Can work as a solution to #2889, #145

### DIFF
--- a/src/kOS.Safe/Binding/BoundVariable.cs
+++ b/src/kOS.Safe/Binding/BoundVariable.cs
@@ -1,4 +1,4 @@
-﻿using kOS.Safe.Execution;
+using kOS.Safe.Execution;
 using kOS.Safe.Encapsulation;
 
 namespace kOS.Safe.Binding
@@ -7,6 +7,8 @@ namespace kOS.Safe.Binding
     {
         public BindingSetDlg Set;
         public BindingGetDlg Get;
+
+        public bool Volatile = false;
 
         private object currentValue;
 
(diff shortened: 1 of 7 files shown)


Dunbaratu:
I don't think it does everything desired by those issues, but it is a fine idea to have this.
I think the documenation should make it clear whether it's "instructions that were left prior to this call" or "instructions that are left afer this call".

Reading when ``shared.Cpu.instructionsPerUpdate`` gets incremented, it appears to get incremented *after* executing an instruction, so when the current instruction is evaluating the suffix and putting the answer on the stack, it's seeing the value minus one.

I think we can just move the increment line from CPU.cs to just before, rather than just after, the ExecuteInstruction and it won't mess anything up.  It would make the result of this easier to nail down.

i.e. currently it does this on line 1505 of CPU.cs:
```
                    executeNext = ExecuteInstruction(currentContext, doProfiling);
                    ++InstructionsThisUpdate;
```
and it could be changed to this without introducing a problem, I think:
```
                    ++InstructionsThisUpdate;
                    executeNext = ExecuteInstruction(currentContext, doProfiling);
```
This would mean that semantically "InstructionsThisUpdate" is now counting how many instructions have happened *including* the current one, when before it was counting how many instructions have happened *excluding* the current one.

src/kOS.Safe/Binding/BoundVariable.cs
```diff
@@ -20,6 +22,8 @@ public override object Value
                 // new primitive encapsulation types we instead encapsulate any value returned
                 // by the delegate.  This makes it so that all of the getters for bound variables
                 // don't need to be modified to explicitly return the encapsulated types.
+                if (Volatile)
+                    return Structure.FromPrimitive(Get());
                 return currentValue ?? (currentValue = Structure.FromPrimitive(Get()));
```
Dunbaratu, line 27: I always have a preference for not duplicating a call if there's a way to write it only mentioning the call once. (less likely for a future edit to forget one of the two duplicated places).

how about:
```
if (!Volatile && currentValue != null)
    return currentValue;
return (currentValue = Structure.FromPrimitive(Get()));
```

PR #2900, opened 2021-03-16
Fixes #2899 terrain raycast now ingores parts.

### LINKED ISSUE #2899: Geoposition terrainheight is wrong for LATLNG's that are very close to a rover scanner arm
I swear I thought this bug was already reported, but as I look through the issues, I can't seem to find it, so I'm reporting it anew.

**To reproduce the problem:**

1. Place a rover scanner arm on a vessel and launch it to the scene, by which I mean one of these three parts from the Breaking Ground DLC: https://wiki.kerbalspaceprogram.com/wiki/CRSY_Light_Scanning_Arm, https://wiki.kerbalspaceprogram.com/wiki/SPRT_Medium_Scanning_Arm, https://wiki.kerbalspaceprogram.com/wiki/OP-E_Large_Scanning_Arm
2. Ask kOS to tell you what the ``ship:geoposition:position`` vector is.  Use a vecdraw to visualize it.
3. Notice the position is NOT the terrain under the vessel like it should be.  It's weirdly a few meters up in the sky above the vessel.  If your vessel did NOT have a scanner arm on it, then the vector would be correct, pointing at the ground just under the rover.

**The cause, I think:**

The reason for this happening is something I consider KSP's fault more than kOS's fault, but we may be able to find a means to work around it in the kOS code.

The rover scanner arm parts appear to contain an invisible large spherical collider around them for some reason, and for some reason this collider is on layer 15, which is normally used by KSP for terrain. This sphere appears to be the exact same size and shape as the green "preview range" sphere you can see in the VAB for these scanner arm parts when you rightclick them.  

``GeoCoordinate.GetTerrainAltitude()`` uses a raycast to "find" the real terrain hit for LATLNG's near the active vessel where terrain is loaded, and because this weird phantom sphere is on the same layer as the terrain, it's getting seen as a "hit" for the raycast hitting what it thinks is terrain.

Since it seems likely the stock game won't fix this bizzare feature of a phantom collider on the terrain layer that emphatically is Not terrain, we may have to find a way to work around it.  (Note, this bug affects stock too, as the phantom collider also blocks solar panels from the sun, but again I don't think it likely that this is getting fixed any time soon.)

Dunbaratu:
Fixes #2899

The fix I went with was to detect whether or not a raycast hit
is inside a Part.  If it is, then start a new raycast just past
the hit and keep looking.

The solution involved writing a wrapper around Raycast that
includes this extra logic, such that it could be lifted and
put in a utility library later if any other classes need to
use it.

Using the same example script and craft as in the issue #2899, with this change it now looks like this:
screenshot26

### DIFF
--- a/src/kOS/Screen/Interpreter.cs
+++ b/src/kOS/Screen/Interpreter.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Text;
 using kOS.Execution;
--- a/src/kOS.Safe/Exceptions/KOSYouShouldNeverSeeThisException.cs
+++ b/src/kOS.Safe/Exceptions/KOSYouShouldNeverSeeThisException.cs
@@ -5,7 +5,7 @@ namespace kOS.Safe.Exceptions
 {
     public class KOSYouShouldNeverSeeThisException : KOSException
     {
-        public KOSYouShouldNeverSeeThisException(string message) : base("This is an error endusers should never see, if you see this please report it to the kOS devs:\r\n    " + message)
+        public KOSYouShouldNeverSeeThisException(string message) : base("This is an error end users should never see. If you see this please report it on the kOS github:\r\n    " + message)
         {
         }
     }
--- a/src/kOS/Suffixed/GeoCoordinates.cs
+++ b/src/kOS/Suffixed/GeoCoordinates.cs
@@ -1,6 +1,7 @@
 using UnityEngine;
 using kOS.Safe.Encapsulation;
 using kOS.Safe.Encapsulation.Suffixes;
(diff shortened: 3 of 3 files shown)


src/kOS/Screen/Interpreter.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
Dunbaratu, line 1: I have no idea why git seems to think this line changed.  It didn't.  It's literally identical, even down to the whitespace.

thexa4: A lot of files have an utf byte order mark (BOM) in front of it. Sometimes used to happen when creating new files with older VS versions.

PR #2909, opened 2021-03-27
Fixes #2894 via ToStringIndent()

### LINKED ISSUE #2894: Printing containers (lex, list) is now overly verbose
As a side effect of the work to make single primitives serializable, printing a list or a lexicon on the terminal now takes twice as much vertical screen space as it used to.  This makes it far less likely to see the thing you're looking for and much less usable than before.  For example, I would commonly print a module to see all of its fields and events, but now it takes way too much space and can't fit on a screen.

Dunbaratu:
Fixes #2894

There are going to be two proposals for how to fix #2894. (Read the comments for the issue).

This is one of them.

In this proposal, We stop borrowing the Dump infrastructure for
creating ToString()s.  The Dump infrastructure was designed for
when you want to serialize, meaing every single field must be
visited and written with nothing 'missing', so values can be
saved and restored.  I found that layering on top of that
system while still preserving the notion of ToString() being meant
for human consumption was kind of hard to wrap my head around
so I went with this idea.

This PR is being made before this is fully tested, so it can be
shown to other people who are proposing alternate solutions.  I
also want to show what this output ends up looking like when you
try it.  Note the code still has some "eraseme" things in it
at this point, which is how I remind myself that an idea wasn't
done yet.  (I have a check I do before a final PR, usually, where
I 'grep' all my code for the string "eraseme" to highlight places
I stuck something in temporarily during testing.  Here I'm using
that to remind myself of stuff that isn't done yet in this PR.

While I think this produces good output, I am willing to be convinced
other approaches might work too, so I'm open to trying to preserve
using Dumper for indented ToString()s if it can be done.

### DIFF
--- a/src/kOS.Safe/Encapsulation/EnumerableValue.cs
+++ b/src/kOS.Safe/Encapsulation/EnumerableValue.cs
@@ -1,5 +1,6 @@
 using kOS.Safe.Encapsulation.Suffixes;
 using kOS.Safe.Serialization;
+using System.Text;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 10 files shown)


Dunbaratu:
Here is an example program and the output it makes with this PR.

Example program:
```
set L1 to LIST(100,200,300,400,500,600,LIST("a","b","c"),"Hello").
set L2 to LIST(10, 20, 30, 40, 50, 60, "inner 1", "inner 2", red, green, blue, range(0,50,2)).
// Let's nest L2 inside L3:
set L3 to LEX("key 1", "value 1", "key 2", "value 2", "key 3", L2).
set L4 to QUEUE("A", "B", "C").
set L5 to STACK("A", "B", "C").
L1:add(L3).
L1:add(L4).
L1:add(L5).
L1:add(LIST()).
L1:add(LIST("one thing")).
L1:add("just one last thing").
print L1.
```

With output like this:
screenshot27

Dunbaratu:
Hah I just realized I used the terminology "top ->" to mean the start of a Queue and "front ->" to mean the start of a Stack.  Those terms should be swapped.

PR #2910, opened 2021-03-27
Make printed structures slightly more readable

thexa4:
Could be a possible solution for #2894

Example output with the same script as #2909:
image

### DIFF
--- a/src/kOS.Safe/Encapsulation/QueueValue.cs
+++ b/src/kOS.Safe/Encapsulation/QueueValue.cs
@@ -37,6 +37,12 @@ namespace kOS.Safe.Encapsulation
             InnerEnumerable.Enqueue(val);
         }
 
+        public override Dump Dump()
+        {
+            Dump result = base.Dump();
+            result.Annotations.Add(0, "<-- front");
+            return result;
+        }
         public override void LoadDump(Dump dump)
         {
             InnerEnumerable.Clear();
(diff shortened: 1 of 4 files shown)


Dunbaratu:
I'm not at my dev computer to test this right now, but can you increase the max depth a bit so the RGBA() and RANGE() values get printed out in the example script?  They were in there to test the case of something that *does* have a ToString() defined but is also Serializable so it has Dump logic, to see if the Dump printout ends up overriding the ToString printout like it does with the current code.  (i.e. the problem in the current implemention, prior to either PR, is that instead of seeing this as the value in the list:
```
RGBA(1, 0, 0, 0)
```
it showed this:
```
[8] = RGBA
["r"] = 1
["g"] = 0
["b"] = 0
["a"] = 0
```
(Which led to the strange effect that the 8th element when you ``print L2.`` looked nothing like what you get when you ``print L2[8].`` directly.  The latter was going through ToString() and the former was doing a more raw dump of fields.)

thexa4:
Right, I did not add anything for this. Lets see what I can do

image

thexa4:
Added support for summarizing an object in one line.

image

Dunbaratu:
Unrelated note - how are you running that in the VAB?

thexa4:
> Unrelated note - how are you running that in the VAB?

https://github.com/KSP-KOS/KOS/compare/develop...thexa4:feature/ksc-mainframe (has quite a lot of changes and isn't well tested, didn't want to create a PR for it yet)

thexa4:
TODO:
- [ ] Refactor so that printing control flow is done locally, not in the formatters
- [ ] Try to look at prefix annotations

thexa4:
Initial implementation doesn't crash instantly, testing with it now.

Issue #2905, opened 2021-03-25
BOUNDS returns wrong values when Waterfall mod is installed.


Dunbaratu:
I was noticing my landing script thought my vessel was over 18 meters tall, when clearly it was only about 3 or 4 meters high.

Diagnosing this, it turns out that engines return wrong bounds when the Waterfall mod is installed.  It seems to be that the exhaust plume is being counted as part of the engine's dimensions (even when the throttle is zero so the engine isn't firing).

To reproduce:
1. Have these two mods installed: Waterfall, and Stock Waterfall Configs.
2. Make a ship that uses a Terrier engine (may happen with other engines too, but this is the one I noticed it with).
3. ``for p in ship:parts { print "bounds for " + p + " is " + p:bounds. }``
4. Note the ``:RELMIN`` for the Terrier engine part has a negative Z of about 18.1 meters.

To fix this we should probably have some better logic within bounds calculations to know which components are "not physical" and therefore should not count.  Something that's purely visual with no colliders probably shouldn't count.

Dunbaratu:
Probable cause: ``PartValue.GetBoundsValue()`` iterates over all Meshes in the component with this:
```
MeshFilter[] meshes = Part.GetComponentsInChildren<MeshFilter>();
```
Some of those meshes will be visual-only (and will be unnecessarily detailed with more vertices than the object "physically" has in its collider).
It should probably be iterating over just the colliders, not all the meshes, using this:
```
Collider[] colliders = Part.GetComponentsInChildren<Collider>();
```
The ``Collider`` type has a ``.bounds`` member just like the mesh type does so presumably the algorithm from there down shouldn't need to change that much.

Dunbaratu:
Closing to redirect it over to issue #2942.

PR #2945, opened 2021-05-21
Change bounds creation to use colliders instead. #2942

### LINKED ISSUE #2942: Bounds should calculate using Colliders instead of MeshFilters.
kOS Bounds is based on iterating over these MeshFilters:
```
Part.GetComponentsInChildren<MeshFilter>();
```
But it turns out MeshFilters are all the Rendering shapes, not all the collider shapes.  Generally the same algorithm could be used but have it iterate over the part's colliders rather than its ``MeshFilter``s, i.e. these:
```
Part.GetComponentsInChildren<Colllider>();
```

Using the MeshFilters causes these three issues:

1. It causes issue #2905 with the Waterfall mod (which I will close and redirect here).
2. Because the rendered mesh is typically more high-res than the collider, there's a discrepancy between the answers.  The collider is the one that physically matters so it's the one kOS should return.
3. Because the rendered mesh is typically more high-res than the collider, using the collider will give it fewer vertices to iterate over, making the algorithm less expensive.

### Ignore triggers

IMPORTANT: When using Colliders, bypass any that have ``Collider.isTrigger == true``.  Those don't count as a real physical object (the physics engine does not make it hit things, it passes right though things).  A trigger is a special case sort of collider that merely fires off a callback hook when it touches things, so the programmer can react to that callback in whatever arbitrary way.  An example of where this is used in KSP is the invisible collider spheres around robot scanner arms in the Breaking Ground DLC.  Those colliders are "trigger" colliders and ignoring them will prevent them from falsifying the bounds.

CodeLeopard:
Changes the way a part's Bounds are computed from using the part's meshes to using colliders.
Colliders that are triggers are excluded.

Fixes #2942

See #2905 for additional information.

### DIFF
--- a/src/kOS/Suffixed/Part/PartValue.cs
+++ b/src/kOS/Suffixed/Part/PartValue.cs
@@ -101,7 +101,72 @@ namespace kOS.Suffixed.Part
                 // They merely fire an event when they intersect with something, but have no physics effect.
                 if (collider.isTrigger) continue;
 
-                Bounds bounds = collider.bounds;
+                // Colliders report their bounds in world space, as AABB. But we need the bounds in Part local space.
+                // AABB also means that there may be (significant) empty space inside the bounds.
+                // To get the local bounds:
+                // For a MeshCollider we get the sharedMesh.bounds which does use local space.
+                // For the other colliders we need special cases.
+
+                Vector3 center;
+                Vector3 extents;
+
+                if (collider is MeshCollider meshCollider)
+                {
+                    center = meshCollider.sharedMesh.bounds.center;
+                    extents = meshCollider.sharedMesh.bounds.extents;
+                }
+                else if (collider is BoxCollider boxCollider)
+                {
+                    center = boxCollider.center;
+                    extents = boxCollider.size / 2;
+                }
+                else if (collider is SphereCollider sphereCollider)
+                {
+                    center = sphereCollider.center;
+                    extents = Vector3.one * sphereCollider.radius;
+                }
+                else if (collider is CapsuleCollider capsuleCollider)
+                {
+                    center = capsuleCollider.center;
+
+                    extents = Vector3.one * capsuleCollider.radius;
+                    // The Capsule could face in one of 3 directions, overwrite extent on that axis.
+                    switch (capsuleCollider.direction)
+                    {
+                        case 0:
+                        {
+                            extents.x = capsuleCollider.height / 2;
+                            break;
+                        }
+                        case 1:
+                        {
+                            extents.y = capsuleCollider.height / 2;
+                            break;
+                        }
+                        case 2:
+                        {
+                            extents.z = capsuleCollider.height / 2;
+                            break;
+                        }
+                        default:
+                        {
+                            throw new Exception("Unknown CapsuleCollider direction: " + capsuleCollider.direction);
+                        }
+                    }
+                }
+                else if (collider is WheelCollider wheelCollider)
+                {
+                    center = wheelCollider.center;
+                    extents = Vector3.one * wheelCollider.radius;
+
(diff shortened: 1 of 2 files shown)


Dunbaratu:
I was just playing with this installed today and I noticed the bounds are still including the realfuels' engine plumes.  I'm not sure why those are "colliders" but it may be that it's not enough to check for triggers.  Maybe it's also necessary to filter some colliders out by layer number?

mgalyean:
This is an interesting problem.  A plume, for realism, would need colliders.  But this gets into the realm of "flavors" of collision.  One certainly wouldn't want the same collision rules for smoke as for a 3kt engine.  Is there some kind of mask that tells a collision mesh what it is allowed to collide with?  I'm gathering that the plume is perhaps considered a part of the craft and so brings its collision mesh along with when trying to determine bounds.  If not the layer number one would have to find something different about the plume meshes to filter them out by for sure and I'm curious what that ends up being, but if the plume is considered a craft part, maybe filtering it out on the part level prior to looking at meshes might work.  And I don't even know if that makes sense

Dunbaratu:
Okay, so the plumes are NOT the problem.
Verified the problem also happens on Stock.
Something about this PR changing things from Meshes to Colliders also made the bounds all bugged out.
If you use the TestBounds script from here, on stock, you can see this PR messed things up - and I can't figure out why unless colliders exist in some other alternate reference frame that's different from the reference frame Meshes use.  Because literally all that changed is the collider bounds instead of the mesh bounds are used.

Dunbaratu:
I think this is the issue:

The documentation for Mesh.bounds in Unity says this:

> The bounding volume of the Mesh.
>
>This is the axis-aligned bounding box of the mesh in its local space (that is, not affected by the transform)

The Unity documentation for Collider.bound says this:

> The world space bounding volume of the collider (Read Only).

The Mesh bounds is in local space, which we then find the 8 corners of it and transform that into World space afterward.

The Collider bounds starts out already in world space, which means:

(A) It's wrong to do that transform on them, and

(B) They'll have extra space around the corners because world axes don't align with the part's orientation.

It looks like there is a solution but it's messier than just a drop-in replacement.  It goes like so:

Get the Type of the Collider.  Colliders come in 4 types: Sphere, Capsule, Box, and Mesh.

If the type is MeshCollider, then use the code already written for Meshes, but you have to get the ``MeshCollider.sharedMesh.bounds`` (which is local space like the code expects) and not the ``MeshCollider.bounds`` (which is world space).

If the type is ``BoxCollider``, ``CapsuleCollider``, or ``SphereCollider`` then they have no direct API to return the local bounds, but kOS can manually work out the what the local bounds would be based on the relatively simple parameters of these shapes.

CodeLeopard:
Thanks for looking in to this and identifying what exactly was wrong, I'll update this PR to include the suggested fix.

CodeLeopard:
I have implemented the suggested solution. Some notes though:
- I came across WheelCollider and had to add a reference to be able to deal with it.
- Currently if an unexpected collider type appears an exception will be thrown. Should that be handled differently, perhaps with a default value or best guess?

Dunbaratu:
> Currently if an unexpected collider type appears an exception will be thrown. Should that be handled differently, perhaps with a default value or best guess?

Good question.  I think throwing an exception here is probably inappropriate because it's not something the user can fix in their script.  Maybe just a SafeHouse.Logger.Log() message rather than an exception.  Then we can see it is happening but it won't break the script execution.

CodeLeopard:
Okay I changed it to log and skip.
Log message includes the Part's toString and collider index to help diagnosing.

Dunbaratu:
I started playing with this installed today and now it seems to be working properly.  Good Job!

PR #2955, opened 2021-06-04
Improved RP-1 tech progression

RCrockford:
Changes RP-1 kOS cores to have tech progression that follows the user selected avionics tech.

This requires changes in RP-1 in addition for correct behaviour, but should be safe to merge before that as it will do nothing if the changes are not present.

### DIFF
--- a/src/kOS/Module/kOSProcessor.cs
+++ b/src/kOS/Module/kOSProcessor.cs
@@ -302,6 +302,83 @@ namespace kOS.Module
             additionalMassGui = AdditionalMass * 1000;
         }
 
+        private PartModule RP1AvionicsModule = null;
+
+        private void FindRP1Modules()
+        {
+            RP1AvionicsModule = null;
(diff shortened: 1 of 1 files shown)


Dunbaratu:
I just tried loading this into my RP-1 save (without the associated RP-1 PR, so I could verity that it doesn't break things if this is in kOS but the RP-1 part of it isn't there) and immediately ran into this problem - I cannot get the numbers to appear.  It looks like this (two screenshots included):
screenshot21
screenshot22

Dunbaratu (changes requested):
I think I may have found why when I tested this on an existing RP-1 installation it caused the problems.

RCrockford:
The exception was in some debug logging code that isn't really required ironically. I have removed it, so it should be more robust now.

RCrockford:
The required changes in RP-1 have now been released, so that should make testing this a little easier.

Issue #1617, opened 2016-04-20
Add KSPedia integration


hvacengi:
We've discussed how best to do a KSPedia page.  The issue is that we don't know of any easy way to export our existing documentation into a KSPedia friendly format.  So we're probably looking at making a couple of basic pages, with links to the full online documentation.

Someone on the forum composed a walk through on how to add this support, though I have not had time for a thorough review of it yet: http://forum.kerbalspaceprogram.com/index.php?/topic/135434-how-to-creating-new-kspedia-entries-updated-with-full-walkthrough/

PR #2958, opened 2021-06-06
KSPedia Panels for kOS

kna27:
This PR creates KSPedia panels for kOS. The panels are split into two sections, Tutorials, and Documentation. Each has very brief summaries of the most commonly used features of kOS and is shrunk down. Many panels have a link to the full page on the kOS docs so the reader can get more information. The panels are intended to get the user of kOS started using the mod, and for them to look at the docs or other external resources to learn more. If someone could please review the layout, the order, and the text/content of the panels, it would be greatly appreciated. To add the panels to KSP, copy the `KOS/src/kOS KSPedia/AssetBundles/kos.ksp` file to `GameData/kOS`. The kos.ksp file can be placed anywhere within the GameData folder.

#1617


Dunbaratu:
To be honest I have no clue what to do with these 623 files in the PR.  Some of them look like generated content instead of source content.  There's DLL's and EXE's in here.  There's shader files (did you actually make new shaders yourself or are these just stock shader files that got added to the commit?), .meta files, etc.

kna27:
Hello,
Many of these files were generated by Unity, but are specific to the project and are necessary to the project. All the .meta files are used to determine GameObject's layers and sorting orders, and thus need to be committed to prevent any errors in the build. The shaders are used by the KSP Part Tools, and I am pretty sure removing them would also cause compiling errors. There are many folders that are not necessary and not edited by me (Including sln files, csproj files, Libraries folder, etc) that I did not commit and have added/were already added to the .gitignore. This is normal for Unity. To collaborate on a Unity project, the Assets folder and the ProjectSettings folder have to be committed, which is what I did. I also included the AssetsBundle folder because that is where the .ksp file is built to and is specific towards KSP Part tools. I already removed files that don't have to be committed.

mgalyean:
This is a lot of additional overhead.  Is anyone else thinking this is more of a separate, but related mod, kind of thing?  I'm fairly sure most ppl would want this to be optional to kOS, but maybe I'm wrong.  I could definitely see kOS playing smoothly with this type of thing, but don't see it as logically being a non-optional aspect of kOS.  Unity/Mono is challenged enough as it is for many of us.  If I'm misunderstanding I'm open to enlightenment

kna27:
> Is anyone else thinking this is more of a separate, but related mod, kind of thing?

I mean all that this is adding to kOS is some panels in the KSPedia, I think a separate mod or repo would be a little too much considering this. This is a fairly small Unity Project that has been added to a single folder, so I don't think there is too much overhead.

> Unity/Mono is challenged enough as it is for many of us. If I'm misunderstanding I'm open to enlightenment

This took exactly zero scripting to make. All you need to do is use the Unity UI to place panels, prefabs, text, and images the way you want, and to use KSP Part Tools to build it. I followed [this](https://forum.kerbalspaceprogram.com/index.php?/topic/137628-kspedia-creation-tutorial/) tutorial on the KSP Forums to create it.

mgalyean:
So it would not cause more non-optional memory usage on kOS' part?  If that is the case then my worries are unfounded.  I just don't want increased load time or increased memory usage for something that given I have a web browser and the KSP screen is already too crowded I know I wouldn't make use of even if it carried no overhead. Others may think it useful though and I'm not trying to step on that, just not wanting any additional overhead to be non-optional

kna27:
> So it would not cause more non-optional memory usage on kOS' part?

Yes, this adds less than 550KB to kOS, making the mod only 6% bigger. Plus, it is only a .ksp file, which is handled well memory-wise since it is only loaded once at start iirc.

kna27:
Hello @Dunbaratu, I removed every single Unity project file and other files (including .meta), so now only the prefabs, the images, fonts, and the build remain. Because of this, if any wants to work on it they need to make their own Unity project and set up KSP Part Tools, but now there are much fewer files that are going to be added (623 to 36).

Dunbaratu:
@kna27 I wasn't necessarily objecting to the existence of the files, I just wanted to know what they were and to verify they were "source" rather than "generated".  (Seeing .DLLs and .EXEs made me assume they were generated, and worse yet, not cross-platform.  But if those are Mono DLLs and Mono EXEs such that they work on all platforms developers may be using, I think it should be okay to have it in the repo).

I was looking at the github repos for other KSP mods that have KSPedia pages in them, and I believe having that large pile of files is indeed the typical standard way other modders are doing it, so I think they can come back into the PR.  Sorry for making you think you had to do the extra work to remove them.  I was merely asking for further info first.

kna27:
Okay, I am sorry for the miscommunication. I added around half of them back, with the other half being TMP example assets that I somehow missed before. Note that before this is ready to be merged, someone should really proofread and just check if they like the text on the panels or if anything at all should be changed. I have a question though, is there some sort of automatic deploy system for kOS, or is the GameData folder put together by hand? Because if so, the person building it needs to add the `KOS/src/kOS KSPedia/AssetBundles/kos.ksp` to the gamedata folder. If there is some sort of automatic build/deploy system, the `kos.ksp` file needs to be copied to the GameData folder for anything to actually happen in-game. Thanks!

Dunbaratu:
No problem.  It's mostly on me for being unfamiliar with using Unity editor.
What I plan to do is take this PR and test that all the Unity files needed are there by trying to load it into Unity myself.

PR #2972, opened 2021-06-16
Enables some formerly debug suffixes for steering manager

RCrockford:
They are handy!


nuggreat:
Not all of these suffixes should be exposed though the steering manager and some of them are set to be exposed on SHIP through PR #2881 specifically MOI and torque.

I will also comment here as I did there that any exposed vectors should be in `ship-raw` frame or `ship-ship` frame and should be structured such that they are convertible between by multiplying by `SHIP:FACING` or `SHIP:FACING:INVERSE` depending on the initial frame.

Also the usual note about "new" suffixes needing documentation.

PR #2975, opened 2021-06-22
Comma-list declarations issue 2856

### LINKED ISSUE #2856: Support comma-list declarations
e.g. this:
```
local a is 3, b is 20, c is "hello".
```
and this:
```
set x to 3.21, y to 50.12.
```
This would be consistent with ``parameter`` which already allows this.

DanTGL:
Added comma-list declarations.

Fixes #2856 

It is not finished yet as I have only added them for `SET` right now.

### DIFF
--- a/doc/source/language/variables.rst
+++ b/doc/source/language/variables.rst
@@ -92,6 +92,18 @@ Any variable declared with ``DECLARE``, ``DECLARE LOCAL``, or ``LOCAL``
 will only exist inside the code block section it was created in.
 After that code block is finished, the variable will no longer exist.
 
+It is also possible to declare multiple variables in a single ``DECLARE`` statement,
+separated by commas, as shown below::
+
+    // These all do the exact same thing - make local variables:
+    DECLARE A IS 5, B TO 1, C TO "O".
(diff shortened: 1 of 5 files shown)


marianoapp:
If you made changes to the grammar you need to commit the kRISC.tpg file too. I assume you did since the PR includes changes to the Parser class.

thexa4:
Does this allow `set foo to 1, bar to foo.`?

DanTGL:
@marianoapp I didn't realize that I was supposed to modify the kRISC.tpg file, but I've done it now and generated the code with TinyPG.

@thexa4 Yes, that works.

DanTGL:
This basically works by splitting them up in separate operations, so the only actual changes are how the statement is parsed and compiled. So it basically still functions the same way as having separate statements.

This is basically the only thing I've changed in the code aside from kRISC.tpg:
https://github.com/KSP-KOS/KOS/blob/87e9c021b0bae6cb53749932484577780d3649b2/src/kOS.Safe/Compilation/KS/Compiler.cs#L2175-L2183

PR #3007, opened 2021-08-21
Provide a mechanism for addons to customize partmodulefields

JonnyOThan:
This works for me...it's a little lacking in that "partmodule as XXX" can better handle module inheritance.  Right now an addon would have to explicitly specify all child classes it knows about (and would miss any added by other mods).  Alternatively we could just have a list of construction functions and call ALL of them, returning the first one that doesn't return null.  But that's a hypothetical issue.


mgalyean:
Is there a way to not run until all other patches have been applied and so catch other mod' patches?  I really don't know

JonnyOThan:
> Is there a way to not run until all other patches have been applied and so catch other mod' patches? I really don't know

I don't think that solves the problem.  The issue is this:

Addon A provides ModuleA and registers itself to customize the partmodulefields for that module.  Mod B adds ModuleB which inherits from ModuleA.  Mod B might not even care about KOS.  In theory we'd still want the customized partmodulefields for ModuleB, but they won't work because no one registered a handler for ModuleB.

PR #3016, opened 2021-09-13
Prevent clobbering built-ins.

Dunbaratu:
(TODO: Read through the issues backlog and find issues this is connected to and put the reference number here so this will close them.  There's possibly more than one such issue.)

This implements a compile-time check that makes it illegal for a script (or the interpreter) to create an identifier that clashes with any of the following:

- The built-in variable names (SHIP, NORTH, ALTITUDE, VELOCITY, etc)
- The system locks (THROTTLE, STEERING, etc)
- The built-in function names (PRINT(), HEADING(), V(), R(), etc)

Note that this is a compile-time check, so it detects the problem and complains even if the offending line of script code is never reached and executed.

**The need for this PR is this**:

Once one of these built-in identifiers gets clobbered by the script, it often cannot be accessible again for the rest of that program.  (or the entire interpreter context).

**The danger with this PR is this**:

Because this was never enforced before, There's probably lots of scripts out there that may have accidentally used a name that masked a built-in, like the example below:

```
// R() is the built-in rotation constructor, and this next line
// replaces it with a new user-defined R() "function":
lock r to 1.

// But if the script never tries to use R() from here on,
// it might have worked anyway, blissfully unaware that it wiped out
// a built-in function.
```

This enforcement in this PR, while better practice, would break existing scripts that have things like the situation above, and require having them rename such variables.  If it's a script the user is actively working on, that's no big deal.  If it's a script the user found out on the internet, it is.

### DIFF
--- a/kerboscript_tests/user_functions/functest34.ks
+++ b/kerboscript_tests/user_functions/functest34.ks
@@ -1,0 +1,57 @@
+//
+// Test SETSUFFIX on the left side of a SET.
+//
+
+// as a function:
+function aaa {
+  return ship.
+}
+set aaa:name to "A".
+set aaa():name to aaa:name + "B".
+set aaa:name to aaa:name + "C".
+if aaa:name = "ABC" {
+  print "SUCCESS: " + aaa:name + " = " + "ABC".
+} else {
+  print "FAILURE: " + aaa:name + " <> " + "ABC".
+}
+
+// as a lock:
+lock bbb to ship.
+set bbb:name to "A".
+set bbb():name to bbb:name + "B".
+set bbb:name to bbb:name + "C".
+if bbb:name = "ABC" {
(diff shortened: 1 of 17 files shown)


Dunbaratu:
Note: one problem to try to fix - the exceptions are coming out saying the line number and column but not the offending filename.  It's being reported in a different format than usual because it's coming from the compiler and not using the normal compiler exception handling - I need to fix that too.

Dunbaratu:
Problem found: Today I was playing with this and tried to do this:
```
lock heading to 3.
```
This (correctly) issued the error message telling me that I can't lock the heading because that would clobber the built-in `heading()`.
BUT, it (incorrectly) still broke the ``heading()`` function anyway, despite the error message telling me it caught the problem and claimed it refused to do it. from then on, I could not ``lock steering to heading(90,0).``

PR #3045, opened 2022-02-15
Found a better API call that works when the navbvall is hidden.

Dunbaratu:
This API call works properly regardless of whether the navball is hidden or not.


Dunbaratu:
Edited because I forgot to include the issue # to auto-close the issue.

PR #3093, opened 2023-08-06
Add new CrewMember suffixes

JonnyOThan:
Sorry about the messy commit history here...should be fine to squash & merge, or if you'd rather I clean this up before sending a PR I can do that.


src/kOS/Suffixed/CrewMember.cs
```diff
@@ -7,7 +7,7 @@ namespace kOS.Suffixed
     [kOS.Safe.Utilities.KOSNomenclature("CrewMember")]
     public class CrewMember : Structure
     {
-        private readonly ProtoCrewMember crewMember;
+        public readonly ProtoCrewMember crewMember;
```
luxgladius, line 10: Did you mean to change this or was it just for debug.

JonnyOThan: My changes to kos-career use it, but I ended up just using krafs publicizer so that I didn't have to wait for this to be merged.  So I don't really care if that change is included, but it might be nice as a marker to say that external code is now relying on it.  Or maybe it would be better to have a public get/private set property.

PR #3125, opened 2024-11-24
Stop setting release_status

HebaruSan:
This metanetkan currently hard-codes a `release_status` of `stable`, which isn't correct because this repo has pre-releases.

Now this is removed.

Hey @JonnyOThan, I see you've pushed some commits to this repo, probably including the one that led to that pre-release. Could you please review this?


JonnyOThan:
Works for me!  Thanks!

PR #3126, opened 2024-11-24
Actually stop setting release_status

HebaruSan:
Hi @JonnyOThan, I didn't notice that the default branch here is develop, whereas the netkan uses master, so the fix from #3125 hasn't been applied yet. Now it's fixed on master as well.


JonnyOThan:
thanks, filed this too: https://github.com/KSP-KOS/KOS/issues/3127

PR #3129, opened 2025-01-30
Fix #3119. Don't expose ListValue<T> lists to users

### LINKED ISSUE #3119: Can't add items to the list returned by ship:dockingports [BUG]
When you get the list `SHIP:DOCKINGPORTS` you can't add items to that list as you get a "specified cast is not valid" error.
Code to reproduce:
```
SET dl TO SHIP:DOCKINGPORTS.
dl:ADD(0).
```
There is also an assoceated entry in the KSP log from the C# side of the mod
```
[LOG 15:29:23.789] System.InvalidCastException: Specified cast is not valid.
  at (wrapper castclass) System.Object.__castclass_with_cache(object,intptr,intptr)
  at kOS.Safe.Encapsulation.Suffixes.OneArgsSuffix`1[TParam].Call (System.Object[] args) [0x00000] in <44fd6c0ba1d84aeaa0b357d09ee83648>:0 
  at kOS.Safe.Encapsulation.Suffixes.DelegateSuffixResult.Invoke (kOS.Safe.Execution.ICpu cpu) [0x00226] in <44fd6c0ba1d84aeaa0b357d09ee83648>:0 
  at kOS.Safe.Compilation.OpcodeCall.StaticExecute (kOS.Safe.Execution.ICpu cpu, System.Boolean direct, System.Object destination, System.Boolean calledFromKOSDelegateCall) [0x002b2] in <44fd6c0ba1d84aeaa0b357d09ee83648>:0 
  at kOS.Safe.Compilation.OpcodeCall.Execute (kOS.Safe.Execution.ICpu cpu) [0x0000d] in <44fd6c0ba1d84aeaa0b357d09ee83648>:0 
  at kOS.Safe.Execution.CPU.ExecuteInstruction (kOS.Safe.Execution.ProgramContext context, System.Boolean doProfiling) [0x00203] in <44fd6c0ba1d84aeaa0b357d09ee83648>:0 
  at kOS.Safe.Execution.CPU.ContinueExecution (System.Boolean doProfiling) [0x00040] in <44fd6c0ba1d84aeaa0b357d09ee83648>:0 
  at kOS.Safe.Execution.CPU.KOSFixedUpdate (System.Double deltaTime) [0x000b0] in <44fd6c0ba1d84aeaa0b357d09ee83648>:0 
  ```
  which combined with the lack of verbosity makes me thing this is a C# problem getting back into kerboScript some what similar to issue #3006
  
  There is a simple work around where one simply creates a new list of there own and populates it with everything from the dockingports list but that should not be necessary.

### PR DESCRIPTION
**Pull request by sug44**
Use `Collections.Generic.List<T>` instead of `ListValue<T>`. Return a `ListValue` copy of these lists to user code.

Changes:
FileContent:binary
Lexicon:keys
Lexicon:values
Part:children
Stage:resources
Stage:resourceslex
String:split
Structure:suffixnames
Timewarp:ratelist
Timewarp:railsratelist
Timewarp:physicsratelist
Vessel:parts
Vessel:dockingports
Vessel:decouplers (separators)
Vessel:resources
allnodes bound variable
allwaypoints function

Stage:resources, Stage:resourceslex, Vessel:parts, Vessel:dockingports, Vessel:decouplers now return copies instead of references to internal lists. All lists returned by these suffixes don't get modified over time so the only way to tell that they are copies now is to use the equality operator. Stage:resources and Stage:resourceslex returning a reference was a bug.

### DIFF
--- a/src/kOS.Safe/Encapsulation/Lexicon.cs
+++ b/src/kOS.Safe/Encapsulation/Lexicon.cs
@@ -380,9 +380,9 @@ namespace kOS.Safe.Encapsulation
         /// to the list.
         /// </summary>
         /// <returns></returns>
-        public override List<StringValue> GetSuffixNames()
+        public override ListValue GetSuffixNames()
         {
-            List<StringValue> theList = base.GetSuffixNames();
+            ListValue theList = base.GetSuffixNames();
 
             foreach (Structure key in internalDictionary.Keys)
             {
@@ -392,7 +392,7 @@ namespace kOS.Safe.Encapsulation
                     theList.Add(keyStr);
                 }
             }
-            return new List<StringValue>(theList.OrderBy(item => item.ToString()));
+            return new ListValue(theList.OrderBy(item => item.ToString()));
         }
         public override Dump Dump()
         {
--- a/src/kOS.Safe/Encapsulation/Lexicon.cs
+++ b/src/kOS.Safe/Encapsulation/Lexicon.cs
@@ -115,10 +115,10 @@ namespace kOS.Safe.Encapsulation
         private void InitalizeSuffixes()
         {
             AddSuffix("CLEAR", new NoArgsVoidSuffix(Clear, "Removes all items from Lexicon"));
-            AddSuffix("KEYS", new Suffix<ListValue<Structure>>(GetKeys, "Returns the lexicon keys"));
+            AddSuffix("KEYS", new Suffix<ListValue>(GetKeys, "Returns the lexicon keys"));
             AddSuffix("HASKEY", new OneArgsSuffix<BooleanValue, Structure>(HasKey, "Returns true if a key is in the Lexicon"));
             AddSuffix("HASVALUE", new OneArgsSuffix<BooleanValue, Structure>(HasValue, "Returns true if value is in the Lexicon"));
-            AddSuffix("VALUES", new Suffix<ListValue<Structure>>(GetValues, "Returns the lexicon values"));
+            AddSuffix("VALUES", new Suffix<ListValue>(GetValues, "Returns the lexicon values"));
             AddSuffix("COPY", new NoArgsSuffix<Lexicon>(() => new Lexicon(this), "Returns a copy of Lexicon"));
             AddSuffix("LENGTH", new NoArgsSuffix<ScalarValue>(() => internalDictionary.Count, "Returns the number of elements in the collection"));
             AddSuffix("REMOVE", new OneArgsSuffix<BooleanValue, Structure>(one => Remove(one), "Removes the value at the given key"));
@@ -158,12 +158,12 @@ namespace kOS.Safe.Encapsulation
             return internalDictionary.ContainsKey(key);
         }
 
-        public ListValue<Structure> GetValues()
+        public ListValue GetValues()
         {
             return ListValue.CreateList(Values);
         }
 
-        public ListValue<Structure> GetKeys()
+        public ListValue GetKeys()
         {
             return ListValue.CreateList(Keys);
         }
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

**Comment by sug44** [comment]
Yes I considered it too. Basically its a tradeoff for consistency over some performance. I want to avoid people having to stumble on these few suffixes that return readonly references and not copies you can do anything with.

**Comment by darthgently** [comment]
Agree with JonnyOThan. I think this is a documentation fix not a code fix.  The docs just need to make clear they are read only (if they don’t already).  What is the use case for manipulating these lists that can’t be done with a deep copy when needed?  They are used far more often to merely read the current list so not worth the performance hit imho

**Comment by nuggreat** [comment]
It seams to me the better solution from a performance stand point (assuming we want these lists to be modifiable in the first place) would be to preform the cast to a generic list when someone preforms a modification on one of the read only lists as apposed to on the initial list construction, which should only modify the suffixes: `:ADD()`, `:INSERT()`, `:REMOVE()`, `:CLEAR()`. and `:COPY()`.

It is also a rare operation that some one wants to modify these lists as I only ever encountered the one person asking after the problem which is what prompted me to make the issue in the first place.  And it is a far more common operation for someone to do as part of main loops
```
FOR p IN SHIP:PARTS {
  ...
}```
or 
```
IF SHIP:PARTS < something {
  ...
}
```
so avoiding performance impact on these would be good.

JonnyOThan:
Yeah copy-on-write is another method, but more complex to implement and maintain.   No objections though if someone wants to try it.

JonnyOThan:
So is the ListValue<T> -> ListValue conversion still useful?  I'm also a little wary of removing typesafety.

sug44:
It would be perfect to: 
- Not expose `ListValue<T>` to users
- Not create extra copies of lists for performance
- Keep typesafety

But I don't think there is an easy way to achieve all three. Keeping typesafety seems less important

sug44:
Basically in the first commit I was prioritizing the first and the third point, now I prioritize the first and the second point. If there was some middle ground I missed let me know

JonnyOThan:
Why is it good to not expose ListValue<T> to users?

JonnyOThan:
OK I read through #3119 and I think I understand...the type returned by those suffixes have a hidden restriction beyond what one would expect from the KOS `List` type?  And since most of them are already making copies of lists, there's no performance issue.

I think I buy that argument, and I'm not too worried about a performance cost for changing `dockingports` and `decouplers` since those aren't used all that often...but introducing a copy for every access to `ship:parts` seems like it might be an issue.

sug44:
I did revert some changes from the first commit and now these ship suffixes don't create a copy. Is there anything else that needs to be addressed?

8110740497:
Si

sug44:
Yeah you have that right. There are 3 main problems with leaving `parts` suffix `ListValue<Part>`. One is that a `:copy` of it will also return a `ListValue<Part>` list. Two is that C# gets mad at users adding elements of other types before kOS gets mad that they are trying to add an element to a read-only list. Three is that a `:typename` on these lists is bugged and returns `ListValue'1` adding to the confusion. I'm confident there is no bug there right now that is caused by the lack of typesafety. But I do agree that its a ticking time bomb before someone introduces it. Just need to make sure that there are no bugs from the point of creation of the list to the point where its set as read-only. Maybe the solution is to put a warning there? It's also possible that the performance hit of copying is not as bad and it's better to take it reducing complexity for both users and us.

JonnyOThan:
Thanks for explaining that.  I think losing typesafety is the least bad outcome here.

I think maybe we need a refactoring where `ListValue` is a base type that manages possibly heterogeneous lists, and `ListValue<T>` can inherit from it and enforce typesafety for use in C#.  But that can be done later.

PR #3130, opened 2025-01-30
Fix `stage:number` suffix returning stage number of the active vessel

sug44:


sug44:
I had that build fail before. It can be fixed by adding a repository variable "BUILD_ARTIFACTS" equal to "Resources/GameData"

JonnyOThan:
ohhhhh yeah, workflows run from PRs can't access environment variables (for "security")

PR #3150, opened 2025-10-17
Fix #3149: Allow kOS users to read the non-gui KSPField properties

### LINKED ISSUE #3149: kOS doesn't allow to get the module field which is not accessible via UI
Sometimes it's helpful to be able to get some information which was not marked as a gui active by the mod creators.
https://github.com/KSP-KOS/KOS/blob/develop/src/kOS/Suffixed/PartModuleField/PartModuleFields.cs#L446-L447
I understand it's probably a bad idea to be able to edit it, but just reading it should be harmless, isn't it?

AntonKuzin:
I think it would be helpful to allow users to get(but not set) fields which are not accessible via UI. Related issue  #3149 

This is not a final PR, just something to start with.

### DIFF
--- a/doc/source/structures/vessels/partmodule.rst
+++ b/doc/source/structures/vessels/partmodule.rst
@@ -31,13 +31,13 @@ Once you have a :struct:`PartModule`, you can use it to invoke the behaviors tha
           - Accessible fields
         * - :attr:`ALLHIDDENFIELDS`
           - :struct:`List` of strings
-          - Fields not notmally accessible via GUI
+          - Fields not normally accessible via GUI
         * - :attr:`ALLFIELDNAMES`
           - :struct:`List` of strings
           - Accessible fields (name only)
         * - :attr:`ALLHIDDENFIELDNAMES`
           - :struct:`List` of strings
-          - Fields not notmally accessible via GUI (name only)
+          - Fields not normally accessible via GUI (name only)
         * - :attr:`ALLEVENTS`
           - :struct:`List` of strings
           - Triggerable events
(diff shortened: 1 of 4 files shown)


src/kOS/Suffixed/PartModuleField/PartModuleFields.cs
```diff
@@ -413,18 +445,21 @@ private void InitializeSuffixesAfterConstruction()
             AddSuffix("ALLACTIONS", new Suffix<ListValue>(() => AllActions("({0}) {1}, is {2}")));
             AddSuffix("ALLACTIONNAMES", new Suffix<ListValue>(AllActionNames));
             AddSuffix("HASACTION", new OneArgsSuffix<BooleanValue, StringValue>(HasAction));
-            AddSuffix("GETFIELD", new OneArgsSuffix<Structure, StringValue>(GetKSPFieldValue));
+            AddSuffix("GETFIELD", new OneArgsSuffix<Structure, StringValue>(argument => GetKSPFieldValue(argument, field => FieldIsVisible(field))));
             AddSuffix("SETFIELD", new TwoArgsSuffix<StringValue, Structure>(SetKSPFieldValue));
             AddSuffix("DOEVENT", new OneArgsSuffix<StringValue>(CallKSPEvent));
             AddSuffix("DOACTION", new TwoArgsSuffix<StringValue, BooleanValue>(CallKSPAction));
+            AddSuffix("ALLHIDDENFIELDS", new Suffix<ListValue>(() => AllHiddenFields("({0}) {1}, is {2}")));
```
AntonKuzin: Dunno, this suffix is barely usable. For example, ROEngine module has more than 100 hidden fields. Given we can't even scroll the in-game kos window, there's no way to inspect the values which didn't fit in the window.

nuggreat: A lot of people do things like this with fields
```
SET p TO somePart().
SET m TO p:GETMODULE(p:ALLMODULES[someModuleIndex).
SET f TO m:GETFIELD(m:ALLFIELDS[someFieldIndex]).
```
Yes it is fragile as hell and might break if anything about the install changes but that doesn't stop people from viewing this as the "easy" way compared to having to type an exact string for the modules/fields in question.

doc/source/structures/vessels/partmodule.rst
AntonKuzin, line 1: This is a draft version of the changes required for the docs(as I see it). I didn't built it, since they would probably still require some polishing from you because I'm not a native english speaker.

PR #3151, opened 2025-10-23
Allow NoDelegate to **really** do nothing

AntonKuzin:
Just made the actual entry point virtual and overrode it. Tested it - works just fine


JonnyOThan:
Does this solve some kind of problem?

AntonKuzin:
> Does this solve some kind of problem?

Yeah, it does. [See here](https://discord.com/channels/210513998876114944/433682230972252160/1430836196803870741)

PR #3186, opened 2026-05-25
Deploy docs on create-release

sug44:


sug44:
Pretty sure the only thing needed for this to work after merging is going to repo Settings > Environments > Add deployment branch or tag rule > Add "develop" branch if its not there already.

